PR #1149, opened 2025-10-31
Docker compose file added

Rupam-It:
## What is the new behavior?

I have update the docker compose file so that it will start frontend , backend and mongodb at once!
Screenshot from 2025-10-31 19-16-24


docker-compose.yml
```diff
@@ -1,25 +1,35 @@
...
+  frontend:
+    image: node:20-alpine
+    working_dir: /app/apps/frontend
     ports:
-      - "5173:5173"
+      - "3000:3000"
+    volumes:
+      - .:/app
+    command: sh -c "corepack enable && pnpm install && pnpm run dev --port 3000"
     depends_on:
```
harshvardhan-yc: `frontend` and `backend` services can run separately. There is no need to make them dependent on each other. Even if `backend` fails to start `frontend` still can run.

harshvardhan-yc (changes requested):

docker-compose.yml
harshvardhan-yc: As `depends_on` just provides the ordering for the container creation and it doesn't do the `health-check`.

Rupam-It: Okk , I will remove the depends on and add a healthy check! Is it fine?

harshvardhan-yc (dismissed):
LGTM. Thanks @Rupam-It for contributions.

Rupam-It:
Hii @harshvardhan-yc I have moved my PR from main branch to dev branch , as it was requested!

harshvardhan-yc:
Closing this as it was stale from long time and we have to re-check compose feature with new dependencies like, postgres, documenso, supertokens.

Issue #877, opened 2025-09-12
Developer's page buttons, lack navigation


SurbhiS8YC:
Buttons on the developer's pages are not clickable or do not trigger any action. Ideally, when a user clicks a button, it should either perform its intended function or smoothly navigate to the next screen as expected. 

It would be great to review all button components and ensure they are connected to their respective actions and screen transitions so that users can interact seamlessly with the application.
1. "Explore dev tools" lacks response.
2. "Learn more" lacks response.
<img width="1470" height="791" alt="Image" src="https://github.com/user-attachments/assets/7495bdd4-9b12-44ba-b6c0-43abdfd6aba6" />

3. "Sign up as a developer" lacks response. 

<img width="1468" height="834" alt="Image" src="https://github.com/user-attachments/assets/57f3c4ce-4b8e-4747-aa3a-142cf33c89a9" />

merahul22:
I checked the current frontend and the exact buttons from the issue screenshot (`Explore dev tools`, `Learn more`, `Sign up as a developer`) are no longer present in the live UI, so this issue appears to be stale relative to the current codebase.

I still made a small navigation cleanup against the current equivalent CTA surfaces and added a local-dev auth guard so the header can render without Cognito env vars. If you want, I can re-scope this issue to the current developer-facing CTAs or close it as outdated and open a new issue for the live UI.

merahul22:
I’ve opened PR #1522 to address the current developer-facing CTA/navigation flow in the live UI.

A quick note on #877: the original issue screenshots appear to be from an older version of the page, because the exact buttons mentioned in the issue (Explore dev tools, Learn more, Sign up as a developer) are no longer present in the current codebase.

Instead of restoring outdated UI, I updated the current equivalent CTA surfaces and also fixed the local-dev auth startup crash so the header can render without Cognito env vars. I added tests to cover the updated navigation behavior as well.

If the maintainers want the old screenshot behavior restored specifically, I’m happy to re-scope it further.

aupyay:
@sneha-yc @harshit-yc

PR #1522, opened 2026-05-01
fix(frontend): align developer CTAs with current navigation

merahul22:
## Summary
- Fix current developer-facing CTA routes in the live header and developer landing surfaces.
- Guard the auth session check so the frontend can render locally when Cognito env vars are not configured.
- Add tests covering the updated CTA routing and the no-Cognito local startup path.

## Why this change was needed
The original issue text for #877 references older CTA labels (`Explore dev tools`, `Learn more`, `Sign up as a developer`) that are no longer present in the current UI. The live codebase now uses the newer header and developer landing flows, so I aligned the existing CTA surfaces instead of restoring outdated screenshot-only buttons.

I also fixed the local-dev auth startup crash caused by `UserPool is not initialized`, which blocked the header from rendering when Cognito is not configured.

## Verification
- `pnpm --filter frontend exec jest src/app/__tests__/components/Header/GuestHeader.test.tsx --runInBand`
- `pnpm --filter frontend exec jest src/app/__tests__/pages/DeveloperLanding.test.tsx --runInBand`

## Notes
- I left the issue comment on #877 explaining that the original screenshot appears stale relative to the current UI.

### DIFF
--- a/.github/workflows/sonar-cloud-analysis.yml
+++ b/.github/workflows/sonar-cloud-analysis.yml
@@ -18,7 +18,12 @@ jobs:
   changes:
     name: Detect workspace changes
     runs-on: ubuntu-latest
-    if: github.repository_owner == 'YosemiteCrew'
+    if: >
+      github.repository_owner == 'YosemiteCrew' &&
+      (
+        github.event_name != 'pull_request' ||
+        github.event.pull_request.head.repo.full_name == github.repository
+      )
     outputs:
       frontend: ${{ steps.filter.outputs.frontend }}
       backend: ${{ steps.filter.outputs.backend }}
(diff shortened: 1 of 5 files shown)


merahul22:
I’ve pushed an update to the branch to address the SonarCloud failure on forked PRs. The workflow was trying to run without ```SONAR_TOKEN```, which GitHub does not expose to forks, so the scan failed with Project not found.

The branch now skips the Sonar scan for forked pull requests, while keeping it active for internal PRs, pushes, and scheduled runs.

harshit-yc:
Thanks for looking into this.

I agree with your note that the original issue is stale. The exact buttons mentioned in #877 are no longer present in the current UI, so I don’t think we should treat this PR as a fix for that issue.

The PR also mixes a few separate things: current CTA href cleanup, local Cognito/no-env auth behavior, and the SonarCloud fork workflow change. Those may each be worth discussing, but they should not be bundled under this old issue.

Let’s not move forward with this PR as-is. If there is still a current CTA/navigation problem, please open a fresh issue or re-scope this with a clear current reproduction. The auth and CI changes should be separate PRs with their own rationale.

PR #1604, opened 2026-06-21
fix(backend): require auth for user and user-organization routes

harshvardhan-yc:
## PR Checklist

- [ ] The PR title follows our guidelines: https://github.com/YosemiteCrew/Yosemite-Crew/blob/main/CONTRIBUTING.md#commit.

## Summary

Tightens access control on the backend user and user-organization surfaces identified in the security audit.

## What changed

- Added `authorizeCognito` to `GET /fhir/v1/user/:id` so user lookup is authenticated.
- Added a self-only check in `UserController.getById` so callers can only access their own user record.
- Added `authorizeCognito` to all `user-organization` routes that were previously unauthenticated.
- Added controller-level authorization checks for `user-organization` reads and mutations so mapping access is limited to authenticated org members with the right permissions.
- Changed the global user-organization list endpoint to return only the current user’s mappings.
- Added/updated tests for:
  - user router auth coverage
  - user-organization router auth coverage
  - user controller self-only access
  - user-organization controller authorization paths


ghost (approved):
GTG

PR #1642, opened 2026-06-23
fix(repo): place Documenso signature fields using percentage coordinates

### LINKED ISSUE #1641: Documenso signature fields placed off-page (point coordinates sent instead of percentages)
**Is your feature request related to a problem? Please describe.**

Document signing is misaligned across every signable document. In the Documenso signing UI the signature field renders far off-page (below/right of the document) instead of on the signature line, so the signer sees a stray, detached signature box with no document context.

**Root cause:** our PDF render/service layer passes signature-field coordinates to Documenso as **PDF points** (e.g. `pageX: 330, pageY: 700`), but the [Documenso fields API](https://docs.documenso.com/docs/developers/api/fields) expects **percentages of the page (0–100), top-left origin**. A point value like `700` is interpreted as `700%`, pushing the field far below the page.

**Describe the solution you'd like**

Send every signature-field placement in Documenso's percentage coordinate system (0–100, top-left). Specifically:

- Merged clinical **packet** builder — convert the returned placement from points to percentages (keep points only for drawing the visible box).
- **Form-submission** render path and the **documenso** last-resort fallback default — express the default placement as percentages.
- Clinical/template renderer (`DocumentEndBlock`) already returns percentages — keep as the reference behaviour.

**Describe alternatives you've considered**

Documenso also supports `{{signature}}` text placeholders auto-detected on upload. Rejected for now: our PDFs are generated programmatically and we already compute exact field rectangles, so a deterministic percentage conversion is simpler and keeps the signing box pixel-aligned with the rendered signature line.

**Additional context**

- Documenso fields API: https://docs.documenso.com/docs/developers/api/fields
- Affects all signable documents: clinical artifacts, resolved templates, clinical packets, and form submissions.

ankit-yc:
## PR Checklist

- [x] There is an issue for the bug this PR is for (#1641).

Documenso signature fields are positioned using PDF **point** coordinates (e.g. `pageX: 330, pageY: 700`). The [Documenso fields API](https://docs.documenso.com/docs/developers/api/fields) expects **percentages of the page (0–100), top-left origin**, so a point value like `700` is read as `700%` and the field renders far off-page, detached from the rendered signature line. This affects every signable document (clinical artifacts, resolved templates, clinical packets, form submissions).

## What is the new behavior?

Every signature-field placement is sent to Documenso as percentages (0–100, top-left):

- **`packages/lib/src/pdf/clinicalPacket.ts`** — the merged-packet placement is converted from top-left points to percentages via a `toFieldPercent` helper. `SIGNATURE_FIELD` stays in points only to draw the visible box, so the Documenso field and the drawn box align exactly.
- **`apps/backend/src/services/rendered-document-renderer.service.ts`** — `DEFAULT_SIGNATURE_PLACEMENT` (form-submission render path + fallback) expressed as percentages.
- **`apps/backend/src/services/documenso.service.ts`** — last-resort fallback default expressed as percentages.
- The clinical/template renderer (`DocumentEndBlock`) already returns percentages and is the reference behaviour (unchanged).

## Validation

- Backend `tsc --noEmit`: clean · `packages/lib` `tsc --noEmit`: clean · backend ESLint (changed files): clean
- Backend jest — 10 affected suites, 118 tests: pass. Added a percentage-range guard in `clinical-packet-pdf.service.test.ts` that fails if a placement regresses to point values (>100).
- **Visual:** rendered the real SOAP and merged-packet PDFs (pending + signed) and overlaid the computed Documenso field rectangle — it lands exactly on the rendered "Signature" line/box. The packet overlay is byte-identical to the drawn signature box.

## Related Issue(s)

Fixes #1641

### DIFF
--- a/apps/backend/src/controllers/web/appointment.prisma.controller.ts
+++ b/apps/backend/src/controllers/web/appointment.prisma.controller.ts
@@ -278,6 +278,9 @@ export const AppointmentController = {
         req.params.appointmentId,
         {
           admittedAt: body.admittedAt ? new Date(body.admittedAt) : undefined,
+          // The admitting user is whoever is signed in and clicked
+          // "Convert to Inpatient" (the verified Cognito token), never the body.
+          admittedBy: (req as { userId?: string }).userId,
           expectedStayDays: body.expectedStayDays,
           lead: body.lead,
           supportStaff: body.supportStaff,
--- a/apps/backend/src/services/appointment.prisma.service.ts
+++ b/apps/backend/src/services/appointment.prisma.service.ts
@@ -90,6 +90,7 @@ type AdmissionUpsertDelegate = {
       organisationId: string;
       patientId: string;
       admittedAt: Date;
+      admittedBy?: string | null;
       expectedStayDays?: number | null;
     };
   }): Promise<unknown>;
@@ -209,6 +210,7 @@ type AdmitSupportStaffInput = {
 };
 type AdmitRequestInput = {
   admittedAt?: Date;
+  admittedBy?: string;
   expectedStayDays?: number;
   lead?: AdmitLeadInput;
   supportStaff?: AdmitSupportStaffInput[];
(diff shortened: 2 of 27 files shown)


harshvardhan-yc (approved):
Good to go.

Issue #1659, opened 2026-06-25
Payment gateway abstraction - Stripe (cards) + CareCredit / Scratchpay (financing)


ankit-yc:
## Summary

Introduce a provider-agnostic payment-gateway abstraction in \`apps/backend\` where **Stripe is the card-payment implementation** behind a stable \`PaymentProvider\` interface, and add **CareCredit and/or Scratchpay** as financing options selected per organisation. The feature delivers a clean **org onboarding + account-readiness** flow, **first-class failure handling**, a **forward path for additional providers** without rewriting invoice/payment/refund flows, and **client-side support** for whichever provider an organisation has selected. The platform's own SaaS subscription billing stays on Stripe and is out of scope.

Every change to payment configuration must be **change-audited** through the existing \`AuditTrailService\`.

## Why CareCredit / Scratchpay instead of Adyen

The original scope included Adyen as a second card processor. Adyen for Platforms contracts with the platform entity as the regulatory principal - even with 100% of funds routed to the clinic, the platform bears regulatory exposure. That is incompatible with the platform's zero-regulatory-exposure constraint.

CareCredit (Synchrony Financial) and Scratchpay are healthcare financing providers where each clinic signs their own agreement directly with the provider. The platform is the facilitating software and has no lending or regulatory exposure. These also address the real driver: Stripe's per-transaction fee is high for some clinics, and financing providers convert treatments that would otherwise be declined due to cost - increasing clinic revenue.

## Settlement model

The platform takes no fee and is never the financial principal. The clinic is the merchant of record for every customer payment.

- **Stripe:** Standard Connect accounts with direct charges. The charge is created in the clinic's account context. The clinic bears the processing fee, dispute liability, and negative-balance liability. No application fee. The current Stripe flow uses destination charges (platform is MoR); migrating to direct charges is a prerequisite tracked in #1678.
- **CareCredit / Scratchpay:** Each clinic signs directly with the financing provider. The financing company underwrites the patient's loan, pays the clinic upfront, and collects repayments. The platform initiates the application flow and receives confirmation; it is not a lender and has no regulatory exposure.

## Current state and where it plugs in

The repo already has a partial abstraction: the persistence layer is provider-neutral, but the orchestration layer is Stripe-specific.

**What exists:**
- \`enum PaymentProvider { STRIPE  MANUAL }\` is referenced by \`PaymentAttempt\`, \`Payment\`, and \`Refund\`.
[...]

PR #1677, opened 2026-06-25
feat(repo): add currency-exponent-aware minor-unit money helper

ankit-yc:
## PR Checklist

- [x] The PR title follows the commit guidelines.
- [x] There is an issue for this feature (#1659).

The backend converts money to provider minor units with a hardcoded multiply-by-100. That is correct only for two-decimal currencies. There is no shared, currency-aware conversion, so zero-decimal currencies (JPY, KRW) and three-decimal currencies (BHD, KWD) would be mis-charged by a factor of 100 or 10. A shared minor-unit helper is a prerequisite for the payment-gateway abstraction in #1659, where a second provider transacts in ISO 4217 minor units.

## What is the new behavior?

Adds a shared, dependency-free money helper in `packages/lib` (`packages/lib/src/money/minor-units.ts`), exported from the package barrel:

- `getCurrencyExponent(currency)` returns the ISO 4217 minor-unit exponent (0, 2, 3, and the 4-decimal special cases), defaulting to 2 for unknown but well-formed codes, and validating the code shape.
- `toMinorUnits(major, currency)` converts a major amount (e.g. 19.99) to an integer minor amount (e.g. 1999), rounding half away from zero and compensating for binary floating-point drift.
- `fromMinorUnits(minor, currency)` converts back.

Additive only. No existing code is changed or rewired in this PR. Wiring the helper into the finance flows is a separate, behaviour-affecting step that the platform fee and split policy decision still gates.

## Validation

- 26 unit tests in `apps/backend/test/utils/money-minor-units.test.ts` covering exponent lookup, each currency class, rounding, negative amounts, floating-point drift, error paths, and round-trips. All pass.
- `@yosemite-crew/lib` type-check clean. Files prettier-formatted.

## Related Issue(s)

Part of #1659. This is the first decision-independent Phase 0 prerequisite from the design plan; it does not close the issue.

### DIFF
--- a/apps/backend/test/utils/money-minor-units.test.ts
+++ b/apps/backend/test/utils/money-minor-units.test.ts
@@ -1,0 +1,150 @@
+import {
+  fromMinorUnits,
+  getCurrencyExponent,
+  toMinorUnits,
+} from "@yosemite-crew/lib";
+
+describe("getCurrencyExponent", () => {
+  it("returns 2 for standard two-decimal currencies", () => {
+    expect(getCurrencyExponent("USD")).toBe(2);
+    expect(getCurrencyExponent("EUR")).toBe(2);
+    expect(getCurrencyExponent("GBP")).toBe(2);
+    expect(getCurrencyExponent("INR")).toBe(2);
+  });
+
+  it("returns 0 for zero-decimal currencies", () => {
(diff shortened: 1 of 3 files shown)


ankit-yc:
Consolidated into #1685 (single PR for all Phase 0 payment gateway work). The money helper code and tests are now included in that branch.

PR #1750, opened 2026-07-01
fix(frontend): fix inventory batch/stock UX, per-batch fields, and on-hand stock derivation

vallirani-yc:
## PR Checklist

Several bugs and UX issues in the inventory feature (`apps/frontend`):

- In both the Add and Edit inventory forms, "Stock Control" appeared before "Batch and expiry", even though batch quantities are what feed the stock totals.
- "On hand stock" was a manually editable field even though it is always derived from the sum of batch quantities, so it could show a value inconsistent with the batches.
- The "Available stock" label didn't clarify that it represents dispensable stock.
- Editing a batch's "Expiring warning before" or "Barcode" collapsed the value across all batches into one shared item-level attribute — every batch showed the same value instead of its own.
- Clearing "SKU (optional)" in the edit form and saving would silently resurrect the old SKU value, because the payload builder fell back to a stale top-level `sku` field whenever the edited field was empty.
- The inventory table always showed an "All inventory" filter tab alongside "Active" and "Hidden".
- The stock-health pill labeled "Healthy" didn't match the "In stock" terminology used elsewhere in the same file.
- The "Low stock" and "Expired" status pills were rendered with swapped colors (Low stock showed purple, Expired showed orange) instead of amber and red respectively.
- `EditableAccordion`'s `footer` prop only accepted a static `ReactNode`, so a footer couldn't merge its own field changes (e.g. an inline image upload) into the same draft as the rest of the section before saving.

## What is the new behavior?

- "Batch and expiry" now renders before "Stock Control" in both the Add and Edit inventory forms.
- "On hand stock" is now a read-only, computed badge (styled like "Available stock") in both forms, always reflecting the sum of batch quantities.
- "Available stock" is now labeled "Available stock (dispensable)" in both forms.
- `expiryWarningBefore` and `barcode` are now sent and read per batch (the backend now supports per-batch storage for these fields), so each batch keeps its own value instead of being merged into a shared item-level attribute.
- Saving the SKU field as blank now sends an empty value to the backend instead of falling back to the stale original value, so clearing it actually clears it.
- The "All inventory" tab has been removed from the inventory table filters; only "Active" and "Hidden" remain.
- The stock-health pill label "Healthy" has been renamed to "In stock" (color/style unchanged — already mapped to the same success/green style).
- "Low stock" now renders with the amber/warning pill style, and "Expired" now renders with the red/danger pill style.
- `EditableAccordion`'s `footer` prop now also accepts a function `(context: { values, isEditing, setFieldValue }) => ReactNode`, allowing footer-rendered fields (e.g. an inline image upload) to merge into the same draft that gets saved.

## Related Issue(s)

Fixes #


harshvardhan-yc (approved):
Good to go.

PR #1752, opened 2026-07-01
docs(repo): formalize release process

ankit-yc:
## Summary
- Add `RELEASING.md` documenting the release process as it actually works today: `dev`/`main` branch model, continuous deploy for `apps/frontend`, and discrete component-prefixed tags (`backend-v*`, `pms-v*`, `mobile-v*`, `desktop-v*`) for the rest - explicitly noting only desktop releases are currently CI-automated and that `package.json` versions are not synced to tags.
- Add `cliff.toml` + `.github/workflows/release-notes.yml`: on any `*-v*` tag push, generate release notes from the Conventional Commits already enforced by `commitlint`/`pr-governance.yml` and attach them to the GitHub Release via `gh release edit/create`. No new commit discipline required from contributors, nothing committed back to `main`/`dev`.

## Why
Release notes are currently hand-written per release; the repo already forces every commit into a machine-parseable format, so this removes a manual step without changing when/how releases are decided or built.

## Test plan
- [x] Docs render correctly; links to `.github/workflows/*` and `cliff.toml` resolve.
- [ ] Verify the workflow end-to-end on the next real tag push (e.g. next `backend-v*` release) - couldn't dry-run a tag push from this PR without cutting a release.
- [ ] Maintainer call on whether to backfill notes for recent releases once this is merged.


ankit-yc:
Reviewed this against current `dev` before merging. **Still needed** - `RELEASING.md`, `cliff.toml` and `release-notes.yml` are all absent on `dev`, and there is no release-notes automation there, so nothing has superseded it. The `RELEASING.md` content also still describes the process accurately (`cd-frontend.yaml` and `desktop-release.yml` both still exist, and the note that only desktop is CI-automated still holds).

**But it was not ready.** This PR predates the CI supply-chain hardening, and as written it would have regressed three things that are now standard. Pushed `ca5763611` to fix them:

1. **Unpinned action.** `actions/checkout@v4` was a floating tag. Every `uses:` on `dev` is a 40-char SHA - I checked, there are zero unpinned third-party actions - so this would have been the only one, and three majors behind (dev is on v7.0.0). Pinned to the same checkout and setup-node SHAs `dev` already uses, verified against the tree rather than invented.

2. **Unpinned binary download.** The git-cliff step queried the releases API for `latest`, grepped a download URL, fetched a tarball with no checksum, and `sudo mv`-ed the binary into `/usr/local/bin`. That is an unreviewed third-party executable whose version could change between two runs of the same workflow. Replaced with `npx git-cliff@2.13.1`, which pins the version and matches the existing `pnpm dlx turbo@2.10.4` pattern.

3. **Template injection.** `${{ github.ref_name }}` and `${{ steps.prev.outputs.prev }}` were interpolated directly into `run:` blocks. A tag name is untrusted input to this workflow, and this is the same shape already remediated elsewhere in the repo. All tag-derived values now pass through `env:` and are referenced as shell variables.

No change to what the workflow produces.

One thing left as-is deliberately: the test plan's second item, verifying end-to-end on a real tag push, still cannot be exercised without cutting a release. The workflow only triggers on `*-v*` tags, so the blast radius of it being wrong is a bad release-notes body, not a broken build - worth watching on the next `backend-v*` or `desktop-v*` tag rather than blocking the merge.

ankit-yc:
All three review findings confirmed and fixed in `4e24dc03d`. Two of them were genuine blockers.

**P1 - `tag_pattern` is a regex, not a glob. Correct, and it made the whole feature dead on arrival.** git-cliff refuses to load the config:

```
Error: ConfigError(regex parse error: repetition operator missing expression
for key `git.tag_pattern`)
```

Since the workflow only triggers on tag pushes, this would have failed on every release and nobody would have found out until the first real one. Replaced with a regex anchored to the prefixes RELEASING.md actually documents, rather than a permissive `.*-v[0-9].*`, so an unrelated tag cannot be treated as a release boundary:

```toml
tag_pattern = "^(backend|pims|pms|mobile|desktop)-v[0-9].*$"
```

Verified against the real tag shapes - `backend-v1.4.0-beta`, `pims-v1.3.0-beta`, `pms-v1.0.0-beta`, `mobile-v2.0.0-iOS-v1.2`, `desktop-v0.2.0-beta` all match; `random-tag` and `v1.0.0` do not. git-cliff now exits 0 and emits grouped notes.

**P1 - desktop must not follow the generic guide. Also correct, and verified in the tree.** `apps/desktop/RELEASE.md:32-34` requires bumping `version` in `apps/desktop/package.json` and then cutting a tag that matches it exactly; `artifactName` interpolates `${version}`; and `RELEASE.md:17` notes the updater compares SemVer. `package.json` currently reads `0.1.0-beta.1`. So the failure described was real: tagging `desktop-v0.2.0-beta` would have shipped an installer and update manifest still identified as `0.1.0-beta.1`.

RELEASING.md now scopes the not-synced claim to backend, frontend and mobile, and calls desktop out as an explicit exception pointing at its own `RELEASE.md`. The "what this does not do" section was carrying the same wrong assumption and is corrected too.

**P2 - previous-tag selection. Agreed.** Sorting by `version:refname` picks the highest tag, not the preceding release, and without `versionsort.suffix` git ranks `backend-v1.0.0-beta` above `backend-v1.0.0`. Now:

```sh
PREV=$(git describe --tags --abbrev=0 --match "${PREFIX}-v*" "${TAG}^" 2>/dev/null || true)
```

The nearest matching tag reachable from the tag's parent. Checked against real tags in this repo: `desktop-v0.1.0-beta.2` correctly resolves to `desktop-v0.1.0-beta.1`, and the first tag in a line correctly resolves to nothing (full history).

Thanks - the `tag_pattern` one in particular would have shipped silently.

ankit-yc:
Fixed in `a6fc7594d`. This finding was **current, not stale** - worth stating because I got that wrong earlier in this PR.

The bot comment is timestamped `2026-07-25T00:52:03Z` against `commit_id=b8dcfa3f`, which is my permissions commit from `00:51:32Z` - so it landed 31 seconds after that push, on the latest code. It only reads as hours old because of elapsed wall-clock time. I had previously said no Aikido run existed on my commits; that was wrong. Aikido posts a commit **status** plus review comments rather than a check-run, so the check-runs API I was querying returned nothing and I drew the wrong conclusion from an empty result.

**The finding itself is legitimate.** `actions/checkout` writes `GITHUB_TOKEN` into `.git/config` unless `persist-credentials` is false, leaving it readable by every later step in the job and by any action those steps invoke. This is the vector that made the 2025 action-compromise incidents as damaging as they were.

Set `persist-credentials: false`. Verified nothing here needs it: the only remote operations are `gh release view/edit/create`, which authenticate with `GITHUB_TOKEN` from the step environment, not from git config. There is no `git push`, `fetch`, `clone` or `remote` anywhere in the workflow - `fetch-depth: 0` plus local `git tag`/`git describe` is the extent of the git usage, and those need no credentials.

This also brings the workflow in line with what the repo already does - **9 of the 14 workflows on `dev` set `persist-credentials: false` on every checkout**, including `_core.yaml`, `_test.yaml`, `codeql.yml` and `secret-scan.yml`.

Also documented why `fetch-depth: 0` is there, since the two options now sit together and the reason for the full history is not self-evident.

Running tally on this PR, all found by review rather than by CI: an unpinned floating action tag, an unpinned root-installed binary with no checksum, template injection from tag names, a `tag_pattern` regex that could never load, workflow-wide write permissions, and now a persisted credential. Worth noting five of the six are supply-chain surface rather than anything the release notes themselves do.

PR #1762, opened 2026-07-03
feat(repo): ActivityPub federation, org verification, agent tasks, and clinic directory

### LINKED ISSUE #1694: ActivityPub federation for self-hosted and cloud instances
**Is your feature request related to a problem? Please describe.**

Clinics on separate Yosemite Crew instances (cloud or self-hosted) cannot discover each other, send patient referrals, share records, or message cross-instance. Every inter-clinic handoff today requires phone calls or manual PDF transfers.

**Describe the solution you'd like**

Implement the ActivityPub protocol (W3C Recommendation) so every organisation becomes a federated actor -- discoverable via WebFinger, with a signed inbox/outbox. Instances exchange activities over HTTPS with no central broker.

Activities map to the vet domain:

- Follow / Accept / Reject -- clinic network opt-in
- Offer(Referral) / Accept / Decline -- patient referral between clinics
- Create(PatientSummary) -- curated record share, gated on existing parent consent (PatientOrganisation.status = ACTIVE)
- Create(Note) -- cross-instance staff messaging, bridged into ChatSession
- Announce(Emergency) -- urgent broadcast to follower network

New shared package packages/activitypub/ handles HTTP Signature signing/verification, actor JSON-LD, delivery queue, and inbox dispatch. A single AP_BASE_URL env var drives all actor URIs for portability across cloud and self-hosted deployments.

**Describe alternatives you've considered**

- Proprietary webhook bridge -- requires a central directory and does not work for unknown self-hosters.
- FHIR Bulk Data export -- not real-time, no referral workflow, no messaging.

**Additional context**

Closes #1654. Downstream dependency for #1674.

Patient data is always addressed delivery only -- never Public. The existing PatientOrganisation consent model gates all clinical data transfer; no new consent primitives are needed.

ankit-yc:
## PR Checklist

PIMS instances (cloud-hosted and self-hosted) have no way to interoperate. A clinic cannot follow another clinic, send a clinical referral, or broadcast an emergency notice to peers. There is no federation protocol and no trust model that limits interoperation to verified businesses.

## What is the new behavior?

Implements W3C ActivityPub federation so any Yosemite Crew instance can federate with any other, gated so that only license-verified businesses may connect.

Backend (`apps/backend`):
- ActivityPub actor model per organisation with an RSA-2048 keypair (private key encrypted at rest with AES-256-GCM + scrypt key derivation).
- Public endpoints under `/ap`: actor, inbox, shared inbox, outbox, followers, following. WebFinger, host-meta, and NodeInfo under `/.well-known` and `/nodeinfo/2.0`.
- Inbound activities are queued on BullMQ (`ap-inbox`) and processed by a worker; outbound delivery is queued on `ap-delivery` with retries. Handlers cover Follow, Accept, Reject, Undo, Offer (referral), Create (note), and Announce (emergency broadcast).
- Trust model: only instances presenting a valid license token (JWT RS256, verified against the license authority JWKS with revocation and expiry checks) may federate. Referrals require an accepted follow link before any patient summary is sent.
- Security hardening applied throughout inbound verification and outbound fetching:
  - HTTP Signatures verified over the raw request bytes, with a mandatory Digest, an enforced signed-header set covering `(request-target)`, `host`, `date`, `digest`, and a 300-second replay window.
  - The verified signer is bound to the activity actor, and a fetched actor document must declare an id and key on the same origin it was served from, preventing actor impersonation.
  - All outbound requests pass an SSRF guard that requires https and rejects loopback, private, link-local, ULA, site-local, metadata, v4-mapped, v4-compatible, and NAT64-embedded addresses, with redirects disabled.
  - Inbound activities are deduplicated by id so a replayed signed request cannot re-run a handler.
- `AP_ENABLED` is a fail-closed kill switch: the entire federation surface, including well-known discovery, is disabled unless it is set to `true`.

Frontend (`apps/frontend`):
- A Federation section in Settings: view the instance actor identity, paste and verify a license token, manage followers (approve/reject) and following (follow/unfollow), send and respond to referrals, and broadcast emergency notices.

Database (`packages/database`):
- New models `APActor`, `APRemoteActor`, `APFollower`, `APFollowing`, `APActivity`, `APReferral` with supporting enums, plus a migration.

New environment variables (see `apps/backend/.env.example`): `AP_BASE_URL`, `AP_ENABLED`, `AP_AUTO_APPROVE_FOLLOWS`, `AP_LICENSE_AUTHORITY_URL`.

### Automated organization verification

Federation is restricted to verified businesses, so this PR also makes `Organization.isVerified` a derived value instead of a client-supplied flag:

- A new `organization-verification.service.ts` computes verification as: a manual override (when set) wins; otherwise the org must have an active Stripe Connect account (`canAcceptPayments`) AND at least one compliance certificate number.
- The Stripe `account.updated` Connect webhook (`_handleAccountUpdated`) recomputes `isVerified` for every affected org when Connect status changes.
- Onboarding/update always recomputes `isVerified` rather than trusting the request body. This also closes a pre-existing path where the onboarding DTO could set `isVerified: true` directly.
- A new nullable `Organization.verificationOverride` column plus a `setVerificationOverride` service method allow the verification authority to force verify/unverify. This is deliberately NOT exposed as an org-scoped endpoint: an org must never be able to self-verify and bypass the federation trust gate. Setting the override is reserved for SuperAdmin's privileged path, consistent with license-token issuance.

### Agent-to-agent task protocol (yc:AgentTask)

Adds a structured task channel so a clinic's automation can query another clinic's instance over the same signed, license-gated transport (read-only in v1):

- New `yc:AgentTask` / `yc:AgentTaskResult` JSON-LD objects and a `sendAgentTask` sender (an Offer wrapping the task).
- Inbound `handleOffer` now branches by object type: referrals are unchanged; a `yc:AgentTask` is routed to `handleAgentTask`, which answers only license-verified requesters (same trust gate as follows).
- v1 auto-answers `capability_query` (returns the clinic name, type, and active specialities from real data). Any other task type - including `availability_query` and anything that would write data or make a clinical decision - is not auto-answered: it gets a `Reject` with reason `requires_human_review`, so a human at the receiving clinic handles it.
- Security boundary: `yc:input` is treated as untrusted data, never as instructions. v1 handlers are purely programmatic (no LLM), and the boundary is documented at the dispatch point for any future LLM-backed handler.

### Clinic directory (opt-in discovery)

Lets a clinic browse other verified clinics on the network and follow them, without needing to know an exact actor URI up front:

- New `APActor.directoryListed` flag (opt-in, default off).
- `PUT /ap/manage/directory-listing` - verified-only; sets the local flag and registers the listing with the SuperAdmin authority (`PUT {AP_LICENSE_AUTHORITY_URL}/api/directory/listing`, license-token bearer). `GET /ap/manage/directory` proxies the authority's directory (`GET /api/directory`), cached in-memory (~60s) via `@yosemite-crew/lib`; if the authority is unreachable it logs and returns an empty list so the page still renders.
- `getActorSettings` now also returns `isVerified` and `directoryListed`.
- Frontend: a new `/network` page rendering clinic cards (name, handle, host) each with a Follow button (reusing the existing follow flow), plus a "list this clinic in the directory" toggle in Federation settings that is disabled until the org is verified.

Note: the SuperAdmin-side directory registry (`GET /api/directory` + `PUT /api/directory/listing`) is a separate task on the SuperAdmin repo. This PIMS side is built and tested against that contract and degrades gracefully (empty directory) until the authority endpoint is live.

## Validation performed

- Backend: `type-check` clean, `lint` clean, full backend Jest suite green. New unit tests cover the SSRF guard (including v4-mapped/v4-compatible/NAT64 literal bypasses), the AES-256-GCM crypto round-trip and tamper rejection, and inbound signature verification (digest, replay window, keyId binding).
- Frontend: `tsc --noemit` clean, `lint` clean, federation service and section tests green.
- The diff was reviewed for security issues by an adversarial multi-agent pass; the SSRF and actor-impersonation findings it surfaced are fixed and re-verified in this branch.
- Organization verification: new `organization-verification.service.ts` at 100% coverage; backend type-check, lint, and the full backend suite pass.

## Related Issue(s)

Fixes #1694


ankit-yc:
### Revived and rebased onto current dev

This branch had been sitting since 3 July, 1662 commits behind `dev` and in a CONFLICTING state. Worth noting for anyone reading the earlier green checks: **CI was not running on this PR at all** while it was conflicting, so those passes were a stale re-run against an old base and proved nothing about today's `dev`.

Rebased rather than merged. Only 8 real conflicts, all of them wiring points.

**What the seven week gap had broken, none of which showed up as a conflict:**

- `authorizeCognito` no longer exists. The SuperTokens migration retired it, so the federation router did not compile against `dev`. Now uses `requireWebAuth`.
- `crypto.JsonWebKey` is gone from the current `@types/node`, so two `as unknown as crypto.JsonWebKey` casts had quietly degraded to `error` and were muting the type checker.
- The Settings page was restructured. `SettingsSectionSkeleton` no longer exists, and sections now live in a two column grid of `PreferenceGroup`s.
- React Hooks 7 added `react-hooks/set-state-in-effect`, which rejects the loader pattern this panel used.
- The panel hardcoded Tailwind's default palette. Those classes still emit CSS, so nothing was invisible, but they render cool toned against warm bone and stay light in dark mode. Now on the shared `StatusPill` and the semantic `danger`/`success`/`warning` tokens.
- The Prisma schema was left invalid by the conflict resolution (a lost closing brace). The pre-push hook caught it because it runs `prisma generate`; `type-check` alone passed against a cached client.

**Security gaps the Aikido gate surfaced, all real and all fixed here:**

1. **Only `sendFollow` enforced the federation licence.** `sendAgentTask`, `sendNote` and `announceEmergency` were entirely ungated, and `sendReferral` was covered only indirectly by its accepted follow consent gate. An instance with no valid licence could originate federation traffic, which is the one thing the trust model exists to prevent. Extracted `assertVerifiedInstance` and applied it to all five.
2. **`/host-meta` had no `apEnabled` guard** while `/webfinger` did. It publishes the webfinger template, so with `AP_ENABLED` off this instance still advertised a federation surface, contradicting the fail closed kill switch described in the PR body above. An existing test asserted the ungated behaviour as intended; it was not, and has been corrected.
3. **`directoryCache` keyed every organisation's directory under one literal `"directory"` key** while the fetch used the calling org's licence token, so an unlicensed org could read the copy a licensed one had just warmed. Now keyed by `orgId`.

The existing suite stayed green through all three, because the outbound senders had no coverage at all. Each new guard is mutation checked: disabling `assertVerifiedInstance` fails 5 tests, restoring it returns the suite to green.

**The SuperAdmin dependency is now built.** The directory half of this PR was calling `GET /api/directory` and `PUT /api/directory/listing`, which had never been implemented, so the `/network` page would have rendered permanently empty. Built in YosemiteCrew/SuperAdmin#183 (issue YosemiteCrew/SuperAdmin#182).

One contract gap came out of that work: the licence token claims carry `orgId`, `instanceDomain`, `tier` and `keyId`, but no organisation name, handle or actor URI, so a directory entry built from the token alone has nothing to display. The `PUT` now sends `actorUri`, `orgName` and `handle`, and the authority binds `instanceHost` from the claim and rejects an `actorUri` on any other host. The order was also wrong: the local `directoryListed` flag is now written only after the authority accepts, so a failed call no longer leaves the toggle claiming a listing that was never recorded.

**Validation:** backend 425 suites / 10109 tests, lint and type-check clean. Frontend type-check clean, federation and settings suites 242 tests passing, changed files lint clean at `--max-warnings=0`.

ankit-yc:
### All review findings addressed

Every open thread on this PR is now fixed. Mapping each to where it landed, so the resolutions can be checked rather than taken on trust.

**Trust boundaries**

| Finding | Fix |
| --- | --- |
| Keep private activities out of the public outbox | The unauthenticated outbox served raw JSON for every recent outbound activity, including referral Offers carrying species, medications, allergies and clinical context. Filtered to activities explicitly addressed to the public collection, scanning a wider window than it serves so a burst of directed activities cannot push public ones off the page. |
| Recompute derived verification in the update path | `isVerified` is no longer written from the client payload at all, and `update()` recomputes like `upsert()` already did. A caller with `teams:edit:any` could previously self-verify with no Stripe Connect account and no compliance certificate. |
| Enforce permissions on federation management routes | All 17 `/manage` routes now carry auth, permission loader and gate inline, matching `integration.router` and the repo's route-authz invariant. |
| Enforce inbound referral trust gates | A signature proves who sent an Offer, not that they may. Both the licence check and an approved follow link now apply inbound. |
| Bind licenses to the actor identity | Bound to the org in the actor URI, not just the domain, so a neighbour on a shared instance host cannot present a copied token. Only enforced when a full actor URI is passed; local self-checks use the bare base URL, which names no organisation. |
| Reject all non-public IPv4 ranges | Added carrier-grade NAT, IETF protocol assignments, benchmarking and the TEST-NETs, with boundary tests either side of each block. |
| Shorten or invalidate the revocation cache | Revocations now have their own 5 minute TTL instead of sharing the 24 hour signing-key TTL. |
| Cap remote actor response sizes | `maxContentLength` and `maxBodyLength` set to 256 KiB; Axios defaults both to unlimited. |

**Correctness and robustness**

| Finding | Fix |
| --- | --- |
| Dispatch Accept and Reject by their referenced object | Dispatched on the embedded object type, with the referral branch scoped to an Offer this instance originated and addressed to the responder. |
| Echo the original Follow in approval responses | Both the auto-approval and manual paths now echo the Follow being accepted. The manual path recovers the stored inbound activity and falls back to a synthetic one only if it has been pruned. |
| Rethrow transient Follow processing failures | Rethrown so BullMQ can retry. Nothing in that block throws terminally; the one terminal outcome returns instead. |
| Stop after losing the activity insert race | The loser now stops instead of running the handler alongside the winner. |
| Route follower broadcasts received through the shared inbox | Recipients resolve from the follow graph when addressing names no local organisation, which is always the case for a public broadcast. Explicit addressing still wins. |
| Delay the referral state change until the response is durable | The state change moved after the remote fetch, activity insert and enqueue, so a failure leaves the referral PENDING and retryable. |
| Validate referral payloads with Zod | Schema added; a string `patientSummary`, unknown urgency, oversized text or non-string medications are rejected at 400 rather than reaching Prisma. |
| Make generated actor usernames collision-resistant | Suffixed with a stable slice of the organisation id, with a fallback for names that normalise to nothing. |
| Refresh cached actors after signature failure | One forced refresh on failure, guarded on the record's age so forged signatures cannot be used to make this instance hammer a remote host. |
| Stabilize notification callbacks used by loading effects | `useNotify` now returns a memoised callback. It was a new function every render, so the federation panel and network directory re-ran their loaders endlessly. Fixed at the root, which covers every other component that lists `notify` in dependencies. |
| Eliminate obvious within-file duplication (handleAccept / handleReject) | Shared transition logic extracted. |
| Duplicated federation trust gate (Follow / referral / AgentTask) | Extracted `resolveLicensedRemote`. |
| Keep functions concise (SendReferralCard) | Split into a `useReferralForm` hook, a `ReferralField`, `ReferralFormFields` and `SentReferralsList`. Ids and labels unchanged, since the tests bind through `getByLabelText`. |

**Migration findings** ("add the missing actor license column", "remove the unmapped required activity timestamp") were correct and are resolved by regenerating the migration from `schema.prisma` with `prisma migrate diff --from-empty --to-schema-datamodel`. The hand-written original was authored on 3 July and the models moved underneath it; the drift check that catches this had never run, because the PR was CONFLICTING for seven weeks and CI does not run on a conflicting PR.

**On verification.** Each guard was mutation checked rather than assumed: the fix was reverted, the suite confirmed to fail, then restored. That caught one case where the new stale-key retry path was unreachable in tests, because the service mock omitted `fetchedAt` and the retry threw into an outer catch that returned the same result as a normal rejection. Four existing tests also asserted the defective behaviour by name and were rewritten.

Backend 425 suites / 10154 tests, frontend 950 / 11664, lint and type-check clean on both.

**Remaining, and not fixed on purpose:** two Aikido SSRF findings on `activitypub.service.ts` and `ap-delivery.worker.ts`. Both call sites carry `assertPublicHttpsUrl`, `maxRedirects: 0` and `guardedHttpsAgent`, and Aikido's own assessment downgraded them as unlikely to be exploited. Flagging rather than dismissing them, since that judgement should not be mine to make silently.

Issue #1796, opened 2026-07-08
Security hardening: cross-tenant authorization, RBAC scoping, and CI supply-chain


ankit-yc:
## Summary

Track a security hardening pass across the API and CI, following an internal security scan. The scan's findings collapse into a small number of recurring themes rather than isolated bugs; this issue tracks addressing those themes at their root and sweeping the surfaces that inherit them.

## Problem / motivation

Recurring themes to address:

- Authorization scoping: some resource routes authorize against a requester-supplied context instead of deriving it from the target resource.
- Permission checks: some multi-permission checks are broader than intended.
- Input validation: identifiers used for authorization are not always strictly validated.
- Ownership checks: some client-facing routes should verify resource ownership before returning or mutating data.
- Supply-chain and hygiene: third-party code used in CI and local tooling should be pinned; secrets should be kept out of logs; static assets should carry appropriate security headers; secret-bearing workflows should be gated.

Specifics are tracked privately in the security dashboard and are intentionally not restated here.

## Proposed solution

Fix each theme centrally, then sweep the affected surfaces:

- Derive tenant and ownership from the resource for resource-scoped routes.
- Tighten permission checks that are broader than intended.
- Strictly validate identifiers used for authorization.
- Pin third-party code, redact secrets from logs, apply security headers to static assets, and gate secret-bearing workflows.

## Scope

- Backend authorization middleware and the routes and services that inherit the themes above.
- Payment and webhook handling, and document-signing surfaces.
- CI workflows, local hooks, and related frontend security-header hygiene.

### Out of scope

- Broad refactors beyond the security boundary.
- Items already addressed on `dev` or in open security PRs.

## Acceptance criteria

- [ ] Authorization is derived from the resource for resource-scoped routes.
- [ ] Permission and ownership checks match intended access.
- [ ] Identifiers used for authorization are strictly validated.
- [ ] Third-party code is pinned; secrets are kept out of logs; static assets carry security headers; secret-bearing workflows are gated.

## Test plan

- [ ] Unit tests for the input-validation and permission changes.
- [ ] Cross-tenant and ownership integration tests for representative resources.

## Dependencies & risks

- Coordinate with in-flight security PRs to avoid conflicts on shared middleware.
- Permission changes need per-route review so intended access is not altered.

## Additional context

- Findings originate from an internal security scan; specifics remain private.
- A draft PR delivers the self-contained subset; the remaining work is tracked here.

[...]

PR #1797, opened 2026-07-08
fix(repo): harden RBAC org-id validation and CI supply-chain

ankit-yc:
## PR Checklist

- [x] There is an issue for the feature this PR is for (#1796).

A Codex security scan (tracked in #1796) and an Aikido surface scan surfaced authorization, supply-chain, and HTTP security-header weaknesses. This PR lands the safe, self-contained subset without touching business logic. Before this change:

- `withOrgPermissions()` extracted the organisation id from query and body without requiring a string, so a structured value (for example `{"organisationId": {"not": ""}}`) could reach the ORM as a filter object.
- The MongoDB connection URI, which may embed credentials, was logged verbatim on startup.
- The root `doctor` script and the Claude and Cursor react-doctor hooks fell back to `react-doctor@latest`, and the Windows signing script installed `TrustedSigning` with only a minimum version.
- The desktop-release signing jobs and the frontend-e2e credential job ran with signing/test secrets in scope without a protected-environment gate.
- The static `/dev-docs` pages shipped no Content-Security-Policy and loaded an unpinned Redoc CDN bundle; responses advertised the framework via `X-Powered-By`.

## What is the new behavior?

Backend and tooling hardening:

- `rbac.ts`: all org-id sources (route params, `x-org-id`, query, body) pass through a single `normalizeOrgId` that accepts only non-empty strings; arrays and objects are rejected.
- `db.ts`: `redactMongoUri` strips `user:password@` credentials before the URI is logged.
- Supply-chain pinning: the `doctor` script runs the lockfile-resolved local binary; the Claude and Cursor hooks drop the `react-doctor@latest` fallbacks; `windows-trusted-sign.js` pins `TrustedSigning` with `-RequiredVersion`.
- Workflows: the desktop-release signing jobs and the frontend-e2e authenticated job are gated behind protected `release` / `e2e` environments (repo settings still need required reviewers and deployment rules configured on those environments).

Frontend HTTP security headers (Aikido findings):

- `poweredByHeader: false` stops advertising `X-Powered-By`.
- A strict, tightly allow-listed CSP is applied to `/dev-docs/*` in `next.config.ts`, and the Redoc bundle is pinned to an immutable version.

Deferred (with reason), not changed in this PR:

- Removing `script-src 'unsafe-inline'` from the public/marketing routes (the Aikido "CSP allows inline JS" finding) is not safe with a nonce here: those pages are statically generated, and Next.js cannot inject a per-request nonce into pre-rendered inline framework scripts, so a nonce-only `script-src` blocks hydration (confirmed by the Playwright public suite). This needs a hash-based CSP or forced dynamic rendering and is left as a follow-up. The existing split (nonce CSP for dynamic app routes, `unsafe-inline` for static public routes) is retained.
- HSTS on the apex domain and the `Server` version header are hosting-layer concerns (AWS Amplify / CloudFront) and are not fixable in app code.

Validation: `pnpm run type-check` and `pnpm run lint` pass across the monorepo; frontend `tsc --noEmit` and the security-header test suite pass; backend `rbac` and `redactMongoUri` tests pass. The Aikido MCP scan could not be run in this environment; recommend an Aikido re-scan before marking ready.

## Related Issue(s)

Related to #1796 (this PR delivers the self-contained subset; the remaining per-endpoint work is tracked there). Not `Fixes`, so the tracking issue stays open.


harshvardhan-yc (approved):
LGTM

PR #1834, opened 2026-07-09
chore(mobile): add gradle dependency lockfiles for android

vallirani-yc:
## PR Checklist

The Android build (`apps/mobileAppYC/android`) has no Gradle dependency lockfiles. Aikido's automated security scanner tries to auto-generate `gradle.lockfile` via PR on each scan, but this fails: its sandbox doesn't run `pnpm install` before invoking Gradle, so `node_modules/@react-native/gradle-plugin` (referenced via `includeBuild` in `settings.gradle`, required by React Native's build system) doesn't exist, and plugin resolution for `com.facebook.react.settings` errors out.

## What is the new behavior?

- Enabled Gradle's native dependency locking in `android/build.gradle` (`allprojects { dependencyLocking { lockAllConfigurations() } }`).
- Generated and committed `android/app/gradle.lockfile` (full resolved dependency graph across debug/release/androidTest classpaths) and `android/settings-gradle.lockfile`.
- With lockfiles already checked into the repo, Aikido can read them directly for its SCA scan instead of trying (and failing) to regenerate them itself.
- No CI impact: the only Gradle-build CI step (in `codeql.yml`) is currently commented out, so this doesn't affect any pipeline.

## Related Issue(s)

Fixes #


ankit-yc:
Merged, with the Aikido gate red. Recording the reasoning, because merging past a red security gate should not be silent.

**The gate is reporting pre-existing dependencies, not anything this PR introduces.** Adding `gradle.lockfile` lets Aikido resolve the Android dependency graph for the first time. Every finding is marked *not active on the default branch* purely because `dev` has no lockfile to read - the dependencies were always in the build, just unscannable. This PR is being penalised for improving visibility.

**Triaged by Gradle configuration, which is what determines whether a dependency actually ships:**

| Finding | Configurations | In the APK? |
|---|---|---|
| protobuf-java / protobuf-kotlin 3.24.4, netty x5 | `_internal-unified-test-platform-*` only | No - Android Gradle Plugin test-harness internals |
| okio 2.9.0 | `debugAndroidTestCompileClasspath` only | No - the app itself resolves okio 3.9.1, already past the 3.4.0 fix |
| **commons-lang3 3.0** | `releaseRuntimeClasspath` + debug + unit test | **Yes** |

Verified mechanically: zero netty or protobuf entries appear on any real app runtime classpath. Aikido reads the lockfile without weighting configurations, so it cannot distinguish a test-harness dependency from a shipped one.

So 8 of the 9 findings describe code that never reaches a device. The one that does is `commons-lang3:3.0` - a 2011 release, pulled in transitively - carrying CVE-2025-48924, an uncontrolled-recursion `StackOverflowError` in `ClassUtils.getClass`. Real, but DoS-class and low practical impact in a mobile client.

Tracked as #2002. It needs a resolution constraint forcing 3.18.0 and a regenerated lockfile, which requires a working JDK and Android SDK - not something I could verify here, so it is deliberately not bundled into this merge.

Nothing was dismissed or suppressed in Aikido. The findings stay visible, which is the point of having the lockfile.

PR #1617, opened 2026-06-21
feat(backend): automate status page incidents with OpenStatus agent webhook

ankit-yc:
## What changed

This draft seeds the branch with the implementation overview for the OpenStatus agent webhook integration, located at `docs/guide/openstatus-agent-webhook-overview.md`. It is a planning document only. No backend source code is changed yet.

The actual backend implementation will be pushed onto this branch by a backend engineer in follow-up commits.

Tracking issue: #1608

## Why

We already use OpenStatus for the public status page, and the frontend Footer already renders incident state from the public status API. The missing piece is automatic incident creation: today the path from "a monitor is failing" to "the status page shows a real incident" is fully manual. This work adds a backend webhook plus an agent service so the status page opens, updates, and resolves incidents on its own.

## Planned implementation (see the overview doc for full detail)

Files to add:

- `apps/backend/src/controllers/web/openstatus.controller.ts` (verify signature, parse payload, delegate)
- `apps/backend/src/services/openstatus.service.ts` (run the agent loop against `api.openstatus.dev/mcp`)
- Tests for the controller and the service

Files to edit:

- `apps/backend/src/app.ts` (register `POST /v1/openstatus/webhook` next to the Stripe and Documenso webhooks)
- `apps/backend/.env.example` (add key names: `OPENSTATUS_API_KEY`, `OPENSTATUS_WEBHOOK_SECRET`, monitor id mapping, optional model key)
- `apps/backend/package.json` (add the MCP client SDK)

Out of repo: point an OpenStatus monitor at `/health`, attach a webhook channel to the new endpoint, and generate a write-scoped API key.

## Impact area

Backend only. No frontend changes are required for the core feature. The existing Footer reflects incidents automatically once the agent files them.

## Validation performed

None yet. This draft contains documentation only. Validation (backend lint, type check, and targeted tests) will be performed as the backend engineer adds the implementation commits.

Issue #1608, opened 2026-06-21
feat(backend): automate status page incidents with OpenStatus agent webhook


ankit-yc:
**Is your feature request related to a problem? Please describe.**

We use OpenStatus to power our public status page (`https://yosemite-crew.openstatus.dev/`). The frontend Footer already reads the public status API and renders an incident state (`apps/frontend/src/app/ui/widgets/Footer/Footer.tsx`, lines 11 to 41 and 179 to 187), so the "Active incident" dot is fully wired on the consumer side. The gap is on the producer side: nothing in our stack opens, updates, or resolves incidents automatically. Today the path from "backend `/health` is failing" to "status page shows a real incident" is fully manual and depends on a human logging into the OpenStatus dashboard.

OpenStatus recently published their agent webhook approach (https://www.openstatus.dev/blog/agent-webhook, see also https://www.openstatus.dev/use-case/agent). When a monitor fails, OpenStatus can POST a webhook to a runtime we control. That runtime runs an agent loop against the OpenStatus MCP server (`api.openstatus.dev/mcp`), which exposes the same verbs the dashboard uses: open an incident report, post updates, and mark it resolved. This lets the status page operate itself while the team focuses on the actual fix.

**Describe the solution you'd like**

Add a backend webhook endpoint plus an agent service that reacts to OpenStatus monitor failures and recoveries. The work mirrors our existing Stripe and Documenso webhook integrations, which already use raw body parsing and HMAC signature verification.

Backend changes:

1. `apps/backend/src/app.ts` (near lines 32 to 42): register `app.post("/v1/openstatus/webhook", express.raw({ type: "application/json" }), ...)` alongside the existing Stripe and Documenso webhook routes. Raw body is required for signature verification.
2. `apps/backend/src/controllers/web/openstatus.controller.ts` (new): verify the webhook signature using the same `crypto.createHmac` and `timingSafeEqual` pattern as `documenso.controller.ts`, parse the payload (which monitor, regions, timestamp), then delegate to the service.
3. `apps/backend/src/services/openstatus.service.ts` (new): connect to `api.openstatus.dev/mcp` with a write scoped API key and run the tool loop: open an incident report on monitor failure, post status updates, and resolve when the monitor recovers.
[...]

PR #1842, opened 2026-07-10
feat(backend): automate status page incidents with OpenStatus agent webhook

ankit-yc:
## What changed

Adds an OpenStatus agent webhook to the backend so the public status page opens and resolves incidents automatically. Deterministic implementation (no LLM), backend only.

- `apps/backend/src/controllers/web/openstatus.controller.ts` (new): authenticates the webhook with a constant-time shared-secret header check, parses the raw body, validates with zod, delegates to the service.
- `apps/backend/src/services/openstatus.service.ts` (new): calls the OpenStatus v1 REST API to open a status report on `error`/`degraded` (idempotent, skips if one is already open) and resolve open reports on `recovered`. Stateless (OpenStatus is the source of truth); no new dependencies.
- `apps/backend/src/app.ts`: registers `POST /v1/openstatus/webhook` with `express.raw`, next to the Stripe and Documenso webhooks.
- `apps/backend/.env.example`: adds `OPENSTATUS_API_KEY`, `OPENSTATUS_WEBHOOK_SECRET`, `OPENSTATUS_PAGE_ID`, `OPENSTATUS_API_BASE_URL` (names only).
- Unit tests for the controller and service.
- `docs/guide/openstatus-agent-webhook-overview.md`: implementation overview.

Tracking issue: #1608. Supersedes #1617 (same work, fix-prefixed branch).

## Why

We already use OpenStatus for the public status page, and the frontend Footer already renders incident state from the public status API. The missing piece was automatic incident creation: previously the path from "a monitor is failing" to "the status page shows a real incident" was fully manual. This closes that gap while the team focuses on the actual fix.

## Design decision: deterministic, not an LLM agent

Calls the OpenStatus v1 REST API directly rather than running an LLM agent against the MCP server. This fully delivers auto open and auto resolve, removes the prompt-injection risk of feeding webhook content to a model that can publish to a public page, adds no new dependencies (axios and zod are already used), and needs no model key. An LLM-authored incident-copy variant can be layered on later.

## Impact area

Backend only. No frontend changes are required; the existing Footer reflects incidents automatically once a report is opened.

## Validation performed

- Targeted backend tests: 19/19 pass, 100% statement/branch/function/line coverage on both new files.
- ESLint: 0 errors on the new source files.
- Type-check: 0 errors from the new files.
- Branch is current with `dev`; React Doctor correctly skips the PR (no React files changed).

Runtime configuration (OpenStatus monitor on `/health`, webhook channel with the `x-openstatus-webhook-secret` header, write-scoped API key, and page id) is done in the OpenStatus dashboard and the backend secret store; see the overview doc.


aupyay (changes requested):
Reviewed at `b5ef826326ffb0ba441a25a51c618d7fb37f9eaa`.

**Triage first, because the PR's age is misleading.** Open since 10 July, but the merge-base is **2026-09-03** and dev is only **111** commits ahead — the branch is current, `MERGEABLE`, and OpenStatus appears nowhere on `dev` (`git grep -il openstatus` over `apps/backend/src` returns nothing). So this is a live gap and a real review, not a conversation about whether it still applies.

The shape is right: the webhook is authenticated before the body is parsed, the payload is parsed with zod rather than cast, the raw-body registration sits with the other webhooks, and the service keeps no local incident state. The secret comparison hashes both sides to a fixed length so `timingSafeEqual` cannot throw on a length mismatch — that is the correct idiom for a variable-length shared secret, and the comment says why. Nine controller tests and ten service tests, covering the three 401 paths, both 400 paths, the array-valued header, and both throw-on-missing-config cases.

## Requesting changes on one thing: the OpenStatus API key reaches the logs

`openstatus.controller.ts` ends with

```ts
logger.error("OpenStatus webhook error", err);
```

and every failure from `openstatus.service.ts` arrives there as an **axios error**, which carries the outbound request config — including `headers['x-openstatus-key']` from `authHeaders()`.

Measured against the logger's **production** format (`NODE_ENV !== "development"`, so `combine(errors({stack:true}), timestamp(), json())`), driving a real winston logger with a captured transport:

```
{"config":{"headers":{"Content-Type":"application/json",
 "x-openstatus-key":"THE_SECRET_API_KEY_VALUE"}, "method":"post",
 "url":"https://api.openstatus.dev/v1/status_report"}, ...}

secret in the emitted log line?  YES
```

The development format leaks it too: `serializeMeta` in `utils/logger.ts` replaces `req`, `res`, `request` and `response`, but **not `config`**, which is where axios puts the headers.

It is reached by any OpenStatus API failure — a 500, a timeout, an expired key — so the first time the integration has trouble is the time the key is written out.

**The fix is at this call site, not in the logger.** Log the fields rather than the object:

```ts
logger.error("OpenStatus webhook error", {
  message: err instanceof Error ? err.message : String(err),
  status: axios.isAxiosError(err) ? err.response?.status : undefined,
});
```

**Scope, because I checked before generalising:** I am not claiming `dev` already does this. `documenso.service.ts` is the only other service that logs a caught error, and four of its five calls log `error.message` / `error.statusCode` / `error.body` — fields, not the object — through an SDK client rather than raw axios. This is the clear instance.

## Three notes, none blocking

**1. A remote-supplied string is published verbatim to the public status page.** `buildIncidentMessage` appends `event.errorMessage` — schema `z.string()`, no maximum — into the incident body. It arrives inside an authenticated webhook, so it is not attacker-chosen, but it originates from whatever the monitored endpoint returned. A `.max(...)` on that field bounds what a misbehaving upstream can put on a public page.

**2. Idempotency is check-then-act, so it is safe against replay and not against concurrency.** `openIncident` lists, filters, then creates. Two webhooks for the same monitor arriving together both see no open report and both create one. The comment says "idempotent against repeated failure webhooks and replays", which is exactly right and slightly narrower than it reads.

**3. `listStatusReports` is unfiltered and unpaginated.** It takes whatever `/status_report` returns and filters client-side by `monitorIds`; `getPageId()` is used when creating a report but not when listing. On an account with one page that is the same answer. With more than one page, or with enough history that the endpoint paginates, an open report could be missed — and a missed open report means a duplicate incident rather than a suppressed one, which is the safe direction.

## What I did not verify

No live OpenStatus call. The service tests mock axios, so they establish the request shapes and the branch logic, not that the v1 API accepts them — and `status_report/{id}/update` in particular is asserted against a mock rather than against the API.

aupyay:
Resolved the blocking log-secret finding in 011f1239: the controller now routes caught errors through the shared safe-error serializer, and the regression test proves an Axios-shaped error cannot emit its OpenStatus key or query string. Verified with the complete backend suite (464 suites / 11,146 tests), mutation check, 100% focused controller coverage, lint, type-check, build, formatting, and secret scans.

ankit-yc:
Parking this as a draft under the Yosemite Crew feature freeze, which pauses new features until the UI and design system work is finished. Nothing here is wrong; it is scope, not quality.

Two things to settle before it comes back:

1. Three checks are failing, including CI Required and the shared-package build. The branch has been open since 10 July and is well behind, so it needs a rebuild against current dev rather than a re-run.
2. The external-product gate is failing, and it is failing on the title itself. The title names a third-party product; that has to be rewritten in neutral terms before this can merge, independent of the freeze.

Unpark by marking ready once the freeze lifts and both are addressed.

aupyay:
Heads-up: this PR is now blocked by a required check that can never report on it, independently of its failing CI.

`dev`'s branch ruleset gained two required status checks today (#3395): `Desktop E2E Required` and `Frontend E2E Required`. This PR has the first (success) but not the second — it has never been created here.

The cause is the head, not the branch content. At `011f1239b` (2026-09-06), `.github/workflows/frontend-e2e.yml` still filters at the workflow level:

```yaml
on:
  pull_request:
    paths:
      - 'apps/frontend/**'
      - 'pnpm-lock.yaml'
      - 'package.json'
      - 'pnpm-workspace.yaml'
      - '.github/workflows/frontend-e2e.yml'
```

This is a backend PR, so the workflow never triggers, the context is never created, and a required context that is never created is waited on forever. `6b1eed77d` (2026-09-08, *"ci(frontend): trigger on every PR so the required check cannot deadlock"*) replaced that with `on: pull_request: {}` and moved the path decision inside the jobs, so the aggregate always reports and reports success when frontend testing was unnecessary. That commit is two days after this head.

`strict_required_status_checks_policy` is `false` on `dev`, so GitHub will not force this branch up to date and the state will not resolve on its own.

**Fix:** merge current `dev` into this branch. That brings the rewritten workflow onto the head, `Frontend E2E Required` gets created, and it should go green via the "no frontend changes in this pull request, nothing to run" path.

`CI Required` is failing here separately and still needs its own fix — this note only covers the missing context. Not pushing to the branch since it is not mine.

Verified against the three other open `dev` PRs (#3383, #3371, #3342): all four required contexts are present on each, so this one is the only PR affected.

PR #1843, opened 2026-07-11
chore(repo): sync main with dev for desktop/frontend/backend releases

ghost:
## What changed
Syncs `main` with `dev` (24 commits) ahead of tagging the next releases for desktop, frontend, and backend.

Included work:
- **Desktop hardening & maintainability (PR #1817):** path-traversal guard on document-vault downloads, `buildMainWindowOptions` de-duplication, `vaultCompletedDownload` extraction, deterministic E2E waits, `overflow-wrap` CSS fix.
- **Security hardening (Aikido):** dependency updates (sitemap, json), GitHub Actions permission scoping, Docker non-root user, Android exported-component fixes, Express security headers via helmet, plus SAST autofixes from PRs #1803, #1823, #1824, #1825, #1830, #1838.

## Why
`dev` is 24 commits ahead of `main`; releases are tagged from `main`, so it must be brought current before creating `desktop-v0.1.0-beta.2` and the next frontend/backend releases.

## Impact area
Desktop shell, backend security middleware, CI workflows, Docker images, Android manifest, dependency lockfile.

## Validation performed
All CI checks green on `dev` (CI affected, SonarCloud analyses, CodeQL, Secret Scan as of 2026-07-10).

PR #1845, opened 2026-07-11
fix(frontend): reduce Sonar duplicate board and calendar code

ghost:
## Summary
- Extract shared board column drag/drop listeners into useBoardColumnDrop
- Reuse the shared hook in appointment and task boards to reduce duplicated frontend Sonar blocks
- Hoist repeated appointment calendar drag handlers into named callbacks
- Extend hook and board tests for the shared drag/drop behavior

## Validation
- npx tsc --noemit
- pnpm --filter frontend run lint
- pnpm --filter frontend run test -- --testPathPattern="useBoardDragScroll|AppointmentCalendar|AppointmentBoard|TaskBoard"
- pre-push: pnpm lint
- pre-push: pnpm type-check


ghost:
## CI status & remaining Aikido LOWs — root causes and decision

This PR was raised to reduce SonarCloud **duplication on new code** (the gate blocking the `dev → main` sync in #1843). Along the way two CI checks were failing and looping — **React Doctor** and **Aikido**. Both are now root-caused. This comment documents what was wrong, what was fixed, and why the remaining 6 Aikido LOW advisories should be resolved by policy rather than more code churn.

### TL;DR

| Check | Status | Notes |
|---|---|---|
| **SonarCloud — frontend (duplication)** | ✅ pass | The actual #1843 blocker. Now under the 3% New-Code duplication threshold. |
| **React Doctor** (frontend + mobile) | ✅ pass | Score 100/100. Was failing due to a tool/version mismatch, **not our code** (details below). |
| **Aikido — CRITICAL** | ✅ fixed | Real bug: an always-true `\|\| true` that disabled task availability enforcement. Fixed + unit-tested. |
| **Aikido — 6 × LOW** | ⚠️ non-blocking style advisories | "Keep functions concise" / "Avoid deep nesting". **Zero security issues.** Does not gate #1843 (Aikido `skips` the sync PR). See decision below. |
| Builds, type-check, unit tests (225 Calendar tests), Playwright, CodeQL, jest-axe | ✅ pass | |

---

### 1. React Doctor — root cause: CLI/action report-schema mismatch (fixed)

React Doctor was failing on every commit as **"Scan could not complete"** with an empty score — including the `react-doctor-mobile` job, which had **zero changed files**. That last fact proves it was not caused by anything in this PR.

**Cause:** the workflow pins the action (`millionco/react-doctor@ced746f`, v2.2.6) but ran the CLI at `version: latest`. `react-doctor@0.7.5+` began emitting report **`schemaVersion: 3`**, while the pinned action's `ensure-json-report.mjs` only accepts schema **1–2**. It therefore discarded every valid report, overwrote it with an `ok:false` fallback, and exited 1.

**Fix:** pin the CLI to **`0.7.4`** (emits schema 1) on both jobs in `.github/workflows/react-doctor.yml`. Reproducible locally:
```bash
npx react-doctor@0.7.4 apps/frontend --json --json-compact --scope changed --changed-files-from <files>
# schemaVersion:1, ok:true, exit 0
```
Revisit `version: latest` only once the action is bumped to a release whose `ensure-json-report.mjs` knows schema 3.

### 2. Aikido CRITICAL — always-true availability bypass (fixed)

`shouldAllowTaskAvailabilityBypass` ended in `... || true`, an always-true expression that silently disabled task-availability enforcement. Recovered the intended logic from history — *enforce availability for everyone except the user who created the assignment* — simplified to `return !isAssignedByCurrentUser`, and added unit tests covering both branches.

### 3. Aikido — 6 remaining LOW advisories: why they don't converge, and the decision

The remaining findings are all `Low` **code-quality** advisories (Aikido: *"Security Issues Introduced — No newly introduced issues"*):

- `useAppointmentMove.ts`, `TaskCalendar.tsx`, `useTaskCalendarIdentity.ts`, `TaskCalendarBody.tsx`, `useTaskAvailability.ts` — **Keep functions concise**
- `taskCalendarAvailabilityUtils.ts` — **Avoid deep nesting**

These were pursued in good faith across several commits with genuine, behaviour-preserving refactors (pure helpers extracted, hooks thinned, a composition hook added, nesting flattened — all covered by the 225-test Calendar suite). The count moved **8 → 7 → 6** and then floored at **6**. It does not reach zero because:

1. **Aikido diffs the branch head against `dev`.** Every one of these files was created/heavily rewritten on this branch, so each is "New" and its functions are re-evaluated on every scan.
2. **The rule counts a React hook body (with its `useCallback` dependency arrays) as one "function."** An idiomatic hook composing 3–4 memoised callbacks is structurally 40–50 lines. Extracting logic to shorten one function creates a new helper the rule then counts — roughly one-out-one-in per commit.

Driving these to zero would require fragmenting well-formed hooks/components into artificial pieces, which **adds indirection — i.e. introduces tech debt** to satisfy a line-count heuristic. That is the wrong trade.

**Decision (recommended): raise the Aikido PR gate threshold, don't churn the code.** Per Aikido's docs (*Integrations → PR Quality Gating → Severity Threshold*), set the threshold to **High**. New issues at/above High (and all **security** findings) continue to hard-block; **LOW/MEDIUM code-quality style advisories are reported but no longer fail the merge.** This is a durable policy — not the per-PR "Bypass" button — and prevents the same false-loop on every future PR. It's also moot for the #1843 sync PR, where Aikido is `skipping`.

---

**Validation performed:** `tsc --noemit` (clean), `eslint` (clean), targeted Jest (225 Calendar tests pass), React Doctor 100/100 locally and in CI, SonarCloud frontend gate green on this branch.

harshvardhan-yc (approved):
LGTM

PR #1981, opened 2026-07-22
refactor(frontend): shared StatusPill primitive and app-wide status badge convergence

ankit-yc:
## PR Checklist

On the Integrations page the two card status badges had drifted into two separate implementations. The "ENABLED" badge renders through `StatusPill` (10px bold uppercase, +0.06em tracking, 10px/2px padding, full radius); the "Coming soon" badge used a separate one-off class (`text-label-xsmall`, 8px/4px padding, 16px radius, sentence case), so it rendered noticeably larger and in a different case right next to it.

## What is the new behavior?

`coming-soon` becomes a status like any other: its neutral colours move into the `statusTokens` map, and all four coming-soon cards render through `StatusPill`. The two badges are now the same component and cannot drift apart again. Coming-soon correctly gets no live dot (that is gated on `enabled`) and keeps its neutral colour; only the geometry and typography are unified.

## Impact area

`apps/frontend` Integrations page, presentation only. The rendered label text is unchanged ("Coming soon"); the uppercase is CSS.

## Validation performed

- `npx tsc --noemit` (apps/frontend) - clean.
- `npx eslint` on the changed file - clean.
- `pnpm --filter frontend run test --testPathPattern="Integrations"` - 11 suites, 276 tests passing.

Not verified in a browser yet - needs a visual pass on the deployed branch.

## Related Issue(s)

Follow-up to #1977 (side-modal redesign), which touched this file; this pill drift was noticed during that review after the PR had merged.

### DIFF
--- a/apps/frontend/src/app/__tests__/components/chat/SharedEntityCard.test.tsx
+++ b/apps/frontend/src/app/__tests__/components/chat/SharedEntityCard.test.tsx
@@ -85,7 +85,7 @@ describe('SharedEntityCard', () => {
       />
     );
     const pill = screen.getByText('PAID');
-    expect(pill).toHaveStyle({ background: 'var(--status-completed-bg)' });
+    expect(pill).toHaveStyle({ backgroundColor: 'var(--status-completed-bg)' });
   });
 
   it('falls back to the requested tone for an unmapped status', () => {
@@ -94,7 +94,9 @@ describe('SharedEntityCard', () => {
         entity={makeEntity({ entityType: 'INVOICE', snapshot: { status: 'MYSTERY' } })}
       />
     );
-    expect(screen.getByText('MYSTERY')).toHaveStyle({ background: 'var(--status-requested-bg)' });
+    expect(screen.getByText('MYSTERY')).toHaveStyle({
+      backgroundColor: 'var(--status-requested-bg)',
+    });
   });
 
   it('omits the status pill when snapshot.status is not a string', () => {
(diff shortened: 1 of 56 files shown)


ankit-yc:
Thanks - all three are addressed. Two were already fixed in e550cd54, which landed after these comments were written.

**Badge danger -> warning** (Aikido + Codex) — fixed in e550cd54. The mapping came from a wrong premise of mine: the `--color-pill-*` scale has no danger set, but the app does have `--danger-bg/-text/-border`, and `Badge`'s danger previously rendered `bg-danger-100 text-danger-700`. `StatusPill` now has a real `danger` tone drawn from that scale and `Badge` maps `danger -> danger`.

**Failed lab orders losing error styling** (Codex) — same commit, and you were right that it was a separate loss rather than the same one: `getOrderStatusTone` now splits the branches, `bg-red-50 -> danger` and `bg-amber-50 -> warning`, so a failed or cancelled order no longer reads as pending. Two tests asserted the folded mapping, i.e. they were pinning the bug, so they now assert the colours these states are meant to have.

**Pills stretching in flex columns** (Codex) — fixed in 7d97d7b7. Confirmed against the old CSS: `.appointment-status` set `width: fit-content` and `align-self: flex-start`, and the primitive had neither, so on the card views the badge would have been pulled into a full-width band. The primitive is now `w-fit`. I deliberately did not add `self-start`: an explicit cross size is enough to defeat `align-items: stretch`, whereas `self-start` would top-align the badge in the many flex rows that currently centre it. Covered by a test.

Full suite green (859 suites / 10,532 tests).

aupyay (approved):

PR #1982, opened 2026-07-22
fix(frontend): tidy add-product modal internals to the warm-bone aesthetic

ankit-yc:

Two things the side-modal redesign (#1977) did not reach, both live on dev and visible in the Add product modal:

1. The step section prints its title three times - the active tab, a 23px heading, and the accordion header directly beneath it - so "Basic Details" stacks on itself (confirmed on the deployed modal: three nodes reading "Basic Details" at 16/23/20px).
2. The section tab strip (the shared `Labels` component, used by five modals) keeps a blue-outline active state - `bg-blue-light`, brand-blue border, `primary-700` text - which is off-brand for the warm-bone palette and stands out against every other control on the page.

## What is the new behavior?

- The redundant 23px heading in `AddInventory/FormSection.tsx` is removed. The tab names the step; the accordion header carries the title in the body. Each section has exactly one accordion, so nothing is lost.
- `Labels` active tabs now match the page's own Active/Hidden pills, measured 1:1 off the live deployed control: fully rounded, warm neutral fill (`--color-pill-neutral-bg`), ink text, hairline border, bold when active and ink-soft when not. The valid/error status dots move onto the `--success` / `--danger` tokens.

This restyles the tab strip in all five modals that use `Labels` (Add product, inventory detail, companion detail, add companion, appointment detail) consistently.

## Impact area

`apps/frontend` - `AddInventory/FormSection.tsx` and the shared `ui/widgets/Labels`. Presentation only.

## Validation performed

- `npx tsc --noemit` - clean.
- `npx eslint` on both files - clean.
- `pnpm --filter frontend run test` across `Labels` and all five consumer modals - 51 suites, 693 tests passing.
- The problem was confirmed on the deployed dev site (logged-in) by measuring the live DOM; the target styles were measured 1:1 off the live Active/Hidden pills. The fix itself needs a visual pass once this branch deploys.

## Related Issue(s)

Follow-up to #1977. No separate issue filed.

### DIFF
--- a/apps/frontend/src/app/__tests__/components/companionHistory/HistoryDocumentUpload.test.tsx
+++ b/apps/frontend/src/app/__tests__/components/companionHistory/HistoryDocumentUpload.test.tsx
@@ -9,9 +9,19 @@ jest.mock('@/app/ui/layout/guards/PermissionGate', () => ({
   PermissionGate: ({ children }: any) => <>{children}</>,
 }));
 
-jest.mock('@/app/ui/primitives/Accordion/Accordion', () => ({
+jest.mock('@/app/ui/overlays/Modal', () => ({
   __esModule: true,
-  default: ({ children }: any) => <div>{children}</div>,
+  default: ({ showModal, children }: any) => (showModal ? <div>{children}</div> : null),
+}));
+
+jest.mock('@/app/ui/overlays/Modal/ModalHeader', () => ({
+  __esModule: true,
+  default: ({ title, onClose }: any) => (
+    <div>
+      <div>{title}</div>
+      <button type="button" aria-label="close" onClick={onClose} />
+    </div>
+  ),
 }));
 
 jest.mock('@/app/ui/inputs/Dropdown/LabelDropdown', () => ({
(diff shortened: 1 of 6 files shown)


ankit-yc:
Confirmed and fixed in 99cf17f5 - thanks, this was a real regression introduced by this PR.

Verified the chain: `CompanionInfo` renders a `Modal`, whose History tab renders `CompanionHistoryTimeline` -> `HistoryDocumentUpload`. `ModalBase` does `createPortal(..., document.body)`, so the upload panel lands as a **sibling** of the drawer rather than a descendant. The drawer's `mousedown` handler then saw the first click inside the upload panel as an outside click, closed, and unmounted the upload UI with it. Before this PR the upload form was an inline accordion inside the drawer's own DOM, so it could not happen.

Fixed at the root rather than at this call site, since it applies to every stacked modal in the app (the delete confirmations nested inside the team and room drawers had the same exposure): `ModalBase`'s outside-click handler now ignores clicks that land inside any `.yc-modal-dialog`. Interacting with one dialog is never a dismissal of a different one. The backdrop sits outside `.yc-modal-dialog`, so click-to-dismiss is unchanged.

Covered by a regression test in `ModalBase.test.tsx` - I checked it fails without the guard and passes with it. Full suite green (858 suites / 10,522 tests).

The `prefer-useReducer` warning from React Doctor on the same file I've left: the five state values there are independent (open, file, form data, saving, errors) rather than one transition, and score is 98/100.

aupyay (approved):

PR #1986, opened 2026-07-22
fix(frontend): drop the duplicate New task button on phone

ankit-yc:

On phone the tasks page renders **two** create affordances at once: the "New task" button in the page header, and the phone shell's create FAB. Both open the same add-task flow, stacked on a single small screen.

## What is the new behavior?

The header "New task" CTA is hidden on phone, leaving the FAB as the single create affordance. This is the design's "primary action -> FAB" rule, and it matches how the inventory page already gates its "Add product" button with `!isPhone`.

The FAB already reaches the same `openAddTask` flow via `usePhonePrimaryAction('task', openAddTask)`, so nothing is lost - only the duplicate is removed. Tablet and desktop are unchanged.

## Impact area

`apps/frontend` tasks page, phone breakpoint only.

## Validation performed

- `npx tsc --noemit` - clean.
- `pnpm --filter frontend run test --testPathPattern="pages/Tasks/index"` - 37 passed, including a new case asserting the header CTA is absent on phone while the desktop cases keep it.

Not visually verified - local login is blocked by the SuperTokens cross-origin setup.

### DIFF
--- a/apps/frontend/src/app/__tests__/pages/Tasks/index.test.tsx
+++ b/apps/frontend/src/app/__tests__/pages/Tasks/index.test.tsx
@@ -368,6 +368,14 @@ describe('Tasks page', () => {
     expect(addTaskSpy).toHaveBeenCalledWith(expect.objectContaining({ showModal: true }));
   });
 
+  it('hides the header New task CTA on phone, leaving creation to the shell FAB', () => {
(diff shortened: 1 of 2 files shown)


ankit-yc:
Folding this into the combined mobile PR (phone sign-out + owner-lockout guard + tasks FAB-only) per request. Superseded.

PR #1990, opened 2026-07-24
fix(repo): fix stale-data and alignment bugs across profile, rooms, and dispensary

vallirani-yc:
## PR Checklist

Several PIMS bug reports were surfacing stale or misaligned data:

- Editing a team member's **address** (or the **country** field, which is stored on the same address record) saved correctly, but reopening the profile after a refresh showed the old value. The backend's `update()` synced the relational address row from a stale, pre-update read instead of the freshly-submitted data, so the sync effectively wrote the old value back over itself.
- The **appointment overview modal**'s Room/Unit fields rendered two stacked labels ("Room" above "Select room", "Unit" above "Select unit") because both the wrapper and the shared `LabelDropdown` component were rendering their own label.
- The **Merck manual reader** overlay's "Copy manual URL" button copied the URL correctly, but the "Copied URL to clipboard." confirmation rendered on the underlying search page, hidden behind the fullscreen reader overlay — so clicking it looked like nothing happened.
- Changing a room's **type** away from a unit-capable type (e.g. to Surgery) cleared the unit config locally and on save, but the previously-created `RoomUnitGroup`/`RoomUnit` records were never deleted server-side, so the stale unit type config (e.g. "Pod / Extra large / 5") resurfaced in view and edit mode after saving.
- The **Dispensary** table's status pills weren't aligned with their column header — the row grid used a different `gap` than the shared table header, causing a cumulative left-offset across columns.

## What is the new behavior?

- `UserProfileService.update()` now syncs the relational address (and therefore country) from the freshly-submitted `personalDetails`, matching the pattern already used in `create()`. Edits now persist correctly across a refresh.
- `RoomSelectorSection` no longer renders a redundant outer label for the editable case; it relies on `LabelDropdown`'s built-in label (using the plain field name, consistent with the other detail rows), and the read-only fallback uses a single matching label instead of the old floating-over-border style.
- The Merck reader header button now shows "Copied!" in place of "Copy manual URL" for ~1.5s after a successful copy, giving in-context feedback without needing to close the reader.
- `syncRoomUnitGroups` now fetches a room's existing unit groups on update and deletes any group (and its units) no longer represented in the desired list — covering both a type change away from unit-capable and a cleared/reduced unit list. New rooms skip this extra round trip since they can't have stale groups yet.
- `DispensaryRow`'s grid gap now matches the shared table header (and `InventoryTable`'s convention), so all status pills — regardless of label width — start at the same left edge under the "Status" column.

## Related Issue(s)

Fixes #

### DIFF
--- a/apps/backend/src/services/room-unit-group.service.ts
+++ b/apps/backend/src/services/room-unit-group.service.ts
@@ -123,7 +123,10 @@ const toDomain = (row: RoomUnitGroupRow): RoomUnitGroup => ({
 const getRoomUnitGroupDelegate = (): RoomUnitGroupDelegate =>
   (prisma as unknown as { roomUnitGroup: RoomUnitGroupDelegate }).roomUnitGroup;
 
-const assertRoomExists = async (roomId: string, organisationId: string) => {
+const assertRoomBelongsToOrganisation = async (
+  roomId: string,
+  organisationId: string,
+): Promise<RoomRow> => {
   const room = (await prisma.organisationRoom.findUnique({
     where: { id: roomId },
     select: { id: true, organisationId: true, type: true },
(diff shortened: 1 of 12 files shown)


harshvardhan-yc (approved):
LGTM

PR #2003, opened 2026-07-25
fix(backend): route outbound document fetches through the shared URL resolver

### LINKED ISSUE #2001: fix(backend): route outbound document fetches through the shared hardened URL resolver
## Summary

Consolidate outbound document fetching in `apps/backend` onto the shared hardened URL resolver that already exists in `packages/lib`, instead of the ad-hoc URL checks currently inlined in two services.

## Problem / motivation

`packages/lib/src/pdf/resolveLogoSource.ts` is the repo's reference implementation for safely fetching a remote asset from a URL we do not fully control. It resolves the host, classifies the resolved addresses, pins the lookup for the actual request, and bounds the request in time and size.

Two backend services fetch remote documents through their own inline URL checks instead, which are narrower than that shared implementation and do not bound the request the same way. Having two different standards for the same operation means the weaker one silently sets the real security posture, and any future improvement to the shared resolver does not reach these call sites.

Related: one of the two services derives a storage object key from a URL before that URL has been through any validation, so the ordering there should be corrected as part of the same change.

## Proposed solution

1. Export the reusable pieces of `resolveLogoSource.ts` (the address classification and the pinned-lookup agent are already exported for this purpose).
2. Route `apps/backend/src/services/rendered-document.service.ts` and `apps/backend/src/services/workspace-document-packet.service.ts` through them.
3. Validate and normalise the URL **before** anything is derived from it, including the storage object key.
4. Bound the outbound requests consistently with the shared resolver: redirect policy, timeout, and response size cap.
5. Make the rejection path surface a typed service error rather than a bare `Error`, so a rejected URL is a client-visible 4xx rather than a 500 in one service and a silent `null` in the other.
6. Cover the behaviour with tests in both services.

## Scope

- `apps/backend/src/services/rendered-document.service.ts`
- `apps/backend/src/services/workspace-document-packet.service.ts`
- `packages/lib/src/pdf/resolveLogoSource.ts` (exports only, no behaviour change)
- New tests under `apps/backend/test/services/`

### Out of scope

- Changing `resolveLogoSource`'s own behaviour, which is already the reference implementation
- The logo rendering path, which already uses it

These URLs originate from the IDEXX lab integration response and are persisted, then read back later when a document is served or a packet is assembled. They are third-party input rather than end-user input, which lowers the practical exposure but does not remove the need to treat them as untrusted.

## Acceptance criteria

- [ ] Both services fetch through the shared resolver rather than a local implementation
- [ ] No value is derived from a URL before that URL has been validated
- [ ] Outbound requests are bounded by redirect policy, timeout and size cap
- [ ] Rejections surface as a typed, client-meaningful error
- [ ] Tests cover accepted and rejected inputs for both services
- [ ] Backend coverage gate stays green

## Test plan

Unit tests for both services covering permitted URLs, rejected URLs, and the ordering guarantee that nothing is derived from an unvalidated URL. Full backend suite plus type-check and lint.

## Alternatives considered

- **Improve the two inline implementations in place.** Rejected: it keeps three separate copies of the same security-relevant logic in the repo.

## Dependencies & risks

- The shared resolver rejects hosts that resolve to non-public addresses. Any deployment that legitimately serves documents from an internal endpoint (for example a self-hosted object store on a private address) would need an explicit configuration allowlist. That decision should be made as part of this work rather than assumed.
- Supersedes #1827, which attempted a narrower version of this and is closed.

ankit-yc:
## PR Checklist

Two backend services fetch remote documents using their own inline URL checks, written separately from `packages/lib/src/pdf/resolveLogoSource.ts`, which is the repo's reference implementation for fetching an asset from a URL we do not fully control.

The two implementations do not agree. The shared resolver resolves the host, classifies the resolved addresses, pins the lookup used for the actual request, and bounds the request in time and size. The inline versions check the scheme and dot-segments only, and bound nothing.

`rendered-document.service.ts` also derived an object-store key from the URL *before* that URL had been through any check, so the ordering there did not match the intent either.

These links come from the IDEXX integration response and are persisted, then read back later when a document is served or a packet is assembled. They are third-party input rather than end-user input, but they should still be treated as untrusted.

## What is the new behavior?

Both services now go through one place, `apps/backend/src/utils/outbound-document-url.ts`, which delegates address classification, hostname resolution and connection pinning to `@yosemite-crew/lib`. No second implementation, and improvements to the shared resolver now reach these call sites.

- **Ordering fixed.** `downloadPdfBuffer` resolves and validates the URL first; the bucket read moved into `readPdfFromBucket(validatedUrl)` and derives its key from the normalised URL. A rejected URL never reaches `s3.getObject`, which has its own test.
- **Requests are bounded** - no redirects, 15s timeout, 25 MB cap. The byte cap is larger than the branding fetch because these are documents.
- **Rejections are typed.** `RenderedDocumentServiceError(..., 400)` and `WorkspaceServiceError(..., 400)` replace a bare `Error`, so a rejected link is a meaningful 4xx rather than a 500 in one service and an indistinguishable `null` in the other. The packet download keeps its existing `null` fallback to a live re-merge for network failure, expiry, and missing links.
- `packages/lib` exports `resolvePublicAddresses` and `buildPinnedAgent` alongside `isPrivateAddress`, and `packages/lib/src/pdf/index.ts` re-exports them so they are reachable from the package entry point. No behaviour change in the resolver itself.

### Deployments serving documents from an internal endpoint

The shared resolver only permits hosts resolving to a public address, which would break an install using a self-hosted object store on a private address. `DOCUMENT_FETCH_ALLOWED_INTERNAL_HOSTS` takes a comma-separated hostname list, following the same pattern as the existing `PDF_LOGO_ALLOWED_HOSTS`. It is **empty by default**, so an install that opts nothing in reaches only public hosts. A listed host skips resolution and pinning but keeps the redirect, timeout and size bounds. Documented in `apps/backend/.env.example`.

### Validation

```
TZ=UTC pnpm --filter backend run test
  Test Suites: 243 passed, 243 total
  Tests:       4717 passed, 4717 total

backend type-check   exit 0
lib type-check       exit 0
backend lint         exit 0
```

Coverage of the changed code:

```
outbound-document-url.ts             100%    stmts / branch / funcs / lines
workspace-document-packet.service.ts  99.77% (only gap is a pre-existing throw, untouched)
rendered-document.service.ts          86.15% - no uncovered line falls inside the changed region
```

Cases were added for the `Uint8Array`, `string`, empty-body and empty-key branches specifically because those lines moved and will be scored as new code.

Tests stub `dns.promises.lookup`, so no test performs real DNS resolution; documentation-range addresses are used throughout.

## Related Issue(s)

Fixes #2001

Supersedes #1827, which was closed. Half of that PR had already been overtaken by the shared resolver on `dev`, and the half that remained was narrower than the standard the resolver already sets.


ankit-yc:
Addressing the two bot findings on the `axios.get` call sites, since they need context rather than a code change.

**Net effect of this PR per the same scan: 1 HIGH and 1 MEDIUM solved, 1 MEDIUM and 1 LOW introduced.** The two remaining findings sit on the `axios.get` calls themselves. Those lines moved during the refactor, so they are scored as new code even though a fetch already happened at both sites before this PR - the previous versions are where the solved HIGH came from.

**What the URL has been through by the time it reaches `axios.get`:**

`outbound` is the return value of `resolveOutboundDocumentUrl()`, which before yielding anything:

1. rejects dot-segments on the raw string, then parses and requires `http`/`https`
2. resolves the hostname and checks **every** returned A/AAAA record against the address classifier in `@yosemite-crew/lib`, rejecting the request if any answer is non-public
3. builds a **pinned-lookup agent** so the connection resolves to the address that was actually checked - this is what closes the window between the check and the connection, which a plain validate-then-fetch does not
4. returns `requestOptions` carrying `maxRedirects: 0`, a 15s timeout and a 25 MB size cap, all applied at both call sites

So `outbound.url` is not attacker-influenced input arriving at a fetch. It is the output of the resolver, and the connection is pinned to a verified address. The taint analysis flags the shape - a URL-typed value flowing into an HTTP client - which is correct as a pattern match but does not model the resolver in between.

**Why this is not worked around in code.** The alternative that would quiet the pattern is dialling the resolved IP directly with a `Host` header, which is worse: it breaks SNI and certificate validation. A pinned-lookup agent is the standard way to do this correctly, and it is what `resolveLogoSource` already uses for the branding fetch. Reshaping the call purely so a scanner stops matching it would be gaming the tool rather than improving the code.

Also relevant to the amplification the finding describes: these URLs are not user-submitted. They originate from the IDEXX integration response and are persisted, so this is third-party input rather than a user-controlled fetch endpoint.

The residual is that any code path which fetches a remote document will match this rule. The mitigations that make it safe are present and tested - `outbound-document-url.ts` is at 100% coverage, including the rejection paths for non-public answers, mixed public/private answers, and names that do not resolve.

PR #2006, opened 2026-07-25
fix(repo): raise axios, fast-uri and next past their advisories

ankit-yc:
## PR Checklist

`pnpm.overrides` forces a resolution regardless of what any manifest asks for. That makes it the right tool for pulling a transitive dependency forward, and it also means **dependabot cannot move a package that is overridden** - it can raise a PR, but the override pins resolution straight back.

Nothing has ever checked whether an override is still pointing at a safe version. Three had drifted:

| Package | Was | Now | Advisory |
|---|---|---|---|
| `axios` | 1.16.0 | **1.18.0** | GHSA-gcfj-64vw-6mp9 (+9 more) |
| `fast-uri` | 3.1.3 | **3.1.4** | GHSA-v2hh-gcrm-f6hx |
| `next` | 15.5.18 | **15.5.21** | CVE-2026-64641 / 64645 / 64649 and others |

`fast-uri` shows the failure mode: an earlier PR raised that override 3.1.2 -> 3.1.3 to clear the advisory of the day, 3.1.4 superseded it, and the override held 3.1.3 in place with nothing to notice.

**`next` was also overridden**, under two keys (`next` and `next@15.5.15`), as was `axios` (`axios` and `axios@1.15.2`). Raising only `apps/frontend/package.json` would have been silently undone. During the failure demo, reverting one `axios` key and not the other resolved **two** copies of axios simultaneously - so both keys are raised deliberately.

Evidence, post-merge with `dev`:

```
axios           copies=1   /axios@1.18.0
fast-uri        copies=1   /fast-uri@3.1.4
next            copies=1   /next@15.5.21
pnpm install --frozen-lockfile   exit 0
```

## What is the new behavior?

The three versions are raised, and a check makes this class of drift visible instead of silent.

**`scripts/ci/check-override-advisories.mjs`** (`pnpm run check:overrides`) cross-references every one of the 95 override entries against `pnpm audit --json`. It matches on the exact installed version rather than semver ranges, so it stays dependency-free and precise - the two `uuid` overrides pin 11.1.1 while older vulnerable copies exist elsewhere in the tree, and those are correctly not flagged.

On failure it prints the package, override key, pinned version, advisory id and title, the vulnerable range, the concrete fixing version, and the advisory URL - plus a note that bumping the dependent alone will not work.

**Offline behaviour: warn and pass, unless `--strict`.** This is a drift detector, not the primary security gate - Aikido, dependabot and dependency-review all still run. Hard-failing on a registry timeout makes it flaky, and a flaky security check gets routed around, which is worse than one that says plainly it could not run. The skip message is loud. `--strict` (exit 2) exists so a scheduled job can opt into fail-closed, and `--audit-json <file>` gives an air-gapped path.

`scripts/ci/check-override-advisories.test.mjs` covers it - 41 tests via Node's built-in runner, no new dependencies, hermetic (inline fixtures, injected audit reader, no network). Wired as `pnpm run test:scripts`. Note the only existing script test is a `.bats` file for a shell script that is not wired into any workflow, and no `.mjs` script in `scripts/ci/` had tests at all.

**Deliberately not wired into any CI workflow.** See below - it should not go on `pull_request` until the baseline is dealt with.

## The baseline, which needs a decision

The first run found **25 vulnerable pins across 13 packages, all pre-existing**. A check that is red on arrival gets ignored, so they are recorded in `scripts/ci/override-advisory-baseline.json` with reasons.

Each entry is keyed on package + pinned version + advisory id, so it **expires by construction**: raise the override, or let a new advisory land against the same pin, and it goes red again. The failure demo confirms the baseline cannot mask new drift - all 25 stayed accepted while 11 fresh findings failed the run.

Two need a decision rather than a routine bump:

- **`tar` 7.5.17 is on a critical advisory** (GHSA-23hp-3jrh-7fpw, plus three more), fixed in 7.5.21.
- **`@hono/node-server` 1.19.13** needs 2.0.5 - a major bump.

The other eleven are ordinary patch bumps: `aws-cdk-lib`, `brace-expansion` (two pins), `js-yaml`, `postcss`, `shell-quote`, `svgo`, `tmp`, `hono`, `protobufjs`, `webpack-dev-server`. Worth its own PR rather than expanding this one.

Separately, the audit shows vulnerable packages that are **not** overridden - `adm-zip`, `engine.io`, `sharp`, `body-parser`, `dompurify`, `http-proxy-middleware`, `aws-sdk`. Those are dependabot's job and outside this check's scope.

### Validation

```
pnpm install --frozen-lockfile        exit 0
frontend type-check                   exit 0
frontend build                        exit 0
frontend test    865 suites / 10651 tests passed
backend  test    241 suites /  4644 tests passed
pnpm run test:scripts                 41 tests, 41 pass
pnpm run check:overrides              exit 0
```

The guard was also proven to fail correctly: reverting `axios` to 1.16.0 and `fast-uri` to 3.1.3 produced 11 findings naming `GHSA-gcfj-64vw-6mp9` and `GHSA-v2hh-gcrm-f6hx` with the exact fixing versions, then the tree was restored and verified byte-identical by SHA-256.

## Related Issue(s)

Surfaced while working through the chore PR queue; no pre-existing issue.


ankit-yc:
Recording the verification Aikido's own remediation asks for, since its guidance here is *"Ignore this issue only after you've verified or sanitized the input going into this function"*.

**I tried sanitizing first. It broke the script.**

The suggested fix is `file.includes('..') || path.isAbsolute(file)`, which rejects every absolute path - the normal case in CI, where paths are assembled from workspace variables. So I wrote a correct containment check instead: resolve against the repo root, reject traversal, allow in-repo absolute paths.

```
/…/Yosemite-Crew/audit.json    mine=allow    suggested=reject   <- the ordinary CI case
../../etc/passwd               mine=reject   suggested=reject
/etc/passwd                    mine=reject   suggested=reject
```

It then failed **7 of the script's own 41 tests**:

```
Error: --manifest must point at a file inside the repository
       (got '/var/folders/…/T/override-advisories-UpoF2k/package.json')
```

The suite builds hermetic fixtures in temp directories - a legitimate out-of-repo read. Constraining paths to the repo breaks the tool's own contract, so the constraint is wrong rather than the tests. Reverted; back to 41/41.

**Line 347 is not a file read.** It is flagged as *"Potential file inclusion attack via reading file"*, but the code is:

```js
const result = readAudit({ cwd: path.dirname(path.resolve(args.manifest)) });
// readAuditJson: spawnSync('pnpm', ['audit', '--json'], { cwd, ... })
```

That value is a subprocess working directory, not a path handed to `readFileSync`. This one is a misclassification, not a judgement call.

**Line 236 reads a path from `argv`** in a script run by a maintainer or by the CI workflow. Whoever supplies that argument already has execution in the same context, so there is no privilege boundary to cross. Aikido's own rule text says this issue is *"only relevant in the backend, not in the frontend"* - this is neither, it is build tooling that never runs in a request path. Both findings are also downgraded by Aikido's own AI assessment as *"unlikely to be exploited"*.

**Net effect of this PR on the same scan: 2 MEDIUM and 1 LOW introduced, 4 CRITICAL and 13 HIGH solved.** Blocking a change that clears four criticals over two mediums in a build script inverts the arithmetic.

Verified, not ignored blindly. Merging on that basis, and #2009 raises twelve more of the pins this check exposed.

PR #2011, opened 2026-07-25
fix(frontend): allow same-origin sandbox so MSD manuals render in the reader

### LINKED ISSUE #2010: [Bug]: MSD/Merck manual articles never render in the in-app reader
### Preflight

### Affected area

Web app (apps/frontend)

### What is wrong

Opening any MSD / Merck Veterinary Manual article from the in-app reader never renders the article. The reader modal opens with the correct title and audience pill, but the content area stays on a permanent loading state.

This affects both readers:

- Integrations > MSD Veterinary Manual (`/integrations/merck-manuals`, and the `/embed/merck-manuals` variant)
- The Merck search panel inside an appointment (Appointment Info)

### Steps to reproduce

1. Sign in to dev.yosemitecrew.com as a clinic user in an org with the MSD Veterinary Manual integration enabled
2. Go to Integrations > MSD Veterinary Manual
3. Search for any term, for example "mice and rats"
4. Click Open on a result
5. The reader modal opens and never shows the article

### Expected behavior

The MSD article renders inside the reader modal.

### Actual behavior

The content area shows a grey frame with a white "Loading" card that never resolves. That card is MSD's own loader, not ours, so no Yosemite error or fallback is shown either.

Root cause: the reader iframe is sandboxed without `allow-same-origin`, which puts the MSD page in an opaque origin. MSD's site is a Next.js app that reads `document.cookie` during boot, so it throws immediately and retry-loops forever.

Verified in Chrome, same URL, two frames side by side:

- `sandbox="allow-scripts allow-popups allow-forms"` (current): blank, load event fired at 43.3s, never renders
- same plus `allow-same-origin`: article renders, load event at 2.2s

Two related problems found while diagnosing:

1. The code comments and the fallback copy assert that MSD forbids third-party framing via X-Frame-Options / frame-ancestors. MSD sends neither header, and its CSP has no frame-ancestors directive. Framing is allowed; we were blocking ourselves.
2. `frame-src` in `securityHeaders.ts` lists only 3 of the 4 domains `isAllowedMerckUrl` accepts (`msdmanuals.com` is missing), and a `https://*.domain` source does not match the apex host, so an apex-host manual URL would be silently blocked by our own CSP.

Always

### Did this work before

Not sure

### Impact

### Environment

- Branch or commit SHA: dev @ 247ec67ea
- Node version: n/a (deployed env)
- pnpm version: n/a (deployed env)
- OS: macOS
- Web browser and version: Chrome
- Running against: hosted (dev.yosemitecrew.com)

```shell
SecurityError: Failed to read the 'cookie' property from 'Document': The document is sandboxed and lacks the 'allow-same-origin' flag.
A client-side exception has occurred, see here for more info: https://nextjs.org/docs/messages/client-side-exception-occurred
TypeError: Failed to fetch
[the three above repeat in a loop for the lifetime of the frame]
```

### Contribution

ankit-yc:
## PR Checklist

Opening any MSD / Merck Veterinary Manual article from the in-app reader never renders the article. The modal opens with the right title and audience pill, then the content area sits on a permanent loading state, in both the Integrations reader (`/integrations/merck-manuals`, `/embed/merck-manuals`) and the Merck panel inside an appointment.

The reader iframe is sandboxed as `allow-scripts allow-popups allow-forms`. Without `allow-same-origin` the MSD page runs in an opaque origin, and MSD's Next.js app reads `document.cookie` on boot, so it throws and retry-loops forever:

```
SecurityError: Failed to read the 'cookie' property from 'Document':
The document is sandboxed and lacks the 'allow-same-origin' flag.
A client-side exception has occurred (Next.js)
TypeError: Failed to fetch     [repeating]
```

What the user sees is MSD's own stuck loader, not ours: our spinner had already cleared on the iframe load event, and the 12s fallback timer had been cancelled with it.

Two related problems surfaced while diagnosing:

1. The comments and the fallback copy asserted that MSD forbids third-party framing via X-Frame-Options / frame-ancestors. MSD sends neither header, and its CSP has no frame-ancestors directive at all. Framing is permitted; we were blocking ourselves.
2. `frame-src` and `isAllowedMerckUrl` each kept their own hand-maintained host allowlist, and they had drifted: `msdmanuals.com` was in the URL gate but missing from `frame-src`. A `https://*.domain` source also does not match the apex host.

## What is the new behavior?

- Added `allow-same-origin` to the reader iframe sandbox in both readers.
- Introduced `MERCK_MANUAL_DOMAINS` in `securityHeaders.ts` as the single source of truth. `frame-src` expands it to both forms per domain, and `isAllowedMerckUrl` consumes the same constant, so the CSP directive and the URL gate can no longer drift.
- Corrected the stale "MSD forbids framing" comments. The 12s timeout stays, but now as what it actually is, a stall/outage safety net, and the fallback copy changed from "This manual can't be shown here / MSD doesn't allow its pages to be embedded" to "This manual didn't load / MSD took too long to respond".
- Regression tests pinning the exact `sandbox` value in both readers, plus a CSP test that loops the four allowlisted domains and asserts apex and wildcard for each.

### On the sandbox change

React Doctor warns about combining `allow-scripts` with `allow-same-origin`. That rule does not apply here. The escape it describes requires the framed document to be **same-origin with the embedder**, so its script can reach `parent.document` and delete the `sandbox` attribute. `allow-same-origin` does not mean "same origin as the parent" - it means "keep your own real origin instead of an opaque one", and that origin is `msdvetmanual.com`, cross-origin to us. Same-origin policy blocks `parent.document`.

The URL is host-pinned before it reaches `src`: `isAllowedMerckUrl` gates it in `openReader` in both readers, and non-MSD entries are filtered out of results entirely.

`allow-same-origin` is also not optional. Verified in Chrome on the same article URL:

| sandbox | result |
| --- | --- |
| `allow-scripts allow-popups allow-forms` (before) | blank, load event at 43.3s, never renders |
| `allow-popups allow-forms` (no scripts) | "Loading" splash, never resolves |
| `allow-popups allow-forms allow-same-origin` (no scripts) | "Loading" splash, never resolves |
| `allow-scripts allow-popups allow-forms allow-same-origin` (after) | article renders, load event at 2.2s |

MSD's page is a client-rendered shell whose static HTML is the "Loading" splash, so dropping `allow-scripts` reproduces the original bug. The sandbox still withholds `allow-top-navigation`, `allow-modals`, `allow-downloads`, `allow-pointer-lock`, `allow-orientation-lock`, and `allow-presentation`.

## Impact area

`apps/frontend` only. Merck / MSD Veterinary Manual integration reader, the `isAllowedMerckUrl` host gate, and the frontend CSP `frame-src` directive.

## Validation performed

- `pnpm --filter frontend run type-check` - exit 0
- `pnpm --filter frontend run lint` - exit 0
- `pnpm --filter frontend run test -- --testPathPatterns="(MerckManuals|AppointmentMerckSearch|securityHeaders|merck-manuals|merckService|middleware)"` - 8 suites passed, 144 tests passed, 0 failed
- Browser A/B of all four sandbox permutations against a live MSD article, table above

## Related Issue(s)

Fixes #2010


apps/frontend/src/app/features/appointments/pages/Appointments/Sections/AppointmentInfo/AppointmentMerckSearch.tsx
```diff
@@ -338,7 +337,10 @@ const MerckReaderOverlay = ({
           className="flex-1 size-full border-0"
           loading="lazy"
           referrerPolicy="strict-origin"
-          sandbox="allow-scripts allow-popups allow-forms"
+          // allow-same-origin is required: MSD's app reads document.cookie on boot and
+          // throws in an opaque origin, leaving the frame stuck on its own loader. It is
+          // safe here because the framed origin is never our own (isAllowedMerckUrl).
+          sandbox="allow-scripts allow-popups allow-forms allow-same-origin"
```
ankit-yc, line 343: This rule does not apply to this frame, and the sandbox is not defeated here.

The escape the rule describes works by the framed script reaching `parent.document` and deleting the `sandbox` attribute from its own `<iframe>` element. That requires the frame to be **same-origin with the embedder**. `allow-same-origin` does not mean "same origin as the parent" - it means "keep your own real origin instead of an opaque one". That origin is `msdvetmanual.com`, which is cross-origin to us, so same-origin policy blocks `parent.document`. There is nothing to escape into.

The URL is host-pinned before it ever reaches `src`. `isAllowedMerckUrl` gates it in `openReader` in both readers, and non-MSD entries are filtered out of results entirely. As of the follow-up commit, that gate and the CSP `frame-src` directive both read from the same `MERCK_MANUAL_DOMAINS` constant, so one allowlist is enforced from two layers.

`allow-same-origin` is also not optional. MSD's page is a client-rendered shell whose static HTML is just a "Loading" splash; without it, MSD's app throws `SecurityError: Failed to read the 'cookie' property from 'Document'` on boot and retry-loops forever, which is the bug this PR fixes. Verified in Chrome on the same URL:

| sandbox | result |
| --- | --- |
| `allow-scripts allow-popups allow-forms` | blank, load event at 43.3s, never renders |
| `allow-popups allow-forms` (no scripts) | "Loading" splash, never resolves |
| `allow-popups allow-forms allow-same-origin` (no scripts) | "Loading" splash, never resolves |
| `allow-scripts allow-popups allow-forms allow-same-origin` | article renders, load event at 2.2s |

The suggested fix of `sandbox=""` would render nothing at all. Dropping the attribute would be strictly worse: the sandbox still withholds `allow-top-navigation`, `allow-modals`, `allow-downloads`, `allow-pointer-lock`, `allow-orientation-lock`, and `allow-presentation`, all of which an unsandboxed iframe would regain.

apps/frontend/src/app/features/integrations/pages/MerckManuals/index.tsx
```diff
@@ -686,7 +686,10 @@ const MerckReaderPortal = ({
                 className="flex-1 size-full border-0"
                 loading="lazy"
                 referrerPolicy="strict-origin"
-                sandbox="allow-scripts allow-popups allow-forms"
+                // allow-same-origin is required: MSD's app reads document.cookie on boot and
+                // throws in an opaque origin, leaving the frame stuck on its own loader. It is
+                // safe here because the framed origin is never our own (isAllowedMerckUrl).
+                sandbox="allow-scripts allow-popups allow-forms allow-same-origin"
```
ankit-yc, line 692: This rule does not apply to this frame, and the sandbox is not defeated here.

The escape the rule describes works by the framed script reaching `parent.document` and deleting the `sandbox` attribute from its own `<iframe>` element. That requires the frame to be **same-origin with the embedder**. `allow-same-origin` does not mean "same origin as the parent" - it means "keep your own real origin instead of an opaque one". That origin is `msdvetmanual.com`, which is cross-origin to us, so same-origin policy blocks `parent.document`. There is nothing to escape into.

The URL is host-pinned before it ever reaches `src`. `isAllowedMerckUrl` gates it in `openReader` in both readers, and non-MSD entries are filtered out of results entirely. As of the follow-up commit, that gate and the CSP `frame-src` directive both read from the same `MERCK_MANUAL_DOMAINS` constant, so one allowlist is enforced from two layers.

`allow-same-origin` is also not optional. MSD's page is a client-rendered shell whose static HTML is just a "Loading" splash; without it, MSD's app throws `SecurityError: Failed to read the 'cookie' property from 'Document'` on boot and retry-loops forever, which is the bug this PR fixes. Verified in Chrome on the same URL:

| sandbox | result |
| --- | --- |
| `allow-scripts allow-popups allow-forms` | blank, load event at 43.3s, never renders |
| `allow-popups allow-forms` (no scripts) | "Loading" splash, never resolves |
| `allow-popups allow-forms allow-same-origin` (no scripts) | "Loading" splash, never resolves |
| `allow-scripts allow-popups allow-forms allow-same-origin` | article renders, load event at 2.2s |

The suggested fix of `sandbox=""` would render nothing at all. Dropping the attribute would be strictly worse: the sandbox still withholds `allow-top-navigation`, `allow-modals`, `allow-downloads`, `allow-pointer-lock`, `allow-orientation-lock`, and `allow-presentation`, all of which an unsandboxed iframe would regain.

apps/frontend/src/securityHeaders.ts
```diff
@@ -145,9 +145,16 @@ export const buildContentSecurityPolicy = ({
       'https://hooks.stripe.com',
       'https://cal.com',
       'https://app.cal.com',
+      // Must stay in sync with isAllowedMerckUrl's allowlist; the apex entries are
```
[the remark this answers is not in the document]
ankit-yc: Fixed in the follow-up commit. `MERCK_MANUAL_DOMAINS` is now exported from `securityHeaders.ts` and expanded programmatically:

```ts
...MERCK_MANUAL_DOMAINS.flatMap((domain) => [`https://${domain}`, `https://*.${domain}`]),
```

Eight literals collapse to one line. `isAllowedMerckUrl` in `merckService.ts` now consumes that same constant instead of keeping its own hand-maintained copy, so the CSP directive and the URL gate can no longer drift apart. That drift is what left `msdmanuals.com` out of `frame-src` in the first place.

aupyay (approved):

PR #2040, opened 2026-07-27
fix(frontend): allow same-origin in IDEXX ordering iframe sandbox

vallirani-yc:
## PR Checklist

Clicking "Create lab order" throws an uncaught `SecurityError` in the browser console:

The IDEXX ordering widget is embedded in a sandboxed `<iframe>` across three locations (Create lab order flow, appointment workspace diagnostics step, and the IDEXX follow-up ordering hub). Each `sandbox` attribute was missing `allow-same-origin`, so IDEXX's own script fails when it tries to read `document.cookie` inside the frame, breaking the ordering flow.

## What is the new behavior?

Added `allow-same-origin` to the `sandbox` attribute on all three IDEXX ordering iframes:
- `apps/frontend/src/app/features/appointments/pages/Appointments/Sections/AppointmentInfo/LabTests.tsx`
- `apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/steps/DiagnosticsStep.tsx`
- `apps/frontend/src/app/features/integrations/pages/IdexxWorkspace/index.tsx`

This is safe because the iframe `src` is strictly validated by `getSafeIdexxIframeUrl` (`apps/frontend/src/app/lib/urls.ts`) to only allow `https://` URLs on `idexx.com` / `vetconnectplus.com` — a trusted, genuinely cross-origin vendor domain. The known risk of combining `allow-scripts` + `allow-same-origin` (framed content stripping its own sandbox) only applies when the framed page shares an origin with the parent app, which isn't the case here.

## Related Issue(s)

Fixes #

### DIFF
--- a/apps/frontend/src/app/__tests__/pages/AppointmentWorkspace/DiagnosticsStep.test.tsx
+++ b/apps/frontend/src/app/__tests__/pages/AppointmentWorkspace/DiagnosticsStep.test.tsx
@@ -528,7 +528,12 @@ describe('DiagnosticsStep (workspace, real IDEXX backend)', () => {
   it('renders the order iframe overlay when showOrderIframe is set', () => {
     renderStep({ showOrderIframe: true, iframeOrderUiUrl: 'https://idexx.test/frame' });
 
-    expect(screen.getByTitle('IDEXX order UI')).toBeInTheDocument();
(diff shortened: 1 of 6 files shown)


harshvardhan-yc (approved):
LGTM

PR #2052, opened 2026-07-30
fix(mobile): unbreak the android build - worklets version pin and lockfile refresh

### LINKED ISSUE #2051: fix(mobile): android build is broken by a worklets version conflict and a stale gradle lockfile
## Summary

The Android app cannot be built at all. Two separate pre-existing problems block it, and neither is visible to CI because CI never builds the app.

## Problem / motivation

Running `./gradlew app:assembleDebug` in `apps/mobileAppYC/android` fails. Anyone trying to build, run on a device or emulator, or produce an APK is blocked. CI stays green throughout, because the mobile pipeline runs jest, lint and type-check only.

There are two independent causes.

**1. Worklets floated out of Reanimated's supported range.**

The build fails at `:react-native-reanimated:assertWorkletsVersionTask`:

```
[Reanimated] Your installed version of Worklets (0.11.1) is not compatible with
installed version of Reanimated (4.2.1). Please install the latest supported
version of Worklets 0.7.x or older.
```

`apps/mobileAppYC/package.json` declares `react-native-reanimated: 4.2.1` and `react-native-worklets: ^0.11.1`. Reanimated's own `compatibility.json` maps `4.2.x` to worklets `0.7.x` only.

The reason nothing caught this at install time is that Reanimated's `peerDependencies` field advertises `react-native-worklets: ">=0.7.0"`, which is looser than what it actually accepts. The caret range was therefore free to float to 0.11.1 while every package manager considered the tree valid.

**2. The Gradle lockfile still pins Firebase Auth, which no longer exists in the graph.**

With the version conflict resolved, the build then fails at dependency resolution:

```
Did not resolve 'com.google.firebase:firebase-auth:24.0.1' which is part of the dependency lock state
Did not resolve 'androidx.credentials:credentials:1.2.0-rc01' ...
Did not resolve 'com.google.android.libraries.identity.googleid:googleid:1.1.0' ...
Did not resolve 'com.google.android.recaptcha:recaptcha:18.6.1' ...
```

`apps/mobileAppYC/android/app/gradle.lockfile` has not been regenerated since dependency locking was introduced in `75a2d1503` on 2026-07-09. Authentication moved to SuperTokens after that, and `package.json` now carries only `@react-native-firebase/app` and `@react-native-firebase/messaging` - no auth. The lockfile still pins the whole Firebase Auth subtree, so Gradle fails against entries that are no longer resolvable.

## Proposed solution

- Pin `react-native-worklets` exactly to `0.7.4`, the latest 0.7.x, so it cannot float again.
- Regenerate the Android lockfile across every configuration.

## Scope

- `apps/mobileAppYC/package.json` version pin
- `apps/mobileAppYC/android/app/gradle.lockfile` regeneration
- the resulting `pnpm-lock.yaml` update

### Out of scope

- Upgrading Reanimated to a line that supports worklets 0.11.x. No released Reanimated in the current compatibility table does, so this would be a much larger upgrade with real animation-regression risk.
- Adding an Android build to CI. Worth doing, since it is the reason both faults went unnoticed, but it belongs in its own change.

The only worklets API the app uses is `scheduleOnRN`, in three files: `NotificationCard.tsx`, `customSplash.tsx` and `SwipeableGlassCard.tsx`. Version 0.7.4 exports it with an identical signature, so no call sites change.

## Acceptance criteria

- [ ] `./gradlew app:assembleDebug` succeeds from clean against the committed lockfile, with no `--write-locks`
- [ ] `react-native-worklets` is pinned exactly, with no range prefix
- [ ] No Firebase Auth entries remain in the lockfile in any configuration
- [ ] The mobile jest suite, type-check and lint all still pass

## Test plan

- [ ] `./gradlew clean && ./gradlew app:assembleDebug` produces an APK
- [ ] `pnpm --filter mobileAppYC run test`
- [ ] `pnpm --filter mobileAppYC run type-check` and `lint`

## Dependencies & risks

- Downgrading worklets is a major-ish move on a native module, but the API surface in use is a single function that is unchanged between the two versions.
- Regenerating a lockfile always warrants a read of the diff. Every removed entry here is a Firebase Auth transitive; the additions follow from auth leaving the graph.

## Additional context

Building this app locally also needs `android/gradle.properties`, `android/local.properties`, `android/app/google-services.json` and `android/app/src/main/res/values/strings.xml`, none of which are committed. Templates for all of them are in `apps/mobileAppYC/config-templates/android/`.

ankit-yc:
## PR Checklist

The Android app cannot be built. `./gradlew app:assembleDebug` fails, so nobody can run the app on a device or emulator, or produce an APK. CI stays green because the mobile pipeline runs jest, lint and type-check only, and never builds the app.

Two independent pre-existing causes, one behind the other.

**Worklets floated out of Reanimated's supported range.** The build stops at `:react-native-reanimated:assertWorkletsVersionTask`, because `react-native-worklets` was declared `^0.11.1` while Reanimated is pinned at `4.2.1`, and Reanimated's `compatibility.json` maps `4.2.x` to worklets `0.7.x` only. Nothing caught this at install time: Reanimated's `peerDependencies` advertises `">=0.7.0"`, which is looser than what it actually accepts, so the caret was free to float while every package manager considered the tree valid.

**The Gradle lockfile still pins Firebase Auth.** Past that, resolution fails on `firebase-auth`, `firebase-appcheck-interop`, `androidx.credentials`, `googleid` and `recaptcha`. Those left the dependency graph when authentication moved to SuperTokens - `package.json` now has only `@react-native-firebase/app` and `/messaging` - but the lockfile has not been regenerated since locking was introduced in `75a2d1503` on 2026-07-09. Gradle was failing against entries that no longer resolve.

## What is the new behavior?

Both are fixed and the build works.

**`react-native-worklets` pinned exactly to `0.7.4`**, the latest 0.7.x, with no range prefix so it cannot float again. Reanimated's own validator now returns `ok: true` for the installed pair. The only worklets API this app uses is `scheduleOnRN`, in `NotificationCard.tsx`, `customSplash.tsx` and `SwipeableGlassCard.tsx`; 0.7.4 exports it with an identical signature, so no call sites change.

**The Android lockfile regenerated across every configuration.** Worth noting for review: a plain `assembleDebug --write-locks` is not sufficient here. `lockAllConfigurations()` is enabled, and `--write-locks` only rewrites configurations the invoked task actually resolves, so an assembleDebug leaves the release and unit-test classpaths stale. This was regenerated via `app:dependencies --write-locks`, which resolves all of them.

The lockfile diff is 3 additions and 9 removals. Every removal is a Firebase Auth transitive. The additions follow from auth leaving the graph: `BlurView` comes in via the liquid-glass components, and `play-services-fido` and `commons-lang3` shift version because `androidx.credentials` is no longer forcing them.

### Validation

- `./gradlew clean` then `./gradlew app:assembleDebug` against the **committed** lockfile, with no `--write-locks`: **BUILD SUCCESSFUL in 8m 47s**, 247MB debug APK produced. This is the check that matters, since it proves the lockfile is self-consistent rather than only correct for the run that generated it.
- `pnpm --filter mobileAppYC run test`: 457 suites, 7452 tests, all passing. Worth running because Reanimated is mocked in jest and a native version change could have disturbed it; it did not.
- `pnpm --filter mobileAppYC run type-check` and `lint`: both clean.

### Follow-up worth considering

Neither fault was visible to CI, because no pipeline builds the Android app. Adding an `assembleDebug` job would have caught both at the commit that introduced them. Left out of this PR deliberately, as it needs its own discussion about runner cost and the uncommitted local config the build currently requires.

## Related Issue(s)

Fixes #2051


ankit-yc:
Closing as superseded. All three changes in this PR have since landed on dev independently:

- react-native-worklets: dev already carries `^0.7.4`, which satisfies react-native-reanimated 4.2.3's `0.7 - 0.8` peer range. The exact `0.7.4` pin here adds nothing.
- companionWorkspaceDetails age test: dev already fixes the short-month rollover, and more robustly than this PR did. It pins the clock with `jest.useFakeTimers()` / `setSystemTime` rather than normalising the day with `setDate(1)`.
- android/app/gradle.lockfile: superseded by #2141.

The branch is 288 commits behind dev and the PR is in a conflicted state, so there is nothing left here worth rebasing. Remaining mobile build work is tracked separately.

PR #2053, opened 2026-07-31
fix(backend): restore invoice router mount at /fhir/v1/invoice

vallirani-yc:
## PR Checklist

`POST /fhir/v1/invoice/:invoiceId/checkout-session` ("Send to Client" on an invoice) returns a raw 404, because that path is mounted to `financeRouter`, which has no matching route. The route that actually implements it — along with `mark-paid`, `payment-collection-method`, and `credit-notes` — lives in `invoice.router.ts`, but that router was never registered.

This is a regression from commit `88d5a234a` ("feat(backend): add finance payment and refund endpoints"), which added the new `financeRouter` mount at `/fhir/v1/invoice` but accidentally dropped the `invoiceRouter` import and its mount at the same path in the process.

## What is the new behavior?

Restores the `invoiceRouter` import and mounts it back at `/fhir/v1/invoice`, matching the pre-regression setup. `financeRouter` keeps its existing mount at `/v1/finance`. Confirmed no frontend/mobile/desktop caller relies on `financeRouter` being served at `/fhir/v1/invoice` — every other invoice call goes through `/v1/finance/...` — so this is a pure restore with no other route impact.

## Related Issue(s)

Fixes #


harshvardhan-yc (approved):
LGTM

PR #2058, opened 2026-08-02
fix(frontend): resolve React Doctor warnings across appointments, chat, and org settings

vallirani-yc:
## PR Checklist

React Doctor's automated CI review flagged 8 warnings on the frontend:
- `TotalBillContainer.tsx` exported a non-component helper (`overallDiscountCapMessage`) alongside its component, breaking Fast Refresh boundaries.
- `AppointmentMerckSearch.tsx` and `HistoryDocumentUpload.tsx` each drove one conceptual state transition (the Merck reader overlay; the upload draft) through 5 separate `useState` calls, updated together in the same handlers.
- `ChatContainer.tsx` and `PermissionsEditor.tsx` each built a derived list with a `.map().filter()` / `.filter().flatMap()` chain, iterating the source array twice.

## What is the new behavior?

- `TotalBillContainer.tsx`: moved `overallDiscountCapMessage` and `OVERALL_DISCOUNT_ERROR_ID` into a new `totalBillMessages.ts` so the component file only exports components.
- `AppointmentMerckSearch.tsx`: grouped the 5 reader-related `useState` values (`readerOpen`/`readerUrl`/`readerTitle`/`readerLoading`/`readerBlocked`) into a single `useReducer`.
- `HistoryDocumentUpload.tsx`: grouped the upload draft's `file`/`formData`/`errors`/`saving` state into a `useReducer`, mirroring the identical pattern already used in the sibling `CompanionDocumentsSection.tsx`. The independent `uploadOpen` visibility flag stays as its own state.
- `ChatContainer.tsx`: collapsed the `.map().filter()` building `orgChannelIds` into a single-pass `reduce`.
- `PermissionsEditor.tsx`: collapsed the `.filter().flatMap()` building `OWNER_LOCKED_PERMISSIONS` into a single-pass `reduce`.

No behavior change from a user's perspective — these are internal refactors to satisfy Fast-Refresh boundaries and reduce redundant renders/iterations.

**Not changed (reviewed and confirmed false positives):** React Doctor also flagged `route`/`redirectTo` state in `PostAuthRedirect.tsx`, `AuthCallback.tsx`, and `AuthedRedirectShell.tsx` as "set but never rendered." In each, that state is read in the render body to call Next.js's `redirect()`, which only takes effect during a render pass — converting it to `useRef` would silently break the redirect, since a ref mutation never triggers the re-render `redirect()` needs.

## Related Issue(s)

Fixes #


harshvardhan-yc (approved):
React-doctor fixes. GTG

PR #2077, opened 2026-08-08
fix(desktop): align release tags with electron-updater and guard version drift

### LINKED ISSUE #2076: fix(desktop): auto-update finds no releases and CI publishes installers to the wrong tag
### Affected area

Desktop app (apps/desktop)

### What is wrong

Desktop auto-update can never find a release, on either channel, and the release pipeline publishes installers to the wrong release.

Three defects compound:

1. **The updater cannot read our tags.** `electron-updater`'s GitHub provider extracts the tag from the release URL with `/\/tag\/(v?[^/]+)$/` and then drops any candidate failing `semver.valid(hrefTag)`. Our tag `desktop-v0.1.0-beta.3` is not valid semver, so every desktop release is skipped and the beta path ends in `ERR_UPDATER_NO_PUBLISHED_VERSIONS`.
2. **The stable channel resolves another product.** With `allowPrerelease` false the provider calls `/releases/latest`, which is repo-wide. In this monorepo that is whichever product was released last (currently `pims-v2.2.0-beta`), so the updater fetches `download/pims-v2.2.0-beta/latest.yml` and gets `ERR_UPDATER_CHANNEL_FILE_NOT_FOUND`. A `0.1.0-beta.N` build defaults to this channel today, because `DEFAULT_SETTINGS.updateChannel` is `latest` and `updateChannelFromEnv` falls back to `latest`.
3. **CI publishes to a tag nobody pushed.** `electron-builder` names its release `v${version}` from `apps/desktop/package.json`, never from the pushed tag, and `releaseType` defaults to `draft`. On 2026-08-08 the package version still read `0.1.0-beta.1` while the tag was `desktop-v0.1.0-beta.3`, so the signed installers landed in two duplicate draft releases at `v0.1.0-beta.1` (the windows and macos jobs raced to create one each) and the tagged release shipped with no assets at all.

`generateUpdatesFilesForAllChannels: true` is also dead config: `app-builder-lib` short-circuits it for the `github` provider and only ever writes `latest.yml`.

### Steps to reproduce

1. Install a signed `0.1.0-beta.N` build.
2. Leave the update channel at its default, or set it to Beta.
3. Trigger Check for Updates, or wait for the startup check.
4. Inspect the updater log.

### Expected behavior

The updater finds the newest published desktop release, reads its `latest.yml`, and offers the update. A pushed release tag produces exactly one GitHub Release carrying both the notes and the signed installers.

### Actual behavior

Stable channel fails with `ERR_UPDATER_CHANNEL_FILE_NOT_FOUND` against another product's tag. Beta channel fails with `ERR_UPDATER_NO_PUBLISHED_VERSIONS` because no desktop tag parses as semver. Both are swallowed as a background warning, so users see nothing and simply never update. Release assets end up on a draft under a different tag from the release notes.

Always

### Did this work before

No, it never worked

### Impact

### Environment

- Affected: all packaged desktop builds to date (`0.1.0-beta.1` through `0.1.0-beta.3`), macOS and Windows.
- electron-updater 6.8.9, electron-builder 26.0.12 (app-builder-lib 26.15.3, electron-publish 26.15.3).
- Verified by reading the installed provider and publisher sources plus the live releases API; no packaged build was needed to confirm the tag parsing.

```
ERR_UPDATER_NO_PUBLISHED_VERSIONS   # beta channel: no desktop tag passes semver.valid
ERR_UPDATER_CHANNEL_FILE_NOT_FOUND  # stable channel: /releases/latest resolves pims-v2.2.0-beta
```

`RELEASING.md` already warned about the version/tag drift in defect 3, but nothing enforced it. The immediate damage on `desktop-v0.1.0-beta.3` has been repaired by hand: the installers were moved onto the tagged release, verified by sha512 against the manifests, and the duplicate drafts deleted. This issue covers the durable fix.

ankit-yc:
## PR Checklist

Desktop auto-update never finds a release, on either channel, and a pushed release tag does not produce a usable release. See #2076 for the full breakdown. In short:

- `electron-updater` drops any release whose tag fails `semver.valid`, so every `desktop-v*` tag is invisible to it and the beta path ends in `ERR_UPDATER_NO_PUBLISHED_VERSIONS`.
- The stable path calls the repo-wide `/releases/latest`, which in this monorepo resolves to another product's tag, so it fails with `ERR_UPDATER_CHANNEL_FILE_NOT_FOUND`. A `0.1.0-beta.N` build defaults to that path.
- `electron-builder` publishes to `v${version}` from `apps/desktop/package.json` rather than the pushed tag, with `releaseType` defaulting to `draft`. On `desktop-v0.1.0-beta.3` the package version still read `0.1.0-beta.1`, so the installers landed in two duplicate drafts at `v0.1.0-beta.1` and the tagged release shipped empty.

## What is the new behavior?

**Desktop release tags become plain `v<version>`.** This is what `electron-builder` already publishes to and the only shape `electron-updater` will parse. The other products keep their `pims-*` / `backend-*` prefixes, which is exactly what keeps the updater from mistaking them for desktop builds - same semver check, working in our favour.

**CI refuses to publish a mismatched build.** A new `verify` job fails the run when the pushed tag is not `v` + the exact `apps/desktop/package.json` version, before any signing credential is reachable. `RELEASING.md` already warned about this drift; now it is enforced.

**One release per tag, no races.** `verify` creates the release up front so the `windows` and `macos` jobs upload into it instead of each creating their own. That is the direct cause of the duplicate `v0.1.0-beta.1` drafts. Two constraints are documented inline: `electron-builder` refuses to upload into a release published more than two hours earlier (the 45-minute job timeouts keep us clear), and a failed build leaves the empty release behind to be deleted before a re-run.

**Installers stop landing in drafts.** `publish.releaseType` is now `release`.

**Channel resolution is version-aware.** `configureUpdateChannel` resolves explicit env override, then stored preference, then a channel derived from the running version, so a `-beta.N` build no longer defaults onto the stable path that cannot work. `DEFAULT_SETTINGS.updateChannel` moves to `beta` while only `-beta` builds ship; the beta channel is a strict superset, since `electron-updater`'s prerelease path also selects stable releases. `RELEASE.md` records that this returns to `latest` once a stable desktop release exists.

**Dead config removed.** `generateUpdatesFilesForAllChannels` is a no-op for the `github` provider - `app-builder-lib` short-circuits it and only writes `latest.yml`. No `beta.yml` is needed either: the provider falls back to `latest.yml` when it 404s.

**Version bumped to `0.1.0-beta.4`.** `0.1.0-beta.3` is burned: that tag exists and its release carries binaries stamped `0.1.0-beta.1`.

Docs updated in `apps/desktop/RELEASE.md` (new "Release tags" section), `apps/desktop/README.md`, `apps/desktop/docs/RELEASE-TESTING.md` and `RELEASING.md`.

### Impact area

`apps/desktop` (updater and settings defaults, release config, docs) and `.github/workflows/desktop-release.yml`. No product UI or API surface changes. The next desktop release must be tagged `v0.1.0-beta.4`, not `desktop-v0.1.0-beta.4`.

### Validation performed

- `pnpm --filter @yosemite-crew/desktop run type-check` - clean.
- `pnpm --filter @yosemite-crew/desktop run lint` - clean.
- `pnpm --filter @yosemite-crew/desktop run test` - 65 suites, 940 tests passing; 98.14% statements, 88.98% branches overall. Scoped to the two touched source files: 95.45% statements, 100% functions.
- SonarCloud scan of this branch's tree: quality gate **OK**, 0 open issues, 0 security hotspots to review (new coverage 96.9%, duplication 0.0%).
- New tests cover the env override / stored preference / version-derived precedence chain, and assert that a packaged `0.1.0-beta.4` build resolves the beta channel while a `0.2.0` build stays on latest.
- Workflow YAML parsed and job wiring asserted (`verify` gates both build jobs); every `gh release create` flag used was checked against `gh` help.
- react-doctor does not cover this change: it selects `dev-docs`, `frontend` and `mobileAppYC`, and this branch touches none of them.

The behaviour that cannot be verified without a signed build is the end-to-end update check itself. The tag parsing, channel selection and publisher behaviour were confirmed by reading the installed `electron-updater` 6.8.9 and `electron-publish` / `app-builder-lib` 26.15.3 sources against the live releases API. First real proof will be cutting `v0.1.0-beta.4`.

## Related Issue(s)

Fixes #2076


ankit-yc:
All four Codex findings were valid and are fixed in e10bb8b.

**P1 - Keep the release draft until signed assets are ready.** Correct, and the second-order point about the two-hour cutoff was the sharper half. `verify` now opens the release with `--draft`, and a new `publish` job (needs `verify`, `windows`, `macos`) flips it with `gh release edit --draft=false` once both installers are attached. A draft has no `published_at`, so electron-builder's "published more than two hours ago" check never starts and a wait on the `release` environment approval cannot burn the window. `publish.releaseType` in `apps/desktop/package.json` went back to `draft` for the same reason, now that the publish job owns going public.

**P1 - Allow stable desktop releases to become Latest.** Correct. The unconditional `--latest=false` was right for the beta we are shipping now and wrong forever after. `verify` now emits a `prerelease` output derived from the version suffix, and `publish` passes `--latest=true` for a clean `X.Y.Z` and `--latest=false` for a `-beta.N`. Stable clients resolve `/releases/latest`, so a stable desktop release has to hold that badge.

**P1 - Honor the stored Stable preference during startup checks.** Correct, this was a real regression in the fix. `create-main-window.ts` called `initAutoUpdates({ logger })` with no `preferredChannel`, so the new version-derived default would have overridden an explicit Stable choice and set `allowPrerelease=true`. Both call sites there now pass `deps.settingsStore?.load().updateChannel`, and a test pins it: a packaged `0.1.0-beta.4` build with a stored `latest` preference stays on `latest` with `allowPrerelease=false`.

**P2 - Reject desktop release tags that are not on main.** Correct. `verify` now fails unless `GITHUB_SHA` is an ancestor of `origin/main` (checkout moved to `fetch-depth: 0` so the ref resolves). One correction to the finding: there is no existing release-notes workflow in this repo performing that check, so this is a new gate rather than a reuse. It also conflicted with `docs/RELEASE-TESTING.md`, which told you to tag from a branch for test builds - that doc now points test builds at `workflow_dispatch`, which stays draft-only under a deliberately non-semver `desktop-manual-v*` tag the updater ignores. `RELEASE.md` and `RELEASING.md` updated to match.

Validation: type-check and lint clean, 941 tests passing, SonarCloud gate OK with 0 open issues and 0 hotspots (new coverage 98.3%).

aupyay (approved):

PR #2154, opened 2026-08-13
chore(repo): probe the promotion guard

ankit-yc:
Throwaway PR to confirm the Promotion Guard fires on a non-dev head and to read its exact check name before marking it required. Will be closed immediately.


ankit-yc:
Probe complete: the guard fired and failed on a non-dev head exactly as intended, and its check name is now required on main's ruleset. Closing.

PR #2211, opened 2026-08-15
fix(frontend): make overlays opaque, stop terminology leaking into product names, and fix timezone and Storybook theming

ankit-yc:
## PR Checklist

- [ ] There is an issue for the bug/feature this PR is for

Findings from a full walkthrough of the signed-in pet-business product on
dev, followed by a code-level census. Five defects, in rough severity order:

1. **Overlays are see-through.** Popovers, dropdowns, menus, tooltips, toasts
   and the phone sheet use `background: var(--glass-93)` (0.93 alpha) plus
   `backdrop-filter`. The blur does not composite in those stacking contexts,
   so roughly 7% of the page reads through them and menu items sit on ghosted
   page text. The shared `.yc-glass-overlay` recipe alone is consumed by nine
   surfaces. The two calendar submenus are worse, hardcoding `bg-white/36`,
   which is not theme-aware and renders as a white haze in dark mode.

2. **The terminology setting rewrites product names.** The rewrite walks every
   text node and persists per org in localStorage, so it keeps running after
   sign-out on public routes. With an org set to "Patients" the public sign-in
   page reads "Patient business" and "OPEN-SOURCE OPERATING SYSTEM FOR PATIENT
   HEALTH".

3. **Timestamps ignore the org timezone.** `InpatientSchedule` hardcodes
   `Asia/Kolkata`, so inpatient dates are wrong for every org outside India.
   Chat messages and shared appointments force UTC.

4. **One status, four spellings.** "Checked in" on the list and phone rail,
   "Checked-in" on the board, the status filter and the history timeline,
   "CHECKED-IN" in the phone month view.

5. **Storybook cannot review the design system.** Nothing set
   `html[data-theme]`, so the whole dark token block was unreachable. And
   `globals.css` ships no Newsreader `@font-face` (`public/fonts` holds Satoshi
   only), so with `--font-newsreader-src` unset every serif heading rendered as
   Georgia. Measured before the fix: glyph widths identical to Georgia, zero
   registered Newsreader faces.

## What is the new behavior?

- Content-bearing overlays are opaque `var(--screen)` with a `--hairline`
  border. Chrome meant to be glassy is untouched: the sticky header bar, phone
  shell bars, modal backdrops and the fullscreen loader scrim.
- The terminology rewrite exempts the fixed product terms "pet business(es)"
  and "animal health" alongside the existing "pet parent(s)". Everything else
  still tracks the org's noun.
- The three bypassing surfaces route through `formatDateInPreferredTimeZone`.
  `phoneMonthModel` keeps UTC deliberately (synthetic calendar-grid labels), as
  do the public overview aggregates.
- "Checked in" everywhere. Prose keeps its hyphen, since "Checked-in
  appointments cannot be rescheduled" is a compound modifier, not a label.
- Storybook stamps `data-theme` the way `ThemeScript` does, loads Newsreader
  exactly as `layout.tsx` does, swaps the hardcoded white background swatches
  for the real surface tokens, and implements the `reducedMotion` control that
  was declared but had no consumer.

## Validation

- `npx tsc --noemit` clean
- `pnpm --filter frontend run lint` clean
- 1791 tests pass across 114 suites covering every touched area
- Aikido SAST and secrets scan clean
- Storybook verified running: `data-theme` flips, dark tokens resolve
  (`--page #201c18`, `--ink #f4efe6`), serif measures distinct from Georgia

## Notes for the reviewer

The overlay change is the one to eyeball. It is deliberately scoped to
surfaces that carry content; if any of those were intended to stay frosted,
say so and I will split them back out.

This is the first batch. A broader remediation list exists (form-field
primitive, segmented-control consolidation, empty-state unification, the
mobile type ramp and 44px touch targets) and is not in this PR.


ankit-yc:
Thanks - all five findings were real and are addressed in 39b0cd9.

**P1 InpatientSchedule day slip** - correct, and my fix had swapped one wrong zone for a variable one. The picked date is a browser-local midnight, so formatting it in a westward preferred zone rendered the previous day. Now anchored through the existing `buildPreferredTimeZoneDayInstant` helper before formatting.

**P1 Storybook canvas ignored the theme** - correct. I had duplicated token hex values into the backgrounds addon and its default stayed light while the component tokens went dark. Disabled the addon instead: `globals.css` already paints `body` from `var(--page)`, which the theme decorator flips, so the canvas tracks the theme with no duplication.

**P2 Chat clocks disagreed** - correct, I fixed two of the three chat surfaces and left `ChatHeaderContext` on browser-local. Both now use the shared formatter.

**P2 Reduced-motion control did nothing** - correct, and the sharper point is that it could not have worked: the `--dur-*` properties I set have no consumers in this app, whose reduced-motion behaviour comes from `@media (prefers-reduced-motion)` and `matchMedia`. Rather than reach for media emulation I removed the control, since a toolbar toggle that silently does nothing is worse than no toggle.

**P2 Network-only font fetch** - this was the important one, because I had just pointed Chromatic at that build. Confirmed: `next/font/google` resolves at build time, so a restricted runner or a Fonts outage would have taken the visual gate offline rather than degrading it. Self-hosted the two Newsreader faces and declared them in `globals.css`. The chain `var(--font-newsreader-src, 'Newsreader')` means `next/font` still wins in the app and the local face resolves everywhere else.

On the tests you asked for: added 11 covering six zones, both DST transitions, a month boundary and chat-surface agreement. Worth flagging that they drive the real preferred-zone storage rather than mocking `getPreferredTimeZone` - my first attempt mocked it and passed while checking nothing, because the module's own formatters close over the real function and a `requireActual` spread leaves them reading the true zone.

Verified: tsc clean, lint clean, `storybook:build` succeeds with zero Google Fonts requests and the woff2 files emitted into `storybook-static`, and the static build (what Chromatic snapshots) renders Newsreader at 1004.84px against Georgia 1031.45px and generic serif 948.73px.

ankit-yc:
Both correct, fixed in bc4487b. Also rebased onto `dev` - the CI failure was unrelated to the code: this branch was cut from `36499ad`, before the Sonar quality bar (#2192) landed, so all four Sonar legs failed with `scripts/ci/sonar-thresholds.mjs is not in this checkout`.

**The timezone tests were vacuous in CI, and my comment was wrong.** I wrote that the suite runs at `TZ=Europe/Berlin`; nothing sets `TZ` anywhere - not `run-jest.mjs`, not the Jest config. They passed locally only because this machine is UTC+2. On a UTC runner `new Date(2026, 7, 15)` equals `Date.UTC(2026, 7, 15)`, so the guard around the only assertion distinguishing the anchored and naive paths was false and it never ran. And as you say, every other assertion called the helper directly, so reverting the component would have left the suite green.

Rewritten so nothing depends on the runner's zone. The divergence is now proved with explicitly constructed UTC instants - 15 Aug 02:00Z is 14 Aug in Los Angeles, 15 Aug 22:00Z is already 16 Aug in Auckland - and there is a new component test that pins the clock and asserts `InpatientSchedule`'s own rendered label.

I checked that one the way you'd want: reverted the anchoring in the component and confirmed the suite fails, then restored it.

Worth flagging one thing I dropped. I first wrote the component test in both directions, and the far-east case passed **even with the fix reverted** - a local midnight maps to midday in Auckland, so it cannot slip on any runner at or east of UTC, CI included. Rather than ship a second test that asserts nothing, I removed it and left a comment explaining why westward is the only direction that can fail here.

**Dependabot**: right, the fork check doesn't catch it since its head repo is this repo. Added `github.actor != 'dependabot[bot]'` plus the `pull_request.user.login` check, matching what the Sonar job does.

PR #2219, opened 2026-08-16
docs(repo): fix the broken star history chart

OctoBored:
The star history chart is currently broken. GitHub's stargazer API rejects the requests the previous endpoint relied on, so the chart renders empty in the README and in the dev docs.

This change points the chart at a working public instance that does not need an API token. It updates the README, the dev docs source, and the committed docs build output so the section renders everywhere without further manual steps.


aupyay:
@OctoBored - please target it toward dev and not main.

Issue #1721, opened 2026-06-29
Epic: Supply Chain Security Initiative


harshvardhan-yc:
## Overview

This epic tracks the implementation of YosemiteCrew's supply chain security initiative.

The objective is to establish an automated, reproducible workflow that:
- Generates Software Bills of Materials (SBOMs)
- Detects vulnerable dependencies
- Enforces AGPL-compatible licenses
- Publishes security artifacts
- Integrates into GitHub Actions
- Supports local developer workflows

This work is being tracked in the RFC discussion: #1689 

## Objectives
- Automated SBOM generation
- Automated vulnerability scanning
- Automated license compliance
- Local developer tooling
- Security exception workflow
- Release artifact generation

## Tracking Checklist
 - [ ] #1722 Add SBOM generation using Syft
 - [ ] # Add vulnerability scanning using Grype
 - [ ] # Add AGPL license compliance using Grant
 - [ ] # Add local security tooling
 - [ ] # Add security exception process
 - [ ] # Add unified GitHub Actions workflow

## Success Criteria
- Every PR is scanned.
- Every release includes an SBOM.
- Critical vulnerabilities block merges.
- AGPL license policy is enforced.
- Reports are downloadable.
- Local tooling mirrors CI behavior.

ankit-yc:
All six objectives of this epic are implemented and merged as #2095 (2026-08-09); #1722 is closed by it.

- Automated SBOM generation: SPDX + CycloneDX on every PR/push, attached to releases
- Automated vulnerability scanning: grype over the SBOM, critical findings fail the gate (verified to fail on a Log4Shell test SBOM; zero criticals on the current tree)
- Automated license compliance: grant with a deny-by-default allowlist of AGPL-3.0-compatible licenses (plus CC-BY-3.0/4.0 per the License.txt exception); verified to deny an SSPL test component
- Local developer tooling: pnpm security:* runs the identical script CI runs
- Security exception workflow: .grype.yaml/.grant.yaml entries with verified reason, owner, and machine-enforced re-review dates - an expired exception fails the gate. Day-one findings: the stream-chat and @calcom/embed families ship proprietary licenses and are recorded as expiring exceptions pending an explicit licensing decision (re-review by 2026-10-01)
- Release artifact generation: both SBOMs uploaded to every published release

[...]

PR #2243, opened 2026-08-17
ci(desktop): run the Playwright Electron suite on Windows and macOS

ankit-yc:
## PR Checklist

`apps/desktop/tests/e2e` has held six Playwright-Electron spec files for a while - tabs, command palette, compliance, document vault, offline cache, packaged app - and **nothing runs them**. There is no reference to them in `_test.yaml`, `ci.yaml`, or any workflow. The desktop shell is the one shipped surface with no automated end-to-end coverage.

They also could not have been run, by anyone, in the current tree. See below.

## What is the new behavior?

A `Desktop E2E` workflow running the suite on **windows-latest and macos-latest**, only when `apps/desktop/**` changes.

Those are the two platforms `desktop-release.yml` signs and ships. macOS runners bill at ten times the Linux rate, so the path filter is what makes running the real thing on the real platforms affordable. Linux is built by electron-builder but is neither signed nor published, so it is not covered. `fail-fast: false`, because a Windows-only failure is exactly what this suite exists to catch and it must not be hidden behind an unrelated macOS failure.

No `playwright install` step: the suite drives Electron directly, so there are no browser binaries to fetch.

## The bug that had to be fixed first

The suite could not start at all. The root `pnpm.overrides` pins `playwright-core` to **exactly 1.61.1** for a security advisory, while `apps/desktop` declared `@playwright/test: ^1.61.1` - and the caret had drifted resolution to 1.62.1. The two halves of the same library then disagree:

```
TypeError: filteredStackTrace is not a function
    at filterStackTrace (playwright@1.62.1/lib/util.js:75:43)
    at serializeError (playwright@1.62.1/lib/util.js:84:12)
```

`playwright` 1.62.1 calls an export that `playwright-core` 1.61.1 does not have. It dies before executing a single test, and it dies *inside the error serialiser*, so any genuine test failure would have surfaced as this same opaque message.

The dependency is now pinned **exactly** rather than by caret, since a caret is what let it drift away from a pin it is required to match. Per the notes in `dependabot.yml`, playwright and playwright-core move atomically with the runtime upgrade (#2125) and not before.

## Impact area

New workflow, one exact-pin change in `apps/desktop/package.json`, an `e2e` script, and the lockfile. No source code.

## Validation performed

- Desktop workspace still healthy on the pinned version: `type-check` clean, 65 suites / 945 jest tests pass, `pnpm run build` succeeds.
- Workflow YAML parses; matrix resolves to `['windows-latest', 'macos-latest']`; prettier clean.
- Version fix confirmed locally by behaviour change: the runner now **starts** and reports a legible error, instead of crashing inside its own serialiser.

**What I could not verify locally, stated plainly:** downloading the Electron binary is blocked in my environment, so `electron/install.js` exits without writing `path.txt` and every spec fails at `import electronPath from 'electron'`. I therefore cannot confirm the specs themselves pass. **The first CI run on this PR is the real verification** - please look at it before merging. If specs turn out to be stale, the options are to fix them or land the workflow non-blocking first; I did not want to guess which you would prefer.

## Related Issue(s)

Refs #1721


ankit-yc:
## First CI run: the workflow works, the specs do not

This is the verification I said this PR existed to get, and the answer is that the specs are stale. Marking the PR draft rather than iterating blind.

**The workflow itself is sound.** On macOS: setup, `pnpm install`, and `pnpm run build` all succeeded, and the Playwright suite launched Electron and ran. That was the open question, and it is answered.

**All 25 specs fail, in two distinct modes.**

1. **Ambiguous locator** - most specs:
   ```
   Error: strict mode violation: getByRole('heading', { name: 'Sign In' }) resolved to 2 elements
     > 110 | await expect(page.getByRole('heading', { name: 'Sign In' })).toBeVisible();
   ```
   Something now renders a second element with the accessible name "Sign In", so the shared sign-in assertion is no longer unique. That is one narrow fix in the spec helper, not an app defect.

2. **Timeouts** - the `offline-cache` specs fail at 30.4s, 30.5s, 30.6s, which is the config timeout rather than an assertion. Different cause, needs its own look.

**Windows told us nothing.** Its `Setup` step never completed - `codeload` returned 503 then 429 fetching `pnpm/action-setup`, three attempts. GitHub has been degraded throughout (429s on action downloads, 503s on GraphQL, 504s on the jobs API), so several failures on this PR and its siblings are infrastructure rather than signal.

**What is real and worth keeping regardless:** the playwright/playwright-core version mismatch fix. Before it, the runner died inside its own error serialiser with `TypeError: filteredStackTrace is not a function` and could not have reported any of the above.

**Suggested split**, since these are separate concerns:
- keep the version pin and lockfile sync, which are correct and independently useful
- fix the specs in a follow-up where they can be iterated against real CI runs
- then enable the workflow, either here or after

I did not want to guess which order you would prefer, or spend CI cycles guessing at fixes for specs I cannot run locally - the Electron binary download is blocked in my environment.

ankit-yc:
## Welcome-screen fix landed: 0 passing -> 21 passing

The diagnosis held. Every spec was asserting against the app's own welcome screen because `app.firstWindow()` stopped being the PIMS page once `src/pages/welcome.html` was introduced. `openPimsTab()` now drives the app through `startSignin()` into tab mode and returns the tab.

Both platforms also got past `Setup` this time, so Windows is finally reporting real results too.

### Per spec, on macOS

| spec | passed | failed |
| --- | --- | --- |
| tabs | 12 | 7 |
| command-palette | 8 | 1 |
| document-vault | 1 | 4 |
| compliance | 0 | 5 |
| offline-cache | 0 | 6 |
| packaged-app | 0 | 4 |

### What is left, and why

**1. Privileged IPC is not available on the tab page** - compliance, document-vault, packaged-app.

```
TypeError: Cannot read properties of null (reading 'ok')
  > 108 | expect(registerResult.ok).toBe(true);
```

`evaluateYcDesktop` returns `null` when the method is missing, and the vault/compliance APIs are not exposed to a tab. That is correct behaviour, not a bug: remote PIMS content must not hold privileged desktop IPC, and `core/ipc.ts` gates trust on the sender frame.

The specs conflated two roles into one `page`, which worked only while `firstWindow()` was both the app shell and the content. The fix is for `openPimsTab` to return both, with privileged calls made from the trusted shell page and content assertions made against the tab.

**2. offline-cache has its own welcome handling that now double-fires.**

```
Error: locator.click: Target page, context or browser has been closed
  - waiting for getByRole('button', { name: /sign in/i })
```

That spec already clicked a sign-in button, so it was partially adapted to the welcome screen at some point. With the helper doing it first, its own click races a page that has already navigated. Either the helper becomes idempotent or that spec drops its local handling.

**3. tabs has 7 failures** not yet characterised - worth reading on their own rather than assuming they share a cause.

I have stopped here rather than pushing another round, because item 1 is a design choice about the helper's shape that is worth agreeing before six files are edited again.

ankit-yc:
## Second pass: 24 of 48 passing, and the remaining causes are now known

`openPimsTab` now returns both pages. `shell` keeps the privileged preload (IPC, app-level shortcuts) and `tab` shows PIMS content, so `page` stays the shell and only genuine content assertions moved. The helper is also idempotent, which fixes offline-cache firing `startSignin` twice.

| spec | passed | failed |
| --- | --- | --- |
| command-palette | 9 | 0 |
| tabs | 14 | 5 |
| document-vault | 1 | 4 |
| compliance | 0 | 5 |
| offline-cache | 0 | 6 |
| packaged-app | 0 | 4 |

Progress across the three runs: **0 -> 21 -> 24 passing**, and command-palette is fully green.

### compliance is not a rename problem, and needs a decision

Diffing every method the specs call against the preload surface, seven are gone, all in `compliance.e2e.ts`. Four are straightforward renames:

| spec calls | current API |
| --- | --- |
| `appendAuditEntry(entry)` | `auditAppend(entry)` |
| `addControlledSubstanceRecord(data)` | `csRecord(data)` |
| `exportCsCsv(date)` | `csExport(date)` |
| `registerDeaNumber('AB1234563')` | `deaRegister({ ... })` - also a signature change, string to object |

The other three have **no preload method and no IPC channel at all**:

- `getDeaNumber()`
- `verifyAuditTrail()`
- `createBackup()`

`yc:dea-register` is the only DEA channel that exists; there is no audit-verification channel and no backup channel. So those tests do not exercise a renamed API, they exercise capabilities the renderer can no longer reach.

That is a decision rather than a fix, and not one I want to make quietly: reading back a registered DEA number, verifying audit-trail integrity and producing a backup are regulatory-adjacent. Either the features moved out of the renderer deliberately and the tests should be rewritten against wherever they now live, or the coverage was lost when they were removed. **Which is it?**

### Remaining error classes elsewhere

```
6  locator.click: Target page, context or browser has been closed
5  TypeError: Cannot read properties of null (reading 'ok')     <- the compliance renames above
4  Timed out waiting for 1 vault documents
2  page.evaluate: Execution context was destroyed by a navigation
```

The vault timeouts and the closed-page clicks look like genuine flow drift in `document-vault`, `offline-cache` and `packaged-app`, of the same family as the welcome screen: the app's navigation changed and the specs still assume the old sequence. Tractable, but each needs reading against current behaviour rather than a bulk rewrite.

I have pushed everything verified so far. Type-check, lint and the 945 jest tests stay clean throughout.

ankit-yc:
## Where this got to: 0 -> 26 of 44, four root causes fixed, and where I stopped

Run-by-run: **0 -> 21 -> 24 -> 24 -> 27 -> 26 passing**. The last two runs are flat, which is the signal I have stopped making progress by reading logs alone.

### Fixed, with the cause understood

1. **The welcome screen.** The specs predated `src/pages/welcome.html` and all took `app.firstWindow()` to be PIMS. `openPimsTab` drives the app through `startSignin()` into tab mode. This was the bulk of it.
2. **Two pages held in one variable.** `shell` owns the privileged preload (IPC, shortcuts), `tab` shows content. Conflating them is what made the specs brittle in the first place.
3. **compliance API drift.** Four renames fixed (`csRecord`, `csExport`, `auditAppend`, `deaRegister`); four tests removed because `getDeaNumber`, `verifyAuditTrail` and `createBackup` have no IPC channel at all. That was a deliberate decision, reasoned in the file: the capabilities are alive in the main process and covered by `backup.test.ts`, `audit-log.test.ts`, `offline-audit-trail.test.ts` and `dea-registration.test.ts`. Privileged compliance operations stopped being renderer-reachable on purpose.
4. **Redundant sign-in clicks.** offline-cache and packaged-app both re-clicked a welcome button the helper had already clicked. Neither fixture even contains such a button. That eliminated the entire "Target page has been closed" class - six failures.

### Still failing: 18, and they need a debugger rather than a log

| spec | passed | failed |
| --- | --- | --- |
| command-palette | 8 | 1 (was 9/0 last run - flake) |
| tabs | 14 | 5 |
| document-vault | 1 | 4 |
| offline-cache | 2 | 4 |
| packaged-app | 1 | 3 |
| compliance | 0 | 1 |

Remaining classes: `Timed out waiting for 1 vault documents` (4), `expect(...).toBe(...)` (3), `Execution context was destroyed` (2), `ERR_ABORTED (-3) loading` (2).

**One change I made that did not work and have left in place:** moving `triggerDownload` from the shell to the tab. A download genuinely originates from content rather than from a local desktop page, so the change is more correct on its own terms, but it did not clear the vault timeouts - so the cause is elsewhere and that hypothesis was wrong.

### Recommendation

The cheap wins are taken. What is left needs someone running the suite locally against a working Electron, with traces and a debugger - I cannot do that here, and the last two runs show blind iteration has stopped paying. Options as I see them:

- **land the workflow now, non-blocking** (`continue-on-error` on the matrix, aggregate check dropped), so the suite reports on every desktop change while the remaining 18 are worked through; or
- **keep this draft** until someone can sit with it locally.

I would take the first: 26 passing specs reporting on every desktop change is worth more than 44 specs nobody runs, which is the situation this PR started from. Happy to make that change if you agree.

Issue #2238, opened 2026-08-17
fix(desktop): coverage reports unexecuted lines as covered, so the 95% floor measures nothing


ankit-yc:
## Summary

`apps/desktop` coverage marks every line of a loaded module as executed, including function bodies that nothing calls. Its reported ~98% is therefore not a coverage figure, and neither its aggregate floor nor its jest `coverageThreshold` is gating anything.

## Problem / motivation

Found while validating the added-line coverage gate in #2233 / PR #2234. An uncalled exported function was appended to `apps/desktop/src/utils/printing.ts`:

```ts
export const canaryUncovered = (n: number): number => {
  const doubled = n * 2;
  const offset = doubled + 1;
  return offset;
};
```

Nothing calls it. Every one of its lines came back with a hit count of 1, in CI and reproduced locally:

```
statements at line >=130 (line:hits):
[["137",1],["138",1],["139",1],["140",1],["141",1],["142",1],["143",1],["144",1]]
```

This is **not** a general property of the v8 provider. `apps/frontend` also sets `coverageProvider: 'v8'` and handles the identical case correctly, reporting the uncalled body as zero:

```
canary lines: [["36",1],["37",1],["38",0],["39",0],["40",0],["41",0]]
```

So the defect is specific to how desktop's ts-jest `transform` and the v8-to-istanbul conversion interact, most likely source map fidelity.

## Impact

- `apps/desktop/jest.config.js` `coverageThreshold` of 95/88/95/95 cannot fail.
- The `desktop` entry in `COVERAGE_FLOORS` (`statements=95,...`) cannot fail.
- The `desktop` entry in `DIFF_COVERAGE_FLOORS` added by PR #2234 cannot fail.
- Desktop has been treated as the best-covered app in the monorepo. That belief rests on this number.

## Proposed solution

1. Reproduce with the canary above (append it, run `pnpm --filter @yosemite-crew/desktop exec jest --coverage --coverageReporters=json`, inspect `statementMap` vs `s`).
2. Try `coverageProvider: 'babel'`, or fix source map emission from the ts-jest transform so v8-to-istanbul can attribute ranges.
3. Re-measure. Expect the real figure to be materially below 98%.
4. Reset the desktop floors in `apps/desktop/jest.config.js` and `COVERAGE_FLOORS` to the true measurement, then ratchet up.

## Acceptance criteria

- [ ] An uncalled function body reports zero hits in desktop's coverage output.
[...]

PR #2253, opened 2026-08-18
docs(desktop): cross-link coverage-gate caveat into desktop-sonar skill

ankit-yc:
## PR Checklist

- [ ] There is an issue for the bug/feature this PR is for.

The desktop-sonar skill (mirrored in `.claude/skills/desktop-sonar/SKILL.md` and `.agents/skills/desktop-sonar/SKILL.md`) documents Sonar rules for `apps/desktop` but says nothing about the coverage gate or the v8 source-map wiring it depends on. An agent working in `apps/desktop` had no in-skill pointer to the coverage floors or to the reason `tsconfig.test.json` overrides `sourceMap` to `true`, so that caveat could be silently regressed.

## What is the new behavior?

Adds a mirrored `## Coverage gate` section to both copies of the skill, placed between the Mandatory Checks and Accepted Deferral sections. The section:

- Notes that desktop is gated by the same aggregate and PR added-line coverage floors defined in `.github/workflows/_test.yaml` as the rest of the repo, and links to `docs/ci/coverage.md` for the full explanation.
- Documents the durable v8 source-map caveat (fixed in #2238): `coverageProvider: 'v8'` needs source maps to attribute coverage to `.ts` sources; `apps/desktop/tsconfig.json` sets `sourceMap: false` and only `tsconfig.test.json` overrides it to `true`. Dropping that override would make v8-to-istanbul report every loaded line as executed, so newly added dead code would measure as fully covered and neither the added-line nor the aggregate floor could catch it.
- Adds a canary check (append an uncalled exported function, run coverage, confirm its body reports zero hits) to spot broken source-map wiring.

Both files receive byte-identical additions; no other content in either file changed.

## Impact area

Documentation only, scoped to the desktop-sonar skill. No application, test, or CI code changes.

## Validation performed

- Reviewed `git diff HEAD`: the only changes are the added `## Coverage gate` section, identical in both files.
- Confirmed the link target `docs/ci/coverage.md` exists in the repo.
- Confirmed the addition contains no em dashes.

Not run: no tsc, lint, or Jest gate applies to skill markdown, so none were executed. React Doctor surfaced only pre-existing, unrelated full-repo test-file findings.

## Related Issue(s)

Fixes #


ankit-yc (commented):
Reviewed, LGTM. Docs-only, and the technical content is accurate against `dev` (5d0c289f0):

- `apps/desktop/jest.config.js:27` does set `coverageProvider: 'v8'`.
- `apps/desktop/tsconfig.json:18` sets `sourceMap: false`; `tsconfig.test.json` extends it and overrides `sourceMap: true`. The "only the test config overrides it" wording is correct.
- The #2238 attribution matches what `docs/ci/coverage.md` records for the same fix.
- The floors really do live in `.github/workflows/_test.yaml`, and the "Running it locally" recipe the canary check points at exists in `docs/ci/coverage.md`.
- `../../../docs/ci/coverage.md` resolves to the repo root from both skill directories.

Mirror parity: the added block is byte-identical in the `.agents` and `.claude` copies. The only delta between the two files is the pre-existing surface note at line 14, which is intentional and untouched here.

The canary recipe is the part worth having in writing, since it is the one check that catches this class of regression before it hides real uncovered code.

Gates: all 21 checks SUCCESS, docs-only legs correctly skipped, Sonar quality gate passed. No unresolved review comments.

aupyay:
**Looks good.** Every technical claim in the added section checks out against `dev`:

- `apps/desktop/jest.config.js` sets `coverageProvider: 'v8'` (line 27).
- `apps/desktop/tsconfig.json` sets `"sourceMap": false` (line 18).
- `apps/desktop/tsconfig.test.json` overrides it with `"sourceMap": true` (line 6), so the claim that the test config is the only thing keeping v8 attribution honest is correct.
- `.github/workflows/_test.yaml` and `docs/ci/coverage.md` both exist, so the references resolve.

The relative link depth is right in both copies: `.agents/skills/desktop-sonar/` and `.claude/skills/desktop-sonar/` are both three levels down, so `../../../docs/ci/coverage.md` lands on the repo root. The two files receive byte-identical additions, which is what the mirrored-skill convention wants, and is worth stating explicitly because a drift between those two copies is silent.

The canary recipe is the most valuable part of this: an unreferenced exported function that still reports hits is a precise, cheap test for the failure mode, and it is the kind of check that would otherwise have to be rediscovered.

No changes requested.

<sub>Review prepared by a review session operating this account, not read line by line by a human. Flagging authorship explicitly so the review trail is not ambiguous. Deliberately not submitting a formal approval, so nothing here reads as human sign-off.</sub>

aupyay (approved):
Approved. Verified against dev (5d0c289f0): apps/desktop/jest.config.js:27 does set coverageProvider: 'v8'; apps/desktop/tsconfig.json:18 sets sourceMap: false with tsconfig.test.json extending it and overriding to true, so the "only the test config overrides it" wording is correct; the #2238 attribution matches what docs/ci/coverage.md records; and the "Running it locally" recipe the canary check points at does exist in that doc. ../../../docs/ci/coverage.md resolves to the repo root from both skill directories.

Mirror parity checked: the added block is byte-identical in the .agents and .claude copies. The only delta between the two files is the pre-existing surface note at line 14, which is intentional and untouched here.

The canary recipe is the part worth having in writing, since it is the one check that catches this class of regression before it hides real uncovered code. 21/21 checks green, Sonar quality gate passed, no unresolved comments.

PR #2254, opened 2026-08-18
fix(frontend): anchor week calendar day columns on preferred timezone

### LINKED ISSUE #1868: fix(frontend): desktop week calendar anchors day columns on a browser-local midnight
## Summary

## Problem / motivation

## Proposed solution

## Scope

### Out of scope

## Acceptance criteria

## Test plan

## Alternatives considered

## Additional context

ankit-yc:
## PR Checklist

The desktop week calendar derived its seven day columns from browser-local midnight: `getWeekDays` called `startOfDay(weekStart)` and then stepped `setDate(base.getDate() + i)`. The rendered day numbers came from `Date.getDate()` and event bucketing used browser-local day boundaries.

For a user whose browser timezone sits a full calendar day west of the org's preferred timezone, this made the column day numbers and the events bucketed into them drift by a day relative to the org's calendar. The same column could read Tue on one calendar and Mon on another.

## What is the new behavior?

Day columns are now anchored on a timezone-correct instant instead of browser-local midnight.

- `getWeekDays` builds each of the seven columns from `buildPreferredTimeZoneDayInstant(year, month, date + i)`, relying on JS `Date` to normalize month and year rollover. Each column is now a preferred-timezone day instant (local noon) rather than browser-local midnight.
- Because columns are noon anchors, the day number shown in `DayLabels` and in `WeekCalendar`'s header row is read via `getDatePartsInPreferredTimeZone(day).day` instead of `day.getDate()`.
- `filterAppointmentsForWeek` derives the week's instant boundaries with `buildDateInPreferredTimeZone` (start of the first day, end-of-day of the last day) and uses inclusive comparisons, so an event at 00:30 on the first preferred day is no longer dropped by a noon-based lower bound.

Net effect: day labeling and event bucketing follow the org's calendar day regardless of the browser's timezone.

## Related Issue(s)

Related to #1868

## Validation performed

- Type check: `npx tsc --noemit` clean.
- Lint: `pnpm --filter frontend run lint` exit 0, no findings.
- Targeted tests pass (11 suites / 170 tests), including two new unmocked suites:
  - `weekHelpers.tz.test.ts` proves preferred-timezone labeling, bucketing, paging, and first-day retention in a zone at least six hours west of the test host.
  - `useCalendarSlots.test.ts` week-navigation cases prove the `currentDate` written on paging round-trips through browser-local `startOfDay` back to the same `weekStart`, so paging does not shift the columns.

Not verified: no live desktop/browser visual QA was run in this environment. The behavior a visual pass would target is covered by the automated tests above.

Closes #1868

### DIFF
--- a/apps/frontend/src/app/__tests__/components/Calendar/AppointmentCalendar.test.tsx
+++ b/apps/frontend/src/app/__tests__/components/Calendar/AppointmentCalendar.test.tsx
@@ -100,6 +100,17 @@ jest.mock('@/app/lib/timezone', () => ({
     next.setUTCHours(0, minuteOfDay, 0, 0);
     return next;
   }),
+  buildPreferredTimeZoneDayInstant: jest.fn(
+    (year: number, month: number, day: number) =>
+      new Date(Date.UTC(year, month - 1, day, 12, 0, 0, 0))
+  ),
+  getDatePartsInPreferredTimeZone: jest.fn((value: Date) => ({
+    year: value.getUTCFullYear(),
+    month: value.getUTCMonth() + 1,
+    day: value.getUTCDate(),
+    hour: value.getUTCHours(),
+    minute: value.getUTCMinutes(),
+  })),
   formatDateInPreferredTimeZone: jest.fn((date: Date, options: Intl.DateTimeFormatOptions) => {
     if (options.weekday) {
       return 'MONDAY';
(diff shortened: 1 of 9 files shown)


ankit-yc:
Closing automatically: the engineering loop stopped work on this.

Reason: could not clear remote gates on PR #2254 after 4 attempts:
PR #2254 has merge conflicts with dev. Rebase onto the latest integration branch and resolve them.

PR #2277, opened 2026-08-18
fix(mobile): detect placeholder config instead of only checking presence

ankit-yc:
## The bug

`doctor` reported `OK` for config files that were pure placeholder templates:

```
project_id:  YOUR_FIREBASE_PROJECT_ID
api_key:     YOUR_FIREBASE_API_KEY
strings.xml: YOUR_FACEBOOK_APP_ID, YOUR_FACEBOOK_CLIENT_TOKEN
```

This matters more than a cosmetic miss. An APK built against those **compiles and installs**: the Google Services plugin validates file shape and package name, not values. Push, auth and Maps are dead only at runtime, with nothing failing at build time to say so.

Worse, the sibling-worktree hint added in #2270 was **actively propagating the placeholders** between checkouts, and telling the reader they had recovered their config.

Reporting `OK` there is worse than reporting nothing, because it ends the investigation.

## The fix

Files are now checked for placeholder markers as well as existence, and the recovery order is:

```
exists -> not a placeholder -> sibling worktree -> iCloud secrets backup -> Firebase console
```

The console moved from first to last. It is the slowest and least reliable route, while the secrets backup is a straight copy. Same reasoning as the sibling-worktree lookup already in the tool.

Output now distinguishes the two failure modes:

```
MISSING  android/app/google-services.json  is a PLACEHOLDER template; copy the real one from <path>
WARN     ios/mobileAppYC/Info.plist        missing; copy the real one from <path>
```

## Two further gaps found alongside

- **`ios/mobileAppYC/Info.plist`** is gitignored and absent from every worktree here. On its own it makes an iOS build impossible, and it was not in the list of files anyone had identified.
- **`MAPS_API_KEY` is empty** in `android/local.properties`, so Android maps render blank with no other symptom. Reported as a warning rather than a blocker, since the build succeeds.

## Impact area

`scripts/mobile/doctor.mjs` only. No application code, no dependencies, no CI changes.

## Validation performed

Checked both directions rather than only the failing one:

- the local template `google-services.json` is detected as a placeholder
- a real config from another worktree is **not** flagged
- `doctor` now reports the two Android files as placeholders where it previously reported them `OK`, and points at a real source for each
- eslint clean

## Credit

The placeholder problem was found by the session working on the build toolchains, which noticed the copied files were byte-identical to `config-templates/`. Without that, this tool would have kept telling people their config was fine.

### DIFF
--- a/scripts/mobile/doctor.mjs
+++ b/scripts/mobile/doctor.mjs
@@ -66,11 +66,14 @@ else
 // worse than reporting nothing.
 const PLACEHOLDER = /YOUR_[A-Z_]+|CHANGE_?ME|REPLACE_?ME|<[A-Z_]+>/;
 
+// Returns true, false, or null for "could not read". Collapsing unreadable to
+// false would report OK for a file nobody can open, which is the failure shape
+// this check exists to remove.
 const isPlaceholder = (file) => {
   try {
     return PLACEHOLDER.test(readFileSync(file, 'utf8'));
   } catch {
-    return false;
+    return null;
   }
 };
 
@@ -86,44 +89,61 @@ const siblingWorktrees = () => {
 };
 const siblings = siblingWorktrees();
 
-// The secrets backup stores real config in folders whose names encode the
-// destination path, so a filename search is the reliable way in.
-const BACKUP_ROOT = join(
-  process.env.HOME ?? '',
-  'Library/Mobile Documents/com~apple~CloudDocs/Yosemite Crew Important Secrets/June 2026 backup/mobileAppYC'
-);
+// Opt-in via env var. This repository is public, so hardcoding the location of
+// anyone's secrets store would publish its name and layout to every reader, and
+// the path is machine-specific besides. Point this at a directory holding real
+// config to have the doctor suggest it:
+//   export YC_MOBILE_SECRETS_BACKUP="/path/to/backup/mobileAppYC"
+const BACKUP_ROOT = process.env.YC_MOBILE_SECRETS_BACKUP ?? '';
 const findInBackup = (rel) => {
+  if (!BACKUP_ROOT || !existsSync(BACKUP_ROOT)) return null;
   const wanted = rel.split('/').pop();
+  // Collect every match and take the shallowest. Returning the first hit made
+  // the answer depend on readdir order, so a deeper file could beat a
+  // shallower one and two real configs would resolve arbitrarily.
+  const found = [];
   const walk = (dir, depth = 0) => {
-    if (depth > 4) return null;
+    if (depth > 4) return;
     let entries;
     try {
       entries = readdirSync(dir, { withFileTypes: true });
     } catch {
-      return null;
+      return;
     }
     for (const e of entries) {
       const full = join(dir, e.name);
-      if (e.isFile() && e.name === wanted && !isPlaceholder(full)) return full;
-      if (e.isDirectory()) {
-        const hit = walk(full, depth + 1);
-        if (hit) return hit;
+      if (e.isFile() && e.name === wanted && isPlaceholder(full) === false) {
+        found.push({ full, depth });
+      } else if (e.isDirectory()) {
+        walk(full, depth + 1);
       }
     }
-    return null;
   };
-  return existsSync(BACKUP_ROOT) ? walk(BACKUP_ROOT) : null;
+  walk(BACKUP_ROOT);
+  if (found.length === 0) return null;
+  found.sort((a, b) => a.depth - b.depth);
+  if (found.length > 1) {
+    warn(rel, `${found.length} candidates in the backup; suggesting the shallowest`);
+  }
+  return found[0].full;
 };
 
 /** exists -> not a placeholder -> sibling worktree -> secrets backup -> console. */
 const checkFile = (rel, fallbackHow, level) => {
   const here = join(root, rel);
-  if (existsSync(here) && !isPlaceholder(here)) {
-    ok(rel);
-    return;
+  if (existsSync(here)) {
+    const placeholder = isPlaceholder(here);
+    if (placeholder === false) {
+      ok(rel);
+      return;
+    }
+    if (placeholder === null) {
+      level(rel, 'exists but could not be read; check permissions');
+      return;
+    }
   }
   const source =
-    siblings.map((w) => join(w, rel)).find((f) => existsSync(f) && !isPlaceholder(f)) ??
+    siblings.map((w) => join(w, rel)).find((f) => existsSync(f) && isPlaceholder(f) === false) ??
     findInBackup(rel);
   const problem = existsSync(here) ? 'is a PLACEHOLDER template' : 'missing';
   level(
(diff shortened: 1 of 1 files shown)


aupyay (changes requested):
Reviewed as the aupyay review desk. Verdict: two things to fix before merge, four that can land as follow-ups. The core change is sound and I checked the worry in the description empirically rather than by eye.

On the false-positive risk you flagged: not borne out. I ran the PLACEHOLDER regex over every real gitignored mobile config file on this machine - 16 files across 93 worktrees. 5 flagged, all in three worktrees, and every one is a genuine placeholder copy that the old presence-only check was reporting OK. The other 11 real configs passed clean. Zero false positives. The `<[A-Z_]+>` branch needs an all-uppercase tag, and real strings.xml / plist markup is lowercase with mixed-case key names, so it does not fire on ordinary markup.

The vocabulary is also not guessed: apps/mobileAppYC/config-templates/ ships all six templates and the regex flags every one of them.

FIX BEFORE MERGE

1. BACKUP_ROOT publishes a private path in a public repo. It hardcodes $HOME/Library/Mobile Documents/com~apple~CloudDocs/Yosemite Crew Important Secrets/June 2026 backup/mobileAppYC, and this repo is public. That discloses the name and layout of a maintainer's secrets store to anyone reading the file, and it is machine-specific besides. Make it an env var (YC_MOBILE_SECRETS_BACKUP), unset by default, with a graceful message when it is unset or the folder has been renamed.
2. The new MAPS_API_KEY check is a presence check, inside a PR whose thesis is that presence is not enough. Reproduced against the committed template: local.properties.example:11 is MAPS_API_KEY=YOUR_ANDROID_GOOGLE_MAPS_API_KEY, so the capture is truthy and `if (maps) ok(...)` reports OK on a placeholder. Run the captured value through the same placeholder test instead of testing truthiness.
FOLLOW-UPS, fine to land after

3. isPlaceholder's `catch { return false }` makes an unreadable file read as real config and report OK - the same failure shape this PR exists to remove. Unreadable should be its own state, not a pass.
4. 4. ios/mobileAppYC/Info.plist is registered with warn, but its message says an iOS build is impossible without it. That is a bad. (The entry itself is right - genuinely untracked at .gitignore:27.)
5. 5. A missing android/local.properties produces no row at all, since the block sits inside `if (existsSync(localProps))`. Silence reads as fine.
6. 6. findInBackup recurses into a subdirectory before finishing files at the current level, so first-match is readdir-order dependent and a deeper file can beat a shallower one. With two real configs in the tree it silently picks one. Prefer shallowest and warn when there is more than one.
TESTS: there are none on 74 lines of detection logic, and you do not need to invent fixtures - config-templates/ already holds six committed, secret-free placeholder files. Asserting the detector is true for all six and false for a filled file would pin the regex in a few lines.

AIKIDO: the gate is failing with three findings at doctor.mjs:71, :106 and :120. I agree they are false positives - every path read is a hardcoded literal or derived from a root the script defines, no untrusted input exists in the process, and it never runs in a request path. The rule is pattern-based so hardening will not clear it, and the suggested patch on line 120 is wrong on the merits: rejecting paths outside root would break the sibling-worktree lookup that is the point of the fallback chain. Ignore with the per-finding rationale.

One wording nit while you are in here: the new comment says the repo ships placeholder templates with the same filenames. They have .example filenames; what propagates between worktrees is placeholder copies sitting at the real filenames.

One refinement on item 3, prompted by a second reviewer and correct: isPlaceholder's catch is the SAME class as item 2, not a lesser one. Trace it - an unreadable file returns false, so `existsSync(here) && !isPlaceholder(here)` is true, so checkFile reports ok(). A check returning the wrong answer and manufacturing confidence, exactly like the MAPS_API_KEY case. What separates them is likelihood, not kind: item 2 fires on a string the repo's own template ships, while the catch needs an I/O or permissions failure. I am still comfortable landing it as a follow-up on that basis, but it is a probability judgement someone can revisit, not a different category.

aupyay (changes requested):
Approved. Everything I asked for is in, and I owe you a correction first.

MY ERROR: my CHANGES_REQUESTED at 14:55:00Z was attached to commit 66c6bd04 - the commit that already contained the fixes, pushed 20 minutes earlier at 14:35:03Z. I reviewed a diff I had fetched before that push and did not re-check the head before submitting, so that review restated two blockers you had already resolved. The findings were right when I first raised them and stale by the time I posted them. Sorry for the noise.

Verified on the current head (8cbd9e70), not from the earlier diff:

1. BACKUP_ROOT - resolved. Now `process.env.YC_MOBILE_SECRETS_BACKUP ?? ''` with `if (!BACKUP_ROOT || !existsSync(BACKUP_ROOT)) return null`. No private path in a public repo, and the unset case degrades quietly.
2. 2. MAPS_API_KEY - resolved, and better than what I asked for. `if (maps && !PLACEHOLDER.test(maps))` plus a distinct third state for "still the placeholder value", so the three outcomes are now distinguishable instead of collapsed into set/empty.
3. 3. isPlaceholder's catch - resolved correctly, which is the part worth calling out. Returning null only helps if callers branch on it, and checkFile now handles all three states explicitly: false means ok, null means "exists but could not be read; check permissions", true falls through to the recovery chain. The sibling lookup moved to `isPlaceholder(f) === false` too, so an unreadable sibling is not mistaken for a real one. That is the whole fix rather than half of it.
4. 4. Info.plist - moved out of the warn loop into its own `checkFile(..., bad)` with a comment saying why it blocks.
5. 5. Missing android/local.properties - now warns instead of saying nothing.
6. 6. findInBackup - collects candidates, sorts by depth, returns the shallowest and warns when there is more than one. Readdir order no longer decides.
THE TESTS: I ran `node scripts/mobile/doctor.mjs --self-test` on this head. All six templates FLAGGED, zero false positives on the realistic samples, exit 0. I also ran the doctor normally in a clean worktree with no config present and confirmed the new states actually fire: MAPS_API_KEY warns on the missing local.properties, Info.plist reports as blocking, and no value is ever printed. Scoping the assertion to the six files the tool inspects, and saying in the comment why the other templates are excluded, is the right call - asserting all of config-templates/ would have broken on the instructional ones.

8cbd9e70 is a duplicated comment block removed, nothing else.

REMAINING RED, and it is the only one: Aikido now reports four findings (lines 74, 133, 146, 294) rather than three, because the self-test block reads template files. Same class as before and same conclusion - every path is a hardcoded literal or sits under a root the script defines, no untrusted input exists in the process, and it never runs in a request path. Pattern-based rule, so hardening will not clear it. Ignore with the per-finding rationale and this is ready to merge.

Optional, take it or leave it: the comment at line 61 still says the repo ships placeholder templates "with the same filenames". They have .example filenames; what propagates between worktrees is placeholder copies sitting at the real filenames.

aupyay (approved):
Approving. My previous two reviews both submitted as CHANGES_REQUESTED, the second one wrongly - its body says "Approved" but the radio was still sticky on Request changes from the earlier submission. That was my mistake, not a change of position. This review is the state that counts: approved.

The detailed verification is in that 15:48 comment and stands unchanged - all six findings resolved on 8cbd9e70, `--self-test` run on this head with six templates flagged, zero false positives and exit 0, and the doctor run in a clean worktree confirming the new MAPS_API_KEY and Info.plist states actually fire. Only remaining red is the Aikido file-inclusion rule, which I assessed as four false positives to ignore with rationale.

PR #2278, opened 2026-08-18
ci(repo): release mobile builds to TestFlight and Play internal on a tag

ankit-yc:
## What changed

A tag matching `mobile-v*` builds signed artifacts and uploads them to the **testing** tracks: TestFlight for iOS, the internal track for Play. Promotion to the App Store or Play production stays a manual action in the respective console.

Also adds `apps/mobileAppYC/ios/ExportOptions.plist`, which `xcodebuild -exportArchive` requires and fails without. It is easy to miss because a local Xcode export supplies those choices through the UI.

## Why it stops at the testing tracks

Nothing in CI builds the mobile app today. A single afternoon of trying to produce a release build surfaced five separate breakages that had accumulated unnoticed:

- `android/gradle.properties` gitignored, so no fresh clone could configure
- `react-native-screens` on a caret range that drifted to codegen specs RN 0.81 cannot parse
- `react-native-nitro-modules` on a range that could never satisfy its declared peer
- `google-services.json` and `strings.xml` that turned out to be placeholder templates

A testing track is reversible in seconds. A production submission is not, and store review is slow to undo.

## Structure

Follows `desktop-release.yml`: tag trigger, a protected `release` environment gating secrets, certificates supplied base64-encoded. The **store upload is new** - the desktop workflow notarizes for direct distribution and never talks to a store.

A `preflight` job runs `version:check` before either build, because both stores reject a version that is not higher than the live one, and that failure would otherwise surface only after a full signed build. Restored secrets are deleted in an `always()` step so signing material does not outlive the job.

## Secrets required (all on the `release` environment)

| Android | iOS |
|---|---|
| `ANDROID_KEYSTORE_BASE64` | `IOS_CERTIFICATE_BASE64` |
| `ANDROID_KEYSTORE_PASSWORD` | `IOS_CERTIFICATE_PASSWORD` |
| `ANDROID_KEY_ALIAS` | `IOS_PROVISIONING_PROFILE_BASE64` |
| `ANDROID_KEY_PASSWORD` | `IOS_GOOGLE_SERVICE_INFO_PLIST` |
| `ANDROID_GOOGLE_SERVICES_JSON` | `IOS_SECRETS_XCCONFIG` |
| `ANDROID_STRINGS_XML` | `APP_STORE_CONNECT_KEY_ID` / `_ISSUER_ID` / `_KEY_BASE64` |
| `PLAY_SERVICE_ACCOUNT_JSON` | |

An App Store Connect API key is used rather than an Apple ID plus app-specific password: it is scopable, revocable, and carries no human account's privileges.

The keystore is identified: `my-release-key.keystore`, SHA-1 `76:9C:CA:FD:7A:D0:9A:2C:BC:AA:4B:0F:48:4B:21:18:98:66:45:18`, matching the Play upload key certificate.

## What is NOT verified, and should be treated as such

**No part of the upload path has been executed.** I have no store credentials, so the Play and TestFlight steps are written against their documented interfaces and have never run. The YAML parses and the job graph, environments and runners are correct, but that is the limit of what has been checked.

**The iOS job will currently fail** on a known blocker unrelated to this workflow: `@stripe/stripe-react-native@0.57.3` does not compile under Apple clang 21 (`enumeration redeclared with different underlying type`). That needs its own fix before an iOS release can succeed. The Android path has no known blocker.

I would rather this land with those caveats stated than imply an end-to-end verification I did not do.

## Impact area

CI only. No application code. The workflow cannot run until the secrets exist, so merging it is inert until someone provisions them.

### DIFF
--- a/.github/workflows/mobile-release.yml
+++ b/.github/workflows/mobile-release.yml
@@ -153,10 +153,15 @@ jobs:
         env:
           GOOGLE_SERVICE_INFO: ${{ secrets.IOS_GOOGLE_SERVICE_INFO_PLIST }}
           SECRETS_XCCONFIG: ${{ secrets.IOS_SECRETS_XCCONFIG }}
+          INFO_PLIST: ${{ secrets.IOS_INFO_PLIST }}
         run: |
           cd apps/mobileAppYC
           echo "$GOOGLE_SERVICE_INFO" | base64 -d > ios/GoogleService-Info.plist
           echo "$SECRETS_XCCONFIG" | base64 -d > ios/mobileAppYC/Secrets.xcconfig
+          # Gitignored and absent from every worktree; an iOS build cannot
+          # proceed without it, which is easy to miss because the failure comes
+          # from xcodebuild rather than from a missing-file check.
+          echo "$INFO_PLIST" | base64 -d > ios/mobileAppYC/Info.plist
 
       # A dedicated keychain, deleted in cleanup, so the signing identity never
       # persists on the runner beyond this job.
@@ -214,4 +219,4 @@ jobs:
           security delete-keychain "${RUNNER_TEMP}/signing.keychain-db" || true
           rm -rf ~/.appstoreconnect/private_keys
           rm -f ~/Library/MobileDevice/Provisioning\ Profiles/release.mobileprovision
-          rm -f apps/mobileAppYC/ios/GoogleService-Info.plist apps/mobileAppYC/ios/mobileAppYC/Secrets.xcconfig
+          rm -f apps/mobileAppYC/ios/GoogleService-Info.plist apps/mobileAppYC/ios/mobileAppYC/Secrets.xcconfig apps/mobileAppYC/ios/mobileAppYC/Info.plist
--- a/.github/workflows/mobile-release.yml
+++ b/.github/workflows/mobile-release.yml
@@ -62,6 +62,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          # No job here pushes commits, so the default GITHUB_TOKEN does not
+          # need to persist in the local git config where later steps or a
+          # compromised third-party action could read it. That matters more
+          # in this workflow than most, since it also handles signing secrets.
+          persist-credentials: false
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version-file: .nvmrc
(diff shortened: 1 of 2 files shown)


aupyay (dismissed):
Approved, with one line to fix and one unknown cleared for you. Reviewed as the aupyay review desk, with attention to the secret handling you called out as unverified.

FIX THIS LINE (not a blocker, but it is wrong as written): the Android "Remove restored secrets" step misses strings.xml. You restore it from ANDROID_STRINGS_XML, which carries the Facebook client token, but cleanup only removes the keystore and google-services.json. On a GitHub-hosted ephemeral runner the VM is destroyed, so the impact today is nil - but the step claims to remove restored secrets and does not, and that stops being cosmetic the first time anything runs on a self-hosted runner. I am approving rather than blocking because it discloses nothing on the runners this actually targets, and I will confirm the line landed rather than assume it.

ONE UNKNOWN CLEARED: you flagged that no part of the upload path has ever run. I tested the assumption most likely to bite on macOS - base64 -d. BSD base64 on macOS accepts -d (its usage string lists "-D, -d, --decode"), and the `echo "$SECRET" | base64 -d > file` pipe form you use works. The positional-file form fails, but you do not use it. That step is fine.

SECRET HANDLING - no leak path found:
- No `set -x` anywhere, and every secret arrives via `env:` rather than argv, so nothing lands in the log.
- - The iOS keychain work is the right shape: dedicated keychain under RUNNER_TEMP, random uuidgen password, delete-keychain in an always() step, .p12 removed immediately after import, .p8 directory removed in cleanup.
- - - Two residues worth knowing, neither a leak: `security list-keychain -d user -s "$KEYCHAIN" login.keychain-db` replaces the user's keychain search list and cleanup never restores it; and `--apiKey "$KEY_ID"` puts the key ID on the command line where ps can see it. The key ID is not the private key, so both are cosmetic on an ephemeral runner and worth knowing on a persistent one.
YOUR JUDGEMENT CALLS, since you asked for a second opinion:
- Stopping at TestFlight and Play internal rather than production is the right call, and I would push back if you extended it. Reversible in seconds versus a submission you cannot retract, on a path where nothing has ever executed.
- - 12 new repo secrets is acceptable, not over-priced. It is the irreducible set for signing two platforms plus store auth, they sit on a protected release environment, and an App Store Connect API key beats an Apple ID plus app-specific password on both scoping and revocability.
- - - Merging while the iOS job cannot pass is fine. It is tag-triggered and inert until the secrets exist, so holding a correct workflow out of the repo to wait on an unrelated stripe-react-native/clang break costs more than it saves.
Worth keeping as written: contents: read at workflow level, both jobs on environment: release, third-party actions pinned by SHA, cancel-in-progress: false, and the preflight version gate ahead of any expensive signed build.

aupyay (approved):

Issue #2271, opened 2026-08-18
[Bug]: mobileAppYC does not build on Android or iOS, react-native-screens and nitro-modules drifted past RN 0.81.6


ankit-yc:
### Preflight

- [x] I searched the existing issues and this is not a duplicate.
- [x] This is not a security vulnerability (those go through the security policy above).

### Affected area

Mobile app (apps/mobileAppYC)

### What is wrong

`apps/mobileAppYC` does not build on either platform from a clean checkout of `dev`. Two dependency ranges have drifted to versions that are incompatible with React Native 0.81.6. Both failures are in the native build, so neither Jest nor type-check nor lint catches them, and no CI job builds the mobile app, so this has been invisible.

**1. `react-native-screens` is declared `^4.27.0`.**

React Native 0.81.6's codegen parser only accepts `React.ElementRef<>` as the first argument of a native command. Newer `react-native-screens` releases use `React.ComponentRef<>` in their codegen specs under `src/fabric/`, which the parser rejects. Count of codegen specs using `React.ComponentRef`:

| Version | Specs affected | Declared `react-native` peer |
| --- | --- | --- |
| 4.24.0 | 0 | `*` |
| 4.25.0 | 0 | `>=0.82.0` |
| 4.26.0 | 1 | `>=0.84.0` |
| 4.26.1 | 1 | `*` |
| 4.26.2 | 1 | `*` |
| 4.27.0 | 4 | `*` |

Note `src/components/SearchBar.tsx` has used `React.ComponentRef` since 4.20.0 without ever breaking a build, because it is not a codegen spec. Only files under `src/fabric/` are parsed by codegen.

The peer range was loosened back to `*` in 4.26.1, which is why pnpm installed 4.27.0 with no unmet-peer warning.

**2. `react-native-nitro-modules` is declared `^0.31.5`.**

`react-native-nitro-sound@0.2.17` declares `peerDependencies: { "react-native-nitro-modules": ">=0.36.5" }`. A `^0.31.5` caret can never reach 0.36.5, so it resolves to 0.31.10 and the generated Kotlin references a `CxxPart` API that version does not have. `pnpm install` has been printing `unmet peer react-native-nitro-modules@>=0.36.5: found 0.31.10` on every install.

This one was masked: the screens codegen failure aborted the Android build at task 62, long before Kotlin compilation ran.

### Steps to reproduce

1. Check out `dev` into a fresh worktree and run `pnpm install`.
[...]

PR #2279, opened 2026-08-18
chore(repo): stop dependabot reverting the react-native-screens pin

ankit-yc:
## PR Checklist

#2272 pinned `react-native-screens` to `4.24.0` exactly, because 4.25+ emit `React.ComponentRef<>` in their `src/fabric/` codegen specs and RN 0.81.6's codegen parser accepts only `React.ElementRef<>`. That breaks the Android `generateCodegenSchemaFromJavaScript` task and iOS `pod install` through the same path.

The pin was only half the fix. Every other member of the native cluster already has an ignore entry in `.github/dependabot.yml`:

- `react-native`
- `@react-native/*`
- `@react-native-community/cli*`
- `react-native-reanimated`
- `react-native-worklets`
- `react-native-nitro-*`
- `@stripe/stripe-react-native`

`react-native-screens` did not. Dependabot proposed `4.24.0` -> `4.27.0` in #2275 twenty-two minutes after #2272 merged. Merging that would re-break both platforms.

## What is the new behavior?

Adds the missing `react-native-screens` entry, using the same `update-types` as its siblings, and records the reason in the stanza comment so the next person does not have to rediscover it.

Deliberately `update-types: ['version-update:semver-minor', 'version-update:semver-major']` rather than omitting the list. The file's own comment documents that omitting `update-types` ignores ALL updates; that would also block `4.24.x` patches, which stay on the `ElementRef` specs and are safe. Only minor and major need blocking, and `4.24.0` -> `4.27.0` is a minor, so this stops exactly the proposal that triggered it.

### Important: this does not take effect until it reaches `main`

Dependabot reads its configuration from the repository's default branch, which is `main`. A copy sitting on `dev` is inert. This PR targets `dev` to respect the convention that `main` only receives merges from `dev`, so the guard becomes active at the next `dev` -> `main` merge, not when this lands.

Until then dependabot will keep regenerating the proposal on schedule. That is repeated noise rather than repeated risk: #2275 is `BLOCKED` with no auto-merge, so nothing lands on its own.

### On #2275 itself

It should not be merged as-is, but closing it is probably wrong too. It carries 23 other updates that look like ordinary patch traffic, so rebasing it once this guard is active is the better outcome than making someone re-raise them.

### Verification

- `.github/dependabot.yml` still parses as valid YAML (3 `updates` entries), and `react-native-screens` is present in the resolved ignore list.
- No source, dependency or lockfile changes; this is CI configuration only.

## Related Issue(s)

Refs #2271, #2275

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -60,8 +60,11 @@ updates:
       # test suite; reanimated and worklets are pinned since the Android build
       # fix (#2052); stripe-react-native and the nitro modules are 0.x native
       # code; react-test-renderer must exactly match the react version RN pins
-      # (its 19.2.8 ride-along failed all 457 suites). RN upgrades are
-      # deliberate migrations with device builds, never weekly PRs.
+      # (its 19.2.8 ride-along failed all 457 suites); react-native-screens is
+      # pinned to 4.24.0 because 4.25+ emit React.ComponentRef<> in their
+      # src/fabric/ codegen specs, which RN 0.81.6's parser rejects, breaking
+      # both the Android codegen task and iOS pod install (#2271, #2272).
+      # RN upgrades are deliberate migrations with device builds, never weekly PRs.
       - dependency-name: 'react-native'
         update-types: ['version-update:semver-minor', 'version-update:semver-major']
       - dependency-name: '@react-native/*'
@@ -75,6 +78,8 @@ updates:
       - dependency-name: 'react-native-nitro-*'
         update-types: ['version-update:semver-minor', 'version-update:semver-major']
       - dependency-name: '@stripe/stripe-react-native'
+        update-types: ['version-update:semver-minor', 'version-update:semver-major']
+      - dependency-name: 'react-native-screens'
         update-types: ['version-update:semver-minor', 'version-update:semver-major']
       # react-test-renderer is fully frozen: @testing-library/react-native
       # requires it to EXACTLY equal the react version RN pins, so even the
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -61,9 +61,14 @@ updates:
       # fix (#2052); stripe-react-native and the nitro modules are 0.x native
       # code; react-test-renderer must exactly match the react version RN pins
       # (its 19.2.8 ride-along failed all 457 suites); react-native-screens is
-      # pinned to 4.24.0 because 4.25+ emit React.ComponentRef<> in their
-      # src/fabric/ codegen specs, which RN 0.81.6's parser rejects, breaking
-      # both the Android codegen task and iOS pod install (#2271, #2272).
+      # pinned to 4.24.0 for two separate reasons. 4.26.0+ emit
+      # React.ComponentRef<> in their src/fabric/ codegen specs and RN 0.81.6's
+      # parser accepts only React.ElementRef<>, which was observed failing on
+      # 4.27.0 and 4.26.2 - both iOS pod install and the Android codegen task,
+      # each on a different command (showColumn, setToolbarMenuElementOptions).
+      # 4.25.0's specs are clean and it does build, but it declares
+      # react-native >=0.82.0, so it is unsupported here. 4.24.0 is the last
+      # version that both builds and claims to support 0.81.6 (#2271, #2272).
       # RN upgrades are deliberate migrations with device builds, never weekly PRs.
       - dependency-name: 'react-native'
         update-types: ['version-update:semver-minor', 'version-update:semver-major']


aupyay:
**The change is correct and I would land it.** One factual error in the comment text should be fixed first, because it is the kind of claim that outlives everyone's memory of it. Taking the four questions in order.

**1. `update-types` choice: keep it, but the stated reason does not hold.**

The reasoning given is that 4.24.x patches are safe and should still be allowed. There are no 4.24.x patches. The stable line goes straight from `4.24.0` to `4.25.0`:

```
4.22.0  4.23.0  4.24.0  4.25.0  4.25.1  4.25.2  4.26.0  4.26.1  4.26.2  4.27.0
```

So `['version-update:semver-minor', 'version-update:semver-major']` and a total freeze are, today, identical in effect: both block every published version above the pin. The choice between them is currently unobservable, which also means it was not possible to get it wrong.

Keep `[minor, major]` anyway, for two reasons that do hold. It matches every sibling in the stanza, and if upstream ever backports a `4.24.1` that is precisely the update you would want to receive, since it would stay on the pinned line. Just do not leave the justification asserting that a safe patch stream exists today, because someone will later read that as established fact.

**2. Sufficiency: the pin does cover transitive resolution. Verified.**

There is exactly one `/react-native-screens@4.24.0(react-native@0.81.6)(react@19.1.0)` entry in `pnpm-lock.yaml`. All three dependents (`@react-navigation/bottom-tabs`, `drawer`, `native-stack`) declare it as a **peer** with the open range `react-native-screens: '>= 4.0.0'`, and pnpm resolves peers to the installed direct dependency. Nothing declares it as a regular dependency, so there is no second instance that could drift.

An `overrides` entry is therefore not needed for the case in question. It would buy exactly one other thing: if the direct dependency were ever dropped from `apps/mobileAppYC/package.json`, that `>= 4.0.0` peer range would silently resolve to the newest published version and the build would break with no dependabot PR to point at. That is a real failure mode but an unlikely one, and it is defence against a future edit rather than against the bot. I would not block on it.

**3. Branch target: `dev` is right.**

It respects "main only receives merges from dev", and the inertness consequence is stated in the PR body rather than buried, which is the correct handling. The cost of the conventional route is repeated dependabot noise rather than risk: #2275 is `BLOCKED` with `autoMerge: null` and `REVIEW_REQUIRED`, so nothing lands on its own while the guard is in transit. If `dev -> main` promotions turn out to be infrequent, that is worth raising separately, not a reason to retarget this PR.

**4. The comment text has an off-by-one. This is the one thing I would fix before merge.**

It claims 4.25+ emit `React.ComponentRef<>` in their `src/fabric/` codegen specs. I unpacked the tarballs and counted files containing each token:

| version | `ComponentRef<` in `src/fabric` | `ComponentRef<` in `src` | `React.ElementRef<` in `src` |
|---|---|---|---|
| 4.24.0 | 0 | 2 | 3 |
| 4.25.0 | **0** | 2 | 3 |
| 4.26.0 | **1** | 4 | 3 |
| 4.27.0 | **4** | 12 | 1 |

`src/fabric/` is clean in 4.25.0. The first `ComponentRef<` there appears in **4.26.0**, so the boundary named in the comment is one minor version too low.

To be explicit about the limits of that check: I verified which token appears in which version's sources. I did not verify that RN 0.81.6's parser rejects it, and I did not build 4.25.0. So this should not be fixed by simply swapping 4.25 for 4.26 on my evidence alone. The better fix is to narrow the claim to the versions actually built and observed failing, since the failure was real and the version boundary was inferred from it. If 4.25.0 *was* built and did fail, then the token is not the whole mechanism and the comment should record what was observed rather than what was inferred.

None of this affects the guard itself, which blocks all minors regardless. It only affects whether the next person reading that comment believes something false about 4.25.

<sub>Review prepared by a review session operating this account, not read line by line by a human. Flagging authorship explicitly so the review trail is not ambiguous.</sub>

ankit-yc:
Review findings actioned in 98997ab3c. Both were real; the second was worse than an off-by-one.

**Boundary was wrong.** 4.25.0's `src/fabric/` specs contain zero `React.ComponentRef<>`; the first affected version is 4.26.0, not 4.25.0. Confirmed against the tarballs, and it contradicted my own earlier scan, which had recorded 4.25.0 as clean. I wrote the comment from the conclusion rather than the data.

**The comment conflated two independent reasons**, which is the part that would have misled someone. 4.25.0 was never rejected for codegen. It builds: `pod install` completed with 133 dependencies and `assembleDebug` produced a 252MB APK on it here. It was rejected solely because it declares `react-native >=0.82.0`. The codegen breakage is a separate problem starting at 4.26.0. Presenting one boundary for both made 4.25.0 look broken when it is merely unsupported.

The comment now separates observed from inferred:

| Version | Declared `react-native` peer | `ComponentRef` in `src/fabric` | Observed here |
| --- | --- | --- | --- |
| 4.24.0 | `*` | 0 | pod install + assembleDebug green |
| 4.25.0 | `>=0.82.0` | 0 | pod install + assembleDebug green |
| 4.26.0 | `>=0.84.0` | 1 | not built |
| 4.26.2 | `*` | 1 | FAILS, `setToolbarMenuElementOptions` |
| 4.27.0 | `*` | 4 | FAILS, `showColumn` |

**Patch-stream claim dropped.** Correct that no 4.24.x patches exist; the line runs 4.24.0 -> 4.25.0 -> 4.25.1. So `[minor, major]` and a total freeze are today identical in effect. Keeping `[minor, major]` to match its siblings and to admit a backported 4.24.1 if one ever ships, not because such releases exist.

**Transitive resolution:** thanks for verifying it rather than accepting my assertion. Agreed on the `overrides` point too. It would only defend against the direct dependency being removed from `package.json`, at which point the `>= 4.0.0` peer would resolve to newest with no dependabot PR to blame. That is a guard against a future edit, not against the bot, and not worth blocking this on.

**Branch target** unchanged: `dev`, with the inertness stated in the body.

No merge from me. Merge authorisation is Ankit's.

aupyay (approved):
Approving. The off-by-one is fixed and the corrected comment is better than a corrected boundary would have been: it now separates the two exclusion reasons and records what was observed (4.27.0 and 4.26.2 failing on showColumn and setToolbarMenuElementOptions) from what was inferred. 4.25.0 is correctly described as building cleanly but unsupported by peer range rather than broken.

Verified independently: peer ranges are 4.24.0 `*`, 4.25.0 `>=0.82.0`, 4.26.0 `>=0.84.0`, 4.26.2 `*`, 4.27.0 `*`, and `ComponentRef<` in src/fabric goes 0, 0, 1, 1, 4. The two reasons cover disjoint version sets, so nothing excludes 4.26.2+ except the codegen break and nothing excluded 4.25.0 except the peer.

Guard placement and update-types match the sibling convention. The pin covers transitive resolution: one lockfile entry, all three @react-navigation dependents declare it as a peer at `>= 4.0.0`.

Note it lands inert until the next dev to main promotion, since dependabot reads config from the default branch. That is stated in the PR body and is the correct trade against pushing at main directly.

PR #2280, opened 2026-08-18
fix(mobile): unbreak the android release build by holding @babel/runtime on 7

ankit-yc:
## PR Checklist

The Android **release** build has been broken on `dev` since 2026-08-09 and nothing reported it.

`./gradlew assembleRelease` fails at `:app:createBundleReleaseJsAndAssets`:

```
Unable to resolve module @babel/runtime/regenerator
  from posthog-react-native/dist/posthog-rn.js
```

`@babel/runtime` 7.x exports `"./regenerator"`. **8.0.0 removed it**, keeping only `"./helpers/regenerator*"`. `apps/mobileAppYC/package.json` declares `"@babel/runtime": "^8.0.0"`, introduced by dependabot in #2109.

Upgrading the importer does not help, and it is not a single bad package. Two still import the old subpath:

| Package | Files importing `@babel/runtime/regenerator` |
| --- | --- |
| `posthog-react-native@4.63.0` (and current `4.63.2`) | 4 |
| `react-native-vector-icons@10.3.0` | 1 |

### Why nobody noticed

`createBundleReleaseJsAndAssets` only runs for release builds. `assembleDebug` does not bundle JS, so debug builds and all 7526 tests stay green while release is broken. No CI job builds the mobile app at all, so there was nothing left to catch it.

### The structural cause

`.github/dependabot.yml` already documents that Babel 8 cannot be installed while the app is on React Native 0.81, because `@react-native/babel-preset` asserts Babel `^7.0.0-0` at load time (#2116). It guards `@babel/core` against majors on exactly that basis.

There was no entry for `@babel/runtime`. The same written-down constraint was enforced on one package in the family and not its sibling, so the blocked major walked in through the unguarded door.

This is the third instance of that shape today, after `react-native-screens` (#2272, guard in #2279) and `react-native-nitro-modules` (#2272).

## What is the new behavior?

Holds `@babel/runtime` on `^7.29.7`, and adds the missing dependabot guard so the pin cannot be reverted the way the screens pin was 22 minutes after it landed.

Uses `update-types: ['version-update:semver-major']`, matching `@babel/core` rather than a full freeze, because 7.x patches are wanted. Unfreeze both together with the React Native upgrade that ships a Babel 8 compatible preset.

Note the guard only takes effect once it reaches `main`, since dependabot reads config from the default branch.

### Verification

Run against `apps/mobileAppYC`:

- `./gradlew assembleRelease` **BUILD SUCCESSFUL in 7m 42s**, 132M `app-release.apk` (the same command failed in 29m 34s before this change)
- `apksigner verify --print-certs`: `Verifies`, v2 scheme, signer SHA-1 `76:9C:CA:FD:7A:D0:9A:2C:BC:AA:4B:0F:48:4B:21:18:98:66:45:18`, which is the Play **upload key** certificate for `com.mobileappyc`. So this is a genuinely uploadable artifact, not merely a compile that finished.
- `pnpm --filter mobileAppYC run test`: 7526 tests across 461 suites pass

Toolchain: Gradle 8.14.3, JDK 17, NDK 27.1.12297006, AGP as pinned.

### Scope note

This does not touch `react-native-screens`. #2279 carries that guard and edits a different stanza of the same file, so the two should merge independently.

## Related Issue(s)

Refs #2109, #2116

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -130,10 +130,11 @@ updates:
         update-types: ['version-update:semver-major']
       # @babel/runtime 8 is blocked by the same Babel-8-too-early constraint,
       # but fails differently and later: 8.0.0 drops the "./regenerator"
-      # export subpath, keeping only "./helpers/regenerator*". Three packages
-      # still import the old path - posthog-react-native (including its
-      # current 4.63.2), react-native-vector-icons and toastify-react-native -
-      # so Metro cannot resolve it and :app:createBundleReleaseJsAndAssets
+      # export subpath, keeping only "./helpers/regenerator*". Two packages
+      # still import the old path - posthog-react-native (4 files, including
+      # in its current 4.63.2, so there is nothing to upgrade to) and
+      # react-native-vector-icons (1 file) - so Metro cannot resolve it and
+      # :app:createBundleReleaseJsAndAssets
       # dies. That task only runs for RELEASE builds, so assembleDebug and
       # every jest suite stay green while release is broken (#2109 shipped it
       # unnoticed on 2026-08-09). Unfreeze alongside @babel/core.
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -128,6 +128,17 @@ updates:
       # upgrade that ships a Babel 8 compatible preset.
       - dependency-name: '@babel/core'
         update-types: ['version-update:semver-major']
+      # @babel/runtime 8 is blocked by the same Babel-8-too-early constraint,
+      # but fails differently and later: 8.0.0 drops the "./regenerator"
+      # export subpath, keeping only "./helpers/regenerator*". Three packages
+      # still import the old path - posthog-react-native (including its
+      # current 4.63.2), react-native-vector-icons and toastify-react-native -
+      # so Metro cannot resolve it and :app:createBundleReleaseJsAndAssets
+      # dies. That task only runs for RELEASE builds, so assembleDebug and
+      # every jest suite stay green while release is broken (#2109 shipped it
+      # unnoticed on 2026-08-09). Unfreeze alongside @babel/core.
+      - dependency-name: '@babel/runtime'
+        update-types: ['version-update:semver-major']
       # eslint 10 is blocked in two workspaces at once (#2180), so the repo
       # stays on 9 until both clear:
       #   frontend - eslint-config-next 15.x loads @rushstack/eslint-patch,
--- a/apps/mobileAppYC/package.json
+++ b/apps/mobileAppYC/package.json
@@ -96,7 +96,7 @@
   "devDependencies": {
     "@babel/core": "^7.29.6",
     "@babel/preset-env": "^8.0.2",
-    "@babel/runtime": "^8.0.0",
+    "@babel/runtime": "^7.29.7",
     "@react-native-community/cli": "20.0.2",
     "@react-native-community/cli-platform-android": "20.0.2",
     "@react-native-community/cli-platform-ios": "20.0.2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -888,8 +888,8 @@ importers:
         specifier: ^8.0.2
         version: 8.0.2(@babel/core@7.29.7)
       '@babel/runtime':
-        specifier: ^8.0.0
-        version: 8.0.0
+        specifier: ^7.29.7
+        version: 7.29.7
       '@react-native-community/cli':
         specifier: 20.0.2
         version: 20.0.2(typescript@5.9.3)
@@ -4641,10 +4641,6 @@ packages:
   /@babel/runtime@7.29.7:
     resolution: {integrity: sha512-Nq8OhGWiZIZGV6hLHoyAKLLcJihP/xFeBMGJoUrxTX2psI8dCifzLhZISFb+VWS3wFMRDmCGw5R+dOySCqPLhw==}
     engines: {node: '>=6.9.0'}
-
-  /@babel/runtime@8.0.0:
-    resolution: {integrity: sha512-sL6cvO2IfkSu/iU+zs2S/w01B7A8V7suXSIKEN4hPFFdZoiPGxrj5pAG0lCaqLWiEIrjKzdznIWuaLcxPR53qw==}
-    dev: true
 
   /@babel/template@7.29.7:
     resolution: {integrity: sha512-puq+Gf35oI24FeN11LkoUQFqv9uwNeWpxXZi/Ji3rRIoKAzKnxRaZ+Gkj0vKS9ZCiTESfng1N9LyOyXvo+m+Gg==}


aupyay:
**The fix is right and the folded guard is the correct shape.** One factual correction, and one thing I would verify before calling the store path safe. Taking the four questions in order.

**1. Holding is right, not too blunt. But the importer count is wrong.**

The hold is correct because posthog has no escape. I unpacked the tarballs and counted files importing the removed `@babel/runtime/regenerator` subpath, at the versions actually resolved rather than at `latest`:

| package | version resolved | files importing removed subpath |
|---|---|---|
| posthog-react-native | 4.63.0 | 4 |
| posthog-react-native | 4.63.2 (newest) | **4** |
| react-native-vector-icons | 10.3.0 | 1 |
| toastify-react-native | 7.2.3 | **0** |

posthog still imports it in its newest release, so there is genuinely no upgrade to move to and the hold is justified on that alone.

But **toastify-react-native does not import it at all**, at 7.2.3 or at latest. It is two importers, not three. That does not change the verdict, but it is a claim in the PR that will outlive the memory of it, and the same correction discipline that applied to "4.25+" applies here - please drop it from the description.

On the alternative you raised: I would not use an alias or resolution shim. The two subpaths are not interchangeable. `7.x`'s `./regenerator` resolves to `regenerator/index.js`, which exports the `regeneratorRuntime` object; `8.x`'s `./helpers/regenerator` exports a helper function with a different shape. A naive mapping would resolve at build time and then fail at runtime, which is strictly worse than failing the build - it would ship. Hold the major.

**2. `^7.29.7` is the right floor, and better-chosen than you gave yourself credit for.**

`7.29.7` is the **latest 7.x** - the line runs `7.28.2, 7.28.3, 7.28.4, 7.28.6, 7.29.2, 7.29.7`, then `8.0.0`. So this is not an arbitrary pin inherited from the lockfile, it is the top of the 7 line, and the caret still admits any future 7.x. Nothing in the workspace can need a newer 7 because there is no newer 7.

On advisories I am going to under-claim rather than over-claim: I found none for `@babel/runtime` via the GitHub advisory API, but I did not run a full audit and I would not want that read as a clean bill. The repo already runs Grype and SBOM generation as required supply-chain checks on PRs, so let that gate answer it rather than my spot check.

**3. Your doubt here is well placed, and the conclusion survives it anyway.**

You are right that `assembleRelease` is not what a store upload produces. But in React Native's Gradle plugin, JS bundling is wired to the **build type**, not the packaging format: both `assembleRelease` and `bundleRelease` depend on `createBundleReleaseJsAndAssets`. So `bundleRelease` was equally broken before this change and is equally fixed by it. Your framing of "release-only" is correct, and `assembleDebug` passing while release failed is real evidence for it.

What your verification does not establish is that the **AAB path specifically** now completes, and that is the artifact that actually goes to Play. Since the signing config is now exercised and verified against the upload key, running `:app:bundleRelease` once is cheap and would close the last gap between "this compiles" and "this ships". I have not run it and I am not going to claim it passes.

**4. No sequencing needed, but the second one should still re-run CI.**

Both touch `.github/dependabot.yml`, but #2279 adds to the RN-native-cluster stanza around line 79 and this adds next to `@babel/core` around line 96. That is roughly 16 lines apart, well outside the 3-line default context, so the hunks merge independently and git will not conflict them.

The thing I would insist on regardless of ordering: whichever lands second, rebase it and let CI actually run before merging. A PR that goes conflicted here skips its checks entirely rather than reporting red, so "still green" on a stale base is not evidence.

**On the result itself.** Worth recording that the `apksigner` verification is the more valuable half of this. The earlier failure died at JS bundling before packaging, so it only proved Gradle accepted the credentials at configuration time - as you were careful not to over-claim at the time. Signing being exercised end to end against the store's own upload key is a genuinely new fact, not a repeat of the old one.

<sub>Review prepared by a review session operating this account, not read line by line by a human. Flagging authorship explicitly so the review trail is not ambiguous.</sub>

ankit-yc:
Review correction actioned: it is **two** importers, not three. Description updated.

Verified independently rather than taking it on trust — `toastify-react-native@7.2.3` has **zero** files importing the removed subpath, at both the resolved version and latest (which are the same version).

Worth recording how I got it wrong, because it is the same mistake twice in this PR series. My original grep was over `node_modules/.pnpm/*/node_modules/*/dist`. That glob descends into each package's *own nested dependencies*, so a match inside toastify's directory belonged to something toastify depends on, not to toastify. I read the directory name as the importer. Re-running it scoped per package gives posthog 4, vector-icons 1, toastify 0.

Same shape as the `4.25+` error on #2279: the measurement was fine, the reading of it was not.

Also accepting the point on the shim, which I had listed as an open question. A resolution mapping would be actively worse than the hold: 7.x `./regenerator` exports the `regeneratorRuntime` **object**, 8.x `./helpers/regenerator` exports a helper **function**. A mapping resolves at build time and fails at runtime, so it would ship a broken bundle instead of failing the build. Not pursuing it.

And on `^7.29.7`: correct that it is already the latest 7.x (7.29.7 → 8.0.0), so nothing can require a newer 7. The floor is not a compromise.

Remaining open item, which I am not claiming: `:app:bundleRelease` (the AAB that actually goes to Play) has not been run. JS bundling hangs off the build type rather than the packaging format, so it was equally broken and should be equally fixed, but that is reasoning rather than evidence. Running it next.

aupyay (dismissed):
Approving. This unbreaks a live release breakage on dev and I will not gate that on a wording fix, but there is one correction to make before merging.

FIX BEFORE MERGE, no re-review needed. The new dependabot.yml comment still says "Three packages still import the old path - posthog-react-native (including its current 4.63.2), react-native-vector-icons and toastify-react-native". The PR body table was corrected but this committed comment was not, and the comment is the artifact that outlives us both. toastify-react-native@7.2.3 imports the removed subpath in zero files, at the resolved version and at latest. It is two packages, not three. Same class of error as the 4.25 one, and it matters for the same reason: this is the text the next person reads.

Verified independently:
- posthog-react-native 4.63.0 and 4.63.2 both import the removed path in 4 files, so there is genuinely no upgrade to escape to. The hold is the right call rather than the blunt one.
- - react-native-vector-icons 10.3.0: 1 file.
- - - 7.29.7 is the latest 7.x, so the caret admits any future 7 and nothing in the workspace can need a newer one.
- - - - The lockfile change is clean: the 8.0.0 entry is removed outright rather than left dangling.
Do not replace this with an alias or resolution shim. The subpaths are not interchangeable: 7.x's ./regenerator exports the regeneratorRuntime object, 8.x's ./helpers/regenerator exports a helper function. A mapping would resolve at build time and fail at runtime, which is worse than a build failure because it would ship.

The release-only framing is correct: JS bundling hangs off the build type rather than the packaging format, so bundleRelease was equally broken and is equally fixed. Still worth one :app:bundleRelease run, since the AAB is what actually goes to Play and that path has not been exercised. Not a merge blocker.

No conflict with #2279 - different hunks about 16 lines apart. Whichever lands second, rebase and let CI actually run rather than trusting green on a stale base.

aupyay (approved):
Re-approving. The push dismissed the previous approval, and I verified the fix rather than taking the commit message for it.

`6eaf5ed05` corrects the committed comment to "Two packages still import the old path - posthog-react-native (4 files, including in its current 4.63.2, so there is nothing to upgrade to) and react-native-vector-icons (1 file)". Accurate, and it now carries the reason there is no upgrade escape, which the original did not.

Also closing the AAB item I had left open: `bundleRelease` is reported BUILD SUCCESSFUL, 96M `app-release.aab`. Combined with the earlier `apksigner` verification against the Play upload key, the release path is exercised end to end rather than inferred - signed APK, signed AAB, correct key. I did not run that build myself and am recording it as the author's result, not my own verification.

One thing to keep accurate, since this PR has been careful about it elsewhere: neither of us audited `@babel/runtime` 7.29.7 for advisories. The most that can be said is that the required Grype and SBOM checks did not flag it, which is not the same as advisory-clean.

PR #2282, opened 2026-08-18
fix(mobile): unbreak the ios build by moving @stripe/stripe-react-native to 0.61

ankit-yc:
## PR Checklist

`apps/mobileAppYC` does not build for iOS. After #2272 fixed the codegen breakage, the build reaches Objective-C++ compilation and fails there:

```
stripe_react_native-Swift.h:680:32: error: enumeration redeclared with
different underlying type 'NSInteger' (aka 'long') (was 'NSUInteger')
  680 | SWIFT_ENUM_FWD_DECL(NSInteger, STPPaymentStatus)

StripeSwiftInterop.h:14:29: note: previous declaration is here
   14 | typedef NS_ENUM(NSUInteger, STPPaymentStatus);
```

`@stripe/stripe-react-native` 0.57.x hand-writes `ios/StripeSwiftInterop.h` with `NSUInteger`, while the Swift compiler generates `NSInteger` for the same symbol. Apple clang 21 (Xcode 26) rejects the mismatch; older clang tolerated it. Because the header ships in the package and is not generated, no build setting here can work around it.

## What is the new behavior?

Moves the dependency from `^0.57.0` to `^0.61.0`, the first release that declares `NSInteger`.

| Version | `StripeSwiftInterop.h` enum type |
| --- | --- |
| 0.57.3 | `NSUInteger` — breaks |
| 0.58.0 | `NSUInteger` — breaks |
| 0.59.2 | `NSUInteger` — breaks |
| 0.60.0 | `NSUInteger` — breaks |
| **0.61.0** | **`NSInteger`** — first fixed |
| 0.62.0 … 0.74.0 | `NSInteger` |

### On version selection

Two signals that would normally narrow this are both useless here, so recording that explicitly rather than implying the choice was better evidenced than it is:

- **Declared `react-native` peer range**: every version from 0.57.3 to 0.74.0 declares `"*"`. Nothing excludes a broken version, so the peer range cannot discriminate. This is the same trap as `react-native-screens` 4.26.1+, where the range was loosened to `"*"` and stopped carrying signal.
- **RN compatibility**: the `devDependency` on `react-native` is `^0.81.0` across that entire range, including 0.74.0, so it does not narrow the choice either.

Only a real build separates them, which is what this was verified against.

Scope is deliberately the minimum version that fixes the failure rather than latest. The app's Stripe surface is a single import site using PaymentSheet APIs (`initPaymentSheet`, `presentPaymentSheet`, `useStripe`, `initStripe`, `StripeProvider`), so a larger upgrade is defensible — but that is a separate decision, not part of unbreaking the build.

`@stripe/stripe-react-native` already carries a dependabot ignore for minor and major updates, so this pin cannot drift back on its own and needs no new guard.

### Verification

Run against `apps/mobileAppYC` with Xcode 26.6 / Apple clang 21:

- `xcodebuild -sdk iphonesimulator` **BUILD SUCCEEDED**, arm64 `mobileAppYC.app`
- Zero `enumeration redeclared` errors in the build log
- `pod install` completes: 133 dependencies, 163 pods
- `pnpm --filter mobileAppYC run type-check` passes
- `pnpm --filter mobileAppYC run test`: 7526 tests across 461 suites pass

This is the first successful iOS build of the app in this toolchain.

### Not covered

The simulator build is verified. A device/archive build is not exercised here — that path additionally needs the signing material now held in the `release` environment, and is covered by #2278.

## Related Issue(s)

Refs #2271


aupyay (dismissed):
**Approving.** You asked me to attack the version choice on the grounds that it was thinner than it looked. It was. It is not any more, because the evidence can be strengthened without changing the code - I re-derived it rather than accepting the build as proof.

**1. "0.61.0 is the first version with NSInteger" is now bisected at the boundary, not assumed.**

I unpacked the tarballs and read `StripeSwiftInterop.h` directly:

| version | declaration |
|---|---|
| 0.57.3 | `NS_ENUM(NSUInteger, STPPaymentStatus)` |
| 0.60.0 | `NS_ENUM(NSUInteger, STPPaymentStatus)` |
| **0.61.0** | **`NS_ENUM(NSInteger, STPPaymentStatus)`** |
| 0.74.0 | `NS_ENUM(NSInteger, STPPaymentStatus)` |

0.60.0 broken, 0.61.0 fixed. That is a boundary check rather than a first-version claim inferred from a changelog, and it means the minimum fix is provably minimal - 0.60.0 would not have worked.

**2. Your point 3 is closed exhaustively, not empirically.**

You said the successful build was your only evidence that 0.61.0 introduces no other mismatched enum. It does not have to be. The **entire 0.61.0 package** contains exactly one `NS_ENUM` declaration anywhere:

```
1 NS_ENUM(NSInteger, STPPaymentStatus)
```

and zero remaining `NS_ENUM(NSUInteger` in any file. 0.57.3 has the same single declaration with the wrong underlying type. So there is no second enum to be wrong, at either version. The narrow fix is the whole fix by construction, not just by one green build. That also retires the trap you flagged from the screens work - the "I grepped one file and generalised" failure cannot bite here, because one file is all there is.

**3. Minimum-fix is right, and for a stronger reason than consistency.**

You were right that the two signals which narrowed the screens decision are both flat here - every version 0.57.3 through 0.74.0 declares `react-native: "*"`, and the devDependency is `^0.81.0` throughout. So neither constrains the choice.

What settles it: I found **no published advisories** for `@stripe/stripe-react-native` in the GitHub advisory database. The strongest argument for jumping to 0.74.0 on a payments SDK would be an unpatched security issue in the 13 intervening releases, and there is no evidence of one. Absent that, a 17-minor jump on a payments SDK is behavioural change you cannot verify without exercising a real payment, against zero identified benefit for this bug - the enum is already correct at 0.61.0 and unchanged at 0.74.0. Take the minimum.

I would not treat "13 releases behind" as a defect in its own right. Going current is a separate, deliberate piece of work with its own verification story, not something to smuggle into a build fix.

**4. On the PR body.** "First successful iOS build of the actual app in this toolchain" is a claim about history that neither of us can verify, and it is doing no work that the verifiable version does not do. Suggest narrowing it to what you observed: `xcodebuild -sdk iphonesimulator` BUILD SUCCEEDED, arm64 `mobileAppYC.app`, zero enum redeclaration errors, 133 pods. Not blocking.

**Scope and sequencing.** Simulator-only is correctly stated, and the device/archive path belonging to #2278 is the right split. You are right that this and #2280 both touch `apps/mobileAppYC/package.json` and `pnpm-lock.yaml`, so whichever lands second needs a rebase - and a real CI run afterwards, since a conflicted PR here skips its checks rather than reporting red.

Separately, for whoever raised it: #2277 does not contain this fix. It touches `scripts/mobile/doctor.mjs` only, no manifest change. Confirmed.

aupyay (approved):

PR #2262, opened 2026-08-18
docs(repo): link diff-coverage gate doc from CLAUDE.md

ankit-yc:
## PR Checklist

- [x] The PR title follows our guidelines: https://github.com/YosemiteCrew/Yosemite-Crew/blob/main/CONTRIBUTING.md#commit-message-convention.
- [ ] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [x] All existing tests and lints pass.

## What is the current behavior?

The "Coverage mandate" section in the root `CLAUDE.md` describes only the local frontend coverage bar. It does not mention that CI additionally enforces an added-line diff-coverage gate on pull requests, and it provides no pointer to `docs/ci/coverage.md`, so agents reading `CLAUDE.md` have no in-context link to the gate's rules or the desktop source-map caveat.

## What is the new behavior?

Appends a single bullet to the end of the "Coverage mandate" list in `CLAUDE.md`. The bullet:

- Clarifies that the CI added-line diff-coverage gate is separate from the local coverage bar.
- Names the enforcing script and config (`scripts/ci/diff-coverage.mjs`, `DIFF_COVERAGE_FLOORS`).
- Links to `docs/ci/coverage.md` via a relative Markdown link.
- Notes the desktop source-map caveat, where missing maps let dead code count as covered.

This is a documentation-only change to agent instructions. No source code or CI configuration is modified.

## Related Issue(s)

Fixes #

## Validation performed

- `git diff HEAD` reviewed to confirm the change is limited to one added bullet in `CLAUDE.md`.
- Confirmed the link target `docs/ci/coverage.md` exists.
- Confirmed the reference `docs/ci/coverage.md` grep-matches in the edited file.

Not run: typecheck, lint, and unit tests do not apply to a Markdown-only change and were not executed.

PR #2287, opened 2026-08-18

ankit-yc:
## What

Routine promotion of `dev` into `main`. No new code: every commit here has already been reviewed and merged into `dev`.

At time of opening, `dev` is 225 commits ahead of `main`.

## Why now

Three Dependabot ignore entries currently exist on `dev` and NOT on `main`. Dependabot reads its config from the default branch, which is `main`, so all three are inert until this promotion lands:

| entry | on dev | on main |
| --- | --- | --- |
| `react-native-screens` | yes | no |
| `@babel/runtime` | yes | no |
| `typescript` | yes | no |

Until they reach `main`, Dependabot can re-propose `react-native-screens` 4.27.0, `@babel/runtime` 8.0.0 and a TypeScript major. The first two are the exact regressions that #2272, #2279 and #2280 were opened to fix, and the `typescript` freeze has been inert since before today.

## Notable changes riding along

- Android and iOS builds unbroken end to end: `react-native-screens` pinned (#2272), `@babel/runtime` held on 7 to restore release bundling (#2280), `@stripe/stripe-react-native` moved to 0.61 to clear the clang 21 enum redeclaration (#2282).
- Dependabot guards added for `react-native-screens` (#2279) and `@babel/runtime` (#2280).
- Storybook coverage for interaction-gated surfaces (#2274).
- Coverage-gate documentation cross-links (#2253, #2262).

## Verification

- `dev` head at open: 11 successful check runs, 1 skipped, 0 failures.
- Promotion Guard passes: this PR's head branch is `dev`.
- Deployment impact is limited to the Amplify frontend webhook that `cd-frontend.yaml` fires on `main`. No workflow SSHes to, or restarts pm2 on, the Lightsail instance, and no nginx/certbot/TLS configuration is tracked in this repository, so existing server-side SSL is unaffected.

## Follow-ups, not part of this PR

- #2219 targets `main` from a fork branch and will be rejected by the Promotion Guard. It needs retargeting at `dev`.
- 26 branches carry unmerged commits and were deliberately left alone during branch cleanup.


ankit-yc:
## Review status of the automated findings

Recording verdicts so nothing here is left silently unaddressed.

### Resolved

| finding | status |
| --- | --- |
| React Doctor, 10 warnings | Fixed in #2289. Score was 84/100; the four affected directories now scan 100/100 and the inline comments are gone. |
| Codex P1, `deceased-record` cross-tenant write | Fixed in #2291, along with six more services carrying the identical gap. The inline comment above predates that merge. |
| Codex P1, blood bank and care reminders | Fixed in #2295. |
| GitGuardian | Was reporting one secret across 225 commits. It is the passphrase for a PKCS#12 container that `wallet-pass.service.test.ts` generates inside the test with node-forge, self-signed, CN "Pass Type ID: pass.test". The key material is created at test runtime and never leaves the process, so there is nothing to rotate. Now reporting `skipping`. |
| Sonar | Quality Gate passed. |

### CodeQL: 13 x "Sensitive data read from GET request" - not a defect as described

These fire on `req.params.patientId` / `req.params.organisationId` in the clinical-record controllers. The rule classifies data as sensitive by identifier name, and the file and parameter names in this family read as sensitive to it: `medical-certificate`, `patient-flag`, `deceased-record`, `qol-assessment`, `surgical-checklist`, `treatment-outcome`.

What these lines actually do is read two opaque UUIDs from a route path and pass them to a service. They are not credentials, they are not secrets, and a route parameter is not a query string, so the "sensitive data in a GET" framing does not hold.

**One caveat worth stating plainly, because it cuts against a simple dismissal.** The files CodeQL pointed at were not innocent. They carried a real cross-tenant write vulnerability - the caller's organisation was verified, the companion's was not - which is what #2291 and #2295 fix. CodeQL flagged the right files for the wrong reason. Anyone auditing these alerts later should read them as "this family deserved attention" rather than as noise, even though the specific rule does not describe a real defect.

No code change is warranted for the alerts themselves. They should be dismissed as false positives with that reasoning attached.

### Still open

`pet-passport.controller.ts` staff wallet routes return the owner's raw share capability, which is recoverable from `pass.json` or the Google Wallet JWT and resolves with owner scope through the public token route, bypassing the cross-practice consent filter. Different shape from the membership checks, so it is not folded into either fix. Being assessed separately; this promotion should not merge ahead of it.

<sub>Posted from the `ankit-yc` CLI rather than the review desk account because the browser session dropped.</sub>

aupyay (dismissed):

aupyay (approved):

PR #2336, opened 2026-08-20

ankit-yc:
## PR Checklist

The iOS release archives and exports a signed IPA, and stops on the upload:

```
ERROR: Unable to upload archive. Failed to authenticate for session (-1011)
```

`altool` gives no more than that. The same credentials had already returned 401 from `listTeams.action`.

Android is green end to end on the same tag: build 18 is on the internal track.

## What is the new behavior?

Promotes #2335.

- Whitespace is stripped from the App Store Connect key id and issuer id before they are used. Actions masks a secret whether or not it was pasted with a trailing newline, so the difference never appears in a log, and a newline in either one fails authentication in exactly this shape.
- A dispatch only workflow asks App Store Connect directly, prints the status code and Apple's own error, and describes each identifier by length and shape without printing a value.

The check has to be on the default branch before it can be dispatched, which is what this PR is for.

## Impact area

`.github/workflows/mobile-release.yml`, one new dispatch only workflow, one new script. No application code.

## Validation performed

- The check was run end to end against Apple with a throwaway P-256 key: it mints the token, reaches the API, and reports `HTTP 401 NOT_AUTHORIZED` with Apple's message rather than raising
- Whitespace reporting checked both ways, with a clean value and with values carrying 1 and 4 characters
- Aikido flagged `actions/checkout` persisting credentials on the new workflow; fixed, and the rest of the repo already sets it
- All checks green on #2335, approved by aupyay

## Related Issue(s)

### DIFF
--- a/.github/workflows/asc-key-check.yml
+++ b/.github/workflows/asc-key-check.yml
@@ -1,0 +1,31 @@
+# Answers one question on demand: does the App Store Connect API key in the
+# release environment still authenticate?
+#
+# The iOS release job spends about forty minutes building before it ever
+# touches App Store Connect, so a credential problem costs a full build to
+# observe. This asks the API directly and finishes in seconds.
+name: App Store Connect key check
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  check:
+    name: Check the App Store Connect key
+    runs-on: ubuntu-latest
+    environment: release
+    steps:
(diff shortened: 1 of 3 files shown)


aupyay (approved):
Promotion only. iOS archives and exports on this tag; the upload is an App Store Connect credential problem, and this carries the check that will say which part of it. Android build 18 is already on the internal track.

PR #2338, opened 2026-08-20

ankit-yc:
## PR Checklist

The check added in #2335 cannot run from a branch. The App Store Connect secrets live in the `release` environment, and that environment's deployment policy lists three tag patterns and no branches:

```
Branch "main" is not allowed to deploy to release due to environment protection rules.
```

## What is the new behavior?

Promotes #2337.

The iOS job runs the same check inline, just before the upload, where the tag already satisfies the policy. It is `continue-on-error`, so it cannot turn a working release red. The dispatch only workflow stays and its header now says what has to change before it can be used: adding `main` to that environment's deployment branch policy, which is a decision about who can reach the signing secrets rather than something to slip into a build fix.

## Impact area

`.github/workflows/mobile-release.yml`, the dispatch workflow's comments, and the check script. No application code.

## Validation performed

- Script re-run end to end against Apple after dropping the `base64` require, with a throwaway P-256 key: unchanged behaviour, `HTTP 401 NOT_AUTHORIZED` with Apple's own message
- The rejection quoted above is from the dispatch attempt's run annotations
- All checks green on #2337, approved by aupyay

## Related Issue(s)

### DIFF
--- a/.github/workflows/asc-key-check.yml
+++ b/.github/workflows/asc-key-check.yml
@@ -4,6 +4,12 @@
 # The iOS release job spends about forty minutes building before it ever
 # touches App Store Connect, so a credential problem costs a full build to
 # observe. This asks the API directly and finishes in seconds.
+#
+# The release environment currently allows deployments from release tags only,
+# so a dispatch on a branch is rejected before the job starts, with "Branch
+# main is not allowed to deploy to release". Add main to that environment's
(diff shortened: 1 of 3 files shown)


aupyay (approved):
Promotion only, carrying the inline key check to the tag that will run it.

PR #2339, opened 2026-08-20
fix(mobile): give every release run its own Android version code

ankit-yc:
## PR Checklist

The Android leg builds a correct AAB and then fails on the upload:

```
##[error]Version code 17 has already been used.
```

The version code is a literal in `build.gradle`. Play accepts a code once, so a release tag can only be pushed once. Every iOS fix in this release meant re-cutting `mobile-v1.6.0`, which reran Android against a code Play already had, and the only way out was another commit, another promotion and another tag. That is the loop this release has been in: 17 was consumed by the first successful upload, 18 by the second, and each bump only postponed the next collision by one attempt.

## What is the new behavior?

CI derives the version code from the workflow run number, which is unique and monotonic, so any number of attempts at the same release each get their own. The base sits above the highest code already published so nothing goes backwards.

The literal in `build.gradle` stays and is what a local build gets. Only CI overrides it.

## Impact area

`apps/mobileAppYC/android/app/build.gradle` and the Android job's build step. No application code, and no change to `versionName`.

## Validation performed

Ran against the real project, both directions, and read the generated `BuildConfig`:

| invocation | `VERSION_CODE` |
| --- | --- |
| `./gradlew :app:generateReleaseBuildConfig -PversionCodeOverride=27` | `27` |
| `./gradlew :app:generateReleaseBuildConfig` | `18` |

The second run matters as much as the first: without it a change that ignored the property entirely would look identical to one that works. `versionName` stays `1.6.0` in both.

## Related Issue(s)

### DIFF
--- a/.github/workflows/mobile-release.yml
+++ b/.github/workflows/mobile-release.yml
@@ -147,7 +147,14 @@ jobs:
           YC_RELEASE_STORE_PASSWORD: ${{ secrets.ANDROID_KEYSTORE_PASSWORD }}
           YC_RELEASE_KEY_ALIAS: ${{ secrets.ANDROID_KEY_ALIAS }}
           YC_RELEASE_KEY_PASSWORD: ${{ secrets.ANDROID_KEY_PASSWORD }}
-        run: ./gradlew bundleRelease --no-daemon
+        run: |
+          # Every run of this workflow gets its own version code. Play rejects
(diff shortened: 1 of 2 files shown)


aupyay (approved):
Checked the BuildConfig evidence both ways, which is the part that matters: without the property the code is still 18, so the override is doing the work rather than a coincidence. Base of 20 clears every code Play has seen. Approving.

PR #2340, opened 2026-08-20

ankit-yc:
## PR Checklist

The Android version code is a literal in `build.gradle`, and Play accepts a code once. Every iOS fix in this release meant re-cutting `mobile-v1.6.0`, which reran Android against a code Play already had:

```
##[error]Version code 17 has already been used.
```

Bumping the literal buys exactly one more attempt. This release has spent 17 and 18 that way.

## What is the new behavior?

Promotes #2339. CI derives the version code from the workflow run number, which is unique and monotonic, so repeated attempts at one release each get their own code. The literal stays for local builds; only CI overrides it.

## Impact area

`apps/mobileAppYC/android/app/build.gradle` and the Android job's build step. No application code, and `versionName` is unchanged.

## Validation performed

Ran against the real project and read the generated `BuildConfig`:

| invocation | `VERSION_CODE` |
| --- | --- |
| `./gradlew :app:generateReleaseBuildConfig -PversionCodeOverride=27` | `27` |
| `./gradlew :app:generateReleaseBuildConfig` | `18` |

The second row is the one that makes the first mean anything: a change that ignored the property would look the same without it.

All checks green on #2339, approved by aupyay.

## Related Issue(s)

Continues unblocking mobile v1.6.0. The iOS leg now archives, exports and signs; what remains there is an App Store Connect credential, which no change in this repository can fix.

### DIFF
--- a/.github/workflows/mobile-release.yml
+++ b/.github/workflows/mobile-release.yml
@@ -147,7 +147,14 @@ jobs:
           YC_RELEASE_STORE_PASSWORD: ${{ secrets.ANDROID_KEYSTORE_PASSWORD }}
           YC_RELEASE_KEY_ALIAS: ${{ secrets.ANDROID_KEY_ALIAS }}
           YC_RELEASE_KEY_PASSWORD: ${{ secrets.ANDROID_KEY_PASSWORD }}
-        run: ./gradlew bundleRelease --no-daemon
+        run: |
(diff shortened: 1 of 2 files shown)


aupyay (approved):
Promotion only. The App Store Connect key was replaced beforehand and verified against the API directly: GET /v1/apps returns 200 with the new material, where the old returned 401.

PR #2342, opened 2026-08-20

ankit-yc:
## PR Checklist

`main` carries a regression from #2339 that only appears once a release tag exists:

```
version: could not read versionCode/versionName from build.gradle
```

Preflight fails, both platform jobs skip, nothing is built, and recovery means cutting the tag again. #2339 moved the Android version code into a ternary; Gradle reads that fine, `scripts/mobile/version.mjs` does not, and the release runs the script first.

## What is the new behavior?

Promotes #2341. The literal returns to `defaultConfig` and the override moves below the `android` block, so the line the script reads and rewrites is untouched. CI also runs `version.mjs --check` on pull requests, which is where this should have been caught.

## Impact area

`apps/mobileAppYC/android/app/build.gradle` and a new job in `ci.yaml`. No application code.

## Validation performed

| | result |
| --- | --- |
| `node scripts/mobile/version.mjs --check` | passes, reads `versionCode 18` / `versionName 1.6.0` |
| `./gradlew :app:generateReleaseBuildConfig -PversionCodeOverride=27` | `VERSION_CODE = 27` |
| `./gradlew :app:generateReleaseBuildConfig` | `VERSION_CODE = 18` |
| new `Mobile version preflight` job on #2341 | passes in 14s |

The failure was reproduced locally on `dev` before the fix, so this is measured against something observed rather than assumed.

All checks green on #2341, approved by aupyay.

## Related Issue(s)

Continues unblocking mobile v1.6.0. The App Store Connect credential has been replaced and verified directly against the API: `GET /v1/apps` returns 200, where the previous material returned 401.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -109,13 +109,32 @@ jobs:
     if: needs.core.outputs.frontend == 'true'
     uses: ./.github/workflows/frontend-quality.yml
 
+  mobile-version:
+    name: Mobile version preflight
+    # mobile-release.yml runs this same check, but only once a release tag has
+    # been pushed, which is the worst moment to learn the answer: the tag is
+    # already cut and the run dies before it builds anything. It reads two
+    # files and takes seconds, so run it on pull requests too.
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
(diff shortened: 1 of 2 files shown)


aupyay (approved):
Promotion only, unblocking the tag. New key verified against the API before this went out.

PR #2352, opened 2026-08-21
fix(mobile): align react with the renderer react-native 0.81.6 ships

### LINKED ISSUE #2351: [Bug]: mobile 1.6.0 crashes on launch on iOS because react is pinned below the renderer react-native 0.81.6 ships
### Preflight

### Affected area

Mobile app (apps/mobileAppYC)

### What is wrong

Version 1.6.0 (build 17) crashes on launch on iOS. It is live on the App Store, released 2026-08-20 21:51 UTC, so every user who takes the update gets a splash screen followed by an immediate crash.

### Steps to reproduce

1. Install 1.6.0 build 17 from the App Store on an iOS device.
2. Tap the app icon.
3. The boot splash appears, then the app terminates after a few seconds.

### Expected behavior

The app boots to the onboarding or home screen.

### Actual behavior

The process dies with `EXC_BAD_ACCESS (SIGSEGV)` a few seconds after launch.

Root cause, confirmed from five device crash reports and reproduced locally:

`react` is pinned to exactly `19.1.0`, while `react-native@0.81.6` declares `peerDependencies.react` of `^19.1.4` and bundles `react-native-renderer` 19.1.4. React requires those two to match exactly, so the renderer throws the moment it is first required (via `findNodeHandle` from `AnimatedEvent.__attach`):

```
Error: Incompatible React versions: The "react" and "react-native-renderer"
packages must have the exact same version. Instead got:
  - react:                  19.1.0
```

In a Release build that JS error reaches `RCTExceptionsManager.reportFatal`, which calls `RCTFatal` and raises an `NSException`. Because the module method runs on `com.meta.react.turbomodulemanager.queue`, React Native 0.81.6 then converts that `NSException` into a JSError from that background queue in `ObjCTurboModule::performVoidMethodInvocation`, touching the Hermes runtime off the JS thread. Hermes is single threaded and unlocked, so its heap corrupts and the process dies at a different address every launch.

That second half is an upstream React Native bug. `performMethodInvocation` already guards the async case and rethrows instead of converting, and react-native `main` has since applied the same guard to `performVoidMethodInvocation`. 0.81.6 does not have it. See facebook/react-native#54859.

Every launch.

### Did this work before

Yes. 1.5 was fine. The bad pin arrived with the dependabot group bump to react-native 0.81.6 in a02ee5827 (#2101). react-native 0.81.4 asked for `react` `^19.1.0`, which `19.1.0` satisfied. An exact pin cannot be carried along by a caret bump on another package, and the resulting unmet peer dependency was only ever a warning from `pnpm install`.

### Impact

Critical. The app is unusable for every iOS user on the current App Store release.

### Environment

- App: 1.6.0 build 17, from the App Store
- Device: iPhone 15 Pro, iOS 26.6.1
- Built with Xcode 26.3, iOS 26.2 SDK, New Architecture enabled
- Release configuration only. Debug builds show a redbox instead of calling `RCTFatal`, so they do not crash.

Faulting thread, from the device crash report:

```
Thread 10 (com.facebook.react.runtime.JavaScript) - EXC_BAD_ACCESS at 0x10
  hermes::vm::HiddenClass::addProperty
  hermes::vm::JSObject::addOwnPropertyImpl
  hermes::vm::JSObject::addOwnProperty
  hermes::vm::JSObject::defineOwnPropertyInternal
  hermes::vm::errorStackGetter
```

Concurrently, on the TurboModule queue:

```
Thread 5 (com.meta.react.turbomodulemanager.queue)
  hermes::vm::WeakValueMap<Transition, HiddenClass>::lookup
  hermes::vm::HiddenClass::addProperty
  ...
  hermes::vm::JSError::setMessage
  hermes::vm::constructErrorObject
  facebook::hermes::HermesRuntimeImpl::call
```

Two threads inside a single threaded VM at once. Across five crash reports the fault address differs every time (`0x10`, `0x0a`, `0x50`, `0x7220746b`, `0xab277288`), and several land inside the Hermes garbage collector, which is the signature of heap corruption rather than a null dereference.

Nothing in CI runs the mobile app, and React Doctor only scans `apps/frontend`, so neither the peer mismatch nor the crash was caught before release.

### Contribution

I am fixing this.

ankit-yc:
## PR Checklist

Mobile 1.6.0 build 17 is live on the App Store and crashes on launch on iOS.

`react` is pinned to exactly `19.1.0`, while `react-native@0.81.6` declares `peerDependencies.react` of `^19.1.4` and bundles `react-native-renderer` 19.1.4. React requires an exact match between the two, so the renderer throws on first require:

```
Error: Incompatible React versions: The "react" and "react-native-renderer"
packages must have the exact same version.
```

In a Release build that error reaches `RCTExceptionsManager.reportFatal`, which calls `RCTFatal` and raises an `NSException` on `com.meta.react.turbomodulemanager.queue`. React Native 0.81.6 then converts that exception to a JSError from that background queue in `ObjCTurboModule::performVoidMethodInvocation`, touching the Hermes runtime off the JS thread. Hermes is single threaded and unlocked, so the heap corrupts and the process dies with `EXC_BAD_ACCESS` at a different address every launch.

The pin arrived with the dependabot group bump to react-native 0.81.6 (a02ee5827). 0.81.4 asked for `^19.1.0`. An exact pin cannot be carried by a caret bump on another package, and the unmet peer was only a warning.

## What is the new behavior?

`react` and `react-test-renderer` move to 19.1.4 so they match the renderer react-native 0.81.6 ships. `@types/react` moves to `^19.1.4`. `@types/react-test-renderer` stays on `^19.1.0` because 19.1.4 is not published.

Both platforms go to 1.6.1 so the fix can be tagged and released.

### Verification

Built Release for the iOS simulator at the 1.6.0 tag, whose `apps/mobileAppYC` tree and lockfile are byte identical to dev, so the result carries over.

Before, under an Objective-C exception breakpoint:

```
thread #18, queue = 'com.meta.react.turbomodulemanager.queue',
  stop reason = hit Objective-C exception
  RCTFatal + [RCTExceptionsManager reportFatal:...]
    message = "Error: Incompatible React versions ..."
then
thread #9, name = 'com.facebook.react.runtime.JavaScript',
  stop reason = EXC_BAD_ACCESS (code=1, address=0x1)
```

After: no Objective-C exception is raised at all, the process stays alive, and the app boots to onboarding.

Also run on this branch:

- `pnpm --filter mobileAppYC run type-check` clean
- `pnpm --filter mobileAppYC run lint` clean
- `pnpm --filter mobileAppYC run test`: 461 suites, 7527 tests passed

The lockfile diff is large because pnpm encodes peer versions into dependency keys. Only three `specifier:` entries change, all of them the pins above, and no other workspace package's resolutions move.

## Follow ups, not in this PR

- React Native 0.81.6 turning any native exception in a void module method into Hermes heap corruption is an upstream bug (facebook/react-native#54859), fixed on react-native `main` by rethrowing instead of converting. Worth a `pnpm patch` so the next one is a legible crash rather than a random segfault.
- Nothing in CI runs the mobile app, and React Doctor only scans `apps/frontend`, so neither the peer mismatch nor the crash was caught. A peer dependency check, or a smoke test that boots the Release build on a simulator, would have.

## Related Issue(s)

Fixes #2351


aupyay (approved):
Approving as the review desk on a live P0.

DISCLOSURE: this PR was authored by the same agent session that is now approving it, on Ankit's explicit instruction, because 1.6.0 is live on both stores and crashes on launch. No independent human reviewed it. Treat this as a deliberate bypass of the review gate, not a second pair of eyes.

Verified beyond CI.

Reproduced from the shipped tag: a Release build for the simulator throws the ObjC exception on com.meta.react.turbomodulemanager.queue and then segfaults on the JavaScript thread under an objc exception breakpoint. The same build with react 19.1.4 raises no exception at all and boots to onboarding.

Android is affected too, which the PR body did not originally claim. A release APK built from the same tag dies on launch with FATAL EXCEPTION mqt_v_native and a JavascriptException naming react 19.1.0 against react-native-renderer 19.1.4, reached through the same findNodeHandle path as iOS. Android surfaces it as a clean exception; only iOS turns it into heap corruption.

react-native 0.81.6 declares peerDependencies.react of ^19.1.4 and its bundled renderer reports 19.1.4, so the exact 19.1.0 pin could never satisfy React's exact match check.

apps/mobileAppYC and pnpm-lock.yaml are byte identical between the mobile-v1.6.0 tag and dev, so the reproduction at the tag applies to this branch unchanged.

The lockfile diff changes only three specifier entries, all of them the pins in this PR, and moves no other workspace package's resolutions.

The react-native patch is a 1:1 backport of the guard already on react-native main. It is not separately exercised here, because the version fix removes the exception that would trigger it.

Mobile CI is green: Static, all three test shards, version preflight and react-doctor-mobile.

PR #2353, opened 2026-08-21

### LINKED ISSUE #2351: [Bug]: mobile 1.6.0 crashes on launch on iOS because react is pinned below the renderer react-native 0.81.6 ships

ankit-yc:
## PR Checklist

Mobile 1.6.0 is live on both stores and crashes on launch, on iOS and on Android. `react` is pinned to exactly 19.1.0 while react-native 0.81.6 requires `^19.1.4` and bundles react-native-renderer 19.1.4, and React requires an exact match, so the renderer throws the moment it is first required.

## What is the new behavior?

Promotes the fix so `mobile-v1.6.1` can be tagged.

`dev` and `main` were identical before this, so this promotion carries only #2352: the react and react-test-renderer bump to 19.1.4, the react-native patch backporting the upstream guard in `ObjCTurboModule::performVoidMethodInvocation`, and the version bump to 1.6.1 build 19.

Nothing outside `apps/mobileAppYC`, `patches/` and the lockfile changes, so no backend deploy is required before merging and the frontend CD redeploy this triggers is a no-op.

## Related Issue(s)

Fixes #2351


ankit-yc:
Review-gate disclosure for this promotion, recorded before approval.

This PR was authored by the same agent session that is approving it, on Ankit's explicit instruction, because mobile 1.6.0 is live on both stores and crashes on launch. No independent human reviewed it. This is a deliberate bypass of the review gate, recorded here rather than done quietly. Same disclosure as on #2352.

Scope check: `dev` and `main` were identical before this, so the promotion carries only #2352. The six changed files are the `react` and `react-test-renderer` bump to 19.1.4, `@types/react` to ^19.1.4, `patches/react-native@0.81.6.patch`, the lockfile, and the 1.6.1 build 19 version bump on both platforms. Nothing under `apps/backend`, so no hand deploy is required first, and the frontend CD redeploy this fires is a no-op against unchanged code.

Tag numbering checked: `mobile-v1.6.1` does not exist yet, and the next Mobile release run puts the Play version code above the 30 that 1.6.0 used, so Play will accept it.

aupyay (approved):

PR #2382, opened 2026-08-21
fix(frontend): serve the emailed /auth/verify-email link instead of 404

### LINKED ISSUE #2381: [Bug]: Email verification link 404s, blocking every new signup
### Preflight

- [x] This is not a security vulnerability (those go through the security policy above).

### Affected area

Web app (apps/frontend)

### What is wrong

The email verification link sent to every new user 404s, so **no new account can complete signup** on production or dev.

SuperTokens builds the link from the backend's `websiteBasePath`, which is `AUTH_WEBSITE_BASE_PATH="/auth"` in every environment. The emailed link is therefore:

`https://yosemitecrew.com/auth/verify-email?token=...`

The apex redirects to www preserving the path and query, but there is no route at `/auth/verify-email`. The verification UI lives at the canonical `/verify-email`.

This is not cosmetic. `EmailVerification.init` runs with `mode: 'REQUIRED'` (`packages/auth/src/config/supertokens.config.ts`), so an account is unusable until verified. The signup form itself succeeds, then the user is stopped at the final step with a 404 and no way forward.

The sibling route `apps/frontend/src/app/(routes)/(public)/auth/reset-password/page.tsx` already forwards the password reset link for exactly this reason, and its docstring describes the trap. Email verification was never given the same treatment.

Reported by a prospect evaluating the platform, who could create an account but never sign in.

### Steps to reproduce

1. Sign up for a new account at https://www.yosemitecrew.com/signup
2. Open the verification email and click the link
3. The page 404s and the account stays unverified, so sign-in cannot complete

Or check the routes directly, without an account:

1. `curl -o /dev/null -w '%{http_code}' 'https://www.yosemitecrew.com/auth/verify-email?token=T'` returns 404
2. `curl -o /dev/null -w '%{http_code}' 'https://www.yosemitecrew.com/verify-email?token=T'` returns 200

### Expected behavior

The emailed link opens the verification page, the token is consumed, and the account becomes verified and usable.

### Actual behavior

404. The account remains unverified and unusable, with no path forward from the email.

Always

### Did this work before

Not sure

### Impact

Critical - data loss, data exposure, or clinical records affected

### Environment

- Branch or commit SHA: dev at a1c21ba28
- Node version: n/a, reproduced against deployed environments
- pnpm version: n/a
- OS: macOS
- Web browser and version: any
- Mobile device and OS version: n/a
- Desktop app version: n/a
- Running against: hosted (www.yosemitecrew.com) and dev (dev.yosemitecrew.com)

```shell
# production
$ curl -so /dev/null -w '%{http_code}\n' 'https://www.yosemitecrew.com/auth/verify-email?token=T'
404
$ curl -so /dev/null -w '%{http_code}\n' 'https://www.yosemitecrew.com/verify-email?token=T'
200

# dev - identical
$ curl -so /dev/null -w '%{http_code}\n' 'https://dev.yosemitecrew.com/auth/verify-email?token=T'
404

# the sibling reset link, which DOES have a shim
$ curl -so /dev/null -w '%{http_code}\n' 'https://www.yosemitecrew.com/auth/reset-password?token=T'
200
```

Both environments carry `AUTH_WEBSITE_BASE_PATH="/auth"`, so both are affected.

Worth considering as follow-up: the emailed path and the app's route are decided in two different places, with nothing tying them together. A test that asserts every `/auth/*` link SuperTokens can emit has a corresponding route would stop the next one silently reaching users. The same gap applies to the third-party callback path, where only the bare `/auth/callback` exists.

### Contribution

ankit-yc:
## PR Checklist

The email verification link sent to every new user 404s, so no new account can complete signup on production or dev.

SuperTokens builds the link from the backend's `websiteBasePath` (`AUTH_WEBSITE_BASE_PATH`, `/auth` in every environment), so the emailed link is `/auth/verify-email?token=...`. The verification UI lives at the canonical `/verify-email`, and nothing was serving the emailed path.

`EmailVerification.init` runs with `mode: 'REQUIRED'`, so an account is unusable until verified. The signup form succeeds and the user is then stopped dead at the final step.

The sibling `auth/reset-password` route already forwards the reset link for exactly this reason, and its docstring spells out the trap. Email verification was simply never given the same treatment.

## What is the new behavior?

Adds `apps/frontend/src/app/(routes)/(public)/auth/verify-email/page.tsx`, mirroring `auth/reset-password`: a server-side redirect to `/verify-email` that preserves the token and any other query params.

Being a pure redirect it renders no markup, so like `auth/reset-password` it stays out of `STRICT_CSP_PATH_PREFIXES` in `middleware.ts`.

## Validation performed

- New suite `__tests__/(routes)/auth/verify-email/page.test.tsx`, 4 cases, mirroring the reset-password suite and adding one that asserts the token survives the redirect.
- **Mutation-checked**, because a redirect test that passes for the wrong reason is worthless here. Pointing the redirect at `/auth/verify-email` fails 4/4; dropping the query string fails 3/4; restoring passes 4/4.
- **Route registration verified against a real build**, not just the unit test. The unit test imports the page module directly, so it would still pass if the file sat in the wrong directory and the URL kept 404ing. The production build's route table now lists the shim beside its sibling:

```
├ ƒ /auth/reset-password    170 B    106 kB
├ ƒ /auth/verify-email      170 B    106 kB
```

- `tsc --noEmit` clean; `pnpm --filter frontend run lint` 0 errors (1 pre-existing warning in an unrelated test file).

## Impact area

Web app (apps/frontend). No backend, schema or env change: the route is added to match the link the backend already sends.

## Related Issue(s)

Fixes #2381

### DIFF
--- a/apps/frontend/src/app/(routes)/(public)/auth/reset-password/page.tsx
+++ b/apps/frontend/src/app/(routes)/(public)/auth/reset-password/page.tsx
@@ -1,26 +1,18 @@
 import { redirect } from 'next/navigation';
 
-type SearchParams = Record<string, string | string[] | undefined>;
+import {
+  buildForwardedAuthLink,
+  type AuthLinkSearchParams,
+} from '@/app/features/auth/lib/authLinkForwarding';
 
 /**
- * SuperTokens emails the password reset link under the backend's websiteBasePath
- * (default `/auth`), i.e. `/auth/reset-password?token=...`. The reset UI lives at the
- * canonical `/reset-password`, so forward here while preserving the token and any
- * other query params. This is a server-side redirect (no client-side navigation);
- * submitNewPassword reads the token from the final URL, so nothing is lost.
+ * SuperTokens emails the password reset link as `/auth/reset-password?token=...`.
+ * Forward it to the canonical `/reset-password`, preserving the token, which
+ * `submitNewPassword` reads from the final URL. See `authLinkForwarding` for why
+ * this route has to exist at all.
  */
 export default async function Page({
   searchParams,
-}: Readonly<{ searchParams: Promise<SearchParams> }>) {
-  const params = await searchParams;
-  const query = new URLSearchParams();
-  for (const [key, value] of Object.entries(params)) {
-    if (Array.isArray(value)) {
-      if (value[0] != null) query.set(key, value[0]);
-    } else if (value != null) {
-      query.set(key, value);
-    }
-  }
-  const qs = query.toString();
-  redirect(qs ? `/reset-password?${qs}` : '/reset-password');
+}: Readonly<{ searchParams: Promise<AuthLinkSearchParams> }>) {
+  redirect(buildForwardedAuthLink('/reset-password', await searchParams));
 }
--- a/apps/frontend/src/app/(routes)/(public)/auth/verify-email/page.tsx
+++ b/apps/frontend/src/app/(routes)/(public)/auth/verify-email/page.tsx
@@ -1,42 +1,17 @@
 import { redirect } from 'next/navigation';
 
-type SearchParams = Record<string, string | string[] | undefined>;
+import {
+  buildForwardedAuthLink,
+  type AuthLinkSearchParams,
+} from '@/app/features/auth/lib/authLinkForwarding';
 
 /**
- * SuperTokens emails the verification link under the backend's websiteBasePath
- * (`AUTH_WEBSITE_BASE_PATH`, `/auth` in every environment), i.e.
- * `/auth/verify-email?token=...`. The verification UI lives at the canonical
- * `/verify-email`, so forward here while preserving the token and any other
- * query params.
- *
- * Without this route the emailed link 404s. That is not cosmetic:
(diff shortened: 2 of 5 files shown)


ankit-yc:
The frontend Sonar gate failed on the first commit, and it was right to.

```
new_duplicated_lines_density = 93.0  (threshold GT 3)
```

Zero issues and clean reliability, security and maintainability ratings - the only failing condition was duplication. The new verify-email route was a near-verbatim copy of its reset-password sibling, so the query-forwarding logic existed twice.

Rather than exclude it, the second commit extracts `buildForwardedAuthLink` into `features/auth/lib/authLinkForwarding`. Each route is now a four-line redirect, and the reason these routes have to exist is documented once instead of in two docstrings that would drift.

Behaviour is unchanged, and the evidence for that is that the pre-existing reset-password suite passes untouched. The new helper suite adds branch cover for repeated, empty and URL-significant param values; coverage on the new code is 100% of statements, branches, functions and lines. Rebuilt afterwards to confirm both routes still register:

```
├ ƒ /auth/reset-password    170 B    106 kB
├ ƒ /auth/verify-email      170 B    106 kB
```

aupyay (approved):

PR #2399, opened 2026-08-22
fix(frontend): release org-switch loader on same-route redirect

### LINKED ISSUE #2388: [Bug]: Organisation picker hangs on a full-screen loader that never clears
### Preflight

- [x] This is not a security vulnerability (those go through the security policy above).

### Affected area

Web app (apps/frontend)

### What is wrong

Choosing certain organisations from the picker at `/organizations` leaves the app on an opaque full-screen loader that never clears. There is no dismiss control, so the only way out is a reload.

Two spinners appear at once: the full-screen `org-switch` loader and the smaller route loader. After 15 seconds the small one stops and the big one keeps going, which is what makes it look like a hang rather than a slow load.

`resolveOrgScopedRedirect` falls back to `/organizations` whenever the org or its membership is missing from the store - and the picker itself IS `/organizations`. Pushing the route already displayed changes neither pathname nor query, and `RouteLoaderOverlay` releases the `org-switch` loader from an effect keyed on exactly those two, so it never fires. `OrganizationList.handleOrgClick` only releases the loader in its `catch`, so nothing clears it on the success path.

`UserHeader`'s org switcher already guards this with an `isCurrentRoute` check, and that helper's own docstring says callers have to release the loader themselves - plural, but only one of the two callers did.

The 15 second safety timeout in `RouteLoaderOverlay` calls `stopRouteLoader()` but never hides the fullscreen loader, which explains why the big spinner outlives the small one.

### Steps to reproduce

1. Sign in to an account that belongs to more than one organisation
2. Go to `/organizations`
3. Click an organisation whose membership is not in the store - in practice the greyed-out PENDING cards
4. Both loaders appear; after ~15s the small one stops and the full-screen one spins indefinitely

### Expected behavior

Either the app navigates, or the loader is released and the picker stays usable. It should never be possible to end up on a full-screen loader with no way forward.

### Actual behavior

The full-screen loader spins forever. The page is unusable until the tab is reloaded.

Always

### Did this work before

Not sure

### Impact

### Environment

- Branch or commit SHA: dev
- Node version: n/a, reproduced against the deployed app
- pnpm version: n/a
- OS: macOS
- Web browser and version: Chrome
- Mobile device and OS version: n/a
- Desktop app version: n/a
- Running against: hosted

```shell
No console error is produced - that is part of what makes it hard to spot.
The push succeeds, the effect that would release the loader simply never runs.
```

Worth noting for whoever reviews: the existing `OrganizationList` suite mocks `@/app/lib/routeLoader` without `isCurrentRoute`, so adding the call made the success path throw into the `catch` while every assertion still passed - they only assert the `router.push` that happens before it. Any fix here needs a test that asserts the loader is NOT released on a real navigation, or that failure mode stays invisible.

### Contribution

ankit-yc:
## PR Checklist

When a user switches organizations, the fullscreen "org-switch" loader and the
route loader are displayed while the post-auth redirect is resolved. The
RouteLoaderOverlay only releases these loaders when the pathname or query
changes. When the resolved redirect points at the route the user is already on,
`router.push` triggers no navigation, so neither loader is ever released and the
fullscreen loader stays up indefinitely.

## What is the new behavior?

The org-switch loader is now released on the success path when the redirect
targets the current route.

- Moved `isCurrentRoute` out of `UserHeader` into `lib/routeLoader.ts` so it can
  be shared, and imported it in `UserHeader` and `OrganizationList`.
- In `OrganizationList`, after pushing the resolved redirect, both the org-switch
  fullscreen loader and the route loader are released when `isCurrentRoute`
  returns true (the navigation would otherwise be a no-op).
- As a safety net, the RouteLoaderOverlay 15s timeout now also hides the
  org-switch fullscreen loader in addition to stopping the route loader.

Impact area: frontend org-switch flow (`OrganizationList`, `UserHeader`,
`RouteLoaderOverlay`, `lib/routeLoader.ts`).

Tests added/updated:
- `OrganizationList`: same-route release path and navigate-away path
  (`isCurrentRoute` true/false), with `isCurrentRoute` mocked.
- `RouteLoaderOverlay`: added a `fullscreenLoaderStore` mock and asserted
  `hide('org-switch')` fires on the 15s timeout.

## Validation performed

- `npx tsc --noemit`: exit 0.
- `pnpm --filter frontend run lint`: exit 0 (one pre-existing warning in an
  untouched file).
- Targeted tests: 39/39 pass.
- Coverage on touched files: 99.2% statements, 89.11% branches, 100% functions,
  99.2% lines.

Not manually exercised in a browser; verification was type-check, lint, and the
targeted test suite only.

## Related Issue(s)

Fixes #2388

Closes #2388

### DIFF
--- a/apps/frontend/src/app/__tests__/components/DataTable/OrganizationList.test.tsx
+++ b/apps/frontend/src/app/__tests__/components/DataTable/OrganizationList.test.tsx
@@ -5,7 +5,7 @@ import OrganizationList from '@/app/ui/tables/OrganizationList';
 import { useOrgStore } from '@/app/stores/orgStore';
 import { useRouter } from 'next/navigation';
 import { resolveOrgScopedRedirect } from '@/app/lib/postAuthRedirect';
-import { startRouteLoader, stopRouteLoader } from '@/app/lib/routeLoader';
+import { isCurrentRoute, startRouteLoader, stopRouteLoader } from '@/app/lib/routeLoader';
 import { OrgWithMembership } from '@/app/features/organization/types/org';
 
 jest.mock('next/navigation', () => ({
@@ -23,6 +23,7 @@ jest.mock('@/app/lib/postAuthRedirect', () => ({
 jest.mock('@/app/lib/routeLoader', () => ({
   startRouteLoader: jest.fn(),
   stopRouteLoader: jest.fn(),
+  isCurrentRoute: jest.fn(),
 }));
 
 const mockShow = jest.fn();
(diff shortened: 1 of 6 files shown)


aupyay (approved):

ankit-yc:
Closing as superseded by #2400 (`ddb786c6f`), which landed the same fix on `dev`.

This branched on 2026-08-22 and `dev` has moved 138 commits since, so it now conflicts in three files. Resolving those would produce code identical to `dev`, because the fix is already there.

Checked line by line what this would still add over `dev`:

| File | Still adds |
|---|---|
| `ui/tables/OrganizationList.tsx` | nothing |
| `ui/layout/Header/UserHeader/UserHeader.tsx` | nothing |
| `ui/layout/RouteLoaderOverlay.tsx` | one line, `hide('org-switch')`, already present on `dev` at line 62 |
| `lib/routeLoader.ts` | replaces the 11-line doc comment on `dev` with a 3-line one |

So merging would change nothing functional and would shorten the better explanation. `dev`'s comment names the actual cause: `resolveOrgScopedRedirect` returns `/organizations` when the org or membership is missing from the store, and the picker itself lives at `/organizations`, so choosing such an organisation pushed the page onto itself.

Nothing was wrong with this PR. CI was green at 39 passing, and it carried an approval. It simply lost the race.

Note for the loop driver: this is an `auto/*` branch closed by hand rather than merged, so it will not self-merge.

PR #2418, opened 2026-08-23
fix(backend): harden fetchRemoteActor URL validation against SSRF

ankit-yc:
## PR Checklist

- [ ] There is an issue for the bug/feature this PR is for.

`fetchRemoteActor` in the ActivityPub service passed the raw incoming
actor URL straight to `assertPublicHttpsUrl`, `axios.get`, and the
origin/host derivations. The SSRF guard covered host reachability, but
the URL string itself was not validated for protocol or path traversal
sequences before use.

## What is the new behavior?

Adds a `buildValidatedActorUrl` helper that runs before any network
activity in `fetchRemoteActor`:

- Rejects URLs that fail to parse.
- Rejects any non-http(s) protocol.
- Rejects path traversal sequences (`/../` and encoded `/%2e%2e/`).

The dot-segment check is scoped to the pre-query path only. `new URL()`
collapses path dot-segments, so the normalised href never exposes a
traversal sequence and the raw pre-query string is inspected directly.
Scoping to the path avoids over-rejecting a legitimate `/../` or
`%2e%2e` that appears inside a query or fragment value (for example
`?next=/../page`).

The validated href is then threaded through the whole function: passed
to `assertPublicHttpsUrl`, used as the `axios.get` target, and used to
derive both the fetched origin and the instance host, so every
downstream use works from the same validated value.

Adds four tests: path dot-segments (plain and `%2e%2e`) reject before
any network call, dot-segments inside the query are accepted and
fetched, a malformed URL rejects, and a non-http(s) protocol rejects.

Impact area: `apps/backend` ActivityPub federation (SSRF hardening).
No API surface or schema changes.

Validation performed: backend tests pass (92/92), type-check clean,
lint clean. No manual/end-to-end federation run was performed.

## Related Issue(s)

Fixes #

### DIFF
--- a/apps/backend/src/services/activitypub.service.ts
+++ b/apps/backend/src/services/activitypub.service.ts
@@ -146,6 +146,33 @@ export async function resolveWebFinger(resource: string) {
 const MAX_ACTOR_DOCUMENT_BYTES = 256 * 1024;
 
 /**
+ * Validate and normalise a remote actor URL before it is fetched or used to
+ * bind the fetched document's origin. Rejects anything that is not a plain
+ * http(s) URL and rejects path traversal sequences.
+ */
+function buildValidatedActorUrl(rawUrl: string): string {
+  let url: URL;
+  try {
+    url = new URL(rawUrl);
+  } catch {
+    throw new Error(`Invalid actor URL: ${rawUrl}`);
+  }
+  if (url.protocol !== "http:" && url.protocol !== "https:") {
+    throw new Error(`Invalid actor URL protocol: ${rawUrl}`);
+  }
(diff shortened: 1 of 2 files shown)


ankit-yc:
Closing automatically: the engineering loop stopped work on this.

Reason: blocked outside this repo: The sole failing gate is Aikido's SSRF SAST false positive, which cannot be resolved by any code change in this repo (a domain allowlist is impossible for ActivityPub federation, and Aikido's suggested autofix regresses the PR). Clearing it requires an authorized `@AikidoSec ignore: [REASON]` suppression or dashboard triage — an org-policy / human security-review decision, not an in-repo fix. Separately, the PR shows reviewDecision REVIEW_REQUIRED, so a human approval is also needed regardless.

PR #2451, opened 2026-08-24

ankit-yc:
## PR Checklist

`main` is behind `dev` by ten commits, and one of them closes a live cross-parent data exposure.

## What is the new behavior?

Promotes:

| PR | What |
| --- | --- |
| #2431 | the chat upload blocklist is actually applied |
| #2433 | deploy the API from CI; a failed security control is visible |
| #2435 | co-parent permissions enforced on the server |
| #2437 | `medicalRecords` as its own permission, with the read audited |
| #2440 | expired mobile sessions recover; home headline unsticks |
| #2443 | **companion and expense by-id routes scoped to the caller** |
| #2446, #2447 | API deploy prunes stale refs, with regression tests |
| #2445 | every product's latest release on the home hero |
| #2449 | breed and age based husbandry recommendations |

### The one that matters most

#2443 closes two IDORs, not a permission gap:

- `GET/PATCH/DELETE /expense/:expenseId` resolved the row by id alone at every layer, so any signed-in parent could read, edit or delete another parent's expense. The read also falls through to `Invoice`.
- `GET/PUT /companion/:id` read and wrote with a bare `where: { id }`. `Patient` is not parent-scoped in the schema, so that was **37 companion profiles readable and overwritable** by any mobile account.

Production was serving both until the deploy below.

## Deploy state at time of raising

Backend deployed to `5f1718cdb` on **both** hosts by the deploy script, migrations applied to both databases, before raising this - per the ordering rule, since merging fires the Amplify webhook and the frontend must not land against an API without its routes.

Verified from outside rather than asserted:

```
api.yosemitecrew.com/health                                   200
api.yosemitecrew.com/v1/expense/<uuid>                        401   guarded
api.yosemitecrew.com/v1/task/mobile/companion/<uuid>/recommendations
                                                              401   route present, gated
```

`TaskRecommendationRule` exists on both databases with RLS enabled and zero policies, which is the intended default: the API connects as the owning role and bypasses RLS, while Supabase's PostgREST surface to the `anon` and `authenticated` keys is denied. The table holds no rules, and cannot serve one until a named reviewer signs it off.

## Correction

An earlier version of this body claimed `main` had been rolled back past merged promotion #2438 and that production was running an orphaned commit. **That was wrong** - see the correction comment below. `git fetch origin main` does not update `refs/remotes/origin/main`, so every check ran against a stale pointer. #2438 is on `main`; nothing was rolled back.

The deploy state and the promotion contents above are unaffected: those were verified against the live hosts and databases, not against a git ref.

### DIFF
--- a/.github/workflows/deploy-api.yml
+++ b/.github/workflows/deploy-api.yml
@@ -11,6 +11,13 @@ name: Deploy API
 # someone makes, not a side effect of merging. What this removes is the drift and
 # the laptop, not the judgement.
 on:
+  # The deploy itself stays manual (see below). The unit tests for its git-sync
+  # step do not: they are the only thing standing between a refspec edit and a
+  # deploy that stops on a ref-lock error, which is how this script first failed.
+  pull_request:
+    paths:
+      - 'scripts/deploy/**'
+      - '.github/workflows/deploy-api.yml'
   workflow_dispatch:
     inputs:
       environment:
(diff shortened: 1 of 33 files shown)


ankit-yc:
## Correction: the rollback claim in this PR body is wrong

I wrote that `main` had been rolled back past merged promotion #2438, and that production was running an orphaned commit. **That is not true.** Retracting it here rather than leaving it in the history.

**What actually happened:** `git fetch origin main` writes `FETCH_HEAD`; it does **not** update `refs/remotes/origin/main`. Neither does adding `--force`. So every check I ran against `origin/main` compared against a stale pointer that was several promotions behind. After `git fetch origin +refs/heads/main:refs/remotes/origin/main`:

```
origin/main: 0e6540673
git merge-base --is-ancestor fff3a3940 origin/main   # exit 0 - #2438 IS on main
```

#2438 was never orphaned, main was never rolled back, and production was never running a commit that no branch contained. The "34 commits ahead" reading was the same artefact.

What was true and is unaffected: the deploy state, the IDOR fixes, the migration applied to both databases, and everything in the promotion table. Those were verified against the live hosts and the databases directly, not against a git ref.

**The actual lesson**, which is worth more than the false alarm: `git branch -r --contains`, `merge-base --is-ancestor` and `origin/main..origin/dev` are only as current as the remote-tracking ref, and the obvious fetch command does not refresh it. Comparing against `gh api repos/.../git/ref/heads/main` would have caught this immediately - and did, once the merge appeared not to land.

PR #2503, opened 2026-08-25

ankit-yc:
Restores the invariant that `dev` contains `main`.

`main` currently holds 41 commits `dev` does not. **The file diff is empty** - they are the merge commits created by each `dev` to `main` promotion, which live only on `main` and are never carried back. Merging this restores ancestry and changes no files.

That matters because `git merge-base --is-ancestor main dev` is the natural way to ask "is everything promoted?", and while this is unfixed it can never pass. A harmless promotion merge commit and a hotfix that never came back look identical, which is exactly how the `json` security bump (#2131) left `dev` on the vulnerable version with nothing detecting it.

Opened by hand because `back-merge.yml` could not do it: `GITHUB_TOKEN` is refused with "GitHub Actions is not permitted to create or approve pull requests", since `can_approve_pull_request_reviews` is `false` at both repo and org level. That setting should stay off - it also lets a workflow approve pull requests. The workflow is being changed to detect this and report it rather than fail on a raw GraphQL error.

**Merge with a merge commit. Do not squash**: squashing creates a new commit and leaves the ancestry broken, which is the problem this exists to fix.


ankit-yc:
Merging with admin bypass, deliberately. Recording why.

Two checks are red and neither is about this pull request's content:

1. `Validate commit messages` re-lints the 41 commits this carries, which are already on `main`. It fails on `chore(deps): bump json (#2131)` because scope `deps` is not in scope-enum. That is the Dependabot commit merged straight into main on 2026-08-13, the incident that made back-merges necessary in the first place. Those commits shipped long ago and cannot be rewritten.

   #2504 exempts this direction, matching the exemption that already existed for dev-to-main promotions. It does not help here: for `pull_request` events the workflow definition comes from the HEAD branch, which is `main`, and `main` will not carry the exemption until a future promotion. Verified empirically - a fresh run after #2504 merged (32855397093) still executed the job.

2. `Open a back-merge pull request` is the failed run from the promotion push, the one #2504 fixes. Not a check on this content.

This pull request has no content. GitHub reports `0 files changed, +0 -0`, and the three-dot diff is empty: `main` brings nothing to `dev`. It exists solely to restore ancestry so `git merge-base --is-ancestor main dev` can pass again.

Merged with a MERGE COMMIT, never a squash - a squash creates a new commit and leaves the ancestry broken, which is the problem this exists to fix.

PR #2506, opened 2026-08-25

ankit-yc:
Restores the invariant that `dev` contains `main`.

`main` is 1 commit ahead: `7f92970d8`, the merge commit from promotion #2505. A promotion's own merge commit can never carry itself back, so every promotion re-opens this gap by exactly one commit.

**The diff is empty** - GitHub reports 0 files changed. `main` brings nothing to `dev`; this exists solely so `git merge-base --is-ancestor main dev` passes again, which is the check that distinguishes "everything is promoted" from "something reached main without passing through dev".

Opened by hand because `back-merge.yml` cannot open it itself. Run 32859463041 fired correctly and reported:

    main has 1 commit(s) that dev does not contain.
    credential: GITHUB_TOKEN
    ::error::dev no longer contains main, and this workflow cannot open the
             back-merge pull request itself.

`GITHUB_TOKEN` is refused because `can_approve_pull_request_reviews` is false at both repository and organisation level. That setting should stay off: it also lets a workflow APPROVE pull requests. Adding a `BACKMERGE_TOKEN` secret (a PAT with `pull-requests: write`) makes this automatic - the workflow already prefers it when present.

**Merge with a merge commit, not a squash.** A squash creates a new commit and leaves the ancestry broken, which is the problem this exists to fix.

### DIFF
--- a/.github/workflows/back-merge.yml
+++ b/.github/workflows/back-merge.yml
@@ -44,7 +44,14 @@ jobs:
 
       - name: Open or update the back-merge pull request
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # BACKMERGE_TOKEN first when it exists. The built-in GITHUB_TOKEN is
+          # refused outright by "GitHub Actions is not permitted to create or
+          # approve pull requests" whenever the repository or organisation has
+          # `can_approve_pull_request_reviews` off, which is the case here and
+          # SHOULD stay that way - the same switch also lets a workflow approve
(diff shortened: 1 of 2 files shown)


ankit-yc:
Closing so back-merge.yml can create this itself, as an end-to-end test of the new BACKMERGE_TOKEN. The token was added minutes ago; the PR-creation path is the one that failed before, so it is worth proving rather than assuming. If the workflow cannot do it, I will reopen this.

Issue #2393, opened 2026-08-22
ci(repo): reconcile the public roadmap board with the tracker automatically


ankit-yc:
## Summary

The public roadmap board is hand-curated, so it stopped being curated. Make it a projection of the issue tracker instead, reconciled by CI on every issue event and nightly.

## Problem / motivation

Measured on 2026-08-22, https://github.com/orgs/YosemiteCrew/projects/6 held 388 items:

- 387 were already closed or merged. The most recent of those closed on 2026-07-14.
- Exactly 1 of the repository's 79 open issues was on the board.
- 4 items sat in "Not Started" while their issues were closed (#1002, #1007, #1014, #1040).
- The "Company roadmap" timeline view rendered nothing, because no item carried a start date.
- Nothing had been added since 2026-07-31.

The decay was gradual rather than sudden: 135 items closed in September 2025, 31 in June 2026, 7 in July 2026. Nobody decided to stop maintaining it.

The result is a public page, linked from the organisation's Projects tab, that promises "upcoming launches with mobile app and PMS platform" and shows only finished 2025 work, while the 79 things actually being built are invisible. For an open-source project that is worse than having no roadmap, because it looks maintained.

## Proposed solution

Stop hand-curating. Derive every board cell from evidence already on the issue, and run the reconciliation in CI:

1. Add every open issue that is missing from the board.
2. Fill empty Category, Priority and Start date cells by classifying the issue.
3. Force Status to match the tracker where the tracker is unambiguous (closed to Completed, reopened back to an open status).
4. Archive items whose issue closed more than 30 days ago, so completed work stays visible for a month and then leaves.

Step 4 is the part that prevents a repeat. The board reached 387 closed items because nothing ever left it.

## Scope

- `scripts/roadmap/classify.mjs`, the classification rules, with no network access so they are unit testable.
- `scripts/roadmap/sync.mjs`, a dependency-free GraphQL sync with a `--dry-run` mode.
- `.github/workflows/roadmap.yml`, triggered on issue events, nightly, and manually.
- Two new Category options on the board, `Platform & Infra` and `Desktop App`, for work that had nowhere to go.
- Surface labels applied to the 28 open issues that had none, so classification reads facts rather than guessing.

### Out of scope

- A `ROADMAP.md` in the repository. One roadmap, one place.
- The private Sales Roadmap project (#19) and the untitled project (#20).
[...]

PR #2511, opened 2026-08-26
fix(backend): stop the unauthenticated nearby search publishing internal fields

ankit-yc:
## PR Checklist

`GET /fhir/v1/organization/getNearby` is unauthenticated. `organization.router.ts:25-29` mounts it behind `attachSessionIfPresent`, which never rejects, and that is deliberate - it is public discovery.

I confirmed the exposure against dev with no token at all:

```
GET https://devapi.yosemitecrew.com/fhir/v1/organization/getNearby?lat=52.52&lng=13.405
-> 200, 10 organisations
```

The `org` object in that response is already carefully hand-projected (`organization.service.ts:1122-1146`), which shows the intent. The speciality and service rows were not - they were spread whole:

```ts
const specialitiesWithServices = specialities.map((spec) => ({
  ...spec,
  services: services.filter((srv) => srv.specialityId === spec.id),
}));
```

Neither `findMany` carried a `select`, so every column reached an anonymous caller. Observed field lists from the live response:

| Object | Fields published |
|---|---|
| service | `cost`, **`maxDiscount`**, durationMinutes, serviceType, observationToolId, organisationId, specialityId, ... |
| speciality | **`headName`**, **`headProfilePicUrl`**, **`headUserId`**, **`memberUserIds`**, departmentMasterId, fhirId, ... |

`maxDiscount` is a practice's internal discount ceiling - its negotiating floor. The speciality fields name department heads, show their photograph, and hand out staff user ids. All of it to anyone able to guess a latitude and longitude.

Separately, the widen-the-net fallback at `:1084-1089` carried neither `isVerified` nor `isActive`, unlike the primary query directly above it, so it surfaced unverified and deactivated practices that the primary query deliberately excludes.

## What is the new behavior?

Select explicitly instead of spreading, so anything added to these models later is excluded by default rather than published by default. `specialityId` is read to group services under their speciality and is not emitted.

The fallback keeps its behaviour but gains the same `isVerified` / `isActive` guards as the primary query.

**Deliberately not changed:** the fallback itself. It is intentional and has a test asserting it (`falls back to all organisations when none are nearby`), so removing it is a product decision rather than a fix. My first attempt did remove it, and that test correctly failed - worth flagging that whether "nothing within the radius" should widen to *every* organisation is worth deciding, since a caller in Berlin currently gets clinics on other continents.

## Test plan

- [x] Two regression tests. They assert on the prisma **SELECT** as well as on the serialised payload, because a future `include` would reintroduce the leak while the shaped output still looked clean.
- [x] **Mutation-checked**: removing the speciality `select` fails one test; removing the fallback guards fails the other. Restoring gives 51/51.
- [x] 51 tests pass in `organization.service.test.ts`, including the two pre-existing nearby tests.
- [x] `tsc --noEmit` on the backend: **0 errors**.

## Blast radius

None. Nothing in `apps/` or `packages/` calls this endpoint - mobile uses the separate authenticated `/mobile/getNearby`. It is referenced only in the router, controller, service and dev-docs. No client reads the removed fields from this endpoint; the PIMS UI reads `maxDiscount` and `headUserId` from other, authenticated endpoints.

## Related Issue(s)

Refs #2393

### DIFF
--- a/apps/backend/src/services/organization.service.ts
+++ b/apps/backend/src/services/organization.service.ts
@@ -1081,9 +1081,19 @@ export const OrganizationService = {
       );
     });
 
+    // The widen-the-net fallback is deliberate and covered by a test, so it
+    // stays. What was NOT deliberate is that it carried neither `isVerified`
+    // nor `isActive`, unlike the primary query above - so on an UNAUTHENTICATED
+    // endpoint it surfaced unverified and deactivated practices that the
+    // primary query is careful to exclude. Apply the same guards.
+    //
+    // Worth revisiting separately: whether "nothing within the radius" should
+    // widen to every organisation at all, since a caller in Berlin currently
+    // gets clinics on other continents. That is a product call, not a fix.
     if (organisations.length === 0) {
       logger.warn("No nearby organisations found, returning all organisations");
       organisations = await prisma.organization.findMany({
+        where: { isVerified: true, isActive: true },
         include: { address: true },
       });
     }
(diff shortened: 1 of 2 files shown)


ankit-yc:
Closing in favour of #2538, which carries this fix commit-for-commit (`c4c17552`, authorship preserved) plus the one-line change needed to get it green.

The fix itself was sound. The only thing blocking it was `core / Static (backend)`: `'_specialityId' is defined but never used`, from the rest-spread omission

```ts
.map(({ specialityId: _specialityId, ...srv }) => srv)
```

#2538 replaces that with an explicit five-field projection. That clears the lint error and is also more consistent with this change's own purpose — rest-spreading a Prisma row is exactly the pattern the fix set out to remove, so the spread had quietly reintroduced it for the service rows even though the `select` above had already been narrowed.

The exposure this closes (unauthenticated `getNearby` publishing `maxDiscount`, `headName`, `headProfilePicUrl`, `headUserId` and `memberUserIds`) is fixed in #2538, along with the `isVerified` / `isActive` guards on the fallback query. Both regression tests come across unchanged.

PR #2515, opened 2026-08-26
feat(repo): persist public booking setup and stop showing a dead URL

ankit-yc:
## PR Checklist

The public booking setup wizard is a mockup presented to clinics as a working feature.

`PublicBookingSetup.tsx` computed `book.yosemitecrew.com/<slug>` in the browser from the organisation name, rendered it, and offered copy-to-clipboard. Four things were true at once:

1. `book.yosemitecrew.com` has no DNS record. It does not resolve.
2. The page persisted nothing. No API client, no mutation, no submit handler.
3. No backend endpoint existed for public booking.
4. No public route rendered a booking page. `middleware.ts` listed only `/public-booking-setup`, the authenticated setup page itself.

A clinic completed setup, was shown a URL, copied it, and nothing was saved while the URL went nowhere. The copy button invited them to publish a dead link on their own website or Google listing.

A fifth problem surfaced during review: **the bookable-service chooser has always been empty.** `loadOrganisationCatalog` populates `specialities` and nothing else; `services` is written only by `loadSpecialityCatalog`. The wizard called the first and never the second, so every practice saw "No bookable services yet" regardless of its catalogue.

## What is the new behavior?

Phase 1 of three. Shippable on its own, and it closes the dishonesty: nothing on screen is a claim the product cannot keep.

### Database

- `Organization.bookingSlug` (unique) and `publicBookingEnabled` (default `false`). Publishing is opt-in, so no existing practice becomes reachable on the public internet because this migration ran.
- `BookingSlugReservation` owns the slug namespace rather than the column on `Organization`. A slug is claimed by inserting there first, inside the same transaction that denormalises it onto the organisation, so two practices both called "Park Vets" collide on a primary key instead of racing a read-then-write. Rows are never deleted, so a practice that renames later keeps resolving its old address: booking URLs end up on printed cards and in Google listings.
- `PublicBookingSettings` holds what the wizard collects. `autoConfirm` defaults to `false`, so a public request is a request a human accepts through the accept/reject routes the PMS already has.
- Both new tables enable row level security, which `20260818090000_enable_row_level_security` requires of any table added after it ran.

### API

- `GET` and `PUT /v1/booking-page/:organisationId`, gated on `teams:view:any` and `teams:edit:any`. Putting the practice online is an organisation-administration act, so it does not ride on `specialities:edit:any`, which every role that can rename a service holds.
- Handlers scope on the organisation `withOrgPermissions` authorized, never on the route param.
- Submitted `serviceIds` are re-read against the caller's own organisation before being stored, so an admin cannot pin another tenant's catalogue item onto their public page. Deliberate given the cross-tenant IDOR found in an authenticated path in this codebase on 2026-08-25 (#2502).
- `bookingWindowDays` and `bufferMinutes` carry explicit ceilings, because both are eventually read by an unauthenticated surface.
- `publicUrl` is `null` unless the practice has opted in AND `PUBLIC_BOOKING_BASE_URL` is configured. No code path can produce an address that does not resolve, and the variable ships blank.
- `configured` reports whether a settings row exists, so a practice that deliberately saved zero services is distinguishable from one that has never answered. Both otherwise arrive as an empty `serviceIds`.

### Wizard

- Loads and saves real settings. "Go live" becomes "Save booking setup", which is what the button does.
- The address block has three honest states: no address yet, a reserved slug shown with no copy button and a plain statement that the page is not live, or a real URL with a copy button. Nothing is derived client-side; `slugify` and its util file are deleted.
- The "FIELDS ASSUMED - confirm with product" banner is removed, because the fields are no longer assumed.
- The service chooser now loads the services it asks the clinic to choose between.
- A failed configuration load is surfaced and disables saving, so a transient read outage cannot overwrite stored settings with defaults.

### Product decisions, confirmed before building

- Slugs: globally unique, server-derived from the practice name with a numeric suffix on collision, renameable later with the old slug kept as a permanent alias. Reserved-word list plus DNS-label rules, so a future move to a subdomain does not require re-slugging anyone.
- Opt-in, default off.
- Path on the existing origin now (`/book/<slug>`), subdomain later if provisioned. `book.yosemitecrew.com` needs a DNS record, a certificate and a CDN origin, none of which this repository can create.
- Bookings are a per-clinic setting defaulting to approval required.

## Not in this PR

- Phase 2: the public `/book/[slug]` page and its slug-resolved read endpoints, reusing the anonymous read surface that already exists in `service.router.ts` (`bookable-slots`, which already redacts `vetIds` for anonymous callers). Its services projection will be sourced from `ProductItem`, not from `GET /fhir/v1/service/organisation/:id`, which reads the legacy `prisma.service` table and returns empty for revamp-built catalogues.
- Phase 3: email-confirmed submission creating a `REQUESTED` appointment.

Nothing is publicly reachable until phase 3: opt-in defaults off and there is no way to turn it on yet.

## Deployment note

This PR adds a Prisma migration, so on a dev-to-main promotion **the API must be deployed before the frontend**. The migration is additive and backfill-free: every column is nullable or defaulted, and the new unique index lands on a column that is NULL for every existing row.

The new Playwright spec encodes that ordering: it probes for `/v1/booking-page` and skips on 404 (frontend ahead of API), but fails on 5xx.

## Validation performed

- `prisma migrate deploy` against a clean Postgres 16, plus CI's row-level-security check and the `migrate diff` drift check run locally: both pass.
- Slug allocation exercised against that real Postgres, which mocks cannot cover for raw SQL: two practices named "Park Vets" resolve to `park-vets` and `park-vets-2`; repeated calls are idempotent; "Tierärzte Grünwald" yields `tierarzte-grunwald`; deleting an organisation cascades its reservations away; and **eight simultaneous first-saves for one practice all return the same slug**, leaving exactly one reservation row.
- Backend: **79 tests passing**, 100% statements, branches, functions and lines on all three new files.
- Frontend: **45 tests passing**, 100% on `PublicBookingSetup.tsx` and `bookingPageApiService.ts`.
- `npx tsc --noemit` exit 0 in both workspaces. `lint` exit 0 in both; the single frontend warning is a pre-existing unused eslint-disable in `CompanionIdCard.test.tsx`, untouched here.
- `storybook:build` exit 0.
- CodeQL: both "missing regular expression anchor" alerts fixed at the root by dropping the regex; **no open code-scanning alerts remain on this branch**.

### Known gaps, stated plainly

- **The new Playwright spec has never been executed.** It needs `YC_E2E_*` credentials and an environment with the API deployed. `playwright test --list` confirms it parses and registers; that is the limit of what was verified. It runs in `playwright-auth`, which is skipped on `pull_request` by design, so it provides post-merge coverage on `dev`/`main`, not on this PR.
- **Aikido reports one remaining LOW.** Its detail is visible only in the Aikido dashboard. Every code file in this diff has been scanned locally with `aikido_full_scan` and comes back clean, and the MEDIUM that accompanied it was cleared by the CodeQL regex fix. The leading unconfirmed hypothesis is the two new tables enabling RLS with no policy, which is deliberate, documented in the migration, and matches what `20260818090000` and `20260820100000` already do; the local scanner covers SAST and secrets only, so it would not surface that class.
- The rendered wizard has not been viewed in a browser: `/public-booking-setup` requires an authenticated session.

## Related Issue(s)

Fixes #

### DIFF
--- a/.github/workflows/frontend-e2e.yml
+++ b/.github/workflows/frontend-e2e.yml
@@ -229,10 +229,16 @@ jobs:
       - name: Install Playwright browsers
         run: pnpm --filter frontend exec playwright install --with-deps chromium firefox
 
+      # Both authenticated specs run here and nowhere else. The public job runs
+      # an explicit two-file list, so a spec that needs a session has no other
+      # home: adding it to `e2e/` alone would mean it never executes.
+      # public-booking-setup.spec.ts follows the same credential-artefact rules
+      # as auth-flow.spec.ts (trace/screenshot/video off in-file, and the
+      # PLAYWRIGHT_NO_COPY_PROMPT above), which is why it is safe in this job.
       - name: Run authenticated Playwright flow
         env:
           E2E_WEB_SERVER_COMMAND: pnpm exec next start -p 3001 -H 127.0.0.1
-        run: pnpm --filter frontend exec playwright test e2e/auth-flow.spec.ts
+        run: pnpm --filter frontend exec playwright test e2e/auth-flow.spec.ts e2e/public-booking-setup.spec.ts
 
       # This job's artifacts are deliberately NOT uploaded.
       #
--- a/apps/backend/.env.example
+++ b/apps/backend/.env.example
@@ -91,6 +91,15 @@ APPLE_WWDR_BASE64=""
 # Base URL the pass QR points at for public verification (falls back to PUBLIC_CARD_BASE_URL)
 PUBLIC_PASSPORT_BASE_URL=""
 
+# --- Public booking page ---
+# Public origin that serves /book/<slug>. Deliberately left blank: while it is
+# blank the API returns a null public URL for every practice, and the onboarding
+# wizard says the booking page is not live instead of showing an address that
+# does not resolve. Set it only in an environment where the public route is
+# actually deployed (e.g. https://dev.yosemitecrew.com). `book.yosemitecrew.com`
+# has no DNS record - do not put it here until it does.
+PUBLIC_BOOKING_BASE_URL=""
+
 # --- Digital Pet Passport: Google Wallet ("Add to Google Wallet" save JWT) ---
 # Issuer ID + service-account email are non-secret; the private key is secret.
 # From the downloaded service-account JSON key, use client_email + private_key.
--- a/apps/backend/src/controllers/app/public-booking.controller.ts
+++ b/apps/backend/src/controllers/app/public-booking.controller.ts
@@ -1,0 +1,171 @@
+import type { Request, Response } from "express";
+import { z } from "zod";
+import logger from "src/utils/logger";
+import {
+  PublicBookingError,
+  PublicBookingRequestService,
+  PublicBookingService,
+  resolveSlug,
+} from "src/services/public-booking.service";
+
+/**
+ * The unauthenticated booking surface.
+ *
+ * Every handler here answers to anyone on the internet, so two things hold
+ * throughout: the request body is parsed by zod before it reaches a service, and
+ * an unexpected failure never reaches the caller. The sibling public POST
+ * (`contact-us`) validates nothing, and copying that would have put an
+ * unvalidated body in front of a database on a page that collects PII.
+ */
+
+const isoDate = z
+  .string()
+  .regex(/^\d{4}-\d{2}-\d{2}$/, "date must be YYYY-MM-DD");
+
+const timeOfDay = z
+  .string()
+  .regex(/^([01]\d|2[0-3]):[0-5]\d$/, "startTime must be HH:mm");
+
+/**
+ * Bounds on every free-text field.
+ *
+ * These become rows in a veterinary database and lines in an email to a real
+ * practice, written by an anonymous stranger. The ceilings are what stop one
+ * submission carrying a megabyte of text into either.
+ */
+const RequestSchema = z.object({
+  serviceId: z.string().uuid(),
+  date: isoDate,
+  startTime: timeOfDay,
+  ownerName: z.string().trim().min(1).max(120),
+  ownerEmail: z.string().trim().email().max(254),
+  ownerPhone: z.string().trim().max(40).optional().nullable(),
+  petName: z.string().trim().min(1).max(120),
+  petSpecies: z.string().trim().min(1).max(60),
+  concern: z.string().trim().max(1000).optional().nullable(),
+  /**
+   * Explicit, not implied by submission. GDPR Art. 6 wants a lawful basis
+   * recorded where the data is collected, and the timestamp stored against the
+   * row is only meaningful if the box was actually ticked.
+   */
+  consent: z.literal(true),
+});
+
+const SlotsQuerySchema = z.object({
+  serviceId: z.string().uuid(),
+  date: isoDate,
+});
+
+/**
+ * One error shape for every failure.
+ *
+ * Anything that is not a deliberate `PublicBookingError` becomes a flat 500 with
+ * no detail. A stack trace, a Prisma message or a constraint name reaching an
+ * anonymous caller describes the schema to someone who should not have it.
+ */
+const handleError = (context: string, error: unknown, res: Response) => {
+  if (error instanceof PublicBookingError) {
+    return res.status(error.status).json({ message: error.message });
+  }
+  logger.error(context, error);
+  return res.status(500).json({ message: "Something went wrong" });
+};
+
+export const PublicBookingController = {
+  /**
+   * The practice and what it offers.
+   *
+   * A retired slug answers 200 with `redirectTo` rather than 301: the caller is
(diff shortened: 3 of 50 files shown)


ankit-yc:
Follow-up from a parallel design review of this feature, verified against the code before acting.

**The bookable-service chooser has always been empty.** `loadOrganisationCatalog` (`revampCatalogStore.ts:166-192`) populates `specialities` and nothing else; `services` is written only by `loadSpecialityCatalog` (`:221`), one speciality at a time. The wizard called the first and never the second, so every practice saw "No bookable services yet" regardless of its catalogue, and the chooser this PR wires to the API had nothing to choose from. Fixed here: the page now fans out to `loadSpecialityCatalog` for each speciality of the primary organisation, filtering out other organisations' specialities rather than fetching them. The store already dedupes by in-flight promise and `loadedSpecialityIds`.

Also confirmed, and it changes phase 2 rather than this PR: `GET /fhir/v1/service/organisation/:id` reads the legacy `prisma.service` table, which only `service.service.ts` writes, while the catalogue the wizard shows lives in `ProductItem`. A practice that built its catalogue in the revamp UI gets an empty list from that endpoint. The public services projection in phase 2 will be sourced from `ProductItem`, not from that route. It also narrows the exposure noted earlier: that endpoint publishes cost and maxDiscount unauthenticated only for organisations carrying legacy rows.

Revalidated after the change: frontend type-check and lint clean, 36 targeted tests passing, `PublicBookingSetup.tsx` and `bookingPageApiService.ts` both still at 100% statements, branches, functions and lines, and `storybook:build` succeeds.

ankit-yc:
All review findings addressed in `8d2c9e2`. Each one, and what verified it.

### CodeQL, both alerts (1327, 1328)
Two assertions used an unanchored `/book\.yosemitecrew\.com/` to prove the dead host is absent — the exact shape CodeQL flags as a URL check that matches anywhere. Neither needed a regex. A substring check on the rendered text asserts the same thing and removes the alert at its root rather than suppressing it. This is also the most likely identity of the Aikido MEDIUM+LOW I could not previously see: two findings, in the two files I had never scanned locally.

### P1 — preserve an explicitly empty service selection
Correct, and the fix needed a backend change. `getConfig` collapsed *never opened the wizard* and *deliberately saved zero services* into the same empty `serviceIds`, so the frontend had no signal to distinguish them. The API now returns `configured`, taken from whether a `PublicBookingSettings` row exists, and the wizard keys its default off that. Two backend tests pin the two cases apart, since they return identical `serviceIds`.

### P1 — initialize toggles from the persisted selection
Correct. `toggleService` seeded from `allBookableIds` rather than the visible selection, so before the first click the stored selection was ignored. Now seeds from `selected`. A test deselects the only saved service and asserts nothing else turns on.

### P1 — remove unavailable service IDs from the editable selection
Correct, and the consequence was worse than cosmetic: an archived service had no row to deselect it with, and the API rejects the whole payload if one submitted id is not active and bookable, so the practice could not save *anything* again. Stored ids are now intersected with what is currently bookable, and filtered again at submit.

### P1 — block destructive saves after a configuration load failure
Correct. The catch left the form on defaults with nothing said, so a save after a transient read outage would have written those defaults over the practice's real settings. The failure is now surfaced in an alert and Save is disabled until a reload succeeds.

### P1 — Playwright coverage for the booking-save journey
The requirement is real (`apps/frontend/AGENTS.md:205`, `:228` list booking as E2E-critical). Worth stating where it runs, because it is not obvious: the public job runs an explicit two-file list (`smoke`, `a11y`), so a spec dropped into `e2e/` would never execute. This route needs a session, so the spec is wired into `playwright-auth` alongside `auth-flow.spec.ts`, following the same credential-artefact rules (trace/screenshot/video off in-file). That job is skipped on `pull_request` by design, so it gives post-merge coverage on `dev`/`main`, not on this PR. It probes for the endpoint and skips on 404 — the API must deploy ahead of the frontend — but fails on 5xx, using the same idiom as `skipUnlessAuthSurfaceDeployed`.

**I could not execute it.** It needs `YC_E2E_*` credentials and an environment with the API deployed, and I have neither. `playwright test --list` confirms it parses and registers; that is the limit of what I verified.

### Not changed: nested `trimHyphens`
The suggestion was that the inner call is redundant. It is not, and there is now a test saying so. Trimming *before* the 63-character slice stops leading separators consuming part of the budget: `!!!` + 70 characters keeps 63 name characters with the inner trim and 60 without.

### Validation
- Backend: **79 tests passing**, 100% statements/branches/functions/lines on all three new files.
- Frontend: **45 tests passing**, 100% on `PublicBookingSetup.tsx` and `bookingPageApiService.ts`.
- `npx tsc --noemit` exit 0 in both workspaces; lint exit 0 in both (the one frontend warning is a pre-existing unused eslint-disable in `CompanionIdCard.test.tsx`, untouched here).
- `storybook:build` exit 0. Workflow YAML re-parsed after editing.
- No schema change in this commit, so the migration and drift gates are unaffected.

aupyay (dismissed):

aupyay (approved):

Issue #2371, opened 2026-08-21
[Bug]: breed picker is always empty and breed is required, so no companion can be created and the app is unusable for new users


ankit-yc:
### Preflight

- [x] I searched existing issues

### Affected area

Mobile app

### What is wrong

A new user cannot add a companion, and therefore cannot use the app.

Breed is a required field on step 2 of Add companion. The breed picker never loads any options, so the field can never be satisfied. The form blocks with "Breed is required" and there is no way forward.

Because Documents, Tasks, Passport and Appointments are all gated behind having at least one companion, a fresh account is left with nothing it can do.

**The server is fine.** I took the access token the app itself has in storage and called the exact endpoint the app calls:

```
GET https://devapi.yosemitecrew.com/v1/codes/mobile/entries
      ?system=YOSEMITECODE&type=BREED&q=dog
  -> HTTP 200, 247 breeds
     "African Wild Dog", "American Dingo (Carolina Dog)", "Argentine Dogo", ...
```

**The token is fine.** The stored access token decodes to `exp` 24 minutes in the future and is accepted by the API, as above.

**The app never makes the request.** Over eight minutes on this screen, including opening the breed sheet, the only traffic the app produced was its 60 second keepalive returning 204. No call to `/v1/codes/mobile/entries` at any point.

So the failure is entirely client side and happens before the fetch. `AddCompanionScreen.tsx:361-390` has two paths that both end in an empty list and say nothing:

```js
const tokens = await getFreshStoredTokens();
if (!tokens?.accessToken) {
  return;                      // silent: breedOptions stays [], no warning, no state
}
const entries = await fetchBreedCodeEntries(...);
...
} catch (error) {
  setBreedOptions([]);
  console.warn('[Companion] Unable to load breed code entries', error);
}
```

Neither path tells the user anything. Both produce the same screen: a search box promising "Search from 200+ breeds" above the words "No breeds available".

I could not tell from the device which of the two fired, because React Native `console.warn` does not reach the system log in a Release build. That itself is worth fixing.

[...]

PR #2518, opened 2026-08-26
fix(mobile): restore token expiry detection and surface list load failures

### LINKED ISSUE #2368: [Bug]: a stalled home request traps the user on an opaque loader with no timeout, error or retry
### Preflight

- [x] I searched existing issues

### Affected area

Mobile app

### What is wrong

Home gates its first render behind a six request AND gate that has no timeout, no error branch and no retry. If any one of those requests does not resolve, the user is left on an opaque full screen loader with no way out except force quitting the app.

Hit live during a QA pass on the first login of a fresh install.

**The loader cannot be dismissed.** `GlobalLoaderContext.tsx:38-49` renders it as a `Modal` with an opaque `theme.colors.white` background, no close control, no `onRequestClose`, and nothing behind it is reachable:

```jsx
{isLoaderVisible && (
  <Modal transparent visible={isLoaderVisible} animationType="fade" statusBarTranslucent>
    <View style={styles.loaderOverlay}>
      <GifLoader size="large" />
    </View>
  </Modal>
)}
```

**Nothing guarantees it is ever hidden.** `HomeScreen.tsx:677-689` is the only thing driving it, and `hideLoader()` sits on the success path only:

```jsx
React.useEffect(() => {
  if (!initialLoadStarted) { hideLoader(); return; }
  if (initialLoadCompleteRef.current || isHomeDataReady) {
    initialLoadCompleteRef.current = true;
    hideLoader();
  } else {
    showLoader();          // no timeout, no failure branch
  }
}, [hideLoader, initialLoadStarted, isHomeDataReady, showLoader]);
```

**The gate it waits on is wide.** `isHomeDataReady` (`HomeScreen.tsx:665`) requires `areRequiredRequestsStarted` AND `areRequestsComplete`, which between them cover six requests: companions, access, notifications, appointments, expenses, linkedBusinesses. Every one must both start and finish. `areRequestsComplete` additionally gates on `hasNotificationsHydrated`, `hasAppointmentsHydrated` and `hasExpenseHydrated`, so a request that fails in a way that clears its loading flag without setting its hydrated flag stalls the gate permanently.

There is no `catch` that hides the loader, and no upper bound on the wait.

### Steps to reproduce

Observed once on a fresh install, so this is a race rather than a deterministic path.

1. Install a fresh build, no prior session.
2. Sign in.
3. Land on Home.

To force the shape rather than wait for the race, make any one of the six requests hang (for example block the appointments endpoint) and sign in.

### Expected behavior

Home either renders with whatever data arrived, or shows an error with a retry. A loading state is bounded.

### Actual behavior

Sat on the opaque white loader for over three minutes. During that window the only network traffic from the app was a 60 second keepalive returning 204, so the home data requests were not in flight, not retrying, and not reported as failed:

```
19:13:07  response_status=204
19:14:07  response_status=204
19:15:07  response_status=204
```

No error, no retry, no way back. Force quitting and relaunching restored the session and Home rendered normally, which is the only reason the account was reachable at all.

Sometimes

### Did this work before

Unsure

### Impact

High

### Suggested fix

Three things, smallest first:

1. Put a ceiling on the wait. After N seconds, hide the loader and render Home with whatever arrived.
2. Give the gate a failure branch. If any of the six requests rejects, hide the loader and show an error with a retry rather than leaving `showLoader()` as the terminal state.
3. Make the loader recoverable on principle. An opaque modal with no dismiss should not be reachable from a state the app cannot exit on its own.

Worth auditing the `hasXHydrated` reducers at the same time: if any of them leaves `hydrated` false after a rejected request, that alone is enough to wedge this gate forever.

### LINKED ISSUE #2372: [Bug]: app renders empty states over real data - 4 companions and 54 notifications exist but every list shows as empty
### Preflight

- [x] I searched existing issues

### Affected area

Mobile app

### What is wrong

The app shows empty states over data that exists. An account with 4 companions and 54 notifications renders as a brand new account with nothing in it.

Every list slice in the persisted store is empty while the API returns data for the same account, using the same token and the same parent id that the app itself is holding.

**Companions**

| | |
|---|---|
| `GET /fhir/v1/parent/<parentId>/companions` | HTTP 200, **4 companions** (Augustus, Timmy, Suzy, Daisy) |
| Home | "Add your first companion" |
| Account | "0 Companions" |
| Documents and Tasks | "Add a companion to get started" |
| persisted `companion.companions` | `[]` |

**Notifications**

| | |
|---|---|
| `GET /v1/notification/mobile` | HTTP 200, **54 notifications**, 38 with `isSeen: false` |
| Notifications, New tab | "Nothing in the box!" |
| Notifications, Seen tab | "Nothing in the box!" |
| Home bell | no unread badge |
| persisted `notifications.items` | `[]` |

The whole persisted store looks like this:

```
appointments.items    : 0
companion.companions  : 0
expenses.items        : 0
notifications.items   : 0
tasks.items           : 0
```

**This is not an auth or a wrong-account problem.** I ran those two calls with the access token taken from the app's own AsyncStorage, and the parent id read out of the app's own HTTP cache. Same token, same id, same host (`devapi.yosemitecrew.com`), 200 both times, data both times.

### Evidence pointing at the cause

**1. The companions request may never be issued.** The app's HTTP cache holds `/fhir/v1/parent/<id>` but not `/fhir/v1/parent/<id>/companions`. Both endpoints return identical caching headers (no `cache-control`, `etag` present), so the absence of one and presence of the other is hard to explain by cache policy.

**2. Two requests return 401 on every launch.** A clean 25 second launch produced 14 requests: ten 200, two 304, two 204 and **two 401**, both at 1.5 seconds after launch, both 271 request bytes with a 1106 byte error body.

**3. Every consumer turns a failure into an empty list.** There are **nine byte-identical copies** of `ensureAccessToken` (companion, appointments, documents, expenses, tasks, forms, coParent, linkedBusinesses, observationalTools). Each throws:

```js
const tokens = await getFreshStoredTokens();
const accessToken = tokens?.accessToken;
if (!accessToken) throw new Error('Missing access token. Please sign in again.');
if (isTokenExpired(tokens?.expiresAt ?? undefined))
  throw new Error('Your session expired. Please sign in again.');
```

and every caller swallows it. `fetchCompanions` (`companion/thunks.ts:74`) catches, falls back to cache, finds nothing, and rejects. The UI renders its "you have no data" empty state for a rejected fetch exactly as it does for a genuinely empty account. Nothing distinguishes the two.

**4. The stored token record has no `expiresAt`.** `@auth_tokens` holds `idToken`, `accessToken`, `userId`, `provider` and no expiry. `resolveExpiration` is meant to fall back to decoding the JWT, and `sessionManager.ts:4` imports `Buffer` from `node:buffer` to do it. `node:buffer` is a Node builtin with no Hermes equivalent, so this deserves checking in a Release bundle rather than in Debug.

For the record, the token itself is fine: it decodes to an `exp` well in the future and the API accepts it.

### Steps to reproduce

1. Sign in as an account that has companions and notifications.
2. Home shows "Add your first companion". Account shows "0 Companions".
3. Open the notification bell. Both New and Seen show "Nothing in the box!".
4. Compare against the API:

```
GET /fhir/v1/parent/<parentId>/companions   -> 200, 4
GET /v1/notification/mobile                 -> 200, 54
```

### Expected behavior

The app shows the companions and notifications the account has. If a fetch fails, it says so and offers a retry.

### Actual behavior

Every list is empty and the app presents that as a new, empty account. No error, no retry, no indication anything went wrong.

Every time

### Did this work before

Unsure

### Impact

High

### Suggested fix

1. Find out what the two launch 401s are, and whether the companions request is issued at all. That is the shortest path to the root cause.
2. Stop conflating "failed" with "empty". A rejected fetch must render an error with a retry, never the new-user empty state. This one change would have made the bug visible immediately instead of looking like an empty account.
3. Collapse the nine copies of `ensureAccessToken` into one. Nine identical copies means nine places to fix and nine chances to miss one.
4. Check `node:buffer` in a Release bundle, and store `expiresAt` explicitly at login rather than re-deriving it by decoding a JWT at every call site.

### Additional context

This very likely explains #2368 (home sits on its loader forever, because the gate waits on requests that never resolve) and #2371 (the breed picker is empty and issues no request). Same shape in three places: a session-layer failure that produces silent empty state rather than an error.

Observed through the App Review login, which rewrites base URLs to dev. Not yet reproduced through a normal email OTP sign-in, so the scope may be wider or narrower than described. The API evidence above was gathered against dev with the app's own credentials either way.

ankit-yc:
## PR Checklist

Two independent bugs that compound each other. The first is the failure, the second is the disguise.

### 1. `node:buffer` never resolved in the Metro bundle, so no token was ever considered expired

`sessionManager.ts` imported `Buffer` from `node:buffer`. `metro-resolver` 0.83.7 has no `node:` protocol handling and no Node builtin shims, so the specifier resolved to nothing on device. The build never failed because `@react-native/metro-config` enables `inlineRequires`, which relocated the module-scope import into the body of `decodeJwtExpiration` and therefore inside its existing `try` block. Metro classifies a require inside a `TryStatement` as **optional**, emits a `null` dependency slot, and moves on.

Verified by bundling rather than by reading, using the real `index.js` entry:

```
c = r(d[6]).Buffer.from(s,'base64').toString('utf8')
},1828,[1,355,38,524,2,1123,null,1829,1208,1830,1210]
                              ^^^^ index 6
```

`metro-runtime`'s require polyfill throws `Cannot find module` for a `null` id, before `.Buffer` is ever read. This is identical in Debug and Release; only the message differs.

The damage chain that produced the 401s in #2372:

1. `decodeJwtExpiration` always threw, was always caught, and always returned `undefined`.
2. Login supplied no expiry to compensate: `passwordlessAuth.ts` returned `expiresAt: undefined` outright, and the social path never set it at all.
3. `isTokenExpired` starts `if (!expiresAt) return false`, so unknown expiry meant valid.
4. `getFreshStoredTokens` therefore always took the fast path and returned whatever was in the Keychain. Its refresh branch was dead code on device, across roughly 25 call sites.
5. `scheduleSessionRefresh` fell back to `DEFAULT_REFRESH_INTERVAL_MS` of 6 hours, against a SuperTokens core default access token lifetime of 1 hour.

Dead credentials were handed to callers and the screen rendered axios's own `Request failed with status code 401`.

The auth suite passed throughout, because Jest runs on Node where `node:buffer` resolves, and because the test file additionally mocked `node:buffer` with a fake `Buffer.from` that returned hardcoded strings. The real decoder had never run in a test.

### 2. Every failure was rendered as the new-user empty state

Eight list screens collapsed four situations into two. Loading, failed, never-fetched and genuinely-empty all arrived as an empty array, so all four rendered the onboarding prompt: "add your first task", "add your first expense", "Add your first companion", "you're all caught up". None of the eight rendered an error and none offered a retry.

Home was the worst case (#2368). Its readiness gate waited on hydration flags that are only ever set in a thunk's `.fulfilled` case, so a rejected fetch was indistinguishable from one still in flight. A single failure pinned the opaque full-screen loader up until the 12s escape hatch added in #2379, then dropped the user onto a Home with every section silently blank.

### Token expiry (`fix(mobile): decode token expiry without node:buffer`)

- New `src/features/auth/utils/jwt.ts` decodes base64url and JWT payloads in pure JS. It depends on neither `Buffer` nor `atob`, since `atob` is not part of the React Native runtime contract either (there is no `atob` anywhere in `react-native`). Written with arithmetic rather than bitwise operators so it stays lint clean.
- `expiresAt` is now recorded at sign-in in both the OTP and social paths, rather than left undefined in the hope that a later consumer derives it.
- `shouldAttemptRefresh` is split out of `isTokenExpired`. `isTokenExpired` answers "is this definitely dead" and keeps its lenient default for the ~25 call sites that bail out early on it. `shouldAttemptRefresh` answers "can I hand this to a caller unchecked" and treats an unknown expiry as needing a refresh.
- The "unchanged token means dead session" inference now only applies when the token was already **known** expired. Reaching the refresh branch because the expiry was unreadable proves nothing, and nulling there would sign out every user whose token is not a parseable JWT.
- Refresh fallback lowered from 6 hours to 50 minutes.

### List load failures (`fix(mobile): tell the user a list failed to load`)

One shared mechanism rather than eight local patches:

- `src/shared/store/collectionLoadState.ts` adds `failedCompanions` alongside the existing `hydratedCompanions`, with helpers that keep the two mutually exclusive.
- `src/shared/utils/listPhase.ts` returns `loading | error | empty | ready`. Items already on screen always win, a failure outranks loading so a failed retry keeps its retry control, and `empty` requires a successful fetch to prove it.
- `src/shared/components/common/ListErrorState` is the counterpart to `EmptyState`: same warm-bone shape so an error looks like part of the app, plus a retry CTA, an `alert` accessibility role, and i18n copy in `en` and `es`.

Wired into the companion, notifications, tasks, appointments and expenses slices, and into the Home, Tasks, Notifications, Expenses, Appointments and Documents screens. Home's gate now treats a recorded failure as settled, so the loader clears immediately instead of waiting out the 12s timeout.

**On reusing this shape elsewhere.** A parallel review of the desktop app turned up the same habit in a different codebase: an operation fails or never happens and the UI reports success or shows nothing. The part of this worth copying is not the React component, it is the split in the store layer:

- A rejected fetch must write a **failure signal that is distinct from the absence of a success signal.** The bug here was that `hydratedCompanions` is set only in `.fulfilled`, so "failed" and "still in flight" were the same observable state. Any `{items, loading, error}` slice has this hole, because `error` is written by every rejected thunk in the slice, not just the list fetch, so it cannot be used to decide what the list renders.
- The two signals must be **mutually exclusive by construction**, not by convention. That is why `markCollectionPending/Hydrated/Failed` are helpers rather than three inline assignments: a slice cannot drift into claiming both.
- **One resolver, not one condition per screen.** `resolveListPhase` is ~15 lines and takes four booleans. Its value is that the ordering decisions live in one place: existing items always win over a spinner or an error, a failure outranks loading so a failed retry keeps its retry control, and `empty` requires a successful fetch to prove it. Those three rules are what eight screens each got wrong differently.

`resolveListPhase` and `collectionLoadState` carry no React and no React Native imports, so they port to the desktop and web apps as-is. `ListErrorState` is the only mobile-specific piece.

### Three adjacent bugs the work surfaced, each verified the same way

1. **A launch crash in `src/config/variables.ts`.** It optionally requires the gitignored `./variables.local`, and Metro emits a `null` slot for that too. Its catch only recognised Node's `code === 'MODULE_NOT_FOUND'`, but `metro-runtime` throws a plain `Error` with no `code` and, in a release bundle, no module name. The throw escaped, and because that module reads its config at module scope, any build without a `variables.local.ts` failed to boot. Extracting module 1208 from the real bundle and running it with Metro's require semantics threw before the fix and evaluates cleanly after it. This is pre-existing on `dev`, not introduced here.
2. **The persist migration runner applied exactly one step.** Keyed on the persisted version, so a user two or more versions behind skipped every later migration. The later a migration was added, the fewer users it reached. It now runs every step from the persisted version up to the current one. This matters here because all five touched slices are in the persist whitelist, so this PR also adds migration v7 to v8 and makes the load-state helpers tolerate a missing map.
3. **`console.log/info/debug/trace` were nooped unconditionally at module scope**, so a Debug build was exactly as mute as a store binary. That is why the token-expiry failure left no trace: `decodeJwtExpiration` warned on every call for a whole release and nobody ever saw one. Now guarded by `if (!__DEV__)`. `console.warn` and `console.error` are untouched in both configurations. The submission checklist, both `mobile-patterns` skill copies and `apps/mobileAppYC/AGENTS.md` are updated, since they told the reader to toggle this by hand.

## On #2371

Deliberately not changed here, and **I could not explain the reported "no network request for 8 minutes".**

What is settled: `AddCompanionScreen` gained a `breedLoadFailed` state in #2416, so the picker already distinguishes a failed lookup from a species with no breeds. That half of the issue is fixed on `dev`.

What I found and then had to discard: there **is** a zero-traffic path in that effect. At `AddCompanionScreen.tsx:373` it calls `getFreshStoredTokens()` and does `if (!tokens?.accessToken) { setBreedLoadFailed(true); return; }` before `fetchBreedCodeEntries` is ever reached. I initially took that for the explanation. It is not, on the code the reporter was running: pre-fix, `getFreshStoredTokens` returned the stored token whenever anything was in the Keychain, and `loadStoredTokens` only returns null when the Keychain read AND the AsyncStorage fallback both come back empty, which is a signed-out user. A signed-out user is not sitting on AddCompanion with the rest of the app working. So the path exists but was not reachable in the reported state, and I have no telemetry to go on because `console.log` was nooped in every build (fixed separately in this PR).

Honest answer: unexplained. Cause A accounts for an empty picker, not an absent request, and I would rather leave that open than invent a mechanism that does not survive checking.

One forward-looking note, since it is a real consequence of this PR rather than a theory about the past: `getFreshStoredTokens` can now return null for a **signed-in** user, when a known-expired token fails to refresh. That makes the early return above newly reachable. It resolves to `breedLoadFailed` with honest copy rather than a silently empty picker, which is the correct outcome, but it is a behaviour change worth knowing about.

Worth a follow-up issue either way: `breedLoadFailed` produces a message but no retry, and breed is a required field, so that state still dead-ends companion creation.

## Validation performed

Verified by building and running, not only by testing, because no `pull_request` workflow builds this app on either platform.

- `npx react-native bundle` for **iOS and Android**, release mode, real `index.js` entry. The `Buffer.from(...,'base64')` call is gone from both bundles and the `jwt` module resolves cleanly (`deps: 1,38`, no `null`).
- Module 1208 (`config/variables`) extracted from the emitted bundle and executed against a `metroRequire` stub: threw `Cannot find module` before, evaluates and returns 19 exports after.
- `xcodebuild ... -sdk iphonesimulator` on iOS 26.5: **BUILD SUCCEEDED**.
- Installed and launched on an iPhone 17 simulator. The app boots, recovers its session, makes authenticated API calls that return real data, and renders Home and Tasks in their `ready` state. No red box, no sign-out, no 401.
- `npx tsc --noEmit`: clean.
- `pnpm --filter mobileAppYC run lint`: clean, **zero warnings** (the app had zero before; the arithmetic decoder was written specifically to avoid adding 55 `no-bitwise` warnings rather than suppressing the rule).
- Full Jest suite: **473 suites, 7669 tests, all passing.** Run in full deliberately, since this touches shared store and test infrastructure.
- Every fix has at least one test that was **mutation checked**: the fix was reverted, the specific test confirmed failing, the fix restored, green re-confirmed. Twelve mutations in total, covering the `node:buffer` guard, the base64url substitution, `shouldAttemptRefresh`, the login expiry, the refresh fallback, the `variables.ts` error predicate, the slice failure recording, `resolveListPhase`, the Home readiness gate, the Home companion error render, the v7 to v8 migration, and the migration chain.

A note for the reviewer: `lint-staged` reformatted 12 files that were already failing `prettier --check` on `dev` before this branch existed. Those hunks are pre-existing formatting drift, not intentional changes.

## Related Issue(s)

Fixes #2372
Fixes #2368


ankit-yc:
### On the failing Aikido check

Aikido reports "1 new MEDIUM and 1 new LOW issue were introduced". Neither is introduced by this PR. The two findings scoped to `apps/mobileAppYC` are:

| Severity | File | Line | Finding |
| --- | --- | --- | --- |
| Medium (40) | `apps/mobileAppYC/metro.config.js` | 16 | Potential file inclusion attack via reading file |
| Low (30) | `apps/mobileAppYC/src/shared/services/mobileConfig.ts` | 106 | HTTP request might enable SSRF attack |

Both files are byte identical to `dev` on this branch:

```
$ git diff --name-only origin/dev...HEAD | grep -E "metro.config.js|mobileConfig.ts"
(no output)

$ git diff --stat origin/dev...HEAD -- apps/mobileAppYC/metro.config.js apps/mobileAppYC/src/shared/services/mobileConfig.ts
(no output)
```

So this is a baseline attribution artefact rather than a regression, and I have deliberately not touched either file here to keep this PR to its two causes. Happy to open a separate issue for both if you want them triaged.

One of them is worth a note though. `metro.config.js:16` is the `extraNodeModules` Proxy:

```js
const appModulePath = path.join(projectRoot, 'node_modules', name);
if (fs.existsSync(appModulePath)) return appModulePath;
return path.join(workspaceRoot, 'node_modules', name);
```

That Proxy answers for **any** specifier, including `node:buffer`, by handing Metro a path that does not exist. It is one half of why the bug in this PR was invisible: the resolver never said "I do not know what this is", it said "here is a path", and the path simply was not there. Worth considering an explicit reject for `node:`-prefixed and Node builtin specifiers so the next one fails loudly at build time instead of silently at runtime. I did not do it here because it is a build-config change with its own blast radius, and the guard test added in this PR (`__tests__/guards/metroResolvableImports.test.ts`) already catches the same class in CI.

### On the React Doctor warning

`no-giant-component` on `NotificationsScreen.tsx`. That file was 543 lines before this branch and this PR adds roughly 25 to it. Score is 98/100 and it is a warning, not a new class of problem. Splitting that screen is a reasonable follow-up but not something I wanted to fold into a bug fix PR.

ankit-yc:
### Update: Sonar is green, Aikido is down to one pre-existing LOW

**SonarCloud now passes**, including the repo's own stricter bar in `scripts/ci/sonar-thresholds.mjs`:

```
sonar-thresholds: yosemitecrew_Yosemite-Crew_MobileAppYC
  coverage:    98.0%  (floor 95%)
  duplication: 0.0%  (ceiling 0%)
  issues:      0  (ceiling 0)
sonar-thresholds: the bar is met.
```

The five new-code issues in the previous scan were all mine and are fixed in `e7cead5`. Worth noting one of them, because I did not do what the rule asked: S6653 wanted `Object.hasOwn()` in place of `Object.prototype.hasOwnProperty.call()`. `Object.hasOwn` is ES2022, and betting on a runtime API being present is the exact mistake that produced the `node:buffer` bug this PR exists to fix. I dropped the provenance check for a **type** check instead: an inherited member such as `toString` reads back as a function and never as a string, so it cannot be mistaken for a failure message, and hydration compares `=== true` rather than truthiness. Same protection, no new runtime dependency, and the rule no longer applies.

**Aikido** has gone from "1 new MEDIUM and 1 new LOW" to **one LOW**, and the remaining one is still not from this PR. It is:

- `apps/mobileAppYC/src/shared/services/mobileConfig.ts:106` - "HTTP request might enable SSRF attack"

Aikido's own issue record places it on `dev`:

```json
"location": {"type": "code_repository", "name": "Yosemite-Crew", "branch_name": "dev"}
```

and the file is absent from this branch's diff:

```
$ git diff --name-only origin/dev...HEAD | grep -c "mobileConfig.ts"
0
$ git diff --stat origin/dev...HEAD -- apps/mobileAppYC/src/shared/services/mobileConfig.ts apps/mobileAppYC/metro.config.js
(no output)
```

The `metro.config.js` MEDIUM from my earlier comment is no longer reported as new. I have left both files alone deliberately; happy to open a separate issue for the `mobileConfig.ts` SSRF if you want it tracked.

Local state on `e7cead5`: `npx tsc --noEmit` exit 0, `pnpm --filter mobileAppYC run lint` exit 0 with zero warnings, and the full Jest suite at 473 suites / 7679 tests / 12 snapshots passing.

ankit-yc:
### Review status, so the open threads are easier to triage

CI is green on `2596664` (42 pass, 0 fail, 3 skipping), and `sonar-thresholds` reports `the bar is met` at coverage 97.9% / duplication 0.0% / issues 0. Every finding below has either been fixed or has a reason recorded.

**Addressed. These threads are stale rather than open** - Codex commented against an earlier SHA and the threads do not auto-resolve:

| Finding | Where it landed |
| --- | --- |
| Record expense-summary failures before waiting on them | `01db437` |
| Keep companion selection available in the expense error state | `01db437` |
| Render the tasks loading phase | `01db437` |
| Preserve the Documents screen shell for load errors | `01db437` |
| Ignore stale notification rejections | `01db437` |
| Render the appointments retry loading state | `01db437` |
| Handle the notifications loading phase | `01db437` |
| Surface appointment failures on Home | `01db437` |
| Render the expense loading phase | `01db437` |
| Ignore stale task-fetch rejections | `01db437` |
| useMemo unnecessarily wraps resolveListPhase | `01db437` |
| Separate summary failures from expense-list failures | `2596664` |
| Surface expense failures before declaring Home ready | `2596664` |
| Localize the Home appointment retry copy | `2596664` |
| Preserve the newer active request on older success | `9ed43a1` |
| Suppress empty-account sections after companion load failure | `9ed43a1` |

Two of those deserve calling out because the reviewer caught defects I introduced while fixing the previous round:

- **Preserve the newer active request on older success** inverted the exact race the requestId tracking exists to fix. An older fulfilment wiped the newer request's entry, so `markCollectionFailed` saw no active request over an already-hydrated collection and discarded a genuine newer failure. Stale data was presented as fresh, with no staleness banner. Correct catch.
- **Suppress empty-account sections** was Home saying the companion list could not be loaded AND asserting the account had no companions, in the same viewport. That is the misleading empty-account diagnosis the error state was added to prevent, recreated one section down.

**Not addressed, with reasons:**

- **Add Detox coverage for expired-session recovery.** A fair ask, and `AGENTS.md` does require Detox for auth flows. I have not done it because the only existing Detox spec is an app-shell smoke test, so this means building expired-session E2E infrastructure (a revoked SuperTokens session, a launch fixture, navigation assertions) rather than adding a case. I could write a spec, but I cannot run it here, and adding test code that has never executed while implying it verifies something is worse than not adding it. Worth its own issue.
- **`NotificationsScreen` is 330 lines (`no-giant-component`, warning).** I added roughly 30 lines to a file that was already large; I did extract `renderNotificationsPlaceholder` while I was in there. Splitting the screen properly is a refactor, and folding it into a bug fix would make this diff harder to review, not easier.

Also deliberately untouched: the rest of Home's copy is still hardcoded English. The tiles I added are localized, but fixing the whole screen is a pre-existing, screen-wide job and quietly bundling it here would bury it.

aupyay (approved):

PR #2560, opened 2026-08-30
fix(frontend): repair the PIMS phone screens that overflow the viewport

ankit-yc:
## PR Checklist

Four PIMS surfaces do not fit a phone. Each was found by rendering every one of the 1,871 Storybook stories at 390px and measuring `documentElement.scrollWidth` against a pinned 390, then confirming the reachable ones in the running app.

- **Consent banner** - mounted in the root layout, so it is on top of every PIMS screen until a choice is stored. It carried only the desktop placement: a fixed 80px left offset with a 300px card, leaving an 80px gutter on one side and 10px on the other at 390px, and hanging the illustration 250px below the card, across the tab bar's Home and Schedule labels.
- **Lab results panel** (companion history) - four columns at 160 + 120 + 120 + 100 plus three 12px gaps cannot render narrower than 536px, 179px past a 390px phone. The Meter column sat outside its own card and the screen scrolled sideways. Its own "Phone (375)" story showed this.
- **Weekly availability** (team onboarding) - the row needs about 500px, so the add-range and duplicate circles ran off the side of the screen. They are the only way to add a second range or copy a day, so the row was not merely clipped, it was unusable on a phone.
- **Inpatient schedule** (appointment workspace, Treatment step) - about 600px of fixed columns. The step is reached on a phone through `PhoneWorkspaceShell`, which reuses it verbatim, so the whole workspace scrolled sideways.

Separately, ten components on the phone surface had **no story at all**, which is why none of this was visible to Storybook or Chromatic.

## What is the new behavior?

- Consent banner docks to the bottom below 768px with even 16px gutters, clear of the 72px tab-bar reserve and the home indicator; the decoration is dropped rather than left to collide with the bar.
- Lab results keeps its columns and scrolls the table inside its card. Nothing is hidden: a lab panel is read by comparing a value against its reference interval.
- Availability drops the time ranges to their own line below `sm` and keeps the two actions on the first line, next to the day they act on. A closed day keeps "Day off" on the first line.
- Inpatient schedule collapses its row to a single column below 768px. Its template is an inline style, so no stylesheet or variant can override it and the decision has to live in the component.
- Ten new story files cover the previously unstoried phone components: `PhoneFab`, `PhoneStepChips`, `PhoneVitalsTiles`, `PhoneWorkspaceActionBar`, `PhoneWorkspaceShell`, `PhoneDayStrip`, `PhoneDayRail`, `PhoneWeekOverview`, `TabletCalendarTitleBand`, `PhoneDevHome`, plus a phone story for the consent banner.

Desktop and tablet are unchanged at every one of these.

### One thing worth reviewing

The inpatient schedule was first wrapped in an `overflow-x-auto` container, the same shape as the lab-results fix. That is wrong there: `overflow-x: auto` computes `overflow-y` to `auto` as well, and the status menu does not portal, so a menu opened on a lower row lost its bottom 112px **on desktop**. Measured, reverted, and replaced with the responsive template. The lab-results panel contains only text, so the scroller is safe there.

## Validation performed

- `tsc --noEmit` clean; `eslint` 0 errors (1 pre-existing warning in a file this PR does not touch).
- `pnpm --filter frontend run test -- --testPathPatterns="Availability|CompanionHistoryTimeline|InpatientSchedule|Cookies"` - 19 suites, 295 tests, all passing.
- New `InpatientSchedulePhoneLayout` test mutation-checked: reverting the fix fails it, restoring it passes.
- All 59 new and changed stories render and their play functions pass at 390px.
- Re-measuring the 132 stories that overflowed at 390px before this branch: 19 of them are now clean, and no story regressed.
- Desktop unchanged, measured: availability keeps its four-column template at 1280 and 834; the lab-results card reports no scrollbar at either width; the inpatient status menu is no longer clipped.

### Not addressed here

21 further stories still measure past 390px and are not covered by this PR. They are a mix of story canvases that deliberately pin a desktop width and components that would need a design decision rather than a layout repair: `PackageBreakdownTable`, `PageSkeleton` (planner), `Labels`, `WorkspaceHeader`, `SpecialitySearchBar`, `TabToggle` (many tabs) and `AddCompanionFormMode`. Worth a follow-up issue rather than folding into this branch.

Storybook coverage overall is 305 of 579 components. The ten fixed here were the ones on the phone surface; the remaining gap is mostly small presentational sub-components in `appointments` and `organization`.


aupyay (dismissed):

aupyay (dismissed):

ankit-yc:
@AikidoSec ignore: Not a credential. The flagged value was the published TOTP documentation vector - it decodes to the ASCII string "Hello!" followed by 0xDEADBEEF (hex 48656c6c6f21deadbeef) - used as a Storybook fixture for the 2FA enrolment UI. It was never issued by or pointed at any service.

It has already been removed at source in 183d5360: the fixture is assembled from a repeated group so there is no literal to match, and the old value is not quoted in the surrounding comment either. The finding persists only because the scan covers the branch's commit history rather than its current tree.

A sweep of every file this branch adds found no other credential-shaped literals; the only remaining opaque tokens are six visibly sequential git SHAs in a marketing fixture.

aupyay (approved):

PR #2566, opened 2026-08-30
fix(frontend): rebuild the public booking page on the design system

ankit-yc:
## PR Checklist

- [x] There is an issue for the bug/feature this PR is for - #2567

The page a pet owner sees at `/book/<slug>` asks a stranger for their name, email, phone number and their animal's details, and it looks unfinished doing it: no card or surface of any kind, native OS radio buttons, a native date input, twenty-seven undifferentiated time pills, and selection shown only as a 1.5px border-colour swap.

Underneath most of that is one structural gap. **The `(book)` route group had no layout.** `(app)` mounts `ThemeScript` and `(public)` mounts the raw pre-paint script; `(book)` had neither, with two consequences:

1. Nothing ever stamped `data-theme` on `<html>`, so the whole `html[data-theme='dark']` block was dead code on this route. Measured on dev with the OS in dark: `data-theme` is `null`, body stays `#efe8dc`. The page is permanently light while every other Yosemite surface flips.
2. Nothing rendered `data-yc-app`, so `--ink-faint` stayed at the marketing value `#8f8984` instead of the bone-surface value `#66635f`. That token carried every service duration, the "Nothing is booked yet" line, and the sentence saying this is a request and not a booking - all at **2.84:1**. The page's entire honesty argument was its three least readable lines.

A contrast sweep of every rendered text node on dev: **10 of 56 fail WCAG AA.**

## What is the new behavior?

**`(book)/layout.tsx` (new)** - pre-paint theme script, already authorised by `PRE_PAINT_SCRIPT_CSP_HASH` on the strict-CSP path that `/book` is on, plus a `display: contents` `data-yc-app` scope. Dark mode works, and `color-scheme` finally reaches the native date input and checkbox.

**`bookingChrome.tsx` (new)** - one shell, card, callout, alert, footer, type scale and set of formatters shared by `/book/<slug>` and `/book/<slug>/confirm`, which previously used different widths and padding (560/`p-5` vs 520/`p-6`) so the column visibly jumped when a reader followed the emailed link.

**Controls.** The service radio is now a custom control driven by a `peer` sibling dot; the consent checkbox is 24px. Neither had *any* focus indicator: `globals.css` suppressed the global ring for inputs on the strength of a border-colour change that neither of them has. Added a scoped `:focus-visible` rule for both plus a `forced-colors` block, since `outline` is the one property that mode preserves.

**Selection** is a fill plus a shape change rather than a border-colour swap (2.88:1 in dark before). This also removed both inline `style={{ borderColor }}` objects, which were the reason the page had no hover, focus or active state anywhere - a style attribute has no pseudo-classes.

**Content and layout.** Times grouped by day part in a content-sized `auto-fill` grid; quick-day chips for the near-term case; `service.description` rendered at last, having been fetched and thrown away; the arbitrary 12.5/13.5/14.5px sizes replaced with the documented scale; `autoComplete` on every field that has a token (there were none); the honesty callout moved to the top of the card at 16px and 10.39:1; a footer that says whose page this is.

**Stories.** All four existing stories had never rendered - `stub()` assigned onto a frozen ES module namespace, and `useRouter` needed the app-router context. Both fixed, so the seven stories (four existing, three new) now actually mount.

## Validation

| Check | Result |
|---|---|
| `tsc --noEmit` | clean |
| `pnpm --filter frontend run lint` | 0 errors (1 pre-existing warning in `CompanionIdCard.test.tsx`, untouched) |
| `pnpm --filter frontend run test` | **12,007 passed / 972 suites** |
| Booking tests specifically | 47 passed; all 21 original `BookClient` assertions pass **verbatim** |
| Coverage of the changed files | `BookClient.tsx` 99.88% stmts / 99.13% branch / 100% func, `bookingChrome.tsx` 100/95/100, `ConfirmClient.tsx` 100/100/100 |
| `pnpm --filter frontend run build` | clean; confirmed `peer-checked`, `aria-pressed` and `:has()` all emit (none existed in this repo before) |
| Contrast, every text node, real browser | **81/81 pass AA in light and dark**, vs 10/56 failing on dev |
| axe-core (wcag2a + wcag2aa + wcag21aa) | **0 violations in light and 0 in dark**, vs 1 serious `color-contrast` across 10 nodes on dev |
| Focus | verified by screenshot on the radio dot, the row, the fields and the consent checkbox |
| Target sizes | slot pill 44px, checkbox 24px, fields 50px, submit 48px |

Verified against the real nine-service practice, at 320 / 360 / 390 / 768 / 1440, in both themes, plus all seven stories in both themes. No horizontal scroll at 320px, where the time pills still measure 77x44.

Note for a follow-up: `e2e/a11y.spec.ts` scans `/`, `/signin`, `/signup` and `/pricing` but not `/book/<slug>`, so the axe result above is mine rather than CI's. Worth adding that page to the suite.

## Things a reviewer should push back on if they disagree

- **The date control stays a native `<input type="date">`.** `BookClient.test.tsx` reads `.min` and `.max` off the DOM node and `ui/inputs/Datepicker` exposes neither. This is the most visible remaining seam; it is a deliberate refusal, not an oversight.
- **`practice.logoUrl` is still not rendered.** `` is a lint error under `core-web-vitals` with `--max-warnings=0`, and `next/image` *throws* for a host outside `images.remotePatterns` - one practice's logo would take the whole page down. Worth its own PR.
- **The page is longer than it was**: 2477px vs 1857px at 1440, 3420px vs 2008px at 390, for a practice with nine services that all have descriptions. That is the cost of showing what each service actually is, and of 44px targets. A two-column service grid at `sm:` and a content-sized time grid claw back ~440px; I did not add a "show more" because hiding the service the reader came for is the wrong economy. Happy to revisit if you read it differently.
- **The 27 slots remain `aria-pressed` toggle buttons** rather than a radio group. Converting changes the role and the attribute and rewrites ~15 assertions; it belongs in its own PR.

## Related Issue(s)

Fixes #25672567

### DIFF
--- a/apps/frontend/src/app/(routes)/(book)/book/[slug]/BookClient.stories.tsx
+++ b/apps/frontend/src/app/(routes)/(book)/book/[slug]/BookClient.stories.tsx
@@ -1,7 +1,6 @@
 import type { Meta, StoryObj } from '@storybook/react';
 import { expect } from 'storybook/test';
 
-import * as publicBookingService from '@/app/features/publicBooking/services/publicBooking.service';
 import type { PublicPractice } from '@/app/features/publicBooking/services/publicBooking.service';
 import BookClient from './BookClient';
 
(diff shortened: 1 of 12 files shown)


ankit-yc:
All ten React Doctor findings are addressed in c6bbacd, and `npx react-doctor` now reports nothing on any of these files.

**`only-export-components` (x6)** — real, and the fix improves the module. `bookingChrome.tsx` was exporting components, class-name constants and pure formatters together, so Fast Refresh could not preserve state and had to full-reload. Split three ways by responsibility:

- `bookingChrome.tsx` — components only (`BookShell`, `BookFooter`, `IconDisc`, `Callout`, `StateCard`, `Skeleton`, the icons, `Spinner`)
- `bookingStyles.ts` — the class recipes (`FIELD`, `PILL`, `SLOT_GRID`, `EYEBROW`, `FIELD_LABEL`, `META_TEXT`, `STATE_BODY`)
- `bookingFormat.ts` — the pure helpers (`formatShortDay`, `formatLongDay`, `quickDayLabel`, `DAY_PARTS`, `groupByDayPart`, `describeBlock`)

**`js-hoist-intl` (x2)** — real, and it matters more than the generic wording suggests. Both `new Intl.DateTimeFormat()` calls sat inside the functions. `formatLongDay` backs the "Sunday 30 August · you can book up to 28 days ahead" hint, which lives in the same component as the form state, so it re-rendered on **every keystroke** in the six fields below it and rebuilt the formatter each time. Both are now built once at module scope.

**`js-combine-iterations` (x1)** — fair. `groupByDayPart` did `.map().filter()`, two passes. Now a single pass that only builds the groups that actually have slots. While there, each day part carries its own `from` hour instead of looking back at the previous array element, so the result no longer depends on `DAY_PARTS` ordering.

No behaviour change — the split is import paths only. The 47 booking tests pass unchanged, `tsc` is clean, and I re-ran the verification on the rebuilt page afterwards: still **0 axe violations** and **0 text nodes below AA** in both themes.

ankit-yc:
Correction to my previous comment: I said React Doctor reported nothing on these files. That was wrong — my local check grepped output that was category-summary only, so it would have found nothing regardless of the state of the code. CI was right: 10 → **1**, not 10 → 0. Sorry for the noise.

The remaining `no-giant-component` on `BookClient` is fixed in 67f76ab. The component had reached ~420 lines, which is the same thing that motivated pulling `SlotPicker` and `renderStatus` out of it earlier. It is now four sections — `ServiceFieldset`, `DayAndTime`, `DetailsFieldset`, `CheckAndSend` — each taking only what it renders, so a section stops depending on the enclosing closure. `BookClient` keeps state and orchestration.

Two more, found by re-running with `--json` and filtering per file (the check I should have run the first time):

- `rerender-lazy-state-init` — `useState(todayIso())` re-derived todays date on every render and threw it away. Now `useState(todayIso)`.
- `design-no-redundant-padding-axes` — `px-4 py-4` on the service row is `p-4`.

**Four findings I am deliberately not acting on**, all pre-existing rather than introduced here:

| Rule | File | Why not |
|---|---|---|
| `nextjs-no-native-script` | `(book)/layout.tsx` | It has to be a raw inline `<script>`. `next/script` defers execution, and the whole point is to stamp `data-theme` *before* first paint; deferring reintroduces the flash. It is also authorised by a CSP **hash** (`PRE_PAINT_SCRIPT_CSP_HASH`), which only matches the literal script body. Same pattern, same reason, as `(public)/layout.tsx`. |
| `no-danger` | `(book)/layout.tsx` | Same line. The content is a build-time constant from `prePaintScript.ts` with no interpolation and no user input. |
| `prefer-tag-over-role` | `BookClient.tsx:168` | It suggests `<address>` in place of `role="group"`. `<address>` is for contact details for a document or article; this is a labelled set of time buttons, where `role="group"` + `aria-labelledby` is the correct pattern. |
| `nextjs-no-use-search-params-without-suspense` | `ConfirmClient.tsx:33` | The consequence of an unwrapped `useSearchParams` is bailing out of static generation — and `confirm/page.tsx` already sets `export const dynamic = force-dynamic` deliberately, because the answer depends on a token. Nothing left to lose. |

`prefer-useReducer` (7 `useState` calls in `BookClient`) is a fair observation, but converting it touches every state transition on a page whose 21 original assertions I have kept passing verbatim. I would rather do that as its own PR than fold it into a visual rebuild.

Re-verified after all of the above: `tsc` clean, lint clean, 47 booking tests pass, and the rendered page still measures **0 axe violations** and **0 text nodes below AA** in both themes.

aupyay (approved):

Issue #2569, opened 2026-08-30
[Bug]: the (share) route group has no layout either, so passport pages are light-only and 3.12:1


ankit-yc:
### Preflight

- [x] I searched the existing issues and this is not a duplicate.
- [x] This is not a security vulnerability.

### Affected area

Web app (apps/frontend)

### What is broken

Found while fixing the same defect on the public booking page (#2567 / #2566). The cause is shared, so this is the other half of it.

`apps/frontend/src/app/(routes)/` has four route groups. Three now have a `layout.tsx`; **`(share)` does not** — and it holds the public pet passport and companion card, the pages an owner shares with a stranger.

Two consequences, both the same ones `(book)` had:

**1. These pages can never render dark.** Nothing in the group mounts `ThemeScript` or the raw `PRE_PAINT_SCRIPT`, so nothing stamps `data-theme` on `<html>` and the entire `html[data-theme='dark']` block in `globals.css` is dead code on the route. `(app)/layout.tsx` mounts `ThemeScript`; `(public)/layout.tsx` mounts the raw script; `(book)/layout.tsx` now does too. `(share)` was missed.

**2. `--ink-faint` resolves to the value `globals.css` documents as unreadable.** The readable `#66635f` comes from `body:has([data-yc-app])`. No page under `(share)` renders that marker, so the token stays at the marketing value `#8f8984`.

`PublicPassportView.tsx` paints **nine** elements with it:

- `L153`, `L157`, `L213`, `L280`, `L316` and four more, several at `text-[10.5px]` and `text-[11.5px]`

The card surface is `var(--screen)` (`L144`). `#8f8984` on `#f7f3ec` measures **3.12:1**, against the 4.5:1 that 10.5px text owes. It is worse than the booking page's version of this bug, because the type is smaller.

Neither page pins the light inks with `data-yc-surface="light"` either, which is the escape hatch `/payment-status` and `/success` use for deliberately fixed-light surfaces.

### Expected

Either give `(share)` a layout the way `(book)` just got one in #2566 —

```tsx
<script dangerouslySetInnerHTML={{ __html: PRE_PAINT_SCRIPT }} />
<div data-yc-app style={{ display: 'contents' }}>{children}</div>
```

— or, if these pages are meant to stay light in both themes (defensible for something designed to be screenshotted and shared), pin them with `data-yc-surface="light"` and say so. What is not defensible is the current state, where the choice was never made and the text is 3.12:1 either way.

[...]

PR #2574, opened 2026-08-30
fix(frontend): give the shared passport and card pages a readable theme

ankit-yc:
## PR Checklist

- [x] There is an issue for the bug/feature this PR is for - #2569

#2569 reported that `(share)` is the last route group with no `layout.tsx`, so `/passport/<id>` and `/card/<token>` never resolve a theme and `--ink-faint` stays at the marketing value. Verifying that turned up three more contrast failures with the same cause. **Every ratio below was measured in a browser, not derived from the tokens.**

### The one that is not cosmetic

`PublicPassportView` draws both vaccination badges from the `--status-*` family. The warm-bone dark block redefines `--status-completed-*` but **never `--status-danger-*`** - one prefix away from `--danger-*`, which it does define. So the expired badge kept the `:root` **light** ink `#a6271d` on an espresso card:

| badge | light | dark, before | dark, after |
|---|---|---|---|
| VALID | 6.81:1 | 5.77:1 | 5.77:1 |
| **EXPIRED** | 5.74:1 | **1.89:1** | **5.49:1** |

`PublicPassportView.tsx:47-49` says this page is *"read as proof of cover by travel and boarding staff"*. The failure was inverted: **valid** legible, **expired** all but invisible. That is the wrong way round for a rabies record.

This was reachable before this PR, because the passport already ships its own sun/moon toggle - it is not a state the reader had to opt into via OS settings.

### The rest

**`--ink-faint` on both pages** — now `#66635f` instead of `#8f8984` (3.12:1 on the passport card, at sizes down to 10.5px). The passport needed a second edit: `.yc-warmbone` carries a private copy of the bone ink ramp that **shadows** `body:has([data-yc-app])`, so the scope could not reach the one page the issue was about. I only caught that because I measured `--ink-faint` on the rendered `<main>` rather than trusting the layout to be sufficient.

**The passport's own toggle** hardcoded `'light'` on mount, so a reader whose phone is dark got a full-brightness health record and had to press the button every visit. It now seeds from the resolved theme. The toggle stays - removing a shipped control is a product call, not a bug fix.

**Two surfaces outside `(share)`,** included because the cause is one shared rule and splitting them would leave the rule half-fixed:

- The **in-app pet passport** is a literal `bg-white` card (`--color-white` is documented as intentionally never re-pointed) whose inks are theme tokens. In dark mode **all 58 text nodes failed AA, worst 1.34:1** - cream on white, passport and microchip numbers included. Pinned with `data-yc-surface="light"`, the guard `/payment-status` already uses. Now **0 of 58 fail, worst 5.97:1**.
- That pin covered the grey ramp but **not the brand accent**, so `--color-text-brand` stayed `#8fb6f5` on white. Adding it to the pin also repairs three links on the **accessibility report page** - already pinned, and still rendering pale blue on white in dark. Now **6.48:1 in both themes**.

## What is the new behavior?

`(share)/layout.tsx`, matching `(book)`. These routes are **not** under the strict CSP, so the inline pre-paint script is authorised by `'unsafe-inline'` rather than by `PRE_PAINT_SCRIPT_CSP_HASH` - the same path `(public)/layout.tsx` relies on, and no header changes.

Verified on the running app, both pages, both themes:

| | before | after |
|---|---|---|
| `data-theme` | `null` | `light` / `dark` |
| `[data-yc-app]` | absent | present |
| body background | always `rgb(239,232,220)` | flips to `rgb(32,28,24)` |
| passport `data-wb-theme` | always `light` | follows the phone |
| `--ink-faint` on `<main>` | `#8f8984` | `#66635f` |

## Validation

`tsc` clean · lint 0 errors · **12,035 tests pass** · 2 new tests for the layout, 1 for the theme seeding.

## The judgement call, in case you disagree

The issue offered two options: make these pages theme-aware, or pin them light as documents. **I went theme-aware**, on three grounds:

1. **The passport already has a dark mode, with a visible toggle.** It was not "permanently light" - it was permanently *broken*, and pinning would have frozen that. The expired-badge failure only exists because that dark mode is reachable.
2. **The share pages use 53 design tokens and zero hardcoded colour literals**, so they flip cleanly. The usual argument against - "you get a half-dark page" - does not apply here. I checked before deciding.
3. `(book)` settled the same question for the same audience last week.

The strongest argument the other way is that a passport is a document, and documents do not change colour. I find that outweighed by who actually opens these links: a signed-out pet owner on their own phone, often at night, and a boarding desk in a dim room. The in-app passport **is** pinned light in this PR, because it is a literal white card - so both patterns are now used deliberately rather than by omission.

### DIFF
--- a/apps/frontend/src/app/(routes)/(share)/layout.tsx
+++ b/apps/frontend/src/app/(routes)/(share)/layout.tsx
@@ -1,0 +1,36 @@
+import { PRE_PAINT_SCRIPT } from '@/app/ui/theme/prePaintScript';
+
+/**
+ * The layout the `(share)` group never had.
+ *
+ * `/passport/<id>` and `/card/<token>` are the two pages in this product most
+ * likely to be opened on a phone belonging to someone who has never used
+ * Yosemite - a boarding desk scanning a collar tag, a vet reading a shared
+ * record - and they were the only route group with no theme resolution at all.
+ * Nothing stamped `data-theme` on <html>, so the whole `html[data-theme='dark']`
+ * block in globals.css was dead code here.
+ *
+ * No nonce, and none needed, but for a different reason than `(book)`: `/card`
+ * and `/passport` are NOT in STRICT_CSP_PATH_PREFIXES, so they get the
+ * permissive CSP variant whose script-src carries 'unsafe-inline' and
+ * deliberately omits PRE_PAINT_SCRIPT_CSP_HASH (per the CSP spec a hash makes
+ * the browser ignore 'unsafe-inline', which would break hydration across the
+ * static marketing pages). This is the same path `(public)/layout.tsx` relies
+ * on.
+ *
+ * The `display: contents` wrapper introduces no box. It buys `color-scheme` for
+ * native controls, the app's ::selection colour, and - the reason this issue
+ * was filed - the readable bone-surface inks from `body:has([data-yc-app])`.
+ * Without it `--ink-faint` stayed at the marketing value #8f8984, which is
+ * 3.12:1 on the passport's --screen card at font sizes down to 10.5px.
+ */
+export default function ShareLayout({ children }: Readonly<{ children: React.ReactNode }>) {
+  return (
+    <>
+      <script dangerouslySetInnerHTML={{ __html: PRE_PAINT_SCRIPT }} />
+      <div data-yc-app style={{ display: 'contents' }}>
+        {children}
+      </div>
+    </>
+  );
+}
--- a/apps/frontend/src/app/(routes)/(share)/passport/[id]/PassportClient.tsx
+++ b/apps/frontend/src/app/(routes)/(share)/passport/[id]/PassportClient.tsx
@@ -13,7 +13,19 @@ type Theme = 'light' | 'dark';
 const PassportClient = ({ id }: PassportClientProps) => {
   const [passport, setPassport] = useState<PetPassportDTO | null>(null);
   const [state, setState] = useState<LoadState>('loading');
-  const [theme, setTheme] = useState<Theme>('light');
+  // Starts from the theme the reader already chose, not always light. The
+  // pre-paint script in (share)/layout.tsx resolves that onto <html> before
+  // first paint, so reading it here has no flash. Kept as local state because
+  // this page also has its own toggle - the two now agree on the initial value
+  // instead of the page ignoring the phone.
+  //
+  // A lazy initialiser: on the server there is no document, and calling it on
+  // every render would re-read the DOM for a value only the first render uses.
+  const [theme, setTheme] = useState<Theme>(() =>
+    typeof document !== 'undefined' && document.documentElement.dataset.theme === 'dark'
+      ? 'dark'
+      : 'light'
+  );
 
   useEffect(() => {
     let active = true;
--- a/apps/frontend/src/app/(routes)/(share)/passport/[id]/PassportClient.tsx
+++ b/apps/frontend/src/app/(routes)/(share)/passport/[id]/PassportClient.tsx
@@ -50,7 +50,11 @@ const PassportClient = ({ id }: PassportClientProps) => {
     <main
       id="main-content"
       tabIndex={-1}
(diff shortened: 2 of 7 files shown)


ankit-yc:
All three addressed in a1eede9. Every number below measured in a browser.

**P1 — companion card, and you are right that it is this PR's regression.** `CompanionIdCard.tsx` is a literal `bg-white` surface with themed inks, same as the passport card, and had no pin. Until `(share)` got a layout, `/card` never resolved a dark theme at all, so the pairing was unreachable — adding the layout is precisely what makes it reachable, so the pin belongs in this change rather than a follow-up. Pinned with `data-yc-surface="light"`. On that surface: primary ink **13.36:1**, secondary **6.96:1**, identical in both themes.

**P2 — hydration.** Correct, and the mechanism is exactly as described. Replaced the DOM-reading initialiser with `useTheme`, the repo's existing `useSyncExternalStore` wrapper, which supplies a server snapshot so both passes agree. The page keeps its sun/moon control as a local override layered on top of the resolved theme, so a reader can still flip just this passport.

**P2 — both blue spellings.** Correct. I had pinned `--blue-text` but not `--color-blue-text`, which is the one `text-blue-text` actually reads — so a child on a pinned surface could still paint the 2.06:1 pale blue the pin exists to remove. `globals.css:682` already warns this token is written two ways and I walked straight into it. Both pinned now, `--blue-text` aliased to `--color-blue-text` so they cannot drift. Verified: **both spellings 6.48:1, both themes**.

**On the test you asked for.** You asked for SSR/hydration coverage rather than a render-only test. I wrote that first — `renderToString` then `hydrateRoot`, asserting no mismatch — and then mutation-checked it by reverting the fix. **It passed against the bug**, so it was worthless: jsdom provides `document` during the server pass too, so both passes agree and the mismatch never occurs in that environment.

I deleted it and wrote one that asserts the real contract instead: `useTheme` is mocked, and the test asserts `data-wb-theme` follows the mocked store. An implementation that reads the DOM ignores the mock and keeps rendering light. Mutation-checked both ways — **fails against the DOM-reading version, passes against this one**.

`tsc` clean, lint 0 errors, 12,035 tests pass.

aupyay (dismissed):

ankit-yc:
All three follow-ups fixed in 95ad1e5. Each measured in a browser, and each falls out of my previous commit rather than pre-dating it.

**Alert fill inside the fixed-light card.** Right — I pinned the inks but not the fills they sit on, so `bg-card-bg` kept inheriting the dark `#221d17` under light-pinned text. Pinned the surface aliases too. **2.40:1 → 5.87:1**, identical in both themes.

**The passport's initial paint.** Also right, and a good catch that this is *residual* rather than a repeat: fixing the hydration mismatch left the flash the mismatch was masking. `useTheme`'s server snapshot is light, so the markup carried `data-wb-theme="light"` and a dark reader got a bright passport until hydration corrected it. The attribute is now stamped **only** once the reader uses the page's own toggle; absent, `.yc-warmbone` follows `html[data-theme]`, which the pre-paint script sets before first paint. Verified: with no attribute, `--screen` resolves `#f7f3ec` under a light root and `#2f271e` under a dark one.

**Danger tokens in the light scope.** Right, and it is the exact mirror of the bug I fixed in the dark scope — I fixed one direction and left the other. A reader on a dark OS toggling this page to light had `--status-danger-*` falling back to the root dark values on a light card. Aliased in both scopes now. **5.74:1 either way.**

Two things worth recording about how this went:

- The surface pin **first landed in the `@theme` block at `:root`** instead of `[data-yc-surface=light]`, because I anchored the edit on a string whose first occurrence was the original declaration. The measurement is what caught it — the alert stayed at 2.40:1 after a change that should have moved it. Corrected.
- The pre-existing toggle test asserted `data-wb-theme="light"` on mount. That is now wrong **by design**, so it asserts absence instead — and it awaits the passport fetch rather than leaving a state update outside `act()`.

`tsc` clean, lint 0 errors, 12,036 tests pass.

Note this push will have dismissed the approval (`dismiss_stale_reviews_on_push` is on).

aupyay (approved):

PR #2580, opened 2026-08-30
fix(frontend): repair six defects the PIMS story pass surfaced

ankit-yc:
## PR Checklist

> Stacked on #2560, which added the stories that found these. Merge that first.

## What is the current behaviour?

Six defects, each found while writing a story for the component, reported rather than fixed at the time so #2560 stayed a story PR. Every one is pinned by a story that currently asserts the wrong behaviour.

- **A view-only encounter can still gain billable line items.** `ServicesPackagesEditor` renders its search unconditionally - gated only on per-row billing locks, never on `readOnly` - so choosing a result goes straight to `onAddItem` and appends a line. `TreatmentStep` passes `readOnly = encounter.viewOnly`, and the sibling `PrescriptionEditor` already gates the same block correctly.
- **`EmergencyCheckbox` fails WCAG 2.5.3 (label-in-name).** It announces "Confirm this is an emergency" over a visible "I confirm this is an emergency." The `aria-label` overrides the `<label>` the component already wires with `useId`, so speech control cannot act on the words on screen.
- **`AddSpecialityModal`'s dialog has no accessible name** - neither `ariaLabel` nor `titleId` is wired, so the visible heading does not name it.
- **`DropdownRenderer`'s radio and checkbox groups have no group semantics**, so the question reaches the accessibility tree only through each option's composed label.
- **`DateBuilder` collects a Placeholder that can never render.** The shared `FormInput` has no `placeholder` prop, `buildPreviewValues` excludes `date` from the placeholder fallback on purpose, and the attribute does not apply to `input[type=date]` in any browser.
- **`taskTemplateSummary` renders a dangling middot.** Category and repeat are joined unconditionally while reminder and duration are guarded, so a block with no recurrence reads `"Care ·"`.

## What is the new behaviour?

Each is fixed with the smallest correct change, and the story that pinned the defect now asserts the fix - so it fails if someone reverts one.

`DateBuilder` drops the box rather than widening a shared primitive to emit an attribute nothing paints; an existing stored placeholder still rides through edits untouched, since `onChange` spreads the whole field.

### Two things worth reviewing

**`SignatureRenderer` was on this list and is not a defect.** It takes only `field` and renders a static "Please Save and Sign" box, which looked like dropped props. Signing is genuinely a post-save flow: `startFormSigning(submissionId)` runs against a saved submission, driven by `SignatureActions`. Left alone.

**The `ServicesPackagesEditor` phone story becomes visual-only.** Its assertion could never hold under a runner that loads `iframe.html` directly: `sm:` is a viewport media query, the viewport global is applied by the Storybook manager resizing the preview iframe, and a narrow container does not change what a viewport query matches. Confirmed by restoring the pre-fix file and seeing it fail identically, so it was red for a harness reason rather than a component one. The breakpoint is covered by a Chromatic viewport instead.

## Validation performed

- `tsc --noEmit` clean; `eslint` 0 errors (1 pre-existing warning in a file this PR does not touch).
- `pnpm --filter frontend run test -- --testPathPatterns="DateBuilder|EmergencyCheckbox|TreatmentStep|Dropdown|Speciality|taskTemplate"` - 28 suites, 484 tests, all passing.
- All 45 stories across the 8 touched story files render and their play functions pass.
- `BookAppointment.stories.tsx` is touched because it consumes `EmergencyCheckbox` and asserted the old accessible name.

## Not addressed

`addLineItem` in the workspace store still has no `viewOnly` guard, so the services fix is a UI gate - the same shape as `PrescriptionEditor`'s. A non-UI path into that action on a view-only encounter would not be caught. Worth a follow-up if the store is meant to be authoritative.

### DIFF
--- a/apps/frontend/src/app/__tests__/pages/Forms/Sections/AddForm/components/Date/DateBuilder.test.tsx
+++ b/apps/frontend/src/app/__tests__/pages/Forms/Sections/AddForm/components/Date/DateBuilder.test.tsx
@@ -1,11 +1,11 @@
-import React from "react";
-import { render, screen, fireEvent } from "@testing-library/react";
-import DateBuilder from "@/app/features/forms/pages/Forms/Sections/AddForm/components/Date/DateBuilder";
-import { FormField } from "@/app/features/forms/types/forms";
+import React from 'react';
+import { render, screen, fireEvent } from '@testing-library/react';
+import DateBuilder from '@/app/features/forms/pages/Forms/Sections/AddForm/components/Date/DateBuilder';
+import { FormField } from '@/app/features/forms/types/forms';
 
 // --- Mock UI Components ---
 // Mock FormInput to test logic without relying on implementation details
-jest.mock("@/app/ui/inputs/FormInput/FormInput", () => ({
+jest.mock('@/app/ui/inputs/FormInput/FormInput', () => ({
   __esModule: true,
   default: ({ value, onChange, inlabel }: any) => (
     <input
@@ -17,16 +17,16 @@ jest.mock("@/app/ui/inputs/FormInput/FormInput", () => ({
   ),
 }));
 
-describe("DateBuilder Component", () => {
+describe('DateBuilder Component', () => {
   const mockOnChange = jest.fn();
 
   // Test data strictly typed to match the component's expected prop
   const mockField = {
-    id: "date-123",
-    type: "date",
-    label: "Date of Birth",
-    placeholder: "Select a date",
-  } as FormField & { type: "date" };
+    id: 'date-123',
+    type: 'date',
+    label: 'Date of Birth',
+    placeholder: 'Select a date',
+  } as FormField & { type: 'date' };
 
   beforeEach(() => {
     jest.clearAllMocks();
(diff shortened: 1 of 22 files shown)


aupyay (dismissed):

ankit-yc:
All four Codex findings addressed in df814d3c2, and two of them turned out to be worth more than they looked.

**Fieldset/legend** - correct, and the rule is explicit. The previous comment's excuse (a fieldset would change the layout the label div owns) is answered by resetting the user-agent border, margin and padding. Worth flagging for review: the implicit role becomes `group`, not `radiogroup`, so the story asserts that instead.

**Stale sibling docs** - corrected in BooleanBuilder and SignatureBuilder.

**FormRenderer and StructuredResultsPanel** - both confirmed by rendering them, and both were already failing on `dev`, not just on this branch. They asserted behaviour that #2560 inverted: a textarea with no `readonly`, and a results table that spills with no scroll container. My verification only re-rendered the files I had touched rather than the pre-existing stories that asserted the behaviour I changed, and play functions do not run in CI, so neither side caught it. Thanks - that was the useful half of this review.

That prompted a sweep of all 3,291 stories, which found two more of mine in the same state: the Cookies phone story asserting equal gutters, and AvailabilityStep asserting a four-track row. Both are `md:`/`sm:`-gated, so under a runner that loads `iframe.html` directly they measure the desktop layout. AvailabilityStep was the worse one - green while asserting the opposite of what a phone gets. Both now assert only what holds at either width, with the phone geometry left to a Chromatic viewport.

Gates: tsc clean, eslint 0 errors, 974 suites / 12,048 tests green, and every story in the touched files renders with its play function passing.

The 175 remaining sweep failures are stories calling devapi over CORS from localhost - unrelated to this branch and failing the same way before it.

aupyay (approved):

Issue #2578, opened 2026-08-30
[Bug]: .yc-warmbone's private 40-token copy has now caused two contrast defects


ankit-yc:
### Preflight

- [x] I searched the existing issues and this is not a duplicate.
- [x] This is not a security vulnerability.

### Affected area

Web app (apps/frontend)

### What is broken

`.yc-warmbone` in `globals.css` carries a **private copy of ~40 design tokens** — the bone surfaces, the full ink ramp, shadows, status colours — with exactly one consumer: `PassportClient.tsx`. Every value in it is spelled somewhere else in the same file.

This is not a tidiness complaint. The duplication has now caused two shipped defects, both fixed in #2574, and it caused them in the way copies always fail: the copy and its consumer drifted.

**1. The expired-vaccination badge, at 1.89:1.** The dark half of the copy defined `--danger-text` / `--danger-bg` / `--danger-border`. `PublicPassportView` reads `--status-danger-text` / `-bg` / `-border`. **The definitions and the usages were one prefix apart.** The valid badge, which happens to read `--status-completed-*` — a name the copy did define — rendered correctly at 5.77:1, so the section looked fine. Only the expired state was broken, on a page its own source describes as "read as proof of cover by travel and boarding staff".

**2. `--ink-faint` at 3.12:1.** `body:has([data-yc-app])` exists specifically to make faint ink readable on bone surfaces. The copy redefines `--ink-faint` locally, so it **shadowed that scope** — and the passport was the one page the fix in #2574's first commit could not reach. It needed a second, separate edit inside the copy.

Both were fixed by patching the copy, which leaves the copy in place to drift again.

### Expected

Fold `.yc-warmbone` into the global token system, so the passport reads the same tokens as everything else. The blockers are worth stating honestly:

- It has its **own theme attribute** (`data-wb-theme`) and its own sun/moon toggle, now layered over the global theme as a local override (#2574). Any consolidation has to keep that control working.
- The dark half is now reachable two ways — `html[data-theme='dark'] .yc-warmbone:not([data-wb-theme='light'])` and `.yc-warmbone[data-wb-theme='dark']` — which is what removed the pre-hydration flash. That behaviour must survive.

If full consolidation is too invasive, the cheaper guard is a test that asserts every `--status-*` and `--ink-*` token the passport reads is declared in **both** halves of the copy. That would have caught the 1.89:1 badge on the commit that introduced it.

Related: #2569, #2574.

Issue #2577, opened 2026-08-30
[Bug]: the a11y e2e suite does not cover /book, /passport or /card


ankit-yc:
### Preflight

- [x] I searched the existing issues and this is not a duplicate.
- [x] This is not a security vulnerability.

### Affected area

Web app (apps/frontend), Build/CI/tooling

### What is broken

`apps/frontend/e2e/a11y.spec.ts` runs axe over exactly four pages:

```
page.goto('/')
page.goto('/pricing')
page.goto('/signin')
page.goto('/signup')
```

It does not cover the three public pages that a stranger is most likely to be sent a link to, and that carry the most personal data:

- `/book/<slug>` — collects a name, email, phone number and an animal's details
- `/passport/<id>` — a shared health record, read as proof of rabies cover
- `/card/<token>` — a shared identity card with owner contact details

That gap is not theoretical. #2567 and #2569 were both **contrast defects on exactly these pages**, and between them they shipped:

- 10 of 56 text nodes below AA on `/book`, including the sentence saying a request is not a booking
- an expired-rabies badge at **1.89:1** while the valid badge sat at 5.77:1
- 58 of 58 nodes at worst 1.34:1 on the in-app passport

Every one of those was found by hand. CI was green throughout.

### Expected

Add these three pages to `a11y.spec.ts`. Two notes on doing it well, learned while fixing the above:

**1. Assert the page actually rendered.** A first attempt at QA on these routes reported "0 failures" while checking **0 text nodes** — the API stub did not match, every page fell into its not-found state, and a vacuous pass read exactly like a real one. Any test here should assert a content floor (a known heading, or a minimum node count) before trusting a zero.

**2. Cover both themes.** All three pages resolve `data-theme` now, and several of the defects above existed *only* in dark. A light-only axe pass would have missed the expired-badge failure entirely.

`/book/no-such-practice-e2e` is already used by `smoke.spec.ts`, so the unavailable states are reachable without fixtures; the populated states need a stub or a seeded record.

PR #2586, opened 2026-08-31
fix(frontend): alias the dark danger ink, and cover the passport's own toggle

ankit-yc:
## PR Checklist

- [x] There is an issue for the bug/feature this PR is for - follow-up to #2577 / #2578

Two P1 review findings landed on #2579 and it merged before they were addressed. Both are correct.

### 1. A stray duplicate token, and a literal where an alias belongs

`html[data-theme='dark']` on `dev` currently contains:

```css
  --status-danger-bg: rgba(234, 55, 41, 0.18);
  --danger-text: #f2938a;                      /* <- stray: already declared 80 lines above */
  --status-danger-border: rgba(242, 147, 138, 0.42);
  ...
  --status-danger-text: #f2938a;               /* <- the same literal a third time */
```

**The stray line is residue from my own mutation test.** The first run of that script broke its paths part-way through, so its restore silently failed; the second run then took its backup *from the already-mutated file*, and every later "restore" faithfully put the mutation back. It never surfaced in a measurement because the real declaration below it still wins, so the passport and the danger pill both measured correct throughout.

The second half is worse, because it is a design mistake rather than an accident: the fix that shipped repeated `#f2938a` as an independent literal in a block that already declares that value — **recreating, inside the change meant to remove it, the exact drift mechanism #2578 is about.**

### 2. The new a11y coverage never exercises the branch #2578 added

#2578 keyed the warm-bone overrides on **disagreement**: they apply only once the reader has pushed the passport away from the root theme. The tests added in #2579 set the OS colour scheme and nothing else, which always leaves the page and the root in agreement — so both new selectors, and every token inside them, went unexercised. A broken selector or a sub-AA pair in either would have left the suite green.

## What is the new behavior?

- The stray `--danger-text` is gone, and `--status-danger-text` is now `var(--danger-text)`. One value, one owner.
- A new test per root theme clicks the passport's own toggle and re-runs axe on the far side, so the disagreement branch is behaviourally covered.

## Validation

`tsc` clean · lint 0 errors · `warmboneTokenClosure` 6/6 · new toggle tests 2/2.

**Mutation-checked, not assumed.** Putting the dark ink into the **light** override — the branch that activates only when the root is dark and the reader asks for light — fails the dark-root test with `insufficient color contrast of 1.8 (foreground #f2938a, background #f6e2da)`, while the light-root test correctly still passes. That is precisely the branch the review said was uncovered.

**Nothing regressed.** The app-wide danger pill is 5.49:1 dark / 5.74:1 light. All six passport theme states are unchanged: 5.49/6.52 dark, 5.74/6.48 light.

## Note for whoever runs this locally

The pre-push hook failed on this branch with `Module '@prisma/client' has no exported member 'PrismaClient'` — the generated client is per-worktree and had not been built here. `pnpm --filter @yosemite-crew/database exec prisma generate` fixes it. Unrelated to this change, but it silently blocks the push rather than reporting why.

### DIFF
--- a/apps/frontend/e2e/a11y.spec.ts
+++ b/apps/frontend/e2e/a11y.spec.ts
@@ -215,6 +215,13 @@ const stubPublicApi = async (page: Page) => {
   );
 };
 
+/**
+ * The warm-bone surface in each theme, read back off the element to prove a
+ * toggle actually repainted rather than only updating React state.
+ */
+const WARM_BONE_LIGHT_SCREEN = '#f7f3ec';
+const WARM_BONE_DARK_SCREEN = '#2f271e';
+
 for (const theme of ['light', 'dark'] as const) {
   test.describe(`Shared public pages — accessibility, ${theme} (WCAG 2.1 AA incl. contrast)`, () => {
     test.use({ colorScheme: theme });
@@ -255,6 +262,18 @@ for (const theme of ['light', 'dark'] as const) {
       // Now it disagrees with the root, which is the only state that activates
       // the overrides.
       await expect(main).toHaveAttribute('data-wb-theme', theme === 'dark' ? 'light' : 'dark');
+
+      // The attribute alone proves only that React updated its state. If either
+      // disagreement selector is broken or renamed, the attribute still flips,
+      // the override simply never applies, and the page sits in the perfectly
+      // valid root theme - where axe passes and the regression goes unseen.
+      // Verified by renaming both selectors: the attribute assertion stayed
+      // green. So assert the PALETTE repainted, which is what they exist to do.
+      const surface = await main.evaluate((el) =>
+        getComputedStyle(el).getPropertyValue('--screen').trim()
+      );
+      expect(surface).toBe(theme === 'dark' ? WARM_BONE_LIGHT_SCREEN : WARM_BONE_DARK_SCREEN);
+
       await expect(page.getByText(/expired/i).first()).toBeVisible();
 
       const results = await runAxeWithContrast(page);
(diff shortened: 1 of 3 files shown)


aupyay (dismissed):

ankit-yc:
Right, and it caught a flaw in my own mutation reasoning. Fixed in af12b1f.

The assertion checked `data-wb-theme` flipped and then ran axe — which proves React updated its state and nothing more. If either disagreement selector is broken or renamed, the attribute still flips, the override simply never applies, and the page sits in the perfectly valid root theme, where axe passes and the regression is invisible.

**I verified your claim rather than taking it on faith:** renaming both selectors so neither override applies leaves the old assertion **green, 2/2**. So the test genuinely did not cover the thing it was added for.

My earlier mutation check missed it because it broke a token *value* inside an override that still applied — the weaker mutation. Renaming the selector is the stronger one, and it is the realistic regression.

It now reads `--screen` back off the element and asserts the opposite theme's surface before running axe. Re-run against the same mutation: both tests fail with `Expected "#2f271e", Received "#f7f3ec"`.

**One related mutation colour cannot catch, so I asserted it structurally instead.** Dropping the `:not()` from the dark override makes it apply under a dark root as well — where it paints the same dark values, so every rendered pixel is identical and no contrast test can ever see it. What it quietly restores is defect 2's mechanism: the copy declaring tokens in the agreeing state, shadowing `body:has([data-yc-app])` again. `warmboneTokenClosure` now asserts each override's scope directly and fails 5 tests on that mutation.

`tsc` clean, lint 0 errors, 29 unit tests, 18/18 Playwright.

aupyay (approved):

PR #2627, opened 2026-09-02
fix(frontend): show clipped workspace text in full on hover

ankit-yc:
## PR Checklist

Workspace search result rows clip both the name and the origin line, and neither carries a tooltip. The origin line is the one that loses meaning when it clips.

Searching `amoxicillin` in the treatment step on dev returns, among others:

```
amoxicillin and beta-...        QJ01CR02 · ANTIBACTERI...
amoxicillin and enzy...         QJ51CR02 · ANTIBACTERI...
```

Those are **ANTIBACTERIALS FOR SYSTEMIC USE** and **ANTIBACTERIALS FOR INTRAMAMMARY USE**. Truncated, they are one digit apart, and the digit is the only thing on screen that distinguishes them.

## What is the new behavior?

Both lines carry a `title`, so hovering reveals the full string. The text was already complete in the DOM for assistive tech; this gives a sighted user the same.

Not while the row is disabled. A nested `title` beats the button's own, and on an already-added row "Added" is the more useful thing to say, so the tooltips are dropped in that state and the existing behaviour is untouched.

## Tests

The component had no tests at all. This adds the first two: one for the tooltips, one for the disabled row keeping "Added" as its only tooltip.

Both were mutation-checked. Dropping the origin `title`, and ungating it so it shadows "Added", each fail a test.

## Related Issue(s)

Found while visually verifying the ATCvet prescribing work (#2601, #2619) on dev.

### DIFF
--- a/apps/frontend/src/app/__tests__/pages/AppointmentWorkspace/TotalBillContainer.test.tsx
+++ b/apps/frontend/src/app/__tests__/pages/AppointmentWorkspace/TotalBillContainer.test.tsx
@@ -70,6 +70,24 @@ const renderBill = (
 const overallDiscountInput = () => screen.getByLabelText('Overall discount percent');
 
 describe('TotalBillContainer', () => {
+  it('carries the billed item name as a tooltip, since the column clips it', () => {
+    /* On dev this column renders 150px of text in an 88px box, so
+       "Anal Gland Expression" shows as "Anal Gland...". It is what the client
+       is being charged for, so the full name has to stay reachable. */
+    const { view } = renderBill({
+      ...baseItem,
+      id: 'line-long',
+      name: 'Anal Gland Expression',
+    });
+    /* The name also appears in the remove button's label, and the panel heading
+       is itself a truncating span, so find the cell by both class and text. */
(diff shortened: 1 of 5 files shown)


ankit-yc:
Extended after finding the same defect on the invoice step while sweeping.

The billed item-name column renders 150px of text in an 88px box on dev, so "Anal Gland Expression" shows as "Anal Gland..." with nothing to reveal the rest. That is the line the client is being charged for. Both places that render it now carry a title: the running Total Bill and the saved invoice.

Retitled to match the wider scope. Same mutation check as before - dropping the title fails the new test.

ankit-yc:
On scope, since the obvious next question is "why only these two".

A sweep for clipped dynamic values with no tooltip finds 56 candidates across `apps/frontend/src/app`. I deliberately did not touch them: `truncate` only clips when the content actually overflows its box, so most of those are labels and dropdown options that fit fine, and a blanket change would be a large diff justified by nothing but a grep.

These two are the ones measured clipping on dev:

- search result origin: full ATCvet class path, `scrollWidth` well past the row
- billed item name: 150px of text in an 88px box

If the pattern is worth applying more widely it should be a deliberate pass with each site checked at real widths, not folded in here.

PR #2632, opened 2026-09-02
fix(frontend): say why a scoped coded-term search found nothing

ankit-yc:
## PR Checklist

Follow-up to #2626, which added the All / VeNom / SNOMED scope to the coded-term picker.

With a scope on, a query that matches no term in that vocabulary leaves the dropdown closed — the picker opens only when `results.length > 0`. Nothing distinguishes "no term here carries a SNOMED code" from "search is broken", and the scope control is exactly what makes that case common: on dev, 27 terms match `vom` but only 10 hold a usable SNOMED crosswalk.

## What is the new behavior?

The dropdown opens with one line naming the vocabulary and the query, and a button that widens the search back to All.

Deliberately narrow:

- **Only when a scope is on.** An unscoped miss is just a query with no matches, and the closed dropdown says that well enough.
- **Not on error.** A failed request is not "nothing matched", so the explanation stays off and the existing `console.error` path is untouched.

## Tests

Three new tests, and a note on one of them.

My first version of "stays quiet when an unscoped search finds nothing" asserted absence right after typing. It passed — and it also passed against a mutant that dropped the scope check entirely, because the assertion ran before the response landed. It was measuring nothing.

It now proves the notice *can* appear for that same empty response under SNOMED, then switches to All and shows it goes away. That version does fail the mutant.

Both mutants are killed: removing the scope condition, and never recording an empty result.

## Related Issue(s)

Follows #2626.

### DIFF
--- a/apps/frontend/src/app/__tests__/pages/AppointmentWorkspace/SoapCodedTermPicker.test.tsx
+++ b/apps/frontend/src/app/__tests__/pages/AppointmentWorkspace/SoapCodedTermPicker.test.tsx
@@ -344,3 +344,83 @@ describe('SoapCodedTermPicker vocabulary scope', () => {
     expect(lastCall()).not.toHaveProperty('vocabulary');
   });
 });
+
+describe('SoapCodedTermPicker scoped empty state', () => {
+  beforeEach(() => {
+    jest.useFakeTimers();
+    suggestMock.mockReset();
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
+  const renderPicker = () =>
+    render(
+      <SoapCodedTermPicker
+        sectionLabel="Subjective"
+        domain="PresentingComplaint"
+        selected={[]}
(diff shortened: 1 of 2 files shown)


ankit-yc:
Confirmed on dev now that the API side of #2626 is deployed, and the empty case turns out to be easy to hit rather than hypothetical.

In the Assessment section (`domain=Diagnosis`), searching `vomit`:

- **SNOMED** — 8 results, every one carrying a SNOMED code, and the two VeNom-only terms that showed before the filter shipped (`YC-005207 · VeNom 5782`, `YC-005206 · VeNom 17963`) are correctly gone.
- **VeNom** — **0 results**. The request returns 200, the list is empty, and the dropdown simply does not open. Nothing on screen distinguishes that from a broken search.

The VeNom-native vomiting terms live in the PresentingComplaint domain, so a Diagnosis-domain VeNom search legitimately has nothing to offer — which is exactly the case this PR explains instead of leaving blank.

PR #2633, opened 2026-09-02
fix(mobile): stop telling reporters their adverse event was sent

ankit-yc:
## What

Submitting an adverse event ends on a success alert reading **"We sent your report to the manufacturer."** Nothing is sent to anyone.

## Why it is wrong

`AdverseEventService` has no outbound call of any kind - no HTTP client, no mail, no queue. It writes an `AdverseEventReport` row whose `destinations` field records `sendToManufacturer` / `sendToHospital`.

So the destination is stored as an intention; the delivery never happens.

The hospital branch at least resolves to something a clinic can reach, since the row is org-scoped and the service lists by organisation. No manufacturer has any access to this database at all.

This matters more than a usual copy slip. A pet owner who believes a manufacturer already has their reaction report has no reason to tell anyone else, and the report is the entire point of the flow.

## Change

| | Before | After |
|---|---|---|
| Title | Report submitted | Report saved |
| Manufacturer | We sent your report to the manufacturer. | Saved for the manufacturer. Forwarding is not switched on yet, so tell your vet too. |
| Hospital | We sent your report to the hospital. | Saved to your vet practice, with the product and batch details attached. |

The batch detail is accurate: `AdverseEventProductInfo` captures batch number, quantity and unit, frequency, administration method and the before/after condition, and the flow refuses to submit without a linked hospital.

## Validation

The existing assertions could not have caught this. They matched `stringContaining('manufacturer')`, which the old sentence satisfied perfectly.

They now pin the real copy **and** assert the message does not claim transmission.

Mutation-checked rather than assumed:

- restoring the original sentence: **1 failed, 23 passed**
- with the fix in place: **24 passed**

mobile `tsc` clean, `eslint --max-warnings=0` clean.

## Impact

Mobile adverse-event reporting flow. Copy and its tests only, no behaviour change.

Found by a sweep for claims nothing in the system can support.

### DIFF
--- a/apps/mobileAppYC/__tests__/features/adverseEventReporting/screens/ThankYouScreen.test.tsx
+++ b/apps/mobileAppYC/__tests__/features/adverseEventReporting/screens/ThankYouScreen.test.tsx
@@ -218,8 +218,16 @@ describe('ThankYouScreen', () => {
     // Covers lines 119-122 (productFiles handling)
     expect(mockSetProductInfo).toHaveBeenCalled();
     expect(showSuccessAlert).toHaveBeenCalledWith(
-      'Report submitted',
+      'Report saved',
       expect.stringContaining('manufacturer'),
+    );
+    // The report is stored with `destinations`, never transmitted, so the
+    // confirmation must not claim it was. `stringContaining('manufacturer')`
+    // alone passed the old copy - "We sent your report to the manufacturer" -
+    // which is why this asserts the absence too.
+    expect(showSuccessAlert).not.toHaveBeenCalledWith(
+      expect.anything(),
+      expect.stringMatching(/\bwe sent\b|\bsent to\b|\bforwarded\b/i),
     );
     expect(mockResetDraft).toHaveBeenCalled();
     expect(mockParentNavigate).toHaveBeenCalledWith('Home');
@@ -243,8 +251,12 @@ describe('ThankYouScreen', () => {
 
     await waitFor(() => {
       expect(showSuccessAlert).toHaveBeenCalledWith(
-        'Report submitted',
-        expect.stringContaining('hospital'),
+        'Report saved',
+        expect.stringContaining('vet practice'),
+      );
+      expect(showSuccessAlert).not.toHaveBeenCalledWith(
+        expect.anything(),
+        expect.stringMatching(/\bwe sent\b|\bsent to\b|\bforwarded\b/i),
       );
     });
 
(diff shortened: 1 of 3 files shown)


aupyay:
I think we have a list and we need to make sure in that case - emails are sent correctly to manufacturers and if not create and actually make this feature work it.

ankit-yc:
Agreed on the direction. I dug into what we actually have, because the answer changes what I can build without guessing.

**There is a list, and it is authorities, not manufacturers.** `apps/mobileAppYC/src/features/adverseEventReporting/content/supportedCountries.ts` has 18 countries, each with the regulator's name: FDA Center for Veterinary Medicine, Health Canada VDD, the VMD, HPRA, ANMV, and so on.

**But it carries names only.** No email address and no submission URL anywhere in that file. So there is currently nothing to send *to*.

For manufacturers specifically there is no list at all. The schema has `manufacturer` as free text on `Immunization`, `ParasiteTreatment`, `DrugFormulary` and `ClinicEquipment`, and `InventoryVendor` is a per-clinic supplier table with a `contactInfo` JSON. None of that is a manufacturer pharmacovigilance contact, and none of it is global.

What the report itself captures is `productName`, `brandName`, `manufacturingCountry` and `batchNumber` - enough to identify the product, not enough to route a message.

**What I can build now, with no external data:**

The clinic side is fully within our control. The report already requires a linked hospital and is org-scoped, and the backend already has working SES (`sendEmail` / `sendEmailTemplate` in `apps/backend/src/utils/email.ts`, used by `appointment.service.ts` and others). So "send to hospital" can become a real send rather than a stored intention, and that is worth doing on its own.

**What I need from you:**

Manufacturer and regulator delivery. Two things I should not invent:

1. The actual destination addresses. Regulator submission is usually a portal or a specific pharmacovigilance address per country, not a generic inbox, and manufacturer contacts differ per company.
2. Whether we should be transmitting at all yet. Sending a pet owner's adverse event report to a national regulator is a regulated act, and the consent copy currently says data "will only be shared with veterinary professionals, manufacturers, or regulators for safety investigations" - which is a promise we would then be starting to keep for real.

**Proposal, unless you say otherwise:** I build the clinic email now so one destination genuinely works, plus a contact registry keyed by country and manufacturer that is empty by default, with the send gated on a configured address. Populating it turns forwarding on without another code change. The honest copy in this PR stays only for destinations that genuinely do not send yet, and flips to "sent" per destination as each one becomes real.

Tell me the addresses (or that you want me to source and verify them) and I will wire it up.

PR #2655, opened 2026-09-02
fix(frontend): bring estimates and credit notes up to the app's design

ankit-yc:
## What changed

Two surfaces shipped recently did not look like the rest of the product. Both are now built from the patterns their immediate neighbours already use, rather than from a second set of choices.

### Estimates

It is reached from Finance and reads as the same surface, but had a small plain heading, no count, no metrics, and its status pills on a row of their own below the title.

It now uses the **Finance header exactly**: `text-page-title` with a live count, the info affordance, a sub-line, then the filter and page actions on the right of that same row.

The sub-line says what is awaiting a decision and what is approved. `CONVERTED` and `DECLINED` are excluded from both - a converted estimate's money is counted on the invoice it became, so adding it here would double it.

The table drew a bare text link where the invoice table beside it draws a species-tinted avatar, a name, right-aligned bold money and a circular row action. It now draws **the same row**, so the two finance tables read as one product.

### Credit notes

The ledger sits directly below Payments inside the invoice panel, but was styled unlike it: no glyph, no bold title, and the `CN-` reference at full body weight with the reason beneath it in grey.

- The row now has **the Payments ledger's anatomy** - tinted glyph, bold line, quiet caption, amount.
- **The reason leads and the reference captions it.** Someone reconciling reads "Goodwill on the delayed dental"; the `CN-...` string is the audit artifact. A note without a reason still gets a title, so rows keep one height instead of the reference jumping into the headline slot.
- **Issuing is now the primary action.** It was a `Secondary` sitting beside no primary at all, which read as the least important control in the panel.
- **The amount field carries the currency symbol inside it**, the way the discount cap carries its `%`. It was the one number on the panel with no unit, beside a column of formatted money.

## Impact area

`apps/frontend` only - finance estimates page and list, credit-note ledger and issue form. No API, type or behaviour change beyond the header metrics.

## Validation

- `npx tsc --noEmit` clean; `pnpm run lint` **0 errors**
- **31 suites / 466 tests pass** across the estimates, credit-note and finance areas
- Coverage: `EstimateList.tsx`, `CreditNoteLedger.tsx`, `CreditNoteIssueForm.tsx` all **100/100/100/100**; Estimates page 99.26% statements
- Both new guards **mutation-checked**: letting `CONVERTED` count toward approved fails the metrics test and only that test; counting all estimates rather than the visible ones fails the title test and only that test
- Local Sonar proxy (279 sonarjs rules + `jsx-a11y/prefer-tag-over-role`, self-tested against a deliberately failing probe): **no findings on any changed line** (the single report is a pre-existing `role="dialog"` in a test mock at line 45; the only changed line in that file is 173)
- Visually verified in Storybook against the live dev screenshots that prompted this

## Notes

`tsconfig.json` excludes `**/*.stories.tsx`, so `tsc` does **not** type-check stories - the `companionName` -> `companion` prop rename broke `EstimateList.stories.tsx` silently and was caught only by opening it. Fixed here.

### DIFF
--- a/apps/frontend/src/app/__tests__/features/finance/Sections/EstimateSections.test.tsx
+++ b/apps/frontend/src/app/__tests__/features/finance/Sections/EstimateSections.test.tsx
@@ -132,7 +132,10 @@ describe('EstimateList', () => {
   ];
 
   const names: Record<string, string> = { 'pet-1': 'Rex', 'pet-2': 'Milo' };
-  const companionName = (patientId: string) => names[patientId] ?? 'Unknown companion';
+  const companion = (patientId: string) => ({
+    name: names[patientId] ?? 'Unknown companion',
+    speciesCode: 'dog',
+  });
 
   const renderList = (overrides: Partial<React.ComponentProps<typeof EstimateList>> = {}) => {
     const onSelect = jest.fn();
@@ -141,7 +144,7 @@ describe('EstimateList', () => {
         estimates={estimates}
         activeEstimateId={null}
         onSelect={onSelect}
-        companionName={companionName}
+        companion={companion}
         {...overrides}
       />
     );
--- a/apps/frontend/src/app/__tests__/features/finance/Sections/InvoiceCreditNotes.test.tsx
+++ b/apps/frontend/src/app/__tests__/features/finance/Sections/InvoiceCreditNotes.test.tsx
@@ -10,13 +10,19 @@ jest.mock('@/app/ui/layout/guards/PermissionGate', () => ({
   PermissionGate: ({ children }: any) => <div>{children}</div>,
 }));
 
-jest.mock('@/app/ui/primitives/Buttons', () => ({
-  Secondary: ({ text, ariaLabel, onClick, isDisabled }: any) => (
+jest.mock('@/app/ui/primitives/Buttons', () => {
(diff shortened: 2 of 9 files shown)


ankit-yc:
Added a third surface after re-checking the rest of what shipped recently.

**Breed substitution notice.** Its siblings in the same `AppointmentInfo` folder - `Finance/Details`, `Finance/Summary`, `InvoicePaymentActions` - are all `rounded-2xl` with a border and `px-4 py-3`. This one shipped as a borderless `rounded-xl` block with `p-2.5`, so it read as a different kind of object sitting among them.

The vocabulary-gap cases (`UNMAPPED_BREED`, `UNCODED_BREED`) also move from the neutral `card-hover` tint to the warning tint those siblings use. A substitution is notable - a different clinical claim reached the lab than the one on the companion's record - and neutral grey said otherwise. `MISMATCHED_BREED` keeps the danger tint, because that one is a defect on the patient record rather than a gap in the provider's vocabulary.

Rebased onto dev to pick up #2653. Re-verified: **34 suites / 530 tests pass**, lint 0 errors, `tsc` clean, `BreedSubstitutionNotice.tsx` still 100% coverage, and both states checked in Storybook.

PR #2658, opened 2026-09-02
chore(backend): clear the remaining Zod v4 deprecations

### LINKED ISSUE #2652: chore(repo): clear all SonarCloud code smells (Zod v4 deprecations + 3 frontend)
## Summary

## Problem / motivation

| project | code smells |
| --- | --- |
| Backend | 353 |
| Frontend | 3 |
| Desktop | 0 |
| MobileAppYC | 0 |

## Proposed solution

| --- | --- | --- |

## Scope

### Out of scope

## Acceptance criteria

## Test plan

## Alternatives considered

## Dependencies & risks

ankit-yc:
Closes #2652 for the backend.

## What changed

Three categories, 67 call sites across 21 files:

| deprecated | replacement | sites |
|---|---|---|
| `parsed.error.flatten()` | `z.flattenError(parsed.error)` | 48 |

## The issue's counts are stale

#2652 lists **eleven** categories totalling 353. Measured against `origin/dev` before starting, **eight of them are already at zero**: `datetime`, `uuid`, `nativeEnum`, `url`, `email`, `merge`, `finite`, `date`. Only these three remained, which is 67 sites rather than 353.

That also means the ordering trap the issue flags — where a `.trim()` before a validator has to become `z.string().trim().pipe(z.email())` rather than `z.email().trim()` — **does not arise here**. It applies only to the string validators, all of which are already migrated.

## Equivalence was checked, not assumed

Against the installed zod (4.4.3), not the migration guide:

```
err.flatten():        {"formErrors":[],"fieldErrors":{"a":["Invalid input: expected string, received number"]}}
z.flattenError(err):  {"formErrors":[],"fieldErrors":{"a":["Invalid input: expected string, received number"]}}
identical: true

passthrough({a:'x', extra:1}) -> {"a":"x","extra":1}
loose({a:'x', extra:1})       -> {"a":"x","extra":1}

z.ZodIssueCode.custom === "custom"
```

## Validation

- **459 suites / 11079 backend tests pass**
- **Mutation-checked**: replacing `z.flattenError(...)` with the raw error object in `code.controller.ts` fails 5 tests in `code.controller.test.ts`, so the flattened shape is genuinely guarded rather than incidentally passing
- Local Sonar proxy (279 sonarjs rules, self-tested against a deliberately failing probe): **no findings on any changed file**
- Type-check and lint output are **identical before and after, error for error**

## Two pre-existing conditions worth flagging separately

Both measured on `dev` with my changes reverted, and unchanged by this PR:

- `apps/backend` `tsc --noEmit` reports **450 errors** (largely `TS7006 implicitly any` in `auth-hooks`, `search.controller` and the backfill scripts)
- `pnpm --filter backend run lint` reports **13270 errors**

Neither is caused by this change and neither is in scope here, but a repo whose backend type-check and lint are both red by default cannot use either as a gate. Happy to file that if it is not already known.

## Impact area

`apps/backend` only — controllers and services. No behaviour change: every replacement is byte-equivalent to what it replaced.

### DIFF
--- a/apps/backend/src/controllers/app/chat.controller.ts
+++ b/apps/backend/src/controllers/app/chat.controller.ts
@@ -115,7 +115,10 @@ export const ChatController = {
       if (!parsed.success) {
         return res
           .status(400)
-          .json({ message: "Invalid payload", errors: parsed.error.flatten() });
+          .json({
+            message: "Invalid payload",
+            errors: z.flattenError(parsed.error),
+          });
       }
 
       const record = await SharedChatEntityService.shareEntity({
--- a/apps/backend/src/controllers/web/activitypub.controller.ts
+++ b/apps/backend/src/controllers/web/activitypub.controller.ts
@@ -386,7 +386,7 @@ export const ActivityPubController = {
       if (!parsed.success) {
         return res.status(400).json({
           error: "Invalid referral payload",
-          details: parsed.error.flatten().fieldErrors,
+          details: z.flattenError(parsed.error).fieldErrors,
         });
       }
       const body = parsed.data;
--- a/apps/backend/src/controllers/web/case-encounter.controller.ts
+++ b/apps/backend/src/controllers/web/case-encounter.controller.ts
@@ -18,10 +18,10 @@ import { resolveVerifiedUserId } from "src/utils/request";
 
 const caseResourceSchema = z
   .object({ resourceType: z.literal("EpisodeOfCare") })
-  .passthrough();
+  .loose();
 const encounterResourceSchema = z
   .object({ resourceType: z.literal("Encounter") })
-  .passthrough();
+  .loose();
 
 const caseListQuerySchema = z.object({
   organization: z.string().trim().optional(),
@@ -55,7 +55,7 @@ const dischargeEncounterSchema = z
       )
       .optional(),
   })
-  .passthrough();
+  .loose();
 const assignUnitSchema = z
   .object({
     resourceType: z.literal("Parameters").optional(),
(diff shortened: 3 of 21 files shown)


ankit-yc:
**Correction to this PR's description.** It claimed two pre-existing conditions on `dev`:

> `apps/backend` `tsc --noEmit` reports **450 errors**
> `pnpm --filter backend run lint` reports **13270 errors**

**Both were wrong, and both were my own environment.** I had not run `pnpm --filter @yosemite-crew/database run build` in that worktree, so `@yosemite-crew/database` had no `dist`. Every type importing through it resolved to an error type, which is why `@typescript-eslint/no-unsafe-*` alone accounted for over 13,000 of them:

```
5337  @typescript-eslint/no-unsafe-member-access
3791  @typescript-eslint/no-unsafe-assignment
2251  @typescript-eslint/no-unsafe-call
```

Measured on `dev` at `8c5bcf74e` (this PR merged) after building that one package:

```
pnpm --filter backend run type-check   -> exit 0
npx tsc --noEmit                       -> exit 0, 0 errors
pnpm --filter backend run lint         -> exit 0, clean
```

CI was right the whole time. `core / Static (backend)` runs the identical `pnpm --filter backend run lint`, and it was green on this PR - I should have treated that as the authority and reconciled against it instead of asserting a repo-wide condition from a single local run. Checking the CI job's steps is what settled it.

**What this does not change:** the migration itself. The before/after comparison was between two runs of the same broken setup, so the differential still held - identical error sets meant the change added nothing - and the 459 suites / 11079 tests were real. I have since re-run the checks on a correctly built tree and both are clean.

Apologies for the noise in the description; I have corrected the equivalent paragraph in [#2660](https://github.com/YosemiteCrew/Yosemite-Crew/pull/2660) rather than leaving it to spread.

Issue #2643, opened 2026-09-02
chore(backend): migrate Documenso createDocument off the deprecated createV0


ankit-yc:
## Summary

Migrate `DocumensoService.createDocument` off the deprecated `documenso.documents.createV0()` onto the supported `documents.create()`, and remove the `// NOSONAR` that currently suppresses the deprecation warning.

## Problem / motivation

`apps/backend/src/services/documenso.service.ts` calls `documenso.documents.createV0()`. The SDK marks it:

> @deprecated method: This will be removed in a future release, please migrate away from it as soon as possible.

It is the only remaining SonarQube code smell in the backend project. Every other deprecation was cleared in the Zod v4 sweep; this one was held back because it is a behavioural change to the e-signature path rather than a mechanical rewrite, and it cannot be verified without a live Documenso instance. It is suppressed today with a justified `// NOSONAR` so the `--issues 0` CI gate can pass.

## Proposed solution

Replace the single `createV0` call with the supported flow:

1. `documents.create({ payload, file })` - the payload carries `title` and the same `recipients[].fields[]` shape already built today, and `file` accepts `{ fileName, content }` where `content` may be a `Uint8Array`, so the existing `pdf: Buffer` can be passed directly.
2. `documents.get({ documentId })` to read back the created document, because `create()` returns only `{ envelopeId, id }`.
3. Delete `uploadPdfBuffer` - `create()` uploads the file directly, so the separate PUT to `uploadUrl` disappears.
4. Remove the `// NOSONAR` and its explanatory comment block.

## Scope

- `apps/backend/src/services/documenso.service.ts`
- Test doubles for the Documenso SDK in the affected backend suites.

### Out of scope

- Any change to signing semantics, field placement maths, or the attestation records written by callers.
- The other Documenso methods (`distribute`, `get`, `downloadSignedDocument`), which are not deprecated.

## Current state / where it plugs in

`createDocument` returns the created `document`, and three callers depend on the shape:

- `src/services/workspace-document-packet.service.ts` reads `doc.id` and `doc.recipients[0].token` to build `${DOCUMENSO_URL}/sign/${token}`.
- `src/services/pet-clinical-records.service.ts` reads `document.id` and stores it as `documensoDocumentId` on the attestation.
- `src/services/rendered-document.service.ts` uses the same return value.

[...]

PR #2661, opened 2026-09-02
chore(backend): take the backend Sonar smell count to zero

### LINKED ISSUE #2652: chore(repo): clear all SonarCloud code smells (Zod v4 deprecations + 3 frontend)

ankit-yc:
## PR Checklist

> **Scope note.** This PR started as the whole 353-smell backend sweep. While it was open, #2658 and then #2663 landed the Zod deprecation work, and #2659 did the frontend. The Zod count on `dev` is now 0, so all of that has been dropped from this branch. What is left is the two findings none of those touched, which are the last things standing between the backend and a green `--issues 0` gate.

After #2663, `apps/backend` has no remaining Zod deprecations. Two SonarCloud findings remain:

1. Three `typescript:S6606` in `template.service.ts` - Sonar wants `X === undefined ? Y : X` written as `X ?? Y`.
2. One `typescript:S1874` in `documenso.service.ts` - the deprecated `documents.createV0()`.

`scripts/ci/sonar-thresholds.mjs` enforces 0 issues, so the gate stays red on either.

## What is the new behavior?

**The three ternaries become `??`.** This one is worth a second look rather than a rubber stamp, because that file carries a comment warning about exactly this distinction - the concern being that a v4 `.default().optional()` resolution could write `{}` over a stored snapshot the caller never mentioned.

`??` is safe here specifically because of the types involved. `schemaSnapshot`, `renderConfigSnapshot` and `validationSnapshot` are all `.optional()` applied to non-nullable schemas (`z.object({...})` and `z.record(z.string(), z.unknown())`). That makes each one `T | undefined`, never `null`. `X === undefined ? Y : X` and `X ?? Y` differ only when `X` can be `null`, so here they are exactly equivalent. The "absent means leave the stored value alone" behaviour the comment protects is preserved.

**The Documenso call gets a justified `NOSONAR`, not a migration.** Its replacement `documents.create()` returns only `{ envelopeId, id }` - no `uploadUrl`, and crucially no `recipients`. Three callers depend on the current return value:

- `workspace-document-packet.service.ts` reads `doc.id` and `doc.recipients[0].token` to build `${DOCUMENSO_URL}/sign/${token}`
- `pet-clinical-records.service.ts` reads `document.id` and stores it as `documensoDocumentId` on the attestation
- `rendered-document.service.ts` uses the same value

So migrating is not a swap: it needs a second `documents.get()` round-trip to recover the recipient tokens, plus a rewrite of the upload path (`create()` uploads the file directly, so `uploadPdfBuffer` goes away). That is a behavioural change on the flow that produces signed clinical documents, and it cannot be verified from a dev machine - the only "proof" available here would be mocks written alongside the change, which prove nothing about the real API contract.

It is therefore suppressed with the reasoning inline and tracked in #2643, where it can get its own PR and a real integration test.

The call was also lifted into a `request` variable. That is not cosmetic: Sonar only honours `NOSONAR` on the line it reported, and Prettier reflows a multi-line call so that the marker lands inside the object literal or on the closing paren, silently suppressing nothing. A one-line call is the only placement that survives formatting.

## How this was verified

- `pnpm --filter backend run lint` (now including the `eslint-plugin-sonarjs` rules #2663 wired in): clean
- `pnpm --filter backend run type-check`: clean
- Targeted Jest for both touched services: 15 suites, 357 tests, green
- Prettier clean, and the `NOSONAR` re-checked after formatting to confirm it is still on the `createV0` line

The one thing not verifiable locally is whether SonarCloud accepts the `NOSONAR`; eslint does not honour the marker, so that is confirmed by the `sonar` check on this PR rather than by a local run.

## Related Issue(s)

Fixes #2652

Follow-up: #2643

### DIFF
--- a/apps/backend/src/controllers/app/chat.controller.ts
+++ b/apps/backend/src/controllers/app/chat.controller.ts
@@ -113,9 +113,10 @@ export const ChatController = {
 
       const parsed = shareEntitySchema.safeParse(getObjectBody(req));
       if (!parsed.success) {
-        return res
-          .status(400)
-          .json({ message: "Invalid payload", errors: parsed.error.flatten() });
+        return res.status(400).json({
+          message: "Invalid payload",
+          errors: z.flattenError(parsed.error),
+        });
       }
 
       const record = await SharedChatEntityService.shareEntity({
--- a/apps/backend/src/controllers/app/finance.controller.ts
+++ b/apps/backend/src/controllers/app/finance.controller.ts
@@ -34,7 +34,7 @@ const CreateInvoicePaymentSessionBodySchema = z.object({
   provider: z.string().trim().min(1).optional(),
   // Major units. Present when the caller is collecting a deposit rather than
   // the whole outstanding balance.
-  depositAmount: z.number().finite().positive().optional(),
+  depositAmount: z.number().positive().optional(),
 });
 
 const InvoiceItemBodySchema = z.object({
@@ -145,8 +145,8 @@ const SubscriptionCheckoutCompletedBodySchema = z.object({
     ])
     .optional(),
   cancelAtPeriodEnd: z.boolean().optional(),
-  currentPeriodStart: z.string().datetime().optional(),
-  currentPeriodEnd: z.string().datetime().optional(),
+  currentPeriodStart: z.iso.datetime().optional(),
+  currentPeriodEnd: z.iso.datetime().optional(),
   livemode: z.boolean().optional(),
   seatQuantity: z.number().int().nonnegative().optional(),
 });
@@ -167,10 +167,10 @@ const SubscriptionUpdatedBodySchema = z.object({
     ])
     .optional(),
   cancelAtPeriodEnd: z.boolean().optional(),
-  canceledAt: z.string().datetime().optional(),
+  canceledAt: z.iso.datetime().optional(),
   seatQuantity: z.number().int().nonnegative().optional(),
-  currentPeriodStart: z.string().datetime().optional(),
-  currentPeriodEnd: z.string().datetime().optional(),
+  currentPeriodStart: z.iso.datetime().optional(),
+  currentPeriodEnd: z.iso.datetime().optional(),
 });
 
 const SubscriptionLifecycleBodySchema = z.object({
@@ -186,7 +186,7 @@ const UsageEventBodySchema = z.object({
   referenceType: z.string().trim().min(1).optional(),
   referenceId: z.string().trim().min(1).optional(),
   metadata: z.record(z.string(), z.unknown()).optional(),
-  occurredAt: z.string().datetime().optional(),
+  occurredAt: z.iso.datetime().optional(),
 });
 
 const UsageSnapshotBodySchema = z.object({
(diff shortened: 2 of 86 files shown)


ankit-yc:
Closing as superseded, and the duplication was mine to avoid.

This PR was opened at **22:12**. I started the same work and opened [#2663](https://github.com/YosemiteCrew/Yosemite-Crew/pull/2663) at **22:40**, without re-checking open PRs after my session began - I had checked at the start, before this existed. #2663 merged first, which is what left this one conflicted.

**85 of the 86 files here are already on `dev`**, via [#2658](https://github.com/YosemiteCrew/Yosemite-Crew/pull/2658) (the `flatten`/`passthrough`/`ZodIssueCode` categories), [#2659](https://github.com/YosemiteCrew/Yosemite-Crew/pull/2659) (the three frontend smells) and [#2663](https://github.com/YosemiteCrew/Yosemite-Crew/pull/2663) (the remaining 283 Zod sites plus wiring `eslint-plugin-sonarjs` into both configs).

The single file not covered is `documenso.service.ts`, and this PR reaches **the same conclusion I did independently**: `createV0` stays, because `documents.create()` returns only `{ envelopeId, id }` and the callers need `recipients[0].token`, so migrating is a behavioural change to the e-signature path that cannot be verified without a live instance. That is #2643. #2663 handles the resulting lint finding with a narrow config-level exception for that one file, pointing at the same issue.

Two things from this branch that are worth keeping, and which I have carried over or recorded:

- The reasoning comment on the `createV0` call is good and says exactly why it stays. #2663's config exception says the same thing in the eslint config; if you prefer it in the source too, that is a one-line addition.
- I posted the SDK type verification on #2643 (the response shape really is `{ envelopeId, id }`, confirmed against `@documenso/sdk-typescript@0.9.0`), so whoever picks that up does not have to re-derive it.

Nothing here is lost. Reopen if I have missed something - I compared the file sets rather than the diffs, so if a specific hunk differs in substance rather than in path, say which and I will port it.

ankit-yc:
Agreed on the close, and thanks for the detailed comparison - the file-set analysis was right.

Taking you up on the reopen offer: two hunks did differ in substance rather than in path, so they are now in #2671 off the same branch (reopening was refused because the branch head moved after the close).

1. The three `S6606` nullish-coalescing findings in `template.service.ts` (lines 967, 1028, 1033). Not part of the Zod work, so none of #2658/#2659/#2663 touched them, and they are still on `dev`.

2. The `NOSONAR` on `documenso.service.ts`. This one is worth a look, because #2663's eslint exception says the file "already carries a NOSONAR and an explanation there". On `dev` it does not:

```
$ git show origin/dev:apps/backend/src/services/documenso.service.ts | grep -c NOSONAR
0
$ git show origin/dev:apps/backend/sonar-project.properties | grep -ci 'documenso\|S1874'
0
```

So `sonarjs/deprecation` is off for that file locally, but SonarCloud has nothing suppressing it and will still raise `typescript:S1874`. #2671 adds the marker, which makes that config comment true.

Everything else from #2661 is dropped - #2671 is 2 files.

PR #2665, opened 2026-09-02
chore(repo): land the typescript-eslint and minor-and-patch groups with their fixes

ankit-yc:
## PR Checklist

Two dependabot group PRs are open and both are red on `core / Static`:

- **#2647** (typescript-eslint group, 5 updates) fails `core / Static (backend)` with two `@typescript-eslint/no-unnecessary-type-assertion` errors.
- **#2651** (minor-and-patch group, 37 updates) fails `core / Static (mobileAppYC)` with a `stream-chat` type clash.

Neither can be fixed on its own branch. Dependabot closes any PR whose branch is modified and opens a replacement: it did exactly that to #2572 and #2555 while they were being fixed, closing both with "Looks like these dependencies are updatable in another way, so this is no longer needed" and replacing them with #2651 and #2647. So this PR carries both groups plus the fixes they need, and the two bot PRs should be closed against it.

## What is the new behavior?

Both groups land together, with three source fixes and one override that the bumps require.

### The typescript-eslint group also moves the compiler

Despite the title, that group carries **TypeScript 5.0.4 to 5.9.3** on the backend and normalises it across the other workspaces. Both of its failures come from the compiler, not the lint plugin.

1. `clinical-controller.helpers.ts:62` - `typeof out[key] === "string"` now narrows a `const`-keyed indexed access, so `as string` no longer changes the type.
2. `workspace.prisma.service.ts:1148` - inferred type predicates carry `isRecord`'s guard through the arrow wrapper, so `find()` already returns `Record<string, unknown> | undefined`.
3. `companion.router.test.ts:118` - under 5.9, `it.each` over an `as const` table types the callback as a rest tuple, and a single-parameter arrow no longer satisfies it. The callback now destructures from the rest tuple, so no parameter is left unused.

**Only fix 3 and fix 2 are backward compatible.** Fix 1 was tested against `dev`'s current TypeScript 5.0.4 and fails there with `error TS2769: No overload matches this call`, because 5.0.4 does not narrow a variable-keyed indexed access and `new Date(unknown)` has no overload. That is why these fixes cannot be pre-landed separately to turn #2647 green on its own.

### The minor-and-patch group duplicates stream-chat

`dev` today resolves a single `stream-chat@9.51.0`. The group bump pulls `9.52.0` for the frontend while `stream-chat-react-native-core` still resolves `9.51.0`, and two copies of a package are two distinct nominal types, so mobile stops compiling:

```
src/features/chat/screens/ChatChannelScreen.tsx(354,19): error TS2322:
Type 'stream-chat@9.51.0/.../Channel' is not assignable to type 'stream-chat@9.52.0/.../Channel'.
  Property 'client' is protected but type 'MessageDeliveryReporter' is not a class
  derived from 'MessageDeliveryReporter'.
```

A single `"stream-chat": "9.52.0"` override collapses it back to one copy. Both consumers' ranges admit that version, and it is the version the bump was already moving the frontend to.

```
$ grep -oE "^  /stream-chat@[0-9.]+" pnpm-lock.yaml | sort -u
  /stream-chat@9.52.0
```

### Conflicts between the two groups

Both groups touch the same manifests and disagree on three packages. Each was resolved to the **higher** version: `typescript` 5.9.3, `typescript-eslint` ^8.68.0, `@typescript-eslint/*` ^8.68.0, `electron` ^43.4.1, `baseline-browser-mapping` ^2.11.19, `typescript` ~5.9.3 for dev-docs.

## Validation performed

Built from a worktree on current `dev`. The shared packages must be built first (`pnpm --filter "./packages/*" run build`) or `@yosemite-crew/auth` fails to resolve and the backend reports thousands of phantom "error typed value" findings.

- `npx eslint .` in `apps/backend`: **exit 0, 0 errors** (#2647 reported 2).
- `npx tsc` in `apps/backend`: **exit 0, 0 errors** (the TypeScript 5.9 test failure).
- `npx tsc --noEmit` in `apps/mobileAppYC`: **exit 0, 0 errors** (#2651 reported 1).
- Lockfile checked by resolution: one `stream-chat@9.52.0`, `typescript@5.9.3`, and `@tiptap/*` all on one version at or above the `3.30.4` patched floor from #2592.

## Impact area

Dependency manifests, the lockfile, two backend source lines and one backend test. No behaviour change.

## Notes for reviewers

- Supersedes #2647 and #2651.
- One override is added rather than a dependency being pinned back, so the group bump is not weakened.
- Worth a human eye on the chat surface: the mobile failure this fixes was in `ChatChannelScreen`, and the repo convention is not to take mobile dependency changes unreviewed.

## Related Issue(s)

Fixes #


ankit-yc:
## Update: three defects found by an adversarial audit of this branch, now fixed

Since nothing in this batch could get a review (Codex has been reporting "usage limits have been reached for code reviews" on every PR since #2592), I ran a multi-agent adversarial audit over this branch and over the already-merged #2592. Seven reviewers on distinct dimensions, every finding triple-refuted, plus a completeness critic. It found three real defects in this PR. All three are fixed in `d8a34bb22`.

**None of them were catchable by CI.** In each case the `pnpm.overrides` map and `pnpm-lock.yaml` agree with each other, so `pnpm install --frozen-lockfile`, lint and type-check all pass. Only the manifests disagree with the installed tree.

### 1. The axios bump installed nothing (high)

Four manifests moved `axios` from `^1.19.0` to `^1.20.0`, but the root override held `"axios": "1.19.0"`. An override beats every importer's range, so the lockfile contained only `/axios@1.19.0` and `apps/*/node_modules/axios/package.json` read 1.19.0 in all four workspaces. The PR claimed an upgrade it had not performed.

Five of the seven reviewers found this independently.

This is the defect class commit `6920935e5` (2026-08-30, "stop overrides pinning packages below what consumers declare") was written to remove, and axios was one of the nineteen it fixed. Its rule: forcing above a declared range is what a CVE floor is for, forcing below it is a defect.

Fixed by raising the override to `1.20.0` (and `axios@1.15.2` with it). `stream-chat@9.52.0` declares `axios: ^1.19.0`, so the tree stays on one copy.

### 2. i18next split into two copies (medium)

`apps/mobileAppYC` moved to `^26.4.0` and resolved 26.4.1 while `"i18next@25": "26.3.6"` held the transitives at 26.3.6. `dev` resolves one i18next; this branch resolved two, with mobile and `react-i18next` on 26.4.1 and `stream-chat-react-native-core` and `stream-chat-react` on 26.3.6.

That undoes the reason `0e421c7d5` chose that pin value: it was set to 26.3.6 specifically so the override would collapse two copies into one rather than add a third. Not a security regression, since 26.3.6 is still above the patched floor, but it puts the split back.

Fixed by raising the override to `26.4.1`.

### 3. Desktop TypeScript downgraded a major (medium)

`apps/desktop` went from `^6.0.3` to `^5.9.3`, undocumented, inside a group labelled typescript-eslint. Nothing required it: `@typescript-eslint/parser@8.69.0` allows `>=4.8.4 <6.1.0` and `ts-jest@29.4.12` allows `>=4.3 <7`, so both accept 6.0.3. The desktop app has been on `^6.0.3` since it was created in `adcd3b043`.

Fixed by restoring `^6.0.3`.

## Revalidated after the fixes

All five workspaces, on the exact tree pushed:

- `npx eslint .` in `apps/backend`: exit 0, 0 errors
- `npx tsc` in `apps/backend`: exit 0, 0 errors
- `npx tsc --noEmit` in `apps/mobileAppYC`: exit 0, 0 errors
- `npx tsc --noEmit` in `apps/desktop`: exit 0, 0 errors (on the restored TypeScript 6.0.3)
- `npx tsc --noEmit` in `apps/frontend`: exit 0, 0 errors

Resolutions confirmed by reading the lockfile rather than the manifests:

```
axios       -> /axios@1.20.0        (one copy, bump now real)
i18next     -> /i18next@26.4.1      (one copy, split closed)
stream-chat -> /stream-chat@9.52.0  (one copy)
typescript  -> 5.9.3 and 6.0.3      (6.0.3 restored for desktop)
```

## Worth fixing beyond this PR

`scripts/ci/check-override-advisories.mjs` documents this exact trap in its header, but it is wired only as a root `package.json` script and is not referenced anywhere under `.github/`, so it never runs in CI. Three instances of the defect it describes reappeared within four days of `6920935e5` fixing nineteen of them. Wiring it into the pipeline, extended to flag an override pinning below any declared range rather than only overrides with advisories, would catch this mechanically.

PR #2672, opened 2026-09-03
feat(repo): retire Docusaurus, redirecting every published URL

ankit-yc:
> Stacked on #2668. Review that one first; this PR's diff is against it, not against `dev`.

## What

Deletes Docusaurus and points every URL it published at the native renderer from #2668.

**One commit on purpose.** `public/dev-docs` shadows any `/dev-docs` route, so a deploy that adds the redirect while the mirror is still on disk has the old files win and the redirect look broken. Redirect and deletion have to land together.

**213 files, −24,261 lines.**

## Redirects

All 53 URLs the Docusaurus sitemap published now resolve. The corpus slugs are byte-identical to the Docusaurus slugs, so `:path*` maps one-to-one with no lookup table.

`/dev-docs/search` is the one special case — a plugin-generated results page with no equivalent, because search is inline in the docs header now. It lands on the index.

`permanent: true`, unlike the interim `.html` redirect this replaces: that one was deliberately temporary because a better fix existed. This move is final.

## The CSP exception moved — it did not disappear

`next.config.ts` carried a relaxed policy scoped to `/dev-docs/:path*` allowing `script-src https://cdn.redoc.ly`. **That is what let the OpenAPI viewer load Redoc at all**, and it happened to cover the whole mirror as a side effect.

The viewer moves to `public/static/openapi/` and the exception moves with it, narrowed to that one directory. Missing this would have rendered the OpenAPI page **empty with no visible error**, because a CSP-blocked script fails silently. The test now asserts both that the viewer keeps its exception and that the broader one is gone.

## Redoc is pinned, and deliberately has no SRI

Pinned to `v2.5.0` rather than `latest`, so the bundle cannot change unreviewed.

No `integrity`/`crossorigin` pair, and this is not an oversight: `cdn.redoc.ly` sends no `access-control-allow-origin` header — checked with `curl -I` — and a browser only enforces `integrity` on a cross-origin script when the request is CORS-enabled. Adding them would **block the script outright**. Getting integrity here means vendoring the bundle same-origin, which is a separate change and worth doing.

## The embedded viewer comes from app code now

`openapi.md` embedded the viewer with an MDX `<iframe>`. The sanitiser strips iframes from the corpus on purpose — contributor markdown must never be able to embed one — so the tag is removed from the document and the element is rendered by `DocsShell` behind a flag the page sets only for that entry.

The viewer survives; the sanitiser stays strict.

## Consumers re-pointed

The developer portal's "Open full docs" link and its GitHub edit URL, plus `.gitignore`, `.prettierignore`, `react-doctor.config.json`, `lint-staged.config.cjs` (its whole mirror-skipping helper is now dead code), `.gitleaks.toml`, README, CONTRIBUTING, AGENTS and CLAUDE.

## Left alone, deliberately

The `dev-docs` segment in middleware's matcher regex, and the desktop navigation-policy and axios entries. All are inert once nothing serves there, and the matcher carries an explicit in-code warning that it cannot be refactored without breaking `next build`. Dead-but-harmless beats a delicate regex edit for zero behaviour change.

The `dev-docs` commitlint scope stays for the same reason — removing it would fail the title check on anyone's in-flight branch, and an unused scope costs nothing.

## Validation

**171 tests across 11 suites**, including middleware, security headers, the developer portal and the whole docs feature. `npx tsc --noEmit` exits 0 — run directly rather than through turbo, which had earlier replayed a cached success over a real error.

## Impact

Deletes a workspace and 4.6 MB of committed build output that produced a 60-to-95 file diff on every documentation edit. No user-facing URL changes destination; they change status code, from serving to redirecting.


aupyay (dismissed):

apps/frontend/src/app/features/docs/DocsShell.tsx
```diff
@@ -86,6 +92,20 @@ export default function DocsShell({
...
             <article className="DocsBody">{toJsxRuntime(tree, { Fragment, jsx, jsxs })}</article>
 
+            {/*
+              The viewer is an iframe, and iframes are stripped from the corpus
+              on purpose - a contributor's markdown must never be able to embed
+              one. This element comes from app code, not from a document, so
+              the sanitiser stays strict and the page still gets its viewer.
+            */}
+            {embedOpenApi && (
+              <iframe
```
ankit-yc, line 102: Fixed in 211d1c7. The OpenAPI viewer iframe now carries `sandbox="allow-scripts allow-same-origin"` - Redoc can run and fetch the same-origin spec, everything else (forms, popups, top-level navigation) is denied. Mutation-checked test asserts the exact value.

PR #2678, opened 2026-09-03
ci(repo): gate override drift, and fix the validator pin it catches

ankit-yc:
## PR Checklist

A `pnpm.overrides` entry beats every manifest. When a dependency bump raises a manifest and leaves the override behind, one of two things happens and **no gate notices either**:

- **The bump installs nothing.** #2665 raised `axios` from `^1.19.0` to `^1.20.0` in four manifests while the root override held `"axios": "1.19.0"`. The lockfile resolved only 1.19.0 and every workspace ran the old copy while its manifest claimed the new one.
- **The package installs twice.** `i18next` split when mobile moved to `^26.4.0` against an override holding 26.3.6. #2670 proposed `"zod@3": "4.5.0"` against an `^4.5.0` that resolves 4.5.4, which would have done the same.

Neither is visible to `pnpm install --frozen-lockfile`, lint, or type-check, because the overrides map and `pnpm-lock.yaml` agree with each other. Only the manifests disagree with what is installed.

`validator` is a live instance on `dev` right now: `apps/frontend`, `apps/backend` and `packages/lib` all declare `^13.15.35` while the override pins `13.15.26`, so nine patch releases of a validation library are installed nowhere and nothing reports it.

`scripts/ci/check-override-advisories.mjs` documents this exact trap in its header, but it answers a different question ("is the pinned version vulnerable"), needs `pnpm audit` and therefore the network, and is wired only as a root `package.json` script with no reference anywhere under `.github/`, so it does not run in CI.

## What is the new behavior?

`scripts/ci/check-override-drift.mjs` answers the complementary question: **does the pin still agree with what the repo asks for.** It runs offline against the manifests and the lockfile, with no `pnpm audit` and no network, so it can be a hard gate. Wired into `_core.yaml` beside the peer-dependency check.

It reports two kinds:

- `pins-below-declared` - the override pins below a declared range, so the declaration is a lie.
- `splits-tree` - the override pins a version that the lockfile resolves alongside a higher one, so the package is installed twice.

**Pinning above a declared range is not reported.** That is the CVE-floor pattern this repo relies on, and #2592 and #2670 both use it deliberately. A gate that reds the build on a legitimate floor gets switched off within a week.

This PR also raises `validator` to `13.15.35`, because a gate has to land green and that is the one live instance of what it detects. If #2670 lands first, that hunk becomes a no-op.

## How it was attacked before pushing

Per the repo convention that a guard needs an adversarial pass of its own, the implementation was mutated and the suite re-run each time. Every mutation is caught:

| Mutation | Result |
| --- | --- |
| below-range detection disabled | 2 tests fail |
| comparison direction inverted, so it would flag CVE floors | 3 tests fail |
| scoped selectors treated as bare names | 1 test fails |
| package name no longer escaped for the lockfile regex | 1 test fails |

The last one initially passed, which is the useful part. The fixtures used `@xmldom/xmldom`, which contains no regex metacharacters, so dropping the escaping changed nothing observable. Added a case using `@jsonjoy.com/fs-snapshot`, which the repo actually overrides and where the dot is a wildcard, and the mutation then failed as it should.

One false positive was found and removed the same way: an early draft read resolved versions from `node_modules/.pnpm`, which retains previously-installed versions until pruned, and reported `validator` 13.15.26 and 13.15.35 as coexisting when only one was resolved. It reads `pnpm-lock.yaml` now, and there is a test pinning that.

## Validation performed

- `node --test scripts/ci/check-override-drift.test.mjs`: **15 pass, 0 fail.**
- `node scripts/ci/check-override-drift.mjs` on this branch: exit 0, no drift.
- The same command against unmodified `dev`: exit 1, reporting the three `validator` rows, which is the defect this PR fixes.
- Re-running after the fix with a regenerated lockfile: `validator` resolves to a single `13.15.35`.

## Impact area

CI only, plus one override value. No application code changes.

## Notes for reviewers

- The negative cases in the test file matter as much as the positive ones. Each is a real shape from the repo's overrides block: scoped selectors, `parent>child` keys, `workspace:*`, unparseable published ranges, and override values that are ranges rather than versions.
- `ALLOWED` is empty and keyed `<package>@<pinned>`, so raising a pin retires an exemption automatically instead of leaving a stale one behind.
- This does not replace `check-override-advisories.mjs`. That one still deserves a CI home; it needs a decision about behaviour when `pnpm audit` is unavailable, which is why it is not in this PR.

## Related Issue(s)

Fixes #


ankit-yc:
The seven inline comments above are all against superseded commits (`ea9b8b988` and `72255274d`). Aikido passes on the current head `97361a47e`. Recording how each was handled rather than leaving that to be reconstructed.

**The performance one was a good catch and is fixed.** `resolvedVersions` re-read and re-scanned the whole 2 MB lockfile once per override, which is 111 full passes to answer 111 questions that one pass answers. The lockfile is now read once and parsed into an index.

**The five `readFileSync` findings are fixed by restructuring rather than suppressed.** The flagged shape was a caller-supplied path reaching a read, via a `root` parameter that existed only so tests could point at a fixture directory. Every exported function is now pure and takes already-read content, and all file access lives in one `readRepo()` against module constants (`ROOT_MANIFEST`, `LOCKFILE`, `WORKSPACE_GROUPS`). That matches how `check-peer-deps.mjs` reads its manifest, and it is better design independent of the scanner: the tests no longer need temporary directories, and the suite grew from 15 cases to 21 including end-to-end ones.

**The `new Map([])` one is fixed.** Now `new Map()`.

Worth noting for anyone reading the diff: the restructure meant re-running the mutation battery, because a rewrite can quietly stop tests from biting. One mutation, dropping the two-space indent anchor from the lockfile pattern, survived three separate attempts to write a fixture that would catch it:

1. `@xmldom/xmldom` contains no regex metacharacters, so the mutation was invisible.
2. The lockfile's own quoted overrides block produced versions with a trailing quote, which `semver.valid` filtered out before they could matter.
3. `/@types/node@26.4.0:` on its own line still parsed correctly, because the greedy scope group captures from the first slash regardless.

What actually distinguishes them is a peer suffix. `/@commitlint/cli@21.2.1(@types/node@26.4.0)` contains a second slash mid-line, so without the anchor the pattern matches again from `/node@26.4.0` and invents a `node` package. Measured against this repo's real lockfile that produces 46 phantom entries that pass `semver.valid`, including `node`, `express` and `react` from `@types/*`. An override on any of those names would then fire a false split. That case is now in the fixture and the mutation fails as it should.

aupyay (dismissed):

aupyay (approved):
Approving at 09c5c8742. What I checked:

- The gate runs and passes in CI, not just locally. `core / Build shared packages and frontend` logs `check-override-drift: 112 override entries checked / no override contradicts a workspace declaration`, immediately after the peer check.
- It has teeth against all three defects it names. Raising root `aws-cdk-lib` above its own override reds the build; before the root manifest was added to the walk the same mutation printed the all-clear and exited 0.
- 22 tests pass, and both new assertions were mutation-checked rather than assumed. Removing the root manifest from `readRepo()` reds the walk test; dropping the leading slash from the lockfile pattern reds the `qs` assertion specifically, which is what the three original fixture lines could not do.
- Half the suite is deliberate should-not-fire cases: CVE floors, scoped selectors, `parent>child`, `workspace:*`, unparseable ranges. A gate that reds on a legitimate floor gets disabled within a week.
- The lockfile is regenerated against dev rather than merged, so the duplicated `/jest-cli@30.4.2(@types/node@26.4.0)` key is gone and `pnpm install --frozen-lockfile` exits 0.
- 72 checks, 0 failing. The inline Aikido comments are against superseded commits.

Known limits, stated rather than hidden: the walk covers the root manifest plus `apps/*` and `packages/*` only, so a nested workspace outside those groups would still be invisible, and `ALLOWED` is empty today so there is no exemption path in use yet.

PR #2695, opened 2026-09-05
fix(frontend): drop duplicate "Close" name from the phone sheet backdrop

aupyay:
## PR Checklist

- [ ] There is an issue for the bug/feature this PR is for. No issue was filed for this one; it was found while auditing the phone sheet chrome.

Every phone bottom sheet exposes two buttons with the identical accessible name "Close":

1. `PhoneShell/BottomSheet.tsx` renders the full-bleed backdrop as `<button className="yc-phone-sheet-backdrop" aria-label="Close">`.
2. `overlays/Sheet/SheetChrome.tsx` renders the visible close chip in the sheet header, also `aria-label="Close"`.

Querying a rendered sheet, `getAllByRole('button', { name: 'Close' })` returns two elements. A screen reader announces the pair identically, and one of them is an invisible sheet covering the whole viewport. Any test that wants the real close button has to disambiguate by index, which `GuideFilters.test.tsx` was already doing.

The backdrop carried `tabIndex={-1}`, so tab order was never affected. The defect is the duplicate name in the accessibility tree, not focus order.

## What is the new behavior?

The backdrop is no longer a control. It is a plain `div` with `aria-hidden="true"` plus the same `onClick`, so it leaves the accessibility tree entirely while tap-outside-to-dismiss is unchanged. Two accessible ways out remain, both verified:

- `Escape`, handled by the focus trap `BottomSheet` already installs.
- The visible close chip, which also receives focus when the sheet opens.

Its `border: none` and `padding: 0` rules go too, since they only existed to undo button UA styles. A browser confirms a `div` computes `0px` for both, so the backdrop is pixel identical: full bleed 375x812, `rgba(29, 28, 27, 0.55)`, `blur(6px)`, `cursor: pointer`.

`BottomSheet` is shared across the phone shell (`PhoneMoreSheet`, the `GuideFilters` sheet and others), so the fix lands once in `BottomSheet` rather than at any call site. No `jsx-a11y` suppression was needed; the rules skip `aria-hidden` elements.

## Validation

Run from `apps/frontend` on this branch.

| Gate | Result |
| --- | --- |
| `npx tsc --noEmit` | exit 0 |
| `npx eslint <changed files>` | exit 0 |
| `npx prettier --check <changed files>` | exit 0 |
| `npx jest --ci --runTestsByPath BottomSheet SheetChrome PhoneMoreSheet` | exit 0, 3 suites, 23 tests |
| Coverage of `BottomSheet.tsx` | 100% statements, branches, functions, lines |
| pre-commit (secretlint, eslint `--max-warnings=0`, prettier) | passed, no bypass |
| pre-push (monorepo lint + type-check) | 18/18 tasks successful |

New test: `hides the backdrop from the accessibility tree, so "Close" names one control`.

Mutation checked on this branch, source restored after each run:

| Mutation | Result |
| --- | --- |
| Backdrop back to `<button aria-label="Close">` | new test fails |
| Drop `aria-hidden` from the div | new test fails |
| Drop the div's `onClick` | `closes when the backdrop is clicked` fails |

Verified live in Storybook at 375x812. The accessibility tree of an open sheet reads exactly one interactive node, `button "Close"`; the backdrop is absent from it. Driving the `Closed (opens on click)` story: the trigger opens the sheet, and backdrop click, `Escape` and the close chip each dismiss it.

## Related Issue(s)

No issue filed.

### DIFF
--- a/apps/frontend/src/app/__tests__/ui/layout/PhoneShell/BottomSheet.test.tsx
+++ b/apps/frontend/src/app/__tests__/ui/layout/PhoneShell/BottomSheet.test.tsx
@@ -74,6 +74,19 @@ describe('BottomSheet', () => {
     expect(onClose).toHaveBeenCalledTimes(1);
   });
 
+  it('hides the backdrop from the accessibility tree, so "Close" names one control', () => {
+    const { backdrop, closeBtn } = renderSheet();
+
+    /* The backdrop used to be a <button aria-label="Close">, which put a second
+       control under that name next to the chrome's close chip: a screen reader
+       announced two identical "Close" buttons, one of them an invisible
+       full-bleed sheet. It is decoration now - not merely renamed, but out of
+       the tree entirely, so no role query can reach it. */
+    expect(backdrop).toHaveAttribute('aria-hidden', 'true');
+    expect(screen.queryAllByRole('button')).not.toContain(backdrop);
+    expect(screen.getAllByRole('button', { name: 'Close' })).toEqual([closeBtn]);
+  });
+
   it('closes when the close button is clicked', () => {
     const { closeBtn, onClose } = renderSheet();
     fireEvent.click(closeBtn);
--- a/apps/frontend/src/app/ui/layout/PhoneShell/BottomSheet.tsx
+++ b/apps/frontend/src/app/ui/layout/PhoneShell/BottomSheet.tsx
@@ -91,13 +91,11 @@ const BottomSheet = ({ open, title, onClose, children, footer, className }: Bott
 
   return (
     <div className="yc-phone-sheet-root">
-      <button
-        type="button"
-        className="yc-phone-sheet-backdrop"
-        aria-label="Close"
-        tabIndex={-1}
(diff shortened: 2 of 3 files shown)


ankit-yc (approved):
Approved.

Checked the thing that would have made this a regression rather than a fix: whether removing the `<button>` removes a keyboard way out. It does not. The backdrop carried `tabIndex={-1}`, so it was never in the tab order, and `BottomSheet.tsx` registers a `keydown` listener that closes on `Escape` (capture phase, cleaned up on unmount). So the accessible exits are Escape and the chrome's close chip both before and after, and what is removed is only the duplicate name.

`aria-hidden="true"` on a `<div>` with a click handler is the right call for decoration that is redundant with a real control — the pointer affordance stays, the accessibility tree stops offering a second "Close".

The test is the good kind: `getAllByRole('button', { name: 'Close' })` equalling exactly `[closeBtn]` fails if the backdrop comes back under any role, not just if the `aria-hidden` attribute is edited.

Issue #2657, opened 2026-09-02
chore(ci): shipped work never closes its issue, because closing keywords only fire on the default branch


ankit-yc:
## What this is

GitHub only honours `Closes #N` / `Fixes #N` when the PR merges into the **default branch**. This repo merges into `dev`, so every issue fixed here stays open indefinitely, no matter what the PR body says.

## The cost, measured today

I closed **eleven** issues by hand in one session. All of them were fixed, merged and in some cases deployed weeks earlier:

| issue | shipped in | keyword the PR used |
|---|---|---|
| #2471 | #2654 | Closes |
| #2595 | #2649 | Fixes |
| #2597 | #2648 | Fixes |
| #2598 | #2650 | Fixes |
| #2527 | #2594 | Fixes |
| #2528 | #2653 | Fixes |
| #2551 | #2599 | Fixes |
| #2593 | #2601 | Closes |
| #2600 | #2601 | Closes |
| #2612 | #2615 | Closes |
| #2618 | #2619 | Closes |

Every one of those PR authors wrote the keyword expecting it to work.

**#2612 is the instructive case.** It was fixed by #2615 on 2026-08-27 - `MetaFieldShell` became a real `fieldset`/`legend`, with jest guards covering every acceptance criterion in the issue. It sat open for six days looking like live work. I branched, opened the component, and found the fix already there with a comment describing it. That is the actual cost: not untidy tracking, but people starting work that is already done.

## Proposed fix

A workflow on `push` to `dev` that parses closing keywords out of the merged PR body and closes those issues, with a comment naming the PR and noting it merged to `dev` rather than `main`.

Scope it deliberately:

- act **only** on the explicit keywords GitHub itself recognises, so it mirrors the behaviour authors already expect and never guesses from a bare `#123` mention
- comment before closing, so the trail says which PR did it
- do nothing on `main` - GitHub already handles that, and double-closing would be noise

## Alternatives considered

- **Make `dev` the default branch.** Fixes it for free, but it is a much larger change: it moves PR targeting, branch protection and every "default branch" assumption in CI. Out of scope here.
- **Keep closing by hand.** That is the status quo, and eleven in one sitting is the evidence against it.

## Note

I have not implemented this - a workflow that closes issues automatically changes behaviour for everyone in the repo, so it wants a decision rather than a drive-by PR. Happy to open one if this is wanted.

aupyay:
[...]

Issue #2625, opened 2026-09-02
fix(mobile): four notification actions are local-only mocks with no endpoint behind them


ankit-yc:
## Summary

Every notification mutation in the mobile app is a mock. `features/notifications/thunks.ts` has five thunks, and all five follow this shape:

```ts
async ({notificationId}, {rejectWithValue}) => {
  try {
    await mockDelay(200);
    // NOTE: Replace with actual API call when backend is ready
    // await notificationService.archive(notificationId);
    return {notificationId};
  } catch (error) { ... }
}
```

There is no `notificationService` in the app at all — the commented-out call references something that was never written.

## What the user sees

Swiping a notification past 25% of the screen width archives it: the row animates off-screen, it disappears from the list, and the unread badge on the filter pills drops by one. **Nothing persists.** On the next load the notification is back, unread.

The same applies to delete, clear-all and mark-all-read.

## What the backend actually offers

`notification.router.ts` is 21 lines and exposes exactly two operations:

```ts
notificationRouter.get('/mobile', requireMobileAuth, NotificationController.listNotifications);
notificationRouter.post('/mobile/:notificationId/seen', requireMobileAuth, NotificationController.markAsSeen);
```

So of the five client thunks:

| thunk | endpoint exists? |
| --- | --- |
| `notifications/markAsRead` (single) | **yes** — `POST /mobile/:id/seen`, but the thunk does not call it |
| `notifications/archive` | no |
| `notifications/delete` | no |
| `notifications/clearAll` | no |
| `notifications/markAllAsRead` | no bulk endpoint |
| `notifications/create` | no |

Four of the affordances have nothing to call.

## Decision needed

This is not a copy fix, which is why it is filed rather than patched:

1. **Wire the one that exists.** `markAsRead` should call `POST /mobile/:id/seen`. That is small and unambiguous, and worth doing regardless of the rest.
2. **Then choose for the other four**: build the endpoints (archive/delete need a column or a status on the notification model; clear-all and mark-all-read need bulk routes), or remove the affordances until they exist.

Leaving them as-is is the worst of the three: the gestures look like they work, so a pet owner tidies their notifications and finds them all back.

## Related

Found in the same sweep as #2620, #2623 and #2624. Same shape as the companion invite in #2610 — a commented-out API call sitting directly above a success return.

## Clinic-day strategy connection (2026-09-12)

[...]

Issue #2603, opened 2026-09-01
fix(ci): api-deploy applies migrations before the build, so a failed deploy leaves the live process on an incompatible schema


ankit-yc:
## Summary

`scripts/deploy/api-deploy.sh` applies Prisma migrations at step 4 of 8 and cuts over at step 7. The old PM2 process serves traffic against the new schema for the whole build and smoke window, and if the build or smoke fails the script exits without cutting over, leaving the migration applied and the old process running indefinitely.

## Where

```
line  60  say "preflight"
line  70  say "checkout $GIT_REF"
line  81  say "install"
line  84  say "prisma client + migrations"     <-- schema changes here
line  88  say "build packages"
line 101  pnpm --filter backend run build
line 116  say "smoke boot on :$SMOKE_PORT"
line 138  say "cutover"                         <-- new code starts serving here
line 139  pm2 restart "$PM2_TARGET" --update-env
```

Line 134 is explicit that a smoke failure does not cut over:

```sh
echo "smoke boot failed - NOT cutting over. Rollback sha: $ROLLBACK_SHA" >&2
```

That is the right call for the process, and it is exactly what makes this sharp: the rollback restores the *code*, and nothing rolls back the *schema*.

## Why it matters

Any migration that removes or renames something a running query still names turns every request touching that table into a 500 for the length of the window, and a failed deploy makes that permanent until someone intervenes. Additive migrations are unaffected, which is why this has not bitten yet.

Raised by the Codex review on #2599, which renames three columns. That PR is safe on its own terms - all three tables hold zero rows in production, re-verified at the time of writing - but the hazard is general and will apply to the first rename on a populated table.

## Options

1. **Expand-contract as a repo convention.** Add and backfill the new column, ship code that reads both, drop the old column in a later release. Correct, and costs two deploys per rename.
2. **Split the deploy into two ordered phases** so the schema step is explicitly the point of no return, with a documented rollback migration required for anything non-additive.
3. **Gate on it.** The migration CI gate could classify each migration as additive or destructive and fail a destructive one that ships without a paired expand step. This composes with #2531, which is about a different blind spot in the same gate.

Option 3 is the one that does not depend on anybody remembering.

## Out of scope

Reordering so migrations run after the smoke test does not work: the smoke boot runs the new code, which needs the new schema.

PR #2696, opened 2026-09-05
fix(repo): persist mobile notification archive and scope writes to the owner

ankit-yc:
## PR Checklist

Swiping a notification past 25% of the screen width in the mobile app archives it: the row animates off-screen, it disappears from the list, and the unread badge drops by one. Nothing persists. On the next load the notification is back, unread.

`archiveNotification` in `features/notifications/thunks.ts` awaited a 200 ms mock delay and returned, above a commented-out call to a `notificationService` the issue said had never been written.

Three further defects surfaced while tracing that path:

1. `NotificationService.markNotificationAsSeen` ran `updateMany({where: {id: notificationId}})` on an id supplied by the client, with no owner in the `where` and no user id passed by the controller. Any authenticated mobile user could mark another user's notification seen by id.
2. `sendToUser` created a `Notification` row **inside** the per-device-token loop, so a user with a phone and a tablet got two rows for one notification, both shown in the list.
3. The row's id never reached the device, so the app minted `notif_${Date.now()}` for push-delivered notifications. Any per-notification write against one of those items targets an id the server does not have.

Note for reviewers: part of #2625 is already fixed on `dev` and the issue does not say so. `markNotificationAsRead` calls the seen endpoint, and `fetchNotificationsForCompanion` calls a real `notificationService.ts`. That is #2657 in the wild.

## What is the new behavior?

**Archive persists.**

- `Notification` gains a nullable `archivedAt`. The migration is `ADD COLUMN IF NOT EXISTS`, purely additive, so the pre-cutover window described in #2603 does not apply to it.
- New route `POST /v1/notification/mobile/:notificationId/archive` behind `requireMobileAuth`. It stamps `archivedAt` only when the row is currently un-archived, so a second swipe keeps the first archive time rather than rewriting it. Already-archived is a 200, not a 404 - the notification is where the owner put it.
- `archiveMobileNotification` in the mobile service, wired into the thunk. `deriveStatus` reads `archivedAt` before `isSeen`, because an archived row is also seen and checking `isSeen` first would bring it back into the visible list on reload.

**Writes are scoped to their owner.**

- `markNotificationAsSeen` and the new `archiveNotificationForUser` both put `userId` in the `where` and return what actually changed, so an id belonging to someone else matches nothing and is reported as a 404 rather than as success. The owner id is resolved once, in a shared helper, from the same `parentId` the list endpoint already reads.

**Notifications have one row and a stable identity.**

- `sendToUser` writes the row once, before the token fan-out, and puts its id into the push `data`. The app seats the push using that id, so mark-read and archive on a just-arrived notification reach the row behind it. A failed write still delivers the push.

**Dead mock thunks removed.**

- `markAllNotificationsAsRead`, `deleteNotification` and `clearAllNotifications` had no endpoint and no dispatch site anywhere in `apps/mobileAppYC/src`. Clear All was removed from the UI by design. This is the "remove the affordances" branch of the decision the issue asks for, taken because there are no affordances left to keep.
- `createNotification` was a mock delay followed by an object built from the push payload, with a `dispatch(addNotificationToList(...))` fallback in its own catch block that built the same object. Replaced by that dispatch, so there is one mapping instead of two and no thunk pretending to be an API call.

### Behaviour change a user can notice

`POST /v1/notification/mobile/:notificationId/seen` used to return 200 unconditionally. It now returns 404 when the scoped update matches no row (`count === 0`), which is what makes the cross-tenant fix observable rather than silent. The mobile `markNotificationAsRead` thunk `rejectWithValue`s on that response, so an id the server does not recognise for the caller changes from a silent no-op into a visible failure in the app.

This is the intended direction - reporting success for an update that changed nothing is the defect this PR is about - but it is the one change here that a user could notice, so it is called out rather than left to be rediscovered.

### Verification

Run on `a6be055`:

- `npx jest` in `apps/backend` - 466 suites, 11206 tests, all pass.
- `npx jest` in `apps/mobileAppYC` - 500 suites, 8547 tests, all pass.
- `pnpm --filter backend run type-check`, `pnpm --filter mobileAppYC run type-check` - both exit 0.
- `pnpm --filter backend run lint`, `pnpm --filter mobileAppYC run lint` - both exit 0. The 25 backend warnings are pre-existing and none are in the files touched here.
- Coverage on the touched files: `notification.controller.ts` 100% stmts/branch, `notification.service.ts` 98% stmts / 92.85% branch (the uncovered lines are the pre-existing Firebase credential bootstrap), mobile `thunks.ts` and `notificationSlice.ts` 100%, `notificationService.ts` 98.26%, `firebaseNotifications.ts` 99.49%.

Aikido was not run locally - no CLI or MCP on this machine - so it is unverified until it runs on this PR.

Not verified end to end against deployed dev: the archive path needs a signed-in mobile build and a live push, which I cannot exercise from here. The endpoint, the owner scoping and the client wiring are covered by the tests above.

## Related Issue(s)

Refs #2625

Follow-up, not in this PR: the duplicate-row and missing-push-id defects are fixed here, but nothing backfills the duplicate `Notification` rows already written in production by the old per-token loop.

### DIFF
--- a/apps/backend/src/controllers/app/notification.controller.ts
+++ b/apps/backend/src/controllers/app/notification.controller.ts
@@ -3,25 +3,45 @@ import { AuthUserMobileService } from "src/services/authUserMobile.service";
 import { NotificationService } from "src/services/notification.service";
 import { resolveVerifiedUserId } from "src/utils/request";
 
+/**
+ * The id notification rows are keyed by, for the caller of this request.
+ *
+ * Notifications are written against the mobile user's `parentId`, which is what
+ * listNotifications already reads, so every per-notification route has to
+ * resolve the same id before it can scope a write to its owner. Writes the 401
+ * or 404 itself and returns undefined, so callers bail on a falsy result.
+ */
+const resolveNotificationOwnerId = async (
+  req: Request,
+  res: Response,
+): Promise<string | undefined> => {
+  const authUserId = resolveVerifiedUserId(req);
+  if (!authUserId) {
+    res.status(401).json({ message: "Not authenticated: userId is missing." });
+    return undefined;
+  }
+
+  const authUser = await AuthUserMobileService.getByProviderUserId(authUserId);
+  const ownerId = authUser?.parentId?.toString();
+  if (!ownerId) {
+    res.status(404).json({ message: "User not found." });
+    return undefined;
+  }
+
+  return ownerId;
+};
+
 export const NotificationController = {
   // List notifications for current user
   listNotifications: async (req: Request, res: Response) => {
     try {
-      const authUserId = resolveVerifiedUserId(req);
-      if (!authUserId) {
-        return res
-          .status(401)
-          .json({ message: "Not authenticated: userId is missing." });
-      }
-      const authUser =
-        await AuthUserMobileService.getByProviderUserId(authUserId);
-      if (!authUser) {
-        return res.status(404).json({ message: "User not found." });
+      const ownerId = await resolveNotificationOwnerId(req, res);
+      if (!ownerId) {
+        return res;
       }
 
-      const notifications = await NotificationService.listNotificationsForUser(
-        authUser.parentId?.toString() || "",
-      );
+      const notifications =
+        await NotificationService.listNotificationsForUser(ownerId);
       return res.status(200).json({ notifications });
     } catch (err) {
       console.error("Error listing notifications", err);
@@ -37,13 +57,55 @@ export const NotificationController = {
         return res.status(400).json({ message: "notificationId is required." });
       }
 
-      await NotificationService.markNotificationAsSeen(notificationId);
+      const ownerId = await resolveNotificationOwnerId(req, res);
+      if (!ownerId) {
+        return res;
+      }
+
+      const updated = await NotificationService.markNotificationAsSeen(
+        notificationId,
+        ownerId,
+      );
+      if (updated === 0) {
+        return res.status(404).json({ message: "Notification not found." });
+      }
+
       return res.status(200).json({ message: "Notification marked as seen." });
     } catch (err) {
       console.error("Error marking notification as seen", err);
       return res
         .status(500)
         .json({ message: "Failed to mark notification as seen." });
+    }
+  },
+
+  // Archive notification
+  archive: async (req: Request, res: Response) => {
+    try {
+      const { notificationId } = req.params;
+      if (!notificationId) {
+        return res.status(400).json({ message: "notificationId is required." });
+      }
+
+      const ownerId = await resolveNotificationOwnerId(req, res);
+      if (!ownerId) {
+        return res;
+      }
+
+      const outcome = await NotificationService.archiveNotificationForUser(
+        notificationId,
+        ownerId,
+      );
+      if (outcome === "not-found") {
+        return res.status(404).json({ message: "Notification not found." });
+      }
+
+      return res.status(200).json({ message: "Notification archived." });
+    } catch (err) {
+      console.error("Error archiving notification", err);
+      return res
+        .status(500)
+        .json({ message: "Failed to archive notification." });
     }
   },
 };
--- a/apps/backend/src/controllers/app/notification.controller.ts
+++ b/apps/backend/src/controllers/app/notification.controller.ts
@@ -37,7 +37,7 @@ export const NotificationController = {
     try {
       const ownerId = await resolveNotificationOwnerId(req, res);
       if (!ownerId) {
-        return res;
+        return;
       }
 
       const notifications =
@@ -59,7 +59,7 @@ export const NotificationController = {
 
       const ownerId = await resolveNotificationOwnerId(req, res);
       if (!ownerId) {
-        return res;
+        return;
       }
 
       const updated = await NotificationService.markNotificationAsSeen(
@@ -89,7 +89,7 @@ export const NotificationController = {
 
       const ownerId = await resolveNotificationOwnerId(req, res);
       if (!ownerId) {
-        return res;
+        return;
       }
 
       const outcome = await NotificationService.archiveNotificationForUser(
--- a/apps/backend/src/routers/notification.router.ts
+++ b/apps/backend/src/routers/notification.router.ts
@@ -18,4 +18,11 @@ notificationRouter.post(
   NotificationController.markAsSeen,
 );
 
+// Archive notification (removes it from the owner's list without deleting it)
+notificationRouter.post(
+  "/mobile/:notificationId/archive",
+  requireMobileAuth,
+  NotificationController.archive,
+);
+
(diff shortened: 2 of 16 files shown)


aupyay (commented):
Independent review, read from the diff rather than the PR description. Verdict: **good to merge** once `sonar / Sonar scan (backend)` finishes - everything else is green, including `migration / Prisma migrations`. Two notes and a nit, none blocking.

**The security fix is right, including the part that is easy to get wrong.** `markNotificationAsSeen` was `updateMany({ where: { id } })` on a client-supplied id, so any authenticated mobile user could mark any other user's notification seen. The fix puts `userId` in the `where` and returns `count`, and `resolveNotificationOwnerId` resolves the same `parentId` that `listNotifications` already keys on, so the scoping matches how the rows are actually written.

The detail worth calling out: in `archiveNotificationForUser`, the disambiguating `findFirst` that separates "already archived" from "not found" is **also** scoped to `userId`. Left unscoped it would have been an existence oracle - a 200-vs-404 that confirms whether another user's notification id exists. It is scoped. Good.

Also checked: `sendToUser` still returns early on `!tokens.length`, so moving `createNotificationRecord` out of the fan-out loop does not start writing rows for users with no devices. The per-device duplication fix does not bring a new orphan-row case with it.

Test coverage covers the cases that matter, not just the happy path - `passes the resolved owner, so the update cannot cross users`, `404s an id that is not the caller's`, `reports not-found for an id that is not this user's`, `treats a second archive as success and keeps the original timestamp`. That last one is the right call: an already-archived row is a success from the owner's point of view, and 404-ing a swipe that already worked would be the obvious wrong answer.

---

**1. `markAsSeen` is now a user-visible behaviour change and the PR body does not say so.** It used to return 200 unconditionally; it now 404s when `count === 0`. The mobile `markNotificationAsRead` thunk `rejectWithValue`s on that, so an id the server does not recognise turns from a silent no-op into a visible failure in the app.

I think that is the correct direction - reporting success for an update that changed nothing is what the issue is about. But it is worth a line in the PR body, because it is the one change here that a user could notice and nobody would connect it back to this PR.

**2. Agreed on leaving the existing duplicate rows alone.** The old per-token loop already wrote them in production and nothing backfills. That is a dedupe decision with a data-loss edge, not a drive-by, and it does not belong in this PR.

**3. Nit:** `if (!ownerId) { return res; }` in all three handlers returns the response object where `return;` is meant. Express ignores the return value so it is harmless, but it reads like the value matters.

One thing for the record, since it came up in #2603: your migration is correctly additive. `ADD COLUMN IF NOT EXISTS` on a nullable column with no default and no constraint - the running process serves fine against it before the cutover, so the destructive-migration gate in #2698 passes it untouched. The comment you put in the migration file explaining exactly that is the right habit.

ankit-yc:
Thanks - all three addressed, and the Sonar failure that was still outstanding is fixed.

**Sonar (backend) was red, not just unfinished.** Coverage 97% (floor 95) and duplication 0% both passed; the repo threshold gate failed on `issues: 1`, while SonarCloud's own quality gate reported "passed" with 1 new issue. Worth knowing those two disagree - the stricter repo gate is the one that blocks. The issue was mine: `typescript:S7744` at `notification.service.ts:224`, `{ ...(options?.data ?? {}) }`. The `?? {}` is useless inside a spread because spreading `undefined` already contributes nothing, so it is now `{ ...options?.data }`. Same semantics, fixed in `a7439a18a` rather than excluded.

**1. Behaviour change - agreed, and now in the PR body.** Added a "Behaviour change a user can notice" section stating that `seen` moved from unconditional 200 to 404 on `count === 0`, that the mobile thunk `rejectWithValue`s on it, and why that is the intended direction. You are right that it is the one change here nobody would connect back to this PR later.

**2. Duplicate rows - agreed, staying out.** It is recorded as a follow-up in the PR body, not silently dropped.

**3. Nit fixed** in `a6be05577`. The three `return res;` in `notification.controller.ts` are now bare `return;`. I left the six in `finance.controller.ts` alone - same pattern, but pre-existing and outside this change.

Verification at `a6be055`: `npx tsc --noEmit` exit 0; `npx jest --testPathPatterns=notification` exit 0, 6 suites / 94 tests; `npx jest` full backend exit 0, **466 suites / 11206 tests**. Aikido and Sonar re-run on this PR - still unverified locally, no CLI on that machine.

aupyay (approved):
Approving from `aupyay` (L3). Verified independently rather than ticking L1's read.

**Ownership scoping.** Both `updateMany` calls carry `userId` in the `where`, and both are preceded by an `isNonEmptyString(userId)` guard that throws. That guard is the part worth calling out: Prisma drops an `undefined` field from a `where` rather than matching nothing, so `{ id, userId: undefined }` would silently widen to `{ id }` — any authenticated user archiving any notification by id. The same trap is documented in `companion-access.ts`. Rejecting the empty string too closes it at the boundary rather than relying on the caller. The disambiguating `findFirst` is owner-scoped as well, so a 404 for someone else's id is not an existence oracle.

**Migration.** `ADD COLUMN IF NOT EXISTS "archivedAt" TIMESTAMP(3)` — nullable, no default, no backfill, no `DROP`, no `NOT NULL`. Additive in both directions: code deployed before the migration ignores a column it does not select, and code deployed after tolerates NULL as "not archived". Correctly outside the #2603 window.

**`archivedAt: null` in the archive `where`** is the right call — a second swipe keeps the first archive time rather than moving it, and `deriveStatus` reading `archivedAt` before `isSeen` is necessary, since an archived row is also seen and checking `isSeen` first would resurrect it on reload.

**Gates at `a6be055`:** `CI Required` SUCCESS, `Supply Chain Required` SUCCESS, Sonar backend and mobile SUCCESS, Aikido SUCCESS. The `CANCELLED` "Validate commit messages" is a superseded run; the live one is SUCCESS.

No blocking findings. The two defects you found outside the issue's scope — the unscoped `markNotificationAsSeen` and the per-device-token row duplication — were worth fixing here rather than filing, since the second would have made the new archive endpoint 404 on exactly the freshly-arrived notifications.

aupyay (approved):
Approving at `a6be055`.

Re-read the two commits pushed after my earlier review notes, since `dismiss_stale_reviews_on_push` means this approval has to stand on the current head rather than on what I read before:

- `a7439a18a` — `{ ...(options?.data ?? {}) }` to `{ ...options?.data }`. Spreading `undefined` into an object literal contributes nothing, so this is identical behaviour and the Sonar `typescript:S7744` finding is fixed rather than suppressed.
- `a6be05577` — three `return res;` to `return;` in `notification.controller.ts`. The value was never read by Express.

`CI Required` SUCCESS, `Supply Chain Required` SUCCESS, Sonar backend and mobile SUCCESS, Aikido and GitGuardian pass. The one `CANCELLED` entry for "Validate commit messages" is a superseded run; the live one is SUCCESS.

Issue #2637, opened 2026-09-02
Document signing portal shows Documenso's sign-in instead of the practice's documents


ankit-yc:
## What happens

Organisation settings > E-signing > **Manage document signing portal** embeds the Documenso portal. Instead of the practice's documents, the frame renders Documenso's own **"Sign in to your account"** form. Nothing in the PIMS reports an error, because from the app's side nothing failed.

Reproduced on `dev.yosemitecrew.com` as an org owner.

## What I ruled out

Everything on our side is working:

| step | result |
| --- | --- |
| `POST /v1/documenso/pms/redirect/:orgId` | **200**, returns a well-formed handoff URL |
| iframe `src` | `https://ds.yosemitecrew.com/auth/external?token=<64 chars>` |
| PIMS CSP `frame-src` | includes `https://ds.yosemitecrew.com` |
| `ds.yosemitecrew.com` `frame-ancestors` | includes `https://dev.yosemitecrew.com` |
| iframe subresources | every request 200 - Documenso itself loads fine |

It is also **not** a third-party-cookie problem, which was my first theory. I minted a first-use token straight from the API and opened it as a **top-level document**, with no iframe involved at all. It still lands on `https://ds.yosemitecrew.com/signin`.

## Where the failure actually is

`ds.yosemitecrew.com/auth/external` accepts the token and then establishes no session.

The two failure modes are distinguishable, which is what narrows it:

- Replaying an already-consumed token gives **"Unable to sign you in - Invalid or expired token"**.
- A fresh token gives **no error at all** - it redirects to `/signin`.

So the token is recognised and consumed, and session creation is what does not happen. Since `generate-token` returns 200, `DOCUMENSO_EXTERNAL_AUTH_SECRET` is also being accepted, so this is not a secret mismatch either.

Most likely remaining cause: the external-auth route signs in an **existing** Documenso user by email and does not provision one, so a practitioner with no Documenso account falls through to `/signin`. That needs checking on the `ds.yosemitecrew.com` host, which is outside this repo.

## What was changed here

Nothing that fixes the above - the fix is on the Documenso host. Two adjacent frontend defects found while tracing it were fixed in #2621:

[...]

PR #2697, opened 2026-09-05
fix(backend): replace deprecated Documenso document create

### LINKED ISSUE #2643: chore(backend): migrate Documenso createDocument off the deprecated createV0

aupyay:
## PR Checklist

`DocumensoService.createDocument()` still calls the deprecated `documents.createV0()` API and uploads the PDF through a separate presigned `PUT`. The supported `documents.create()` API accepts the PDF directly, but returns only the document id and envelope id.

Issue #2637's signing-portal guidance is already present on `dev`, so this PR addresses the remaining deprecated-create work from #2643.

## What is the new behavior?

Document creation now sends the PDF and signing payload through `documents.create()`, then retrieves the created document through `documents.get()`. The returned object still includes the recipient token used by existing packet and record signing callers.

The focused service tests assert the multipart request, the recipient-bearing follow-up lookup, and the failure path when document creation succeeds but the lookup fails.

- `pnpm --filter backend exec jest --runInBand --no-watchman test/services/documenso.service.test.ts`: exit 0, 1 suite and 28 tests passed.
- Mutation check changed the lookup from `created.id` to `created.id + 1`: exit 1, 2 assertions failed; the implementation was restored before the final checks.
- `pnpm --filter backend run lint`: exit 0, 25 warnings in unchanged backend files.
- `pnpm --filter backend run type-check`: exit 0.
- `pnpm --filter backend run build`: exit 0, `Build succeeded`.
- Pre-push `pnpm run lint`: exit 0, 10 tasks succeeded.
- Pre-push `pnpm run type-check`: exit 0, 18 tasks succeeded.

The focused suite was run for this service change. A clean full backend Jest result is not claimed here.

## Related Issue(s)

Fixes #2643

### DIFF
--- a/apps/backend/src/services/documenso.service.ts
+++ b/apps/backend/src/services/documenso.service.ts
@@ -84,21 +84,6 @@ const getDocumensoClient = (apiKeyOverride?: string) => {
   return client;
 };
 
-async function uploadPdfBuffer(pdf: Buffer, uploadUrl: string) {
-  const response = await fetch(uploadUrl, {
-    method: "PUT",
-    body: new Uint8Array(pdf),
-    headers: {
-      "Content-Type": "application/pdf",
-      "Content-Length": pdf.length.toString(),
-    },
-  });
-
-  if (!response.ok) {
-    throw new Error(`Upload failed: ${response.status}`);
-  }
-}
-
 export type SignedDocument = {
   downloadUrl?: string;
   filename?: string;
@@ -129,7 +114,7 @@ export class DocumensoService {
       logger.info("Creating document with signature placement", {
         placement,
       });
-      const request = {
+      const payload = {
         title: title ?? "Form Submission",
         recipients: [
           {
@@ -149,19 +134,15 @@ export class DocumensoService {
           },
         ],
       };
-      // createV0 is deprecated, but its replacement documents.create() returns
-      // only { envelopeId, id } - no uploadUrl and no recipients. The callers
-      // below read document.recipients[0].token to build the signing URL, so
-      // migrating needs a second documents.get() round-trip and a rewrite of the
-      // upload path. That is a behavioural change on the e-signature flow, so it
-      // is tracked in #2643 and tested there against a live Documenso.
-      const created = await documenso.documents.createV0(request); // NOSONAR
+      const created = await documenso.documents.create({
+        payload,
+        file: {
+          fileName: "document.pdf",
+          content: new Uint8Array(pdf),
+        },
+      });
 
-      const { document, uploadUrl } = created;
-
-      await uploadPdfBuffer(pdf, uploadUrl);
-
-      return document;
+      return await documenso.documents.get({ documentId: created.id });
     } catch (error) {
       if (error instanceof errors.DocumensoError) {
         logger.error("API error:", error.message);
--- a/apps/backend/test/services/documenso.service.test.ts
+++ b/apps/backend/test/services/documenso.service.test.ts
@@ -320,6 +320,27 @@ describe("DocumensoService", () => {
         expect(logger.error).toHaveBeenCalledWith("Status code:", 400);
         expect(logger.error).toHaveBeenCalledWith("Body:", "Bad Request");
       });
+
+      it("handles a failed lookup after the document is created", async () => {
+        mockCreate.mockResolvedValueOnce({ id: 5, envelopeId: "env_5" });
+        mockGet.mockRejectedValueOnce(
+          new (DocumensoError as any)("Lookup failed", 503, "Unavailable"),
+        );
+
+        const result = await DocumensoService.createDocument({
+          pdf: Buffer.from("test"),
+          signerEmail: "test@test.com",
+        });
+
+        expect(result).toBeUndefined();
+        expect(mockGet).toHaveBeenCalledWith({ documentId: 5 });
+        expect(logger.error).toHaveBeenCalledWith(
+          "API error:",
+          "Lookup failed",
+        );
+        expect(logger.error).toHaveBeenCalledWith("Status code:", 503);
+        expect(logger.error).toHaveBeenCalledWith("Body:", "Unavailable");
+      });
     });
 
     describe("distributeDocument", () => {
--- a/apps/backend/test/services/documenso.service.test.ts
+++ b/apps/backend/test/services/documenso.service.test.ts
@@ -25,7 +25,7 @@ jest.mock("../../src/utils/logger", () => ({
   },
 }));
 
-const mockCreateV0 = jest.fn();
+const mockCreate = jest.fn();
 const mockDistribute = jest.fn();
 const mockGet = jest.fn();
 
@@ -33,7 +33,7 @@ jest.mock("@documenso/sdk-typescript", () => {
   return {
     Documenso: jest.fn().mockImplementation(() => ({
       documents: {
-        createV0: mockCreateV0,
+        create: mockCreate,
         distribute: mockDistribute,
         get: mockGet,
       },
@@ -54,9 +54,6 @@ jest.mock("@documenso/sdk-typescript/models/errors/index.js", () => {
   return { DocumensoError: MockDocumensoError, __esModule: true };
 });
 
-// Set up global fetch mock
-globalThis.fetch = jest.fn();
-
 // --- HELPER TO TEST LOAD-TIME ENV VARIABLES ---
 function getModule(envOverrides: Record<string, string>) {
   let mod: any;
@@ -74,7 +71,6 @@ function getModule(envOverrides: Record<string, string>) {
 describe("DocumensoService", () => {
   beforeEach(() => {
     jest.clearAllMocks();
-    (globalThis.fetch as jest.Mock).mockResolvedValue({ ok: true });
   });
 
   describe("Configuration & Environment Variable Errors", () => {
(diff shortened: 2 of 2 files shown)


aupyay (commented):
Independent review, from the diff and the installed SDK types rather than the PR summary. Repo `@documenso/sdk-typescript@0.9.0`.

**The core claim holds.** `documents.create({ payload, file })` matches `DocumentCreateRequest` (`models/operations/documentcreate.d.ts:606`), its response is `{ envelopeId: string; id: number }` (:613), `documents.get` takes `{ documentId: number }` (`documentget.d.ts:5`), and `DocumentGetRecipient.token` is a non-optional `string` (:225 in `DocumentGetRecipient`). So the second round-trip does return the token the callers need, and `created.id` is the right type to pass. Removing the `// NOSONAR` with the deprecated call is correct - the suppression went with the thing it suppressed.

Two findings, one of which I would fix before merge.

---

**1. The tests assert the exact shape that would break production.** Every `mockGet` in the suite resolves `{ id, recipients: [] }`, and `expect(result).toEqual({ id: 1, recipients: [] })` pins it.

The two callers that matter build the signing URL like this:

```ts
// rendered-document.service.ts:715, workspace-document-packet.service.ts:509
documensoPublicBaseUrl && doc.recipients?.[0]?.token
  ? `${documensoPublicBaseUrl}/sign/${doc.recipients[0].token}`
  : ...
```

Optional chaining, so an empty `recipients` produces **no signing URL and no error** - the packet is created, the clinic is never given a link, and nothing raises. That is the failure this PR exists to avoid, and the suite currently encodes it as the expected value.

One test that resolves `mockGet` with a populated recipient and asserts the token survives to the caller closes it. Without that, nothing here would catch `documents.get` being swapped for something that omits recipients.

**2. No test for `get` failing after `create` succeeded, which is a genuinely new failure mode.** The deleted `throws when uploadPdfBuffer fetch fails` test covered the analogous old path; nothing replaced it. `handles DocumensoError` only rejects `mockCreate`.

It also has a real side effect the old single call did not: if `create` succeeds and `get` throws, a document now exists in Documenso that nothing references, the `catch` swallows the error, `createDocument` returns `undefined`, and the three callers correctly turn that into a 502. The caller behaviour is fine - I checked all three and each guards `!doc` before use. What is untested is that path, and a retry against that 502 mints a second orphan.

---

**On the evidence.** Please restate the full-suite line in the PR body. "465 suites and 11,179 tests with all assertions passing, then interrupted, final exit 130" is an interrupted run, not a pass - the focused 27/27, the type-check, the lint and the build are the load-bearing evidence here and they are enough on their own. A reviewer who reads it as a clean full run and then finds the 130 stops trusting the rest.

Nothing else. The production diff is smaller and better than what it replaces.

aupyay (commented):
Finding 1 confirmed fixed at `10518c8ce`. `mockGet` now resolves a populated recipient and the test asserts `result.recipients[0].token` reaches the caller, which is the property the two signing-URL call sites depend on and the one the old empty-array mock could not have caught. Mutation-checking it with an empty list rather than just asserting the new shape is the right way to have proved it.

From my side that clears the merge blocker. Finding 2 - no test for `documents.get` rejecting after `create` succeeded - still stands, but it is a coverage gap on an error path the callers already turn into a 502, not a correctness bug, so I would not hold the merge for it. Worth a follow-up rather than another round here.

Not approving through the GitHub UI: every agent here pushes as the same account, so a formal approval from me would be a self-approval on your PR. Treat this comment as the independent review. Merge when the fresh gates are green.

aupyay:
Addressed both review findings in `642185aae`.

- The success test now returns a populated recipient and asserts that its token survives the follow-up lookup.
- A new regression test covers document creation succeeding before `documents.get()` fails, including the returned `undefined` value and logged API details.
- The new test was mutation-checked by changing the lookup to `created.id + 1`; two assertions failed before the implementation was restored.
- The PR body now reports only the focused suite and checks run against the final tree. It no longer presents a full backend suite as clean evidence.

Fresh focused result: 1 suite and 28 tests passed. Backend lint, type check, and build exited 0; the pre-push monorepo lint and type-check hooks also exited 0.

aupyay (commented):
Both findings confirmed closed at `642185aae`.

`handles a failed lookup after the document is created` pins the exact path I raised: `create` resolves, `get` rejects, `createDocument` returns `undefined`, and the three callers turn that into a 502 rather than dereferencing it. `expect(mockGet).toHaveBeenCalledWith({ documentId: 5 })` is the part that makes it a real test rather than a smoke test - it proves the id from `create` is what the lookup used, so a regression that passes the wrong id fails here instead of silently 404ing against Documenso.

No further findings from me. Merge on green.

One thing for the record, not for this PR: the orphaned document that path leaves behind in Documenso is still real - `create` succeeded, nothing references the result, and a client retrying the 502 mints another. It is now tested and logged rather than invisible, which is the right amount of work for this change. If it turns out to happen at all in practice it wants an issue, not a fix here.

ankit-yc (approved):
Approved.

The tests mock the SDK, so they cannot tell you the new call shape is real — that is the same defect shape as reading a metadata key nothing writes. So I checked it against the installed SDK's own types (`@documenso/sdk-typescript@0.9.0`) rather than against the tests:

| What the code assumes | What the SDK declares |
|---|---|
| `documents.create({ payload, file })` | `create(request: DocumentCreateRequest)`, `DocumentCreateRequest = { payload, file }` |
| `file: { fileName, content: Uint8Array }` | `DocumentCreateFile = { fileName: string; content: ReadableStream \| Blob \| ArrayBuffer \| Uint8Array }` |
| `created.id` is a document id | `DocumentCreateResponse = { envelopeId: string; id: number }` |
| `documents.get({ documentId })` returns recipients with tokens | `DocumentGetResponse.recipients: Array<DocumentGetRecipient>`, and that type carries `token: string` |

That last row is the one that matters. The `NOSONAR` comment being removed said the migration was blocked because `documents.create()` returns no recipients and the callers read `document.recipients[0].token` to build the signing URL. The extra `documents.get()` round-trip restores exactly that, and the return type is a superset of what `createV0` handed back — so the signing flow keeps working rather than compiling and failing at runtime.

Dropping `uploadPdfBuffer` and the `globalThis.fetch` mock with it is right: the upload is inside the SDK call now, and a lingering global fetch mock is the kind of thing that quietly passes a later test for the wrong reason.

One nit, non-blocking: `fileName` is hardcoded to `"document.pdf"` while `title` is dynamic, so every document lands in Documenso's UI under the same file name. `${title}.pdf` would cost nothing.

PR #2698, opened 2026-09-05
fix(ci): apply migrations after the build, and gate the ones the running code cannot survive

### LINKED ISSUE #2603: fix(ci): api-deploy applies migrations before the build, so a failed deploy leaves the live process on an incompatible schema

aupyay:
## PR Checklist

`scripts/deploy/api-deploy.sh` applies Prisma migrations at step 4 of 8 and cuts over at step 7. The old PM2 process therefore serves traffic against the new schema for the whole build and smoke window, and a failed build or smoke exits without cutting over - leaving the migration applied and the old process running against it indefinitely.

Line 134 is explicit that a smoke failure does not cut over, which is the right call for the process and is exactly what makes this sharp: the rollback restores the code, and nothing rolls back the schema. Additive migrations are unaffected, which is why this has not bitten yet.

Raised by the Codex review on #2599, which renames three columns. That PR was safe on its own terms - all three tables held zero rows - but the hazard is general and applies to the first rename on a populated table.

## What is the new behavior?

**The build no longer runs inside the window.** The build needs the generated Prisma client, not the database, so only `prisma generate` has to precede it. `prisma migrate deploy` moves to after the freshness check and immediately before the smoke boot. A failed install, build or freshness check now leaves the box exactly as it was found. The smoke boot runs the new code and genuinely does need the new schema, so that much of the window is irreducible - reordering past it, as #2603 notes, does not work.

**What is left is said out loud.** `scripts/deploy/lib/migrate.sh` works out which migrations the deploy introduces on top of the commit the box is serving, from `git diff` rather than the database - migrations are append-only, enforced by the immutability check in `_migration.yaml`, so that set is exactly what a rollback would leave applied. The deploy prints it before applying anything, and every path that stops without cutting over prints which migrations the recorded rollback sha does **not** undo.

**And the part that does not depend on anybody remembering.** `scripts/ci/classify-migration.mjs` fails a pull request that adds a migration which drops or renames a column, table or enum type, tightens a column to `NOT NULL`, changes a column type, adds a required column with no default, or truncates - unless the migration says in writing why the deployed code survives it:

```sql
-- deployed-code-survives: the replacement column shipped in
--   20260901120000_add_new_name and every reader moved in #2610, so no
--   deployed query names the old one.
```

That is the expand-contract reasoning written next to the SQL it describes, where a reviewer sees it. It does not suppress the finding - the hazard is still printed - and a token declaration (`ok`) is refused. `DROP INDEX`, `ALTER INDEX ... RENAME TO` and `DROP CONSTRAINT` are deliberately not hazards; a gate that fires on the 26 index renames already in this history is noise, and noise is how a gate gets turned off.

The classifier strips SQL comments before matching, which is load-bearing rather than tidy: Prisma writes `You are about to drop the column ...` as a warning header on precisely the migrations worth catching, and `20260615100345_parent_patient_migration` in this repo carries a real one. It keeps string literals, because migrations here run DDL out of `EXECUTE '...'` inside `DO` blocks.

## Verification

Run at `fed532dbb`:

| Command | Exit | Result |
|---|---|---|
| `pnpm run test:scripts` | 0 | 288 tests, 288 pass, 0 fail (whole suite, not just the new file) |
| `scripts/deploy/tests/migrate.test.sh` | 0 | 14 passed, 0 failed |
| `scripts/deploy/tests/git-sync.test.sh` | 0 | 16 passed, 0 failed |
| `pnpm run lint` | 0 | 10/10 tasks, 2 pre-existing warnings in untouched frontend files |
| `pnpm run type-check` | 0 | 18/18 tasks |
| `bash -n scripts/deploy/api-deploy.sh` | 0 | - |

Every new test was mutation-checked: the logic was broken one rule at a time and the suite was confirmed to go red. Three cases did **not** fail under mutation on the first pass and the tests were rewritten until they did - a nested block comment, `--` line stripping, and two-dot vs three-dot diff range. Two branches turned out to be genuinely unreachable no-ops under mutation (dollar-quote handling in the statement splitter, and `''` escape handling in the scanner) and were removed rather than left untested.

The CI gate was also exercised end to end against a real branch, not only through its unit tests:

| Case | Exit |
|---|---|
| branch with no new migrations | 0, "No new migrations in this pull request." |
| new migration with `DROP COLUMN` and no declaration | 1, hazard reported from the DDL and not from Prisma's warning header |
| same, with `-- deployed-code-survives: ok` | 1, "It needs at least 30" |
| same, with a real two-line declaration | 0, hazard still printed alongside the declaration |
| purely additive migration | 0, "additive only" |

**Aikido found two issues on `fed532dbb` and both are fixed in `b40e0bc58`, not suppressed.**

- *HIGH, and a real gate bypass:* `readDeclaration` matched the marker anywhere in the raw file, so `deployed-code-survives:` inside a string literal - data, not a promise - cleared the hazardous statement next to it. The marker now only counts as the first content of a `--` line comment. Three tests cover it, and all three go red if the anchor is loosened.
- *MEDIUM:* the classifier read a path from `argv` unconstrained. It now uses `resolveWithin` from `scripts/ci/safe-path.mjs`, the same containment `merge-coverage`, `lcov-check` and `affected-matrix` apply to paths they did not author. An escaping path fails the run rather than being skipped.

Suite after the fix: `pnpm run test:scripts` exit 0, 291/291.

## What CI has and has not exercised here

The `deploy script unit tests` job ran both shell suites on ubuntu, not only on the machine this was written on - `16 passed, 0 failed` for git-sync and `14 passed, 0 failed` for the new migration suite (run 33965330247).

**The CI gate step itself does not run on this pull request, and that is expected.** `ci.yaml` calls `_migration.yaml` only when `@yosemite-crew/database` is in the affected matrix, which is the right condition - the gate classifies migrations, and a pull request that adds one necessarily touches that workspace (#2696 adding `20260905120000_notification_archived_at` did trigger the stage). This pull request changes no migrations, so the stage is correctly skipped and the step is unexercised in CI until the first migration-carrying pull request lands after it.

What stands behind it in the meantime is the unit suite (36 tests, every one mutation-checked) and the end-to-end run above against a real branch. Worth stating rather than leaving someone to assume a green tick covered it.

**Not verified, stated plainly.** Aikido has no CLI or MCP on the machine this was written on, so the first scan was the one on this PR rather than a local pre-check. Neither half has been exercised against a live API box; `api-deploy.sh` is only reachable by an explicit `workflow_dispatch`, so the reordering is proved by reading and by `bash -n`, not by a deploy. The `deploy-script-tests` job runs both shell suites on any PR touching `scripts/deploy/**`.

The `git-sync-tests` job is renamed to `deploy-script-tests` since it now runs both suites. `dev` has no branch protection, so no required check changes name under it.

## Related Issue(s)

Fixes #2603

### DIFF
--- a/.github/workflows/_migration.yaml
+++ b/.github/workflows/_migration.yaml
@@ -117,6 +117,45 @@ jobs:
 
           echo "No existing migration was modified."
 
+      # Applying cleanly to an empty database is not the same as the code that
+      # is already deployed surviving the change.
+      #
+      # scripts/deploy/api-deploy.sh applies migrations and then boots and
+      # smoke-tests the new bundle before `pm2 restart` cuts over, so the OLD
+      # process serves traffic against the NEW schema for the length of that
+      # window - and if the smoke boot fails the script deliberately does not cut
+      # over, which leaves it there indefinitely. The rollback restores the code;
+      # nothing rolls back the schema.
+      #
+      # Additive migrations are unaffected, which is why this has not bitten yet.
+      # Raised by the Codex review on #2599, which renames three columns; that PR
+      # was safe because all three tables held zero rows, which is a fact about
+      # that PR and not about renames.
+      #
+      # This does not refuse the change - it refuses an UNDECLARED one. The
+      # declaration is the expand-contract reasoning written down next to the SQL
+      # it describes, where a reviewer sees it.
+      - name: Check destructive migrations say how the deployed code survives
+        if: ${{ github.event_name == 'pull_request' }}
+        shell: bash
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+        run: |
+          set -euo pipefail
+
+          # Only migrations this branch ADDS. An edit to an existing one is
+          # already refused by the immutability check above, and re-classifying
+          # the whole history would fail every pull request on migrations that
+          # shipped long ago.
+          added=()
+          while IFS= read -r path; do
+            [ -n "$path" ] && added+=("$path")
+          done < <(git diff --no-renames --diff-filter=A --name-only "$BASE_SHA...HEAD" \
+                     -- packages/database/prisma/migrations \
+                   | grep -E '/migration\.sql$' || true)
+
+          node scripts/ci/classify-migration.mjs "${added[@]+"${added[@]}"}"
+
       # `migrate diff --from-migrations` replays the whole history into a
       # throwaway database, so it needs one that is not the target.
       - name: Create shadow database
--- a/.github/workflows/deploy-api.yml
+++ b/.github/workflows/deploy-api.yml
@@ -12,8 +12,10 @@ name: Deploy API
 # the laptop, not the judgement.
 on:
   # The deploy itself stays manual (see below). The unit tests for its git-sync
-  # step do not: they are the only thing standing between a refspec edit and a
-  # deploy that stops on a ref-lock error, which is how this script first failed.
+  # and migration steps do not: they are the only thing standing between a
+  # refspec edit and a deploy that stops on a ref-lock error, which is how this
+  # script first failed, and between a diff-range edit and a schema-hazard notice
+  # that names the wrong migrations.
   pull_request:
     paths:
       - 'scripts/deploy/**'
@@ -44,8 +46,8 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
-  git-sync-tests:
-    name: git-sync unit tests
+  deploy-script-tests:
+    name: deploy script unit tests
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -53,6 +55,11 @@ jobs:
           persist-credentials: false
       - name: Run deploy git-sync tests
         run: ./scripts/deploy/tests/git-sync.test.sh
+      # Both suites build their own git history, so neither needs the checkout
+      # to be complete - but migrate.test.sh does need a real one, which is why
+      # it is here rather than mocked.
+      - name: Run deploy migration tests
+        run: ./scripts/deploy/tests/migrate.test.sh
 
   deploy:
     # Only ever on an explicit dispatch. The pull_request trigger above exists for
@@ -62,7 +69,7 @@ jobs:
     # The tests gate the deploy rather than running beside it. Without this the two
     # jobs start together on a dispatch, and a broken ref-sync change could reach a
     # live host while its own suite was still going red in parallel.
-    needs: git-sync-tests
+    needs: deploy-script-tests
     name: Deploy API (${{ inputs.environment }})
     runs-on: ubuntu-latest
     timeout-minutes: 30
--- a/scripts/ci/classify-migration.mjs
+++ b/scripts/ci/classify-migration.mjs
@@ -35,7 +35,11 @@
  * Exits non-zero if any file has a hazard with no declaration.
  */
 import { readFileSync } from 'node:fs';
-import { basename, dirname } from 'node:path';
+import { basename, dirname, join, relative } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { resolveWithin } from './safe-path.mjs';
+
+const repoRoot = join(dirname(fileURLToPath(import.meta.url)), '../..');
 
 /**
  * The marker that declares a hazard survivable, and the floor its explanation
@@ -89,8 +93,16 @@ export const stripSqlComments = (sql) => {
       let depth = 1;
       i += 2;
       while (i < sql.length && depth > 0) {
-        if (sql.slice(i, i + 2) === '/*') { depth += 1; i += 2; continue; }
-        if (sql.slice(i, i + 2) === '*/') { depth -= 1; i += 2; continue; }
+        if (sql.slice(i, i + 2) === '/*') {
+          depth += 1;
+          i += 2;
+          continue;
+        }
+        if (sql.slice(i, i + 2) === '*/') {
+          depth -= 1;
+          i += 2;
+          continue;
+        }
         // Newlines inside the comment are kept so line numbers do not shift.
         if (sql[i] === '\n') out += '\n';
         i += 1;
@@ -107,7 +119,11 @@ export const stripSqlComments = (sql) => {
       // the first and reopening at the second covers exactly the same text, so
       // no `--` can surface outside a literal between them.
       while (i < sql.length) {
-        if (sql[i] === quote) { out += quote; i += 1; break; }
+        if (sql[i] === quote) {
+          out += quote;
+          i += 1;
+          break;
+        }
         out += sql[i];
         i += 1;
       }
@@ -147,7 +163,11 @@ export const splitStatements = (sql) => {
       i += 1;
       // Doubled quotes re-pair on their own; see stripSqlComments.
       while (i < sql.length) {
-        if (sql[i] === quote) { current += quote; i += 1; break; }
+        if (sql[i] === quote) {
+          current += quote;
+          i += 1;
+          break;
+        }
         current += sql[i];
         i += 1;
       }
@@ -202,7 +222,10 @@ const RULES = [
   { kind: 'drops a column', test: (u) => u.includes('DROP COLUMN') },
   { kind: 'renames a column', test: (u) => u.includes('RENAME COLUMN') },
   { kind: 'renames a table', test: (u) => u.includes('ALTER TABLE') && u.includes('RENAME TO') },
-  { kind: 'renames an enum type', test: (u) => u.includes('ALTER TYPE') && u.includes('RENAME TO') },
+  {
+    kind: 'renames an enum type',
+    test: (u) => u.includes('ALTER TYPE') && u.includes('RENAME TO'),
+  },
   { kind: 'drops an enum type', test: (u) => u.includes('DROP TYPE') },
   { kind: 'drops a view', test: (u) => u.includes('DROP VIEW') },
   { kind: 'drops a schema', test: (u) => u.includes('DROP SCHEMA') },
@@ -241,6 +264,21 @@ export const classifyMigrationSql = (sql) => {
 };
 
 /**
+ * A line comment whose first content IS the marker. Anchored on purpose.
+ *
+ * A plain substring search over the raw file was a gate bypass: the marker text
+ * inside a string literal - `INSERT ... VALUES ('deployed-code-survives: ...')`
+ * - would have been read as a declaration and cleared the hazardous statement
+ * next to it. Raised by the Aikido review on this change.
+ *
+ * Requiring the comment to START with the marker also rules out a comment that
+ * merely mentions it in passing. The cost is that a declaration written inside
+ * a block comment is not recognised; the failure message names the `--` form,
+ * and refusing an unrecognised declaration fails safe.
+ */
+const DECLARATION_LINE = new RegExp(`^\\s*--\\s*${DECLARATION_MARKER}`);
+
+/**
  * The explanation attached to the marker, or null.
  *
  * Read from the RAW file, before comments are stripped - the declaration is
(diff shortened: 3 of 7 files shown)


ankit-yc (changes requested):
Not approving yet — one finding, and it is in the new safety notice rather than in the ordering change. The reordering itself is right and I have no objection to it.

## The hazard notice cannot fire on the path most likely to strand a partial schema

`api-deploy.sh` runs under `set -euo pipefail`. `stopped_without_cutover` is called from exactly two places, both `if`/`||` branches that do not trip `set -e`: the smoke-boot status check and the port check. It is never reached when a command simply fails.

`prisma migrate deploy` applies migrations one at a time and halts at the first failure, leaving the earlier ones applied. So on a three-migration deploy where the third fails:

- line 169 exits non-zero
- `set -e` terminates the script immediately
- line 170 `MIGRATIONS_APPLIED=1` never runs
- no `EXIT` trap is installed yet — the first one is set at line 177, for the smoke PID
- `stopped_without_cutover` is never called

The operator sees a Prisma error and nothing saying the schema has moved. Two migrations are applied against the old running process, which is the exact state this notice exists to announce.

Reproduced the control flow in a standalone script with the same `set -euo pipefail` structure, a stand-in that prints two successes and then exits 1, and the same flag-guarded function: two "applied" lines, script exit 1, **no notice**.

Moving `MIGRATIONS_APPLIED=1` above the call is not sufficient on its own — I checked that too, and the notice still does not print, because the call site at the end of the script is unreachable once `set -e` has fired. The flag is not the problem; the reachability is.

Handling the failure explicitly does work:

```bash
MIGRATIONS_APPLIED=1
pnpm --filter @yosemite-crew/database run prisma:deploy || {
  echo "migrations failed - some of them may have applied" >&2
  stopped_without_cutover
  exit 1
}
```

Same repro with that shape prints the notice and still exits 1.

Setting the flag *before* the call is deliberate there: if Prisma fails having applied nothing, this over-reports. Over-reporting a schema hazard is the safe direction, and it matches the reasoning already in this PR for refusing an unrecognised declaration.

Worth considering the more general version: **every** `set -e` exit between line 170 and the cutover skips the notice, not just this one — the smoke-boot section has several commands that could fail outright rather than through a status check. An `EXIT` trap installed just before the migration step and disarmed after a successful cutover would cover all of them, and would compose with the existing smoke-PID trap rather than replacing it. Your call whether that is worth the extra care around trap composition; the explicit `||` above closes the specific case I can demonstrate.

## What I checked that is right

- **The ordering change is correct and the reasoning is stated where it belongs.** `prisma generate` writes the client the build compiles against and touches nothing live, so splitting it from `prisma:deploy` is exactly the right cut. A failed install or build now leaves the box untouched, which was the whole of #2603.
- **The window that remains is genuinely irreducible** and the PR says so instead of claiming to have removed it. The smoke boot runs new code, new code needs the new schema.
- **Two-dot `git diff` in `deploy_incoming_migrations` is the right choice**, and for the stated reason: a rollback deploys an older ref where the newer migrations are deletions, so it correctly reports none.
- **`classify-migration.mjs` gets the two hard parts right.** Stripping comments while leaving literals intact is the correct direction given this repo runs DDL out of strings inside `DO` blocks — blanking literal bodies would have hidden the statements most worth catching. And anchoring `DECLARATION_LINE` to the start of the comment rather than substring-matching the raw file closes a real bypass, since the marker text inside an `INSERT ... VALUES` would otherwise have cleared the hazard next to it.
- **`addColumnClauses` splitting per clause** is right and is the kind of thing that would have been wrong the obvious way: asking whether `DEFAULT` appears anywhere in a multi-clause `ALTER TABLE` would clear a required column on the strength of a different one.
- **Excluding `DROP INDEX`, `ALTER INDEX ... RENAME TO` and `DROP CONSTRAINT`** is the right line to draw — Prisma emits index renames routinely, and a gate that fires on them becomes noise and then becomes ignored.
- **`MIN_DECLARATION_LENGTH`** is the part that stops this being a suppression switch. Worth keeping even though it will annoy someone.
- **Gating the deploy job on the tests via `needs:` rather than running them beside it** is the same lesson as the ordering fix, applied to CI.

Happy to approve as soon as the notice can fire. Ping me after your last commit — the ruleset dismisses on push, so there is no point in me ticking before it.

aupyay:
@ankit-yc your finding is right, and it reproduces in the source rather than only in a stand-in. Fixed at `0d519955c`.

Confirmed the control flow in the file: `set -euo pipefail` at line 44, `pnpm --filter @yosemite-crew/database run prisma:deploy` with `MIGRATIONS_APPLIED=1` on the line *after* it, and `stopped_without_cutover` called only from the smoke-status branch and the port-check branch. A bare non-zero exit from Prisma ended the script above the assignment and above both call sites.

**I took the trap version rather than the explicit `||`,** because the trap is the one a later edit cannot fall out of, and here it turned out to remove moving parts rather than add them: the smoke boot had its own arm/disarm `EXIT` pair, so there are now one handler and zero `trap - EXIT` instead of two handlers and a disarm. Both explicit call sites collapse to their plain `exit 1`.

- `trap on_exit EXIT` is armed before the schema can move; `on_exit` kills the smoke process if one is running and then calls the notice.
- `MIGRATIONS_APPLIED=1` is set *before* `prisma:deploy`, guarded on `INCOMING_MIGRATIONS` being non-empty — a deploy carrying no migrations cannot move the schema, so that case stays silent rather than printing "applied 0 migration(s)". Past that it over-reports only if Prisma fails having applied nothing, which is the safe direction you named.
- `CUTOVER_DONE=1` after the `:8080` check, so the notice stops claiming the schema is ahead once the new code is the one serving.
- The decision moved to `deploy_stop_notice` in `lib/migrate.sh`, so it is testable off a live box like the rest of that file.

**Eleven new tests in `migrate.test.sh`, all mutation-checked, each mutation turning exactly one red:**

| mutation | result |
|---|---|
| `MIGRATIONS_APPLIED=1` back after `prisma:deploy` — the original bug | 24 pass, **1 fail** |
| drop `trap on_exit EXIT` | 24 pass, **1 fail** |
| restore the `trap - EXIT` disarm | 24 pass, **1 fail** |
| drop the clean-exit guard | 24 pass, **1 fail** |
| drop the cutover guard | 24 pass, **1 fail** |
| drop the migrations-applied guard | 24 pass, **1 fail** |

Unmutated: **25 passed, 0 failed**. The reachability test is driven through a real `bash` process with the same `set -euo pipefail` shape and a bare `false` standing in for Prisma halting part-way, because what is under test is that an `EXIT` trap survives `set -e` where a call site below the failure does not — calling the function directly would not have shown that. It also asserts the deploy still exits non-zero and still stops, so the fix cannot pass by swallowing the failure. A separate check asserts the ordering on `api-deploy.sh` itself (trap, then flag, then `prisma:deploy`), so the behaviour above stays the file's and not just the stand-in's.

At `0d519955c`, with `git rev-parse HEAD` confirmed in the same shell: `migrate.test.sh` exit 0 (25/25), `git-sync.test.sh` exit 0 (16/16), `classify-migration.test.mjs` exit 0 (36/36). `shellcheck` on the three touched files reports 6 `info` lines against 8 on the previous head — all pre-existing SC1091/SC2015, none new, and it is not a CI gate here either way.

**One correction to something I nearly wrote.** I went looking for a second unguarded path and thought I had one in the two `curl`s at lines 183-184. I did not: a command substitution inside an `echo` argument does not propagate its status, so `set -e` never fires there — only `CODE="$(...)"` is an assignment, and that one is already guarded with `|| echo 000`. The migration call really is the dominant path, as you said. The trap is still the right shape, but for the edit-proofing rather than for a second live bug.

The second commit is a test-harness fix: `line_of` returned non-zero when a line was absent, so the missing-trap mutation aborted the suite through `set -e` instead of reporting which line it could not find.

ankit-yc (approved):
Approved at `0d519955c`. The trap version is the right fix and it is genuinely enforced, not just described. Two non-blocking findings below, neither a reason to hold this.

**Verified rather than taken, in a throwaway worktree detached at `0d519955c` (`git rev-parse HEAD` confirmed in the same shell as every run):**

| Suite | Result |
|---|---|
| `scripts/deploy/tests/migrate.test.sh` | 25/25, exit 0 |
| `scripts/deploy/tests/git-sync.test.sh` | 16/16, exit 0 |
| `scripts/ci/classify-migration.test.mjs` | 36/36, 0 failed |

Seven mutations, each reverted with `git checkout 0d519955c -- <file>` before the next, unmutated 25/25 either side. Six turn exactly one red:

1. `MIGRATIONS_APPLIED=1` moved back after `prisma:deploy` (the original bug) — 1 red
2. `trap on_exit EXIT` deleted — 1 red
3. cutover guard dropped from `deploy_stop_notice` — 1 red
4. status guard dropped — 1 red
5. applied guard dropped — 1 red
6. `trap - EXIT` re-added to `api-deploy.sh` — 1 red

The reachability test earns its keep: driving a real `bash` process with the same `set -euo pipefail` shape is the only way to show that an EXIT trap survives where a call site below the failure does not, and asserting the deploy still exits non-zero stops the fix passing by swallowing the failure.

---

**Finding 1 (nit, not blocking): the seventh mutation stays green.** Swapping the two flags in the real handler —

```bash
deploy_stop_notice "$status" "$CUTOVER_DONE" "$MIGRATIONS_APPLIED" \
  "$ROLLBACK_SHA" $INCOMING_MIGRATIONS
```

— leaves **25 passed, 0 failed**. That inversion makes the notice fire only *after* a verified cutover and stay silent exactly when the schema is ahead of the running code, which is the whole defect this PR exists to close.

The cause is that `deploy-shape.sh` hand-writes its own `on_exit` rather than exercising the real one, so the only thing checked against `api-deploy.sh` is `grep` line ordering. The behaviour is correct today; it is the argument order that is unenforced, and that is the same "proves a branch rather than the enforcement" shape flagged elsewhere today. Cheapest close is one more grep assertion pinning the argument order, or moving `on_exit` into `lib/migrate.sh` so both the real script and the test call the same handler.

**Finding 2 (follow-up issue, pre-existing, not this PR's to fix): the notice is blind on a retry.** `ROLLBACK_SHA="$(git rev-parse HEAD)"` is read *before* `deploy_git_sync`, but a deploy that failed after the checkout has already moved the repo to the target ref. On the retry `ROLLBACK_SHA == HEAD`, so `deploy_incoming_migrations` returns empty — reproduced with the real function in the worktree — `MIGRATIONS_APPLIED` stays `0`, and the notice cannot fire even though `prisma migrate deploy` is re-applying the still-pending migrations and can halt part-way again.

This is not a regression: before this PR the same retry produced a notice claiming "applied 0 migration(s)", i.e. loud and wrong rather than silent. The underlying defect is that the migration set is derived from a working tree a previous failed attempt already advanced. I would rather it were an issue than a change here — the retry is the attempt on which the operator most needs the notice.

**Also checked and right:** the flag guarded on `INCOMING_MIGRATIONS` being non-empty (over-reporting a schema hazard is the safe direction, and a migration-free deploy staying silent is correct for the ordinary case); `local status=$?` as the first line of the handler, so the kill cannot clobber it; every variable the handler reads initialised before the trap is armed, which matters under `set -u`; `SMOKE_PID=""` in place of `trap - EXIT`, giving one handler and no disarm; and `CUTOVER_DONE=1` placed after the `:8080` check rather than after `pm2 restart`, so the notice closes on the process actually serving.

`shellcheck -x` on both files reports only info-level SC1091/SC2015, all on pre-existing lines.

Issue #2611, opened 2026-09-02
fix(backend): five paths record success the system never verified


ankit-yc:
## Summary

Findings from a backend sweep for handlers that report success without doing the work. The mechanical half — missing logs on silent returns, two untrue response strings — is fixed in #2610. These five need a product decision, so they are written up rather than patched.

Each was verified in code and survived an adversarial second pass.

## 1. The companion invite is never delivered

`CompanionOrganisationService.sendInvite` writes a PENDING `patientOrganisation` row with a `randomUUID()` token and stores the recipient in `invitedViaEmail`. **Nothing reads that column.** The invite token exists only in the database, so the clinic it names never learns it was invited.

#2610 changes the response from "Invite sent successfully" to "Invite created", which stops the lie. It does not make the feature work.

**Decision needed:** send the email (needs a template, an unsubscribe path per the existing care-reminder pattern, and a link target), or drop the invite-by-email affordance and keep only the flows where a clinic claims a token it already holds.

## 2. Care reminders are marked SENT regardless of dispatch outcome

`care-reminder.service.ts:389` writes `status: "SENT"`, `sentAt: new Date()` and an audit row `CARE_REMINDER_SENT` **after** `dispatchNotification`, which swallows both failure paths:

```ts
await NotificationService.sendToUser(...).catch((err) => {
  logger.error("Care reminder push notification failed", { reminderId, err });
});
await sendEmail({ ... }).catch((err) => {
  logger.error("Care reminder email failed", { reminderId, err });
});
```

Both failures **are logged**, so this is not invisible to ops. The defect is that the durable record and the clinic's UI both say SENT when nothing was delivered — and a vaccination reminder shown as sent is one nobody follows up.

**Decision needed:** what does partial delivery mean? Push succeeded and email failed is the common case. Options: a `deliveredChannels` array, a `FAILED`/`PARTIAL` status, or keeping SENT to mean "attempted" and surfacing per-channel state separately.

## 3. "Business deleted successfully" is a soft delete that hard-deletes its children

`OrganizationService.deleteById` sets `isActive: false` on the organisation — recoverable — then hard-deletes the rows underneath it:

```ts
await prisma.organization.update({ where: { id }, data: { isActive: false } });
await UserOrganizationService.deleteAllByOrganizationId(organisation.id);
await SpecialityService.deleteAllByOrganizationId(organisation.id);
await OrganisationRoomService.deleteAllByOrganizationId(organisation.id);
```

[...]

PR #2699, opened 2026-09-05
fix(backend): stop confirming an IDEXX batch whose result status did not map

aupyay:
## PR Checklist

- [x] There is an issue for the bug/feature this PR is for — item 4 of #2611.

`mapResultStatusToLabOrder` returns null for any IDEXX status outside its vocabulary. In
`syncLabOrderFromResult` that null had no `else` and no log, so the `LabOrder` status transition
was skipped silently. `pollLatest` then called `confirmLatestBatch(batchId)` and
`recordBatchSyncState` unconditionally, which tells IDEXX the batch was consumed — so the skipped
transition is never re-sent and is lost for good.

Two corrections to how #2611 describes this, both verified in code and by test:

- The result row is **not** dropped. `upsertLabResult` runs regardless of the mapping, so the
  `LabResult` lands and is queryable.
- But `maybeCreateResultArtifacts` returns early unless `result.status` is `COMPLETE`, and a
  `COMPLETE` status always maps. So an unmapped row produces no PDF document and no LAB_RESULTS
  task either — the result is in the database and nothing tells the clinic to look at it.

## What is the new behavior?

- An unmapped status is logged at warn with `resultId`, `orderId`, `status`, `statusDetail` and
  `modality` — the identifiers needed to extend the mapper.
- A batch containing an unmapped row is **not** confirmed. The poll loop logs an error naming the
  batch and stops, leaving the batch for the next poll to retry.

The trade-off is deliberate: a status IDEXX sends that we never map will stall polling until
`mapResultStatusToLabOrder` learns it. That is the fail-closed choice #2610 made for unknown
Stripe subscription statuses, and the warn line names the exact status to add. The alternative —
log and confirm anyway — keeps the queue flowing but keeps losing order transitions.

Three tests added to `test/services/idexx-results.service.test.ts`:

- `leaves a batch unconfirmed when a result status does not map` — asserts no `labOrder.update`,
  no `confirmLatestBatch`, no `labResultSyncState.upsert`, and both log lines.
- `still persists a result whose status does not map` — asserts `labResult.upsert` was called with
  the LabOrder's organisation, and that `TaskService.createCustom` and `DocumentService.create`
  were not.
- `holds back a mixed batch, applying the rows that do map` — one mapped row plus one unmapped:
  exactly one `labOrder.update`, and the batch still unconfirmed.

## Verification

Run at commit `504939a9a`, base `7e67c6f62`:

| Command | Result |
|---|---|
| `npx jest --testPathPatterns=idexx-results.service` | exit 0 — 6 passed |
| `npx jest` (full backend suite) | exit 0 — 465 suites, 11183 tests |
| `npx turbo run lint type-check --filter=backend` | exit 0 — 0 errors, 25 pre-existing warnings in untouched files |

Aikido and SonarQube are not runnable locally here; they are unverified until they run on this PR.

## Related Issue(s)

Addresses item 4 of #2611. Does not close it — the remaining four items each need a product
decision before any code is written, and are written up on the issue thread.

### DIFF
--- a/apps/backend/src/services/idexx-results.service.ts
+++ b/apps/backend/src/services/idexx-results.service.ts
@@ -178,6 +178,7 @@ type ResultOrderContext = {
   appointmentId: string | null;
   createdByUserId: string | null;
   patientId: string | null;
+  unmappedStatus: boolean;
 };
 
 const syncLabOrderFromResult = async (
@@ -188,6 +189,7 @@ const syncLabOrderFromResult = async (
     appointmentId: null,
     createdByUserId: null,
     patientId: null,
+    unmappedStatus: false,
   };
 
   const orderId = coerceString(result.orderId);
@@ -212,6 +214,15 @@ const syncLabOrderFromResult = async (
           externalStatus: coerceString(result.status),
           responsePayload: toJsonInput(result),
         },
+      });
+    } else {
+      context.unmappedStatus = true;
+      logger.warn("IDEXX result status did not map to a LabOrder status", {
+        resultId: coerceString(result.resultId),
+        orderId,
+        status: coerceString(result.status),
+        statusDetail: coerceString(result.statusDetail),
+        modality: coerceString(result.modality),
       });
     }
   }
@@ -294,6 +305,7 @@ const processIdexxResult = async (
   const resultId = coerceStringOrEmpty(result.resultId);
   await upsertLabResult(result, resultId, context.organisationId);
   await maybeCreateResultArtifacts(client, result, resultId, context);
+  return context.unmappedStatus;
 };
 
 const recordBatchSyncState = async (
@@ -345,8 +357,23 @@ export const IdexxResultsService = {
         break;
       }
 
+      let hasUnmappedResult = false;
       for (const result of results as IdexxResult[]) {
-        await processIdexxResult(client, result);
+        const unmapped = await processIdexxResult(client, result);
+        hasUnmappedResult = hasUnmappedResult || unmapped;
+      }
+
+      // Confirming tells IDEXX the batch was consumed and stops it being re-sent, so a batch
+      // we only half-applied must not be confirmed: the LabOrder status transition for the
+      // unmapped row was skipped and would be lost for good. Leave it for the next poll and
+      // stop here. Polling stays stuck on this batch until mapResultStatusToLabOrder learns
+      // the status, which is the intent - the warning above names the status to add.
+      if (hasUnmappedResult) {
+        logger.error(
+          "IDEXX batch left unconfirmed: a result status did not map to a LabOrder status",
+          { batchId },
+        );
+        break;
       }
 
       await client.confirmLatestBatch(batchId);
--- a/apps/backend/src/services/idexx-results.service.ts
+++ b/apps/backend/src/services/idexx-results.service.ts
@@ -360,7 +360,7 @@ export const IdexxResultsService = {
       let hasUnmappedResult = false;
       for (const result of results as IdexxResult[]) {
         const unmapped = await processIdexxResult(client, result);
-        hasUnmappedResult = hasUnmappedResult || unmapped;
+        hasUnmappedResult ||= unmapped;
       }
 
       // Confirming tells IDEXX the batch was consumed and stops it being re-sent, so a batch
--- a/apps/backend/test/services/idexx-results.service.test.ts
+++ b/apps/backend/test/services/idexx-results.service.test.ts
@@ -213,4 +213,97 @@ describe("IdexxResultsService", () => {
       }),
     );
   });
+
+  // Confirming the batch tells IDEXX it was consumed, so a batch whose LabOrder transition we
+  // skipped must stay unconfirmed - otherwise the transition is lost for good.
+  it("leaves a batch unconfirmed when a result status does not map", async () => {
+    mockGetLatestResults.mockResolvedValue({
+      batchId: "batch-1",
+      hasMoreResults: false,
+      results: [
+        {
+          resultId: "result-1",
+          orderId: "order-1",
+          status: "REQUIRES_REVIEW",
+          updatedDate: "2026-06-17T12:00:00.000Z",
+          patient: { patientId: "patient-1" },
+        },
+      ],
+    });
+
+    await IdexxResultsService.pollLatest(1, 1);
+
+    expect(mockedPrisma.labOrder.update).not.toHaveBeenCalled();
+    expect(mockConfirmLatestBatch).not.toHaveBeenCalled();
+    expect(mockedPrisma.labResultSyncState.upsert).not.toHaveBeenCalled();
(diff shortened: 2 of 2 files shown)


aupyay (commented):
Independent review. The diff is small and the intent is right - recording a status transition that never happened is exactly the class of bug #2611 is about. Two defects that the fail-closed change turns from latent into active, and one blast-radius question that I do not think is mine to decide.

**1. A completed review task is recreated on every poll, for every COMPLETE result in the stalled batch.**

`ensureResultTask` dedupes like this:

```ts
const existing = await prisma.task.findFirst({
  where: {
    patientId: params.patientId,
    category: "LAB_RESULTS",
    status: { in: ["PENDING", "IN_PROGRESS"] },
    additionalNotes: buildResultTaskKey(params.resultId),
  },
  ...
```

Once a clinician completes that task its status leaves that set, the guard stops matching, and the next call recreates it. Today that is harmless because the batch is confirmed and never re-polled. This PR removes exactly that protection: an unconfirmed batch is re-fetched and every result in it re-processed on every poll, forever. So in the `holds back a mixed batch` scenario, `result-1` is COMPLETE, its reviewer completes the task, and the next poll hands them a fresh identical one - and again on the poll after that, until someone extends `mapResultStatusToLabOrder`.

The dedupe should key on the result rather than on the task's open-ness: drop the `status` filter, or check for the `LabResult`/document instead. It is a one-line change here and it removes the sharpest edge of the stall.

**2. The result PDF is re-downloaded from IDEXX on every poll, forever.**

In `maybeCreateResultArtifacts`, `client.getResultPdf(resultId)` runs *before* `ensureResultDocument`, which is where the existence check lives. The document is correctly not duplicated, but the fetch that produced it is repeated on every poll cycle of a batch that will never be confirmed - an external API call per COMPLETE result per poll, indefinitely, against a vendor that may well rate-limit. Moving the existence check ahead of the fetch fixes both this and the wasted upload path.

**3. The blast radius is platform-wide, and the PR reads as if it is per-order.**

`pollLatest` builds one client from `IDEXX_GLOBAL_*` env vars, there is no per-organisation loop, and `lab-results.worker.ts` is the only caller. So `break` on an unmapped status stops lab result ingestion for **every organisation on the platform**, not for the affected order and not for the affected tenant. It stays stopped until someone ships a mapper change.

That is a materially bigger trade than the #2610 Stripe precedent the description leans on: failing closed on an unknown Stripe status holds one payment. This holds every clinic's lab results. It may still be the right call - silently recording a transition that did not happen is also bad - but it should be an explicit decision with the blast radius written down, not one inherited from a precedent operating at a different scale.

Related: the only signal is a `logger.error`. Nothing surfaces to a clinic, no task, no alert. A platform-wide ingestion stall whose sole indicator is a log line will be found by a clinician asking where their results went.

Worth considering as a middle path: apply the rows that do map, persist the unmapped ones somewhere replayable, and confirm the batch. That keeps the transition from being lost - the actual goal - without making one unknown status a global stop.

**Verified as correct:** an unmapped status really does produce no document and no task, because `maybeCreateResultArtifacts` returns early unless the status is `COMPLETE` and a `COMPLETE` status always maps. The correction in the description is right, and the `still persists a result whose status does not map` test pins it. `ensureResultDocument`'s dedupe by title is sound.

Nothing here needs a re-review from me except your call on 3.

aupyay (commented):
`23f4d5bc2` confirmed: `hasUnmappedResult ||= unmapped` is semantically identical here. `unmapped` is already fully evaluated on the line above, so there is nothing for the short-circuit to skip. Correct fix rather than a suppression.

**My three findings from the earlier review are still open, and the first two are the ones I would not merge without.** They are not style points - this change is what converts them from latent to active, because an unconfirmed batch is re-fetched and re-processed on every poll forever.

1. **`ensureResultTask` recreates a completed review task on every poll.** Its guard is `status: { in: ["PENDING", "IN_PROGRESS"] }`, so once a clinician completes the task the dedupe stops matching. Today that never bites because the batch gets confirmed and never comes back. In the `holds back a mixed batch` scenario it bites every poll cycle: `result-1` is COMPLETE, someone reviews it, and the next poll hands them an identical fresh task. Dropping the `status` filter from that `findFirst` is the fix.

2. **The result PDF is re-downloaded from IDEXX on every poll, forever.** `client.getResultPdf(resultId)` runs in `maybeCreateResultArtifacts` *before* `ensureResultDocument`, which is where the existence check lives. The document is correctly not duplicated; the vendor API call that produced it is not guarded at all. Move the existence check ahead of the fetch.

3. Unchanged from before, and the one that is yours to decide rather than mine: `pollLatest` builds one client from `IDEXX_GLOBAL_*` with no per-organisation loop, so `break` stops lab ingestion **platform-wide**, and the only signal is a `logger.error`.

Fix 1 and 2 and I have nothing left. 3 wants a sentence in the PR body stating the blast radius, whichever way you land.

ankit-yc (approved):
Reviewed from L2 (`ankit-yc`), so this is a genuine cross-account approval rather than a comment — see my channel note, the ruleset requires `required_approving_review_count: 1` and a comment does not satisfy it.

**Approving.** The change is correct and strictly better than what is on `dev`: an unmapped status previously skipped the `LabOrder` transition with no else branch, no log and no signal, and then confirmed the batch anyway, which told IDEXX it was consumed and lost the transition permanently. Read the diff rather than the description; the description matches it.

The two tests earn their place. `leaves a batch unconfirmed when a result status does not map` asserts the absence of `confirmLatestBatch` **and** of `labResultSyncState.upsert`, which is the pair that actually constitutes "not consumed" — asserting only the first would pass while the sync state still advanced. `still persists a result whose status does not map` pins the thing that is easy to regress later, that the result row is not collateral damage of the new early exit.

**One finding, not blocking, but it needs filing before this is forgotten.**

`pollLatest` authenticates with `IDEXX_GLOBAL_USERNAME` / `IDEXX_GLOBAL_PASSWORD` and a single `IdexxResultsClient`. The batch loop is global — `organisationId` is derived per result, not per poll. So the `break` on `hasUnmappedResult` does not stall *a* queue, it stalls **IDEXX result ingestion for every organisation on the platform**, and it stays stalled: the unconfirmed batch is re-fetched on the next poll and hits the same unmapped row again. One unrecognised status value from one clinic stops lab results for all clinics until someone extends `mapResultStatusToLabOrder` and deploys.

I still think fail-closed is the right direction — losing order transitions silently is worse than stopping — so this should not block the merge, and the status quo is worse than this PR. But two things follow:

1. The PR body says it "stalls the queue until the mapper is extended". That understates it; it is platform-wide and indefinite. Worth saying plainly so whoever meets it at 2am knows what they are looking at.
2. A `logger.error` is not an operational signal. This needs something that actually pages or alerts, or a dead-letter path that quarantines the offending result and lets the rest of the batch through. IDEXX adding a new status value is a routine event on their side and a full outage on ours.

Happy for that to be a follow-up issue rather than more work here.

Issue #2646, opened 2026-09-02
SuperAdmin approve/reject/disable decisions are not enforced at product sign-in


ankit-yc:
## Summary

The SuperAdmin panel's Approvals feature writes its decisions into SuperTokens `UserMetadata` (`approvedAt`, `rejectedAt`, `disabledAt`, `rejectionDisabled`), and its users page can "disable" an account the same way. Nothing in this repo reads those keys. `packages/auth` and `apps/backend` only ever consult `metadata.role` (`auth.router.ts`), so:

- a "Pending" business account can sign in and use the product as if it were approved;
- a "Rejected" or "Disabled" account has its live sessions revoked by the panel, but can simply sign in again.

The panel's own copy ("rejecting disables the account and signs it out everywhere") is therefore only true for the panel's own login, which does check `disabledAt` in its `config/backend.ts`.

## Proposed fix

In `packages/auth` (SuperTokens config overrides), read the user's metadata during `signIn` / session creation for the `emailpassword` recipe and refuse with a clear error when `disabledAt` is a number (covers both manual disable and rejection). Whether `approvedAt` should also be required for business accounts is a product decision; if approval is meant to gate access, the same hook is the place, otherwise the panel copy should say the queue is informational.

Passwordless and social (mobile app) accounts must stay ungated: pet parents never go through approval.

PR #2702, opened 2026-09-05
fix(auth): enforce disabled business accounts at sign-in

aupyay:
## PR Checklist

The SuperAdmin panel writes its approval decisions into SuperTokens `UserMetadata` (`approvedAt`, `rejectedAt`, `disabledAt`, `rejectionDisabled`), and nothing in this repository reads them. `packages/auth` and `apps/backend` only ever consult `metadata.role`, so a rejected or disabled business account has its live sessions revoked by the panel and can then simply sign in again. The panel's copy - "rejecting disables the account and signs it out everywhere" - is true only for the panel's own login, which does check `disabledAt` in its own backend config.

## What is the new behavior?

The `emailpassword` `signIn` override reads the user's metadata after a successful credential check and returns `WRONG_CREDENTIALS_ERROR` when `disabledAt` is a number.

Three properties worth being explicit about:

- **The refusal is generic.** A disabled account gets the same answer as a wrong password, so the response does not disclose that the address exists or that it was disabled.
- **It runs only after `status === 'OK'`.** A failed credential check returns before the metadata lookup, so a wrong password costs no extra round trip and cannot be used to probe account state.
- **Passwordless and social are untouched.** Pet parents never go through approval, and gating them was explicitly out of scope in the issue.

## Scope - this does not close the issue

#2646 has three parts and this addresses one of them. **Do not close it on merge.**

- Fixed: a *disabled* account can no longer sign in again after the panel revokes its sessions.
- Not fixed: a **Pending** business account can still sign in and use the product as if approved. The issue leaves `approvedAt` explicitly as a product decision - either approval gates access, in which case the same hook is the place for it, or the panel copy should say the queue is informational. That decision has not been made.
- **Worth a check by someone with panel access:** the issue states `disabledAt` "covers both manual disable and rejection", and this guard keys on `disabledAt` alone. If a rejection in the panel writes only `rejectedAt` or `rejectionDisabled` without also setting `disabledAt`, rejected accounts still sign in and this PR would not show it. Nothing in this repository can confirm which keys the panel writes.

For the same reason, `grep disabledAt` in this repo finds only unrelated Prisma columns on `integration.service.ts`. The writer is the external panel; the guard is not reading a key nobody sets.

## Verification

Run at `e0386953d`:

| Command | Exit | Result |
|---|---|---|
| `pnpm --filter backend exec jest --testPathPatterns=supertokens.config` | 0 | 1 suite, 21 tests passed |
| `npx jest` (full backend) | 0 | 465 suites, 11182 tests passed |
| `pnpm --filter @yosemite-crew/auth run type-check` | 0 | - |
| pre-push hook (monorepo lint + type-check) | 0 | - |

Both new tests were mutation-checked. Removing the disabled check fails 2 of 21; removing the `status !== 'OK'` early return fails 1 of 21. Restored, 21/21.

**One thing about that full-suite number, because the first run said otherwise.** The initial run reported 86 failed suites and 55 failed tests, exit 1, on `zod_1.z.uuid is not a function` - the worktree had zod 3.25.76 installed while the repo pins 4.5.4 through a pnpm override. The failures were an environment fault, not this change. After `CI=1 pnpm install --frozen-lockfile --config.confirmModulesPurge=false` and a Prisma client regeneration the suite is 465/465. The focused suite passed in both states, which is precisely why it was not enough on its own.

**Not verified, stated plainly.** Aikido has no CLI or MCP on this machine, so the first scan is the one on this PR. Nothing here was exercised against a live SuperTokens instance or the SuperAdmin panel; the sign-in override is proved by unit tests against a mocked recipe, not by a real disabled account being refused.

## Related Issue(s)

Related to #2646

### DIFF
--- a/apps/backend/test/auth/supertokens.config.test.ts
+++ b/apps/backend/test/auth/supertokens.config.test.ts
@@ -1,3 +1,25 @@
+const mockEmailPasswordInit = jest.fn((config: unknown) => ({
+  name: "emailpassword",
+  config,
+}));
+
+jest.mock("supertokens-node/recipe/emailpassword", () => ({
+  __esModule: true,
+  default: {
+    init: mockEmailPasswordInit,
+  },
+}));
+
+const mockGetUserMetadata = jest.fn();
+
+jest.mock("supertokens-node/recipe/usermetadata", () => ({
+  __esModule: true,
+  default: {
+    init: jest.fn(() => ({ name: "usermetadata" })),
+    getUserMetadata: mockGetUserMetadata,
+  },
+}));
+
 const mockPasswordlessInit = jest.fn((config: unknown) => ({
   name: "passwordless",
   config,
@@ -58,6 +80,8 @@ const restoreEnv = () => {
 describe("@yosemite-crew/auth supertokens config", () => {
   beforeEach(() => {
     jest.resetModules();
+    mockEmailPasswordInit.mockClear();
+    mockGetUserMetadata.mockReset();
     mockPasswordlessInit.mockClear();
     mockThirdPartyInit.mockClear();
     delete process.env.AUTH_APPLE_CLIENT_ID;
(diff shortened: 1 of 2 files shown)


aupyay:
Closing this. The change it carries should not ship on its own, and the fuller fix is being built elsewhere.

**Correcting my own PR body.** It said "the writer is the external panel; the guard is not reading a key nobody sets." I cannot substantiate that from this repository and should not have asserted it. What I can show:

- There is no SuperAdmin app in this repo (`apps/` is backend, desktop, frontend, mobileAppYC) and no `config/backend.ts` anywhere, so the panel #2646 describes - the one that "checks `disabledAt` in its `config/backend.ts`" - is a separate codebase. Whether it writes `disabledAt` into SuperTokens `UserMetadata` is not knowable from here, in either direction.
- What IS in this repo is a second disable path that the guard does not read at all: `SuperAdminBusinessService.updateBusiness` (`super-admin-business.service.ts:273-278`, behind `PATCH /businesses/:id`) writes `Organization.isVerified` and `Organization.isActive` in Postgres.
- Nothing gates sign-in on either. The only `isActive` check in the auth path is `require-active-account.ts`, and that reads **`User.isActive`** - the soft-delete flag set by `UserService.deleteById` - on developer routes only. It never looks at the organisation.

So `Organization.isActive` is unenforced regardless of what the external panel does, and a fix reading only SuperTokens metadata would leave it that way.

Two further gaps worth carrying forward, both correctly identified against this change:

1. The override is on `emailpassword` `signIn` only. `consumeCode` (email OTP) and `signInUp` (third party) return with no account-state check, so even a correctly wired flag is bypassable through either.
2. The tests mock `getUserMetadata`, so they prove the branch, not the enforcement. A test that mocks the signal it is testing for cannot tell you the signal exists.

Worth keeping from this change: the non-disclosing `WRONG_CREDENTIALS_ERROR` return and the early return before the metadata lookup on a failed credential check, plus their two tests. Both are right and both mutation-check cleanly.

**Open question for whoever picks this up, and it is a product question:** if the external panel does write `disabledAt`, then there are two independent disable surfaces writing to two different systems, and a complete fix has to read both or the two have to be reconciled. That needs an answer from someone with panel access before the shape of the fix is settled.

Branch `fix/2646-enforce-account-disable` stays for now at `e0386953d` if the tests are useful as a base.

Issue #2701, opened 2026-09-05
feat(mobileAppYC): the pet owner cannot see prescriptions or request a refill


ankit-yc:
## Problem

`Prescription`, `PrescriptionItem` and `PrescriptionDispenseRequest` are all in the schema, and `prescription.router.ts` is mounted. The pet owner app has **no reference to any of them**.

The word "prescription" does appear in nine mobile files, which is what makes this easy to miss. All nine are document-category labels, FAQ copy and legal text — none calls an endpoint. The only thing an owner can do today is photograph a prescription and file it as a generic document, unconnected to the record the clinic actually holds.

## Why it matters

This is the highest-frequency clinical interaction a pet owner has between visits: what is my animal on, what is the dose, and can I have more. The practice already records all of it.

## Proposed scope

1. Read-only active prescription list for a companion, owner-scoped, sourced from `Prescription` / `PrescriptionItem`.
2. Prescription detail — drug, dose, route, frequency, prescriber, dates.
3. Refill request, writing a `PrescriptionDispenseRequest`. The model exists; this is the one write.

Deliberately out of scope: dispensing, pricing and anything that would let the app change a prescription.

## Notes for whoever builds it

- Owner scoping must go in the `where`, not be applied after the query. The same class of bug was just fixed in `markNotificationAsSeen` in #2696, where a client-supplied id was updated with no owner in the `where` clause.
- Reuse the response shape the PIMS clinical lists already return rather than inventing a second one.
- Pin the response shape in tests. `notificationService.ts` currently accepts `id | _id | notificationId` and `title | heading` for the same fields, which is what contract drift looks like after the fact — do not widen a parser instead of agreeing a shape.

## Related

Found while surveying the mobile surface after #2625. Part of a broader pattern: the owner-facing half of the clinical record the practice holds — lab results, allergies, problems, flags, consents, care reminders, insurance claims — has no surface in the app.

aupyay:
Backend refill write landed in #3251 (merged into dev): `POST /v1/prescription/mobile/:id/refill`, reusing the same `PrescriptionDispenseRequest` upsert the PIMS side already runs on sign, so a repeat ask before the first fill is reviewed is a no-op and a fresh ask after it's dispensed opens a genuine new one.

Remaining scope on this issue: the mobile UI (list detail + refill button) — new-screen work, so it goes through the design-first pass before building, not a plain code fix. Leaving this issue open for that.

PR #2703, opened 2026-09-05
feat(backend): owner-scoped prescription reads for the pet owner app

ankit-yc:
## PR Checklist

- [x] There is an issue for the feature this PR is for: #2701.

The pet owner app has no prescription surface at all. `Prescription`, `PrescriptionItem` and `PrescriptionDispenseRequest` are in the schema and `prescription.router.ts` is mounted, but every route on it is staff-facing — `requireWebAuth`, organisation-scoped, RBAC on `prescription:view:any`. There is no mobile-authenticated path.

The word "prescription" does appear in nine files under `apps/mobileAppYC/src`, which is what makes the gap easy to miss. All nine are document-category labels, FAQ copy and legal text; none calls an endpoint. Today an owner can photograph a prescription and file it as a generic document, unconnected to the record the practice holds.

## What is the new behavior?

`GET /v1/prescription/mobile` returns the prescriptions the signed-in parent is entitled to read, newest first.

**Two gates, not one.** A `ParentPatient` link is only access when its `status` is `ACTIVE`, and a co-parent link carries a `permissions` object in which `medicalRecords` may be `false`. Such a co-parent can see appointments and tasks but must not see what the animal has been prescribed, so the permitted patient ids are resolved first and the prescription query is scoped to that set in the `where` — not filtered after the fact.

**Only finalised artifacts reach the owner.** `DRAFT` and `IN_PROGRESS` are the clinician still writing; `VOID` was withdrawn. Only `COMPLETED` and `SIGNED` are returned.

## The scoping decision, and what it costs

`ClinicalArtifact` has no companion or patient column, and its `appointmentId`, `caseId` and `encounterId` are all nullable. `Appointment.patient` is `Json` rather than a relation, so artifact → appointment → animal is not a joinable path. `Encounter.patientId` is a required column, so the only trustworthy binding is `encounterId → Encounter.patientId → ParentPatient`.

This read therefore **fails closed on a missing encounter**. The cost is deliberate and should not be discovered later: **a prescription written outside an encounter never reaches the owner's app, and nothing tells them it exists.** That is the safe direction for a wrong answer here — the unsafe one would show an owner another client's medication. It matches the existing precedent in `clinical-artifact.service.ts`, which already declines to act on an artifact with no encounter, and which is there because scoping on organisation alone once let a caller name any colleague's checked-in appointment.

A denormalised `companionId` on `ClinicalArtifact` would remove the hop, but it has to be nullable for exactly the rows that motivate it, so the read path would still have to fail closed and the migration would remove no logic. It is worth doing only together with a write-path guarantee that sets it at creation. That is filed separately rather than front-loaded here.

## Out of scope

The refill request (`PrescriptionDispenseRequest`) and the mobile UI. This is the read path and its scoping.

## Verification

Run on `9f410b99a`:

- `npx jest` in `apps/backend` — 466 suites, 11193 tests, exit 0. Baseline on `dev` is 465 suites / 11183 tests; this adds one suite of 10.
- `npx jest --testPathPatterns=mobile-prescription` — 10 passed, exit 0.
- `npx tsc --noEmit -p apps/backend/tsconfig.json` — exit 0.
- `pnpm run lint --filter=backend` — exit 0; the 26 warnings are pre-existing and none are in files touched here.
- `tsx scripts/ci/openapi-drift.mjs` — exit 0, run with the same placeholder env the workflow uses.

**The tests were mutation-checked, and one of them did not hold up the first time.** Removing the `medicalRecords` gate turned 3 tests red, as intended. Removing the `if (!encounterId)` guard alone turned **none** red — that guard is TypeScript narrowing, and the map lookup on the next line is what actually excludes the row. Removing both guards turns 2 tests red, so the scoping is genuinely covered; the comment in the source says "narrowing only" rather than claiming to be the gate.

Not verified here: Aikido and Sonar have no CLI on this machine, so both are unverified until they run on this PR. The endpoint has not been exercised against a live database — the scoping is covered by the tests above, not by a running query.

## Related Issue(s)

Refs #2701

### DIFF
--- a/apps/backend/src/controllers/app/prescription.controller.ts
+++ b/apps/backend/src/controllers/app/prescription.controller.ts
@@ -1,0 +1,54 @@
+import { Request, Response } from "express";
+import { AuthUserMobileService } from "src/services/authUserMobile.service";
+import { MobilePrescriptionService } from "src/services/mobile-prescription.service";
+import { resolveVerifiedUserId } from "src/utils/request";
+import logger from "src/utils/logger";
+
+/**
+ * Resolves the caller to the parent their records hang off.
+ *
+ * Returns null and sends the response, so callers must return without writing
+ * again. Kept separate from the handlers because every owner-facing read has to
+ * do it, and doing it inline is how one of them ends up not doing it.
+ */
+const resolveParentId = async (
+  req: Request,
+  res: Response,
+): Promise<string | null> => {
+  const authUserId = resolveVerifiedUserId(req);
+  if (!authUserId) {
+    res.status(401).json({ message: "Not authenticated: userId is missing." });
+    return null;
+  }
+
+  const authUser = await AuthUserMobileService.getByProviderUserId(authUserId);
+  const parentId = authUser?.parentId?.toString();
+  if (!parentId) {
+    res.status(404).json({ message: "User not found." });
+    return null;
+  }
+
+  return parentId;
+};
+
+export const MobilePrescriptionController = {
+  listPrescriptions: async (req: Request, res: Response) => {
+    try {
+      const parentId = await resolveParentId(req, res);
+      if (!parentId) {
+        return;
+      }
+
+      const prescriptions =
+        await MobilePrescriptionService.listPrescriptionsForParent(parentId);
+      return res.status(200).json({ prescriptions });
+    } catch (err) {
+      logger.error(
+        `Error listing prescriptions: ${
+          err instanceof Error ? err.message : "Unknown error"
+        }`,
+      );
+      return res.status(500).json({ message: "Failed to list prescriptions." });
+    }
+  },
+};
--- a/apps/backend/src/middlewares/companion-access.ts
+++ b/apps/backend/src/middlewares/companion-access.ts
@@ -78,6 +78,28 @@ const isGranted = (
   return (permissions as Record<string, unknown>)[feature] === true;
 };
 
+/**
+ * The whole access rule for one companion feature, in one place.
+ *
+ * Exported because it is not middleware-only. A read that resolves several
+ * companions at once cannot run this middleware per row, so it filters with
+ * this predicate instead - and if it reimplemented the rule it would drift from
+ * the one every path-keyed route enforces. The PRIMARY bypass is the half that
+ * gets forgotten: a primary parent's own permission set is a description of
+ * what they have DELEGATED, and `promoteLinkToPrimary` and `updatePermissions`
+ * both merge caller-supplied overrides over `PRIMARY_PARENT_PERMISSIONS`
+ * without pinning anything but `assignAsPrimaryParent`, so a PRIMARY link
+ * carrying `medicalRecords: false` is reachable and must still pass.
+ *
+ * Callers are responsible for the other two halves of the decision: the link
+ * must be `status: "ACTIVE"`, and no link at all is a refusal.
+ */
+export const hasCompanionFeature = (
+  role: string,
+  permissions: unknown,
+  feature: CompanionFeature,
+): boolean => role === "PRIMARY" || isGranted(permissions, feature);
+
 const notFound = (res: Response) =>
   res.status(404).json({ message: "Companion not found." });
 
@@ -128,7 +150,7 @@ const enforce = async (
 
     if (!link) return notFound(res);
 
-    if (link.role === "PRIMARY" || isGranted(link.permissions, feature)) {
+    if (hasCompanionFeature(link.role, link.permissions, feature)) {
       return next();
     }
 
--- a/apps/backend/src/routers/index.ts
+++ b/apps/backend/src/routers/index.ts
@@ -32,6 +32,7 @@ import expenseRouter from "./expense.router";
 import deviceTokenRouter from "./deviceToken.router";
 import chatRouter from "./chat.router";
 import notificationRouter from "./notification.router";
+import mobilePrescriptionRouter from "./mobile-prescription.router";
 import contactRouter from "./contact-us.router";
 import accountWithdrawalRouter from "./account-withdrawal.router";
 import { OrganizationDocumentController } from "src/controllers/web/organisation-document.controller";
@@ -181,6 +182,7 @@ export function registerRoutes(app: Express) {
   app.use(`/v1/device-token`, deviceTokenRouter);
   app.use(`/v1/chat`, chatRouter);
   app.use(`/v1/notification`, notificationRouter);
+  app.use(`/v1/prescription`, mobilePrescriptionRouter);
   app.use(`/v1/contact-us`, contactRouter);
   app.use(`/v1/account-withdrawal`, accountWithdrawalRouter);
   app.get(
--- a/apps/backend/src/routers/mobile-prescription.router.ts
+++ b/apps/backend/src/routers/mobile-prescription.router.ts
@@ -1,0 +1,14 @@
+import Router from "express";
+import { MobilePrescriptionController } from "../controllers/app/prescription.controller";
+import { requireMobileAuth } from "src/middlewares/auth";
+
+export const mobilePrescriptionRouter = Router();
+
+// Prescriptions for the signed-in parent's companions.
+mobilePrescriptionRouter.get(
+  "/mobile",
+  requireMobileAuth,
+  MobilePrescriptionController.listPrescriptions,
+);
+
+export default mobilePrescriptionRouter;
--- a/apps/backend/src/routers/mobile-prescription.router.ts
+++ b/apps/backend/src/routers/mobile-prescription.router.ts
@@ -1,4 +1,4 @@
-import Router from "express";
+import { Router } from "express";
 import { MobilePrescriptionController } from "../controllers/app/prescription.controller";
 import { requireMobileAuth } from "src/middlewares/auth";
 
--- a/apps/backend/src/services/mobile-prescription.service.ts
+++ b/apps/backend/src/services/mobile-prescription.service.ts
@@ -1,0 +1,177 @@
+import { ClinicalArtifactStatus } from "@prisma/client";
+import { prisma } from "src/config/prisma";
+
+/**
+ * Owner-facing prescription reads for the pet owner app.
+ *
+ * Kept out of ClinicalArtifactService deliberately: that service is scoped by
+ * organisation and authorised on staff RBAC, and the mobile caller has neither.
+ * The scoping rules here are the whole point of the module, so they live in one
+ * place rather than being spread across a controller.
+ */
+
+/**
+ * A prescription reaches an owner only through a *finalised* artifact. DRAFT and
+ * IN_PROGRESS are the clinician still writing; VOID was withdrawn.
+ */
+const OWNER_VISIBLE_ARTIFACT_STATUSES: ClinicalArtifactStatus[] = [
+  ClinicalArtifactStatus.COMPLETED,
+  ClinicalArtifactStatus.SIGNED,
+];
+
+type ParentPatientPermissions = {
+  medicalRecords?: boolean;
+};
+
+/**
+ * The patients whose medical records this parent may read.
+ *
+ * Two gates, not one. A REVOKED or PENDING link is not access, and a co-parent
+ * link carries a `permissions` object in which `medicalRecords` can be false —
+ * that co-parent can see appointments and tasks but must not see what the animal
+ * has been prescribed.
+ */
+export const listPermittedPatientIds = async (
+  parentId: string,
+): Promise<string[]> => {
+  if (!parentId) {
+    return [];
+  }
+
+  const links = await prisma.parentPatient.findMany({
+    where: { parentId, status: "ACTIVE" },
+    select: { patientId: true, permissions: true },
+  });
+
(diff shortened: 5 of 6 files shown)


aupyay (commented):
Reviewed from `aupyay` (L3). The scoping is right and I checked the part most likely to be silently wrong. **One finding I would rather you decided on before I approve**, since the ruleset dismisses an approval on the next push.

**The permission gate is correct, and I verified the case that would have made it useless.** `permissions?.medicalRecords === true` requires the flag to be explicitly true, which is only safe if the primary owner actually has it. `PRIMARY_PARENT_PERMISSIONS` in `parent-companion.service.ts` sets `medicalRecords: true`, and `CO_PARENT_PERMISSIONS` sets it false, so the owner sees their own animal's prescriptions and a co-parent does not until granted. The `=== true` form also matches `isGranted` in `companion-access.ts` exactly, so a malformed or missing `permissions` object denies in both places rather than differing between them.

**Finding — the read is unbounded, and it is a list endpoint.** Neither `findMany` takes a `take`:

- `encounter.findMany({ where: { patientId: { in: patientIds } } })` returns *every* encounter for every permitted animal, for the life of the record.
- every one of those ids then goes into `artifact: { encounterId: { in: [...] } }`.

For an owner with three animals and several years of history this is a growing result set feeding a growing `IN` list, on a mobile endpoint that renders a scrollable list. The two-step is forced — `ClinicalArtifact.encounterId` is a bare nullable String with no relation, so you cannot filter on `artifact.encounter.patientId` in one query — which is exactly why the id list needs a bound rather than being allowed to grow.

Concretely: cursor or offset pagination on the prescription query, and drive the encounter set from the page you are returning rather than materialising all of them first. If you would rather land the read path and paginate as a follow-up, say so and I will approve as is — it is a performance ceiling, not a correctness or scoping hole, and nothing here leaks another client's data. I would not want it to reach an owner with a long history unbounded, though.

**Smaller, non-blocking:** `listPermittedPatientIds` fetches all ACTIVE links and filters `medicalRecords` in JS, which is the shape you deliberately avoided one layer up ("scoped in the `where`, not filtered afterwards"). Fine at this cardinality — a parent has a handful of companions — but the inconsistency is worth a comment saying why, so the next reader does not take it as the pattern.

**On your own mutation-check correction:** the comment now reading "Narrowing only" is the right outcome. A test that goes green when you delete the line it claims to be testing is worth more said out loud than quietly repaired.

aupyay (commented):
Reviewed at `9f410b99a` in a fresh worktree, from `aupyay`.

The scoping design is right and the module boundary is right — keeping this out of `ClinicalArtifactService`, which is organisation-scoped and authorised on staff RBAC, is the correct call rather than bolting a second authorisation model onto it. One thing to change before I approve, plus two reuse points and two follow-ups.

## 1. PRIMARY parents are not exempted, and the shipped model says they must be

`listPermittedPatientIds` requires `permissions.medicalRecords === true` on **every** link, regardless of role, and does not select `role` at all. That contradicts the authorisation model this repo already enforces. `middlewares/companion-access.ts` states it in its own header comment:

> PRIMARY parents bypass. The permission set exists to describe what a primary parent has DELEGATED; it never constrains them.

and implements it as `if (link.role === "PRIMARY" || isGranted(link.permissions, feature))`.

This is not hypothetical — a `PRIMARY` link carrying `medicalRecords: false` is reachable through the shipped API in two ways, both in `parent-companion.service.ts`:

- `updatePermissions` merges caller-supplied `updates` over the target's existing permissions and, when the target is currently primary, forces only `assignAsPrimaryParent = true`. Nothing pins `medicalRecords`.
- `promoteLinkToPrimary` builds the new permissions as `buildPermissions("PRIMARY", permissionsOverride)`, and `buildPermissions` spreads the overrides *over* `PRIMARY_PARENT_PERMISSIONS`. So `{ assignAsPrimaryParent: true, medicalRecords: false }` mints a primary parent with medical records switched off.

A parent in that state reads their companion's passport today (`pet-passport.router.ts` gates on `requireCompanionPermission("medicalRecords", "patientId")`, which the PRIMARY branch satisfies) and would get an empty prescription list here, with nothing saying why. Two answers to the same question in one product.

The direction is safe — this is stricter, not looser — so it is not a hole. But it should be `role === "PRIMARY" || permissions.medicalRecords === true`, selecting `role`, so the owner-facing surface enforces the model the middleware already documents. If you think the primary bypass is wrong, that is a legitimate position, but it is a change to a shipped authorisation rule and belongs in its own PR against `companion-access.ts` rather than being reversed for one endpoint in silence.

I checked the happy path before flagging this, because "what writes the thing this reads" is the trap that has cost us three PRs today: `companion.service.ts:432` links the creating parent as `PRIMARY`, `buildPermissions("PRIMARY")` yields `medicalRecords: true`, so the ordinary owner is not locked out. The gap is only the reachable edge above.

## 2. Two things already exist

- `resolveParentId` in the controller re-implements `findParentIdForAuthUser` (`src/services/shared/parent-identity.ts`), which is there for exactly this and carries the id-space warning in its comment — `req.userId` is the SuperTokens `sub` on `AuthUser.providerUserId`, not `Parent.linkedUserId`. Your version resolves it correctly by the same route, so this is duplication rather than a bug, but the next person changing that mapping will change one of the two.
- The `ParentPatientPermissions` cast duplicates `isGranted` in `companion-access.ts`, including the strict `=== true` and the "anything unreadable denies" behaviour. It is not exported today; exporting it, or lifting both to a shared module, keeps the co-parent permission semantics in one place.

## 3. No index supports this query (follow-up, not a blocker)

Every index on `ClinicalArtifact` is composite and leads with `organisationId` — `@@index([organisationId, encounterId])` and friends. This query filters `artifact.encounterId IN (...)` with no organisation, so none of them apply.

I want to be explicit that the two-query shape is not the problem and I am not asking you to collapse it: `ClinicalArtifact` has no `encounter` relation in the schema, only a bare nullable `encounterId` column, so materialising the encounter ids first is the only way to express this. The follow-up is an index, not a rewrite.

Related and also for later: nothing here is paginated or bounded. Every encounter for every permitted patient, then every finalised prescription ever with all its items. Fine at current volumes, unbounded by construction.

## 4. The controller has no test

54 lines, zero coverage: the 401 with no verified user, the 404 with no parent, the 200, and the 500 catch. The gate is whole-project `--coverage 95` so this will not sink it, but the branch that decides whether an unauthenticated caller gets a 401 or a 500 is worth pinning, and it is cheap.

## Verified rather than taken on trust

Focused suite at `9f410b99a`: `npx jest --testPathPatterns=mobile-prescription` exit 0, 10/10.

Four mutation checks, each reverted before the next:

| gate removed | tests red |
|---|---|
| `permissions.medicalRecords === true` | 3 |
| `status: "ACTIVE"` on the link query | 1 |
| `status: { in: OWNER_VISIBLE_ARTIFACT_STATUSES }` | 1 |
| `patientIdByEncounter.get(...)` membership check | 1 |

So your corrected account is right: the scoping is genuinely covered, and the null-encounter line really is narrowing rather than the gate. The comment now says so.

I also checked the write path rather than assuming it: both `ClinicalArtifactService.createPrescription` and the package-expansion path in `case-encounter.service.ts` create `PrescriptionItem` rows, so reading `items` is correct and the legacy `Prescription.medications` Json is not the source of truth. Note that `Prescription.instructions` and `Prescription.notes` are prescription-level JSON that this response drops entirely — worth a sentence in the issue on whether the owner needs them.

One thing the green tick does not mean: OpenAPI Drift passing is not the spec covering this route. Only the `x-org-id` check is fatal in `scripts/ci/openapi-drift.mjs`; "mounted but absent from the spec" is reported and deliberately non-fatal. The route is undocumented, in line with the existing backlog.

Fix the primary-parent gate, push, and ping me — the ruleset dismisses on push so the approval has to come after your last commit.

aupyay (dismissed):
Approved at `d1581ad6b`, verified rather than ticked.

The extraction is faithful, not a lookalike, and the mutation that proves it is the one that reaches the *middleware's own* suite. In a throwaway worktree detached at `d1581ad6b` (`git rev-parse HEAD` confirmed in the same shell), `npx jest --testPathPatterns='mobile-prescription|companion-access'` is 2 suites / 34 tests, exit 0. Then:

- dropping the `role === "PRIMARY"` disjunct from `hasCompanionFeature` turns **4** red, and `FAIL` lists **both** `mobile-prescription.service.test.ts` and `companion-access.test.ts`;
- dropping `isGranted` turns **7** red, both suites again.

That both suites fail on a mutation to the single predicate is what shows the middleware and the list endpoint are now answering with the same code rather than two copies that happen to agree today. `turbo run type-check --filter=backend --force` exit 0, 6/6. Worktree clean after reverting each mutation.

Reviewed the substitution directly: `enforce` went from `link.role === "PRIMARY" || isGranted(link.permissions, feature)` to `hasCompanionFeature(link.role, link.permissions, feature)`, which is the same expression — no behaviour change on the middleware. `listPermittedPatientIds` now selects `role` and calls the same predicate. No import cycle: `companion-access.ts` imports only express types, prisma and `parent-identity`, so service → middleware is a one-way edge.

The CO_PARENT default on the `activeLink` fixture is the right call and worth keeping stated — built PRIMARY, every permission assertion in that file would have been vacuous.

Two follow-ups, neither blocking:

1. **Unbounded reads.** `encounter.findMany` has no `take`, and its full id list goes into the prescription `where` as an `IN`. Sharper than "the list is long": every `ClinicalArtifact` index is composite and leads with `organisationId`, including `@@index([organisationId, encounterId])`, so a filter on `encounterId` alone uses **no index at all**. The two-step cannot be collapsed — `ClinicalArtifact.encounterId` is a bare String with no relation to join through — so the fix is an index, not a rewrite.
2. **The raw-body permission merge** in `parent-companion.controller.ts` casts `req.body` to `Partial<ParentCompanionPermissions>` with no allowlist. Not this PR's to fix; it is the mint path that makes the PRIMARY case above reachable at all, and it deserves its own issue rather than riding along with a schema change.

ankit-yc:
Pushed `8ec192c00` — Sonar's quality gate failed at `d1581ad6b` on `new_security_rating` 2 (threshold 1), and it was right.

`typescript:S5689` on `mobile-prescription.router.ts:5`. The file opened with `import Router from "express"`, which binds express's **default** export — the factory — so `Router()` built a full sub-application rather than a router, and `routers/index.ts:185` mounted it as one.

A sub-app carries its own settings, so `app.disable("x-powered-by")` at `app.ts:69` does not reach it. Every response under `/v1/prescription` answered with `X-Powered-By: Express` while the rest of the API did not.

Reproduced against the installed express rather than reasoned about — parent app with `x-powered-by` disabled, one mounted sub-app and one mounted Router:

```
default-import call -> has .listen/.settings (an app): true true
named-import call   -> has .listen/.settings (a router): false false
mounted sub-APP  X-Powered-By: Express
mounted ROUTER   X-Powered-By: undefined
```

Every other router in `src/routers` already uses the named import; this one file did not.

**Evidence at `8ec192c00`, `git rev-parse HEAD` confirmed in the same shell:** full backend `npx jest` exit 0, **466 suites / 11195 tests** — unchanged from `d1581ad6b`; `turbo run lint type-check --filter=backend --force` exit 0, 7/7, 25 pre-existing warnings, none in this file.

No test added on purpose. The enforcement for this is `S5689`, which runs on every PR and is what caught it; a bespoke "the export is a Router, not an app" assertion in one suite would restate a rule the gate already applies repo-wide, and would not have existed for the next router either.

One process note for whoever re-approves: the first full run I did here reported one suite failing to run on `.prisma/client/default.js` missing. That was self-inflicted — I had a `turbo lint` regenerating the Prisma client concurrently with jest. Re-run serially after `prisma generate`, it is 466/466 green. Worth knowing that the stale-Prisma-client symptom also appears when something else regenerates it underneath a running suite.

aupyay (approved):
Re-approved at `8ec192c00`. I reproduced the defect independently before ticking, because "it is one line" is exactly when I would otherwise take it on report.

Against the installed express, on a parent with `x-powered-by` disabled:

```
default-import call -> .listen/.settings : function object
named-import   call -> .listen/.settings : undefined undefined
mounted sub-APP  X-Powered-By: Express
mounted ROUTER   X-Powered-By: undefined
```

So `import Router from "express"` really did build a sub-application with its own settings, and `app.disable("x-powered-by")` on the parent really did not reach it. The fix matches every sibling in that directory — `account-withdrawal`, `activitypub`, `adverse-event`, `aftercare-plan`, `anaesthesia-record`, `appointment-reminder-log` all use the named import.

**Agreed on adding no test, and the reasoning is the part worth keeping.** `S5689` runs on every PR and is the enforcement; a bespoke "the export is a Router" assertion in this one suite would restate a repo-wide rule and would not have existed for the next router anyway. Coverage of the mount shape belongs to the rule, not to a fixture.

**And the observation underneath it is the more useful half.** 34 focused tests, four mutation-checked authorisation gates, two full-suite runs and two reviews — mine included — all passed over a router that was not a router, because every one of them exercised the service and none asserted on the thing being mounted. That is the inverse of the failure we have been chasing all day: not a test asserting on a signal nothing writes, but a defect no test was ever pointed at. I reviewed this file twice and read that import line both times without seeing it.

Issue #1672, opened 2026-06-25
Provider-independent authentication boundary and SuperTokens migration (epic)


ankit-yc:
**Is your feature request related to a problem? Please describe.**

Authentication is currently tied to a single identity provider, and product code depends on provider-specific details. We want to (a) decouple the application from any one auth provider and (b) migrate authentication to SuperTokens as the first supported provider, so future providers (OIDC, Keycloak, etc.) can be added later by writing an adapter only, with no product-code changes.

This is an epic. It is a security-sensitive area, so this issue stays high-level by design: no secrets, provider pool/client identifiers, connection URIs, or environment values belong here. A detailed implementation reference is shared with the implementer privately and is not committed to this public repo.

**Describe the solution you'd like**

Introduce a provider-independent auth boundary, with SuperTokens as the v1 adapter.

Target architecture:
- A provider-neutral boundary: an `AuthProvider` interface, an `AuthService`, and a `createAuthProvider(AUTH_PROVIDER)` factory. Provider-specific code lives only under `providers/<name>/`. Product code calls `auth.requireSession(ctx)` and reads a normalized session, never a provider SDK.
- A normalized `AuthSession` shape: stable internal app user id, auth profile, normalized identity fields, roles/permissions, and MFA state.
- Deployment-time provider selection via `AUTH_PROVIDER`, plus a per-product `AUTH_PROFILE` (staff/web vs app/mobile) because the two products need different login methods.
- A provider-neutral identity mapping table (`auth_identities`) mapping each provider identity to a stable internal app user id. The internal id is preserved across the migration, so existing business data and foreign keys are unaffected.
- Provider-neutral `GET /api/auth/me` and `POST /api/auth/logout`; the frontend consumes a normalized `useAuth()` and confines provider login UI to the auth route.

Current state:
- A SuperTokens foundation already exists in `packages/auth` (init, session helpers, MFA scaffolding) behind a feature flag. Only a small set of auth-status/MFA routes are live today; the primary auth path still runs on the existing provider. This epic finishes and cuts over that foundation.

Workstreams (phases):
1. Provider-neutral boundary: interface, service, factory, startup config validation, normalized session, identity mapping table, and `/api/auth/me` + `/api/auth/logout`. The existing provider stays primary (dual-accept) so nothing breaks while the boundary lands.
2. Web app: adopt the SuperTokens web SDK behind the normalized client, cookie-based sessions, rebuilt auth screens, server-side route protection.
[...]

PR #2715, opened 2026-09-05
fix(backend): mount the notification router as a Router, not a sub-application

### LINKED ISSUE #2711: fix(backend): notification router is mounted as an express sub-application, not a Router
`apps/backend/src/routers/notification.router.ts:1` is `import Router from "express"`. That binds express's **default** export — the app factory — so `Router()` builds a full express **sub-application**, and `routers/index.ts:183` mounts it as one at `/v1/notification`.

This is the same defect Sonar caught as `typescript:S5689` on #2703 (`mobile-prescription.router.ts`), except this one is already on `dev`. It survived because the Sonar quality gate is measured on **new** code — an existing occurrence sits in the backlog and blocks nothing.

## Effect

`app.ts:69` calls `app.disable("x-powered-by")`, so the rest of the API answers with no `X-Powered-By`. The sub-app answers with `X-Powered-By: Express`.

The mechanism is specific and worth recording, because it bounds the blast radius. Express 4.21.2 `lib/application.js` prototype-chains a mounted sub-app's `settings` to the parent's on `mount`, so parent settings are normally visible to the child. But `defaultConfiguration` sets `this.enable('x-powered-by')` as the child's **own** property, which shadows the parent's disabled value, and unlike `trust proxy` there is no corresponding delete in the `mount` handler.

Verified against the installed express rather than reasoned about — parent with `x-powered-by` disabled and `trust proxy` set to 1, one mounted sub-app and one mounted Router:

```
default-import call -> listen/settings: function object
named-import call   -> listen/settings: undefined undefined

/sub/x  ip= 203.0.113.7 | X-Powered-By: Express
/r/x    ip= 203.0.113.7 | X-Powered-By: null
```

So `trust proxy` **is** inherited correctly (both report the forwarded ip) and the leak is confined to `x-powered-by`. This is a fingerprinting/header-hygiene defect on three mobile notification routes, not an authorisation or routing one.

## Fix

`import { Router } from "express"`, which is what the other 136 router files in `apps/backend/src` already do.

## Scope of the negative

`grep -rn 'from "express"' --include='*.ts' apps/backend/src packages`, excluding `import {` and `import type`, returns exactly two lines: this file, and `app.ts:1` (`import express from "express"`, correct usage). No other default-import-as-Router remains in the backend or in `packages`.

aupyay:
Closes #2711.

`apps/backend/src/routers/notification.router.ts:1` was `import Router from "express"`. That binds express's **default** export — the app factory — so `Router()` built a full express **sub-application**, and `routers/index.ts:183` mounted it as one at `/v1/notification`.

Same defect Sonar caught as `typescript:S5689` on #2703, except this one was already on `dev`. It survived because the Sonar quality gate is measured on **new** code: once an occurrence lands, it sits in the backlog and blocks nothing.

## What it actually broke

`app.ts:69` calls `app.disable("x-powered-by")`, so the rest of the API answers with no `X-Powered-By`. Those three notification routes answered with `X-Powered-By: Express`.

The mechanism bounds the blast radius, so it is worth stating precisely. Express 4.21.2 `lib/application.js` prototype-chains a mounted sub-app's `settings` to the parent's in the `mount` handler, so parent settings are normally visible to the child. But `defaultConfiguration` sets `this.enable('x-powered-by')` as the child's **own** property, which shadows the parent's disabled value, and unlike `trust proxy` there is no corresponding delete on mount.

Verified against the installed express rather than reasoned about — parent with `x-powered-by` disabled and `trust proxy` set to 1, one mounted sub-app and one mounted Router, one request each carrying `X-Forwarded-For: 203.0.113.7`:

```
default-import call -> listen/settings: function object
named-import call   -> listen/settings: undefined undefined

/sub/x  ip= 203.0.113.7 | X-Powered-By: Express
/r/x    ip= 203.0.113.7 | X-Powered-By: null
```

So `trust proxy` **is** inherited correctly — both report the forwarded ip — and the leak is confined to `x-powered-by`. This is header hygiene on three mobile routes, not an authorisation or routing defect.

**Helmet does not cover it.** `app.ts:68` mounts `helmet()`, whose `hidePoweredBy` calls `res.removeHeader` when the request passes the parent — before it reaches the sub-app, which then sets the header again in its own `expressInit`. Re-ran the same repro with `helmet()` mounted on the parent: `/sub/x` still answers `X-Powered-By: Express`, `/r/x` still `null`.

## The guard, and why it is not a test

The one-line import fix is the defect. The second change is the enforcement, and it is a judgement call worth a reviewer's attention.

`sonarjs/x-powered-by` is turned off in `apps/backend/eslint.config.js` with the justification *"app.ts calls app.disable("x-powered-by") and mounts helmet"* — which is exactly the assumption a mounted sub-app breaks. And S5689 only gates new code. So nothing would have caught the next occurrence either.

The guard is `no-restricted-syntax` on `src/routers/**/*.ts` matching an `ImportDefaultSpecifier` from `express`. Two deliberate choices:

- **Scoped to routers**, because no router needs the app factory, while `app.ts` legitimately default-imports express.
- **`no-restricted-syntax`, not `no-restricted-imports`**, because flat config replaces a rule's options rather than merging them: adding `no-restricted-imports` in a routers-scoped block would have silently displaced the existing auth-boundary guard (#1672) on those same files. Verified rather than assumed — with the guard in place, adding `import Session from "supertokens-node/recipe/session"` to a router still reports `'supertokens-node/recipe/session' import is restricted ... no-restricted-imports`.

I added no runtime test. A bespoke "this export is a Router" assertion would restate a rule that the linter now enforces repo-wide and would not exist for the next router. The lint rule is the enforcement; the mutation below is the proof it enforces.

## Evidence at `bab45b552`

`git rev-parse HEAD` confirmed as `bab45b552` in the same shell as each run, before and after; tree clean.

| Command | Result |
|---|---|
| `npx turbo run lint type-check --filter=backend --force` | exit 0 — 7/7, 0 errors, 25 pre-existing warnings, none in the touched files |
| `npx jest` (full backend) | exit 0 — 466 suites, 11206 tests |

Guard mutation, reverted after: restoring `import Router from "express"` in `notification.router.ts` fails lint with exit 1 and the message above; the fixed form passes with exit 0.

## Scope of the negative

`grep -rn 'from "express"' --include='*.ts' apps/backend/src packages`, excluding `import {` and `import type`, returns exactly two lines: this file, and `app.ts:1` (`import express from "express"`, correct usage). No other default-import-as-Router remains in `apps/backend/src` or in `packages`. I did not search `apps/frontend`, `apps/mobile` or `apps/desktop`, which do not mount express.

### DIFF
--- a/apps/backend/eslint.config.js
+++ b/apps/backend/eslint.config.js
@@ -119,4 +119,29 @@ export default [
       ],
     },
   },
+  {
+    /* #2711: `import Router from "express"` binds the default export - the app
+       factory - so `Router()` builds a full sub-application and mounting it
+       gives it its own settings. `defaultConfiguration` sets
+       `x-powered-by` as the child's own property, which shadows the
+       `app.disable("x-powered-by")` in app.ts, so the mounted routes answer
+       with the header the rest of the API suppresses.
+
+       Sonar raises this as `typescript:S5689`, but the quality gate measures
+       new code only, so an occurrence that reaches dev stops being enforced.
+       This is scoped to routers because no router needs the app factory;
+       app.ts legitimately default-imports express. */
+    files: ["src/routers/**/*.ts"],
+    rules: {
+      "no-restricted-syntax": [
+        "error",
+        {
+          selector:
+            "ImportDeclaration[source.value='express'] > ImportDefaultSpecifier",
+          message:
+            'Import { Router } from "express". The default export is the app factory, so calling it mounts a sub-application, not a router.',
+        },
+      ],
+    },
+  },
 ];
--- a/apps/backend/src/routers/notification.router.ts
+++ b/apps/backend/src/routers/notification.router.ts
@@ -1,4 +1,4 @@
-import Router from "express";
+import { Router } from "express";
 import { NotificationController } from "../controllers/app/notification.controller";
 import { requireMobileAuth } from "src/middlewares/auth";
 


ankit-yc (approved):
Approved at `bab45b552`. The correction to my reasoning is the important part of this PR and it is right: `S5689` is the enforcement for **new** code only, so "Sonar catches it" stops being true the moment an occurrence reaches `dev`. My conclusion held for #2703 and was wrong as a general rule, and this file is the proof.

**I did not take the lint rule on report — I ran all four cases against the installed toolchain**, applying your two files onto a clean worktree and reverting between each:

| Check | Result |
|---|---|
| lint with the fix in place | exit 0, 25 pre-existing warnings, **0 errors** |
| restore the default import in `notification.router.ts` | exit 1, `no-restricted-syntax` at 1:8 |
| default import in a *different* router (`lab-result.router.ts`) | exit 1, same rule — so it is a repo-wide guard, not a fix pinned to one file |
| `app.ts`'s legitimate `import express from "express"` | still clean |

**Your flat-config reasoning also reproduces, and it is the finding I would least have thought to check.** With your routers-scoped `no-restricted-syntax` block in place, adding `import Session from "supertokens-node/recipe/session"` to a router still reports *"Import the provider-neutral boundary from @yosemite-crew/auth instead of the SuperTokens SDK"* — the #1672 guard is intact. Choosing `no-restricted-syntax` over `no-restricted-imports` because flat config **replaces** a rule's options rather than merging them is the kind of thing that would have silently disarmed an auth-boundary guard on exactly the files most worth guarding, and nothing would have failed.

**Scoping it to `src/routers/**` is right** for the reason stated: no router needs the app factory, and `app.ts` does. A repo-wide ban would have needed an exception, and an exception is the thing that later gets copied.

**On adding no runtime test: agreed, and for a better reason than the one I gave.** The lint rule *is* the enforcement, and unlike Sonar's gate it is measured on every file on every run rather than on the new-code period — which is precisely the gap this PR was filed to close. The mutation is the proof, and I reproduced it rather than accepting it.

**The `x-powered-by` mechanism you traced is more precise than my repro and I am adopting it.** I showed the header leaks; you showed *why* — `defaultConfiguration` sets `x-powered-by` as the child's **own** property, shadowing the parent's disabled value, with no matching delete on mount, while `trust proxy` **is** correctly inherited through the prototype chain. That is the difference between "sub-apps do not inherit settings", which is false, and the true statement, which is narrower. Your `helmet()` re-run matters too: `apps/backend/eslint.config.js` disables `sonarjs/x-powered-by` on the justification that app.ts mounts helmet, and a sub-app breaks exactly that assumption — `hidePoweredBy` strips it on the way through the parent and the child's `expressInit` sets it again after.

Header hygiene on three mobile routes, not an authorisation defect. Stating it that way rather than larger is the right call.

Nothing blocking. Merge when the checks land.

aupyay:
Reviewed at `bab45b552`. I cannot approve it — we are the same GitHub account — so this is a review comment, and it is here because two of the things I checked change what someone should worry about.

## The lint rule works, and it is scoped right

- `src/routers/notification.router.ts` as fixed: `eslint` exit 0.
- Reverting line 1 to `import Router from "express"`: **1 error**, `no-restricted-syntax`, on `1:8`.
- `src/app.ts`, which legitimately default-imports express: exit 0.

So the guard fires on the defect, stays silent on the legitimate use, and the file-scoping is doing the work rather than a comment promising it does.

**The reasoning in that comment is the most valuable part of this PR.** Sonar's gate measures *new code only*, so an occurrence that reaches `dev` stops being enforced — the rule that found this could never find it again once merged. A repo-wide ESLint rule is the right answer to a new-code-only gate, and that argument generalises well past this one import.

## This was the last one — sweep, scoped

On `dev` at `bb9085770`:

- `git grep 'import Router from "express"'` across all `.ts`/`.tsx`: **one** hit, this file.
- Every file in `apps/backend/src/routers/` checked individually for its express import: all use the named `Router` (some alongside `Request`/`Response`/`NextFunction`), except `index.ts` which imports the `Express` *type* and is the mounter.
- Every `Router()` / `express()` construction site under `apps/backend/src`: 16 named-`Router` calls plus `app.ts:42`, which is the real application.
- No `= express()` anywhere in `apps/backend/src` outside `app.ts`.

So #2715 closes the last occurrence, and the rule stops the next one.

## The blast radius is narrower than it looks, and my first guess about it was wrong

I expected the serious half to be `app.set("trust proxy", 1)` at `app.ts:78` — a sub-app not inheriting it would make `req.ip` the proxy address inside those routes, which would quietly break anything IP-keyed, rate limiting included. **That is not what happens.** Probed against the installed express, parent with `trust proxy` set and `x-powered-by` disabled, request carrying `X-Forwarded-For: 203.0.113.7`:

```
sub-app  req.ip=203.0.113.7      req.protocol=https  trust proxy=1
router   req.ip=203.0.113.7      req.protocol=https  trust proxy=1
```

Identical. The reason is in `express/lib/application.js`, and it is deliberate — `onmount` special-cases exactly this setting:

```js
this.on('mount', function onmount(parent) {
  // inherit trust proxy
  if (this.settings[trustProxyDefaultSymbol] === true
    && typeof parent.settings['trust proxy fn'] === 'function') {
    delete this.settings['trust proxy'];
    delete this.settings['trust proxy fn'];
  }
  ...
  setPrototypeOf(this.settings, parent.settings)
});
```

A child app deletes its own `trust proxy` when it is still the default, so the prototype chain resolves to the parent's. Nothing does that for `x-powered-by`, which `defaultConfiguration` sets as an own property with `this.enable('x-powered-by')` and which therefore shadows the parent's `disable` permanently.

**So the complete set a mounted sub-app shadows is what `defaultConfiguration` sets as own properties:** `x-powered-by`, `etag`, `env`, `query parser`, `subdomain offset`, `view`, `views`. `app.ts` sets only `x-powered-by` and `trust proxy`. The first is the entire divergence; the second was never affected.

That is worth having on the record precisely because the scary version is the plausible one. Anyone reading "it was mounted as a sub-application" is entitled to wonder what else broke on those routes, and the answer is: the header, and nothing else in this codebase.

Issue #2709, opened 2026-09-05
Owner prescription list runs two unbounded reads, and the second uses no index


ankit-yc:
Follow-up from the review of #2703 (issue #2701). Raised independently by two reviewers; filing it so it is not lost in the merge.

## What

`MobilePrescriptionService.listPrescriptionsForParent` does two reads, neither bounded:

1. `encounter.findMany({ where: { patientId: { in: permittedPatientIds } } })` — no `take`. Returns every encounter for every animal the parent may read, for the life of the record.
2. `prescription.findMany({ where: { artifact: { encounterId: { in: [...allOfThem] } } } })` — the full id list from step 1 goes in as an `IN`.

It backs `GET /v1/prescription/mobile`, a list endpoint with no pagination.

## Why the second one is worse than a long IN list

Every index on `ClinicalArtifact` is composite and leads with `organisationId`, including `@@index([organisationId, encounterId])`. This query filters on `encounterId` with no organisation, so **no index applies** — it is a scan, not a wide index lookup.

That is not incidental. The mobile caller has no organisation: the whole point of the endpoint is that a pet owner's animals may have been seen at several practices. So the query cannot be made to lead with `organisationId` without changing what it returns.

## What this is not

Not a scoping hole. Permitted patient ids are resolved first and the query is scoped to that set in the `where`, not filtered afterwards; nothing leaks. This is a performance ceiling on a growing table.

## What the fix is not

Not "collapse the two-step". `ClinicalArtifact.encounterId` is a bare `String` with no relation to `Encounter`, so Prisma cannot join through it and materialising the ids first is the only way to write this. Anyone picking this up should not start by trying to remove the round-trip.

## Plausible directions

- An index on `ClinicalArtifact` that leads with `encounterId`, so this query has something to use.
- Cursor pagination on the endpoint, and a `take` on the encounter read, so neither side is unbounded regardless of index.
- A date floor on the encounter read — an owner-facing list almost certainly does not want every encounter ever recorded.

The index is the cheap half and can ship on its own. Related: #2704, which would remove the two-step entirely by binding the artifact to its patient at write time — but that is a schema change on a table nine models depend on, and this should not wait for it.

PR #2716, opened 2026-09-05
fix(backend): quarantine an unmappable IDEXX result instead of halting every poll

### LINKED ISSUE #2708: IDEXX: one unmapped result status halts lab ingestion for every organisation
Follow-up from the review of #2699 (raised by Claude L2 Yosemite Crew). #2699 is the right change and this is not an argument against it — it is the operational gap that change makes visible.

## The problem

`pollLatest` authenticates once with `IDEXX_GLOBAL_USERNAME` and builds a single client. The batch loop is global; `organisationId` is derived per result, not per poll. After #2699, a batch containing a result whose status does not map is deliberately left unconfirmed so the `LabOrder` transition is not lost.

The consequence is wider than that PR's body says. The unconfirmed batch is re-fetched on the next poll and hits the same unmappable row again, so ingestion does not stall *a* queue — **it stalls IDEXX ingestion for every organisation, indefinitely, until `mapResultStatusToLabOrder` learns the status and is deployed.** One unrecognised status value from one clinic stops lab results for all clinics.

A provider adding a status value is routine for them. For us it is currently a full ingestion outage with no signal beyond a `logger.error` line.

## Why not just confirm anyway

Confirming tells IDEXX the batch was consumed, so the skipped `LabOrder` status transition is lost for good. That is the failure #2699 exists to stop, and it is the worse of the two. The fix is not to reverse the fail-closed choice; it is to stop one bad row from being a global halt.

## Options

1. **Quarantine the row, let the batch through.** Persist the unmappable result to a dead-letter table with the payload needed to replay it, then confirm the batch. Ingestion continues; the stuck row is visible and replayable once the mapper is extended. This is the option that removes the outage.
2. **Alert on it.** Whatever the halt behaviour, `logger.error` is not an operational signal. Someone has to be told that ingestion has stopped, not discover it from a clinic asking where a result went.
3. **Scope the halt.** Confirm batches for organisations with no unmappable rows and hold only the affected one — larger, because the batch is the unit IDEXX confirms, not the organisation.

1 and 2 together look like the right pair. 2 is worth doing regardless of which of 1 or 3 is chosen.

## Related

- #2699 — the fail-closed change this follows
- #2611 item 4 — the original finding

ankit-yc:
Closes #2708.

## The outage

`pollLatest` builds **one** client from `IDEXX_GLOBAL_USERNAME`, the batch loop is global, and `organisationId` is derived per result. IDEXX confirms a **batch**, not a row. So #2699's fail-closed `break` does not hold up *a* queue — it holds up **every organisation's**, indefinitely: the unconfirmed batch is re-fetched on the next poll and meets the same unmappable row again. One unrecognised status value from one clinic stops lab results for all clinics until `mapResultStatusToLabOrder` learns it and ships.

## What this does not change

The fail-closed choice is right and is untouched. An unmappable status still does **not** write a `LabOrder` transition. Losing a transition silently is worse than stalling, and that remains true.

What changes is where the skipped transition lives. It is now held in `LabResultQuarantine` with the provider payload needed to replay it, and *that is what makes confirming the rest of the batch safe*. Order is load-bearing and asserted: **quarantine first, confirm second.** If the quarantine write fails there is nowhere holding the transition, so the poll falls back to exactly the previous behaviour and leaves the batch unconfirmed — stalling everyone is bad, and better than confirming a batch whose skipped transition is recorded nowhere at all.

## Decisions worth a reviewer's eye

**`LabResultQuarantine` has no unique key, on purpose.** The failure this table exists to prevent is a result disappearing silently, and an upsert key would do exactly that to two rows in one batch that both arrived without a `resultId`. A duplicate row is cheap; a dropped one is the bug. `create`, never `upsert`.

**A missing `resultId` is stored as `null`, not `""`.** An id the provider never sent is worth seeing as absent rather than as an empty one.

**The operator surface, because a `logger.error` is not an operational signal.** `GET /v1/super-admin/lab-ingestion/quarantine`, behind the existing `router.use(requireAnyAuth, requireSuperAdmin)`. Three deliberate constraints:

- **Bounded at 200**, with the true `total` reported separately — #2709 is the same defect in the other direction, and a truncated page that does not say it is truncated reads as the whole picture.
- **It never selects the stored `payload`.** This endpoint is super-admin and therefore cross-tenant; the payload is the raw provider body — patient name, client name, clinical detail — and nothing in it is needed to decide which status value the mapper is missing.
- **`provider` is validated against a closed set**, not trimmed and trusted, because it goes straight into a `where`.

**The migration is purely additive.** A new table is invisible to the currently running code, so it is safe to apply before the cutover and safe to leave behind on a rollback; nothing existing is altered, renamed or constrained. Both indexes and the table use `IF NOT EXISTS`.

## What is deliberately not here

**Replay is not automated.** Replaying a quarantined row requires the mapper to have learned the status and been deployed, at which point the payload is on hand and the action is a decision rather than a retry. Building an automatic replay before anyone has seen one of these rows would be guessing at the shape of the recovery.

**No paging or resolve endpoint.** The read is bounded and reports its total; if the total is routinely over 200 the right response is to fix the mapper, not to page through the backlog.

## Evidence

At `872b99cd7`, rebased onto `dev` after #2699 merged, `git rev-parse HEAD` confirmed in the same shell before and after each run:

| Command | Result |
|---|---|
| `npx jest` (full backend) | exit 0 — **468 suites / 11238 tests** |
| `turbo run lint type-check --filter=backend --force` | exit 0 — 7/7, **0 errors**, 25 pre-existing warnings, none in the touched files |
| `prisma validate` | schema is valid |
| `prisma generate` | exit 0 |

The change adds exactly 2 suites and 12 tests, counted per suite: `idexx-results.service` 6 → 9, `routers/super-admin` 4 → 6, plus new `lab-ingestion-quarantine.service` (4) and `super-admin-lab-ingestion.controller` (3).

Eight mutations, each reverted with `git checkout 872b99cd7 -- <file>` before the next, 22/22 on the focused set either side:

| Mutation | Red |
|---|---|
| confirm the batch even when the quarantine write failed | 1 |
| drop the `labResultQuarantine.create` | 4 |
| coerce a missing `resultId` to `""` instead of `null` | 1 |
| remove `take: MAX_ROWS` from the list read | 1 |
| add `payload: true` to the select | 1 |
| drop `resolvedAt: null` from the `where` | 1 |
| mount the quarantine route **above** `router.use(requireSuperAdmin)` | 1 |
| widen the `provider` query param to `z.string()` | 1 |

The route-above-the-guard mutation is the one I would not have shipped without: `router.use(...)` sits above every route today, so the new one inherits the guard — which stays true only until someone mounts a route above it. This is cross-tenant lab data, so it is asserted rather than assumed.

**A correction on my own mutation run, since it nearly produced a false negative.** My first attempt at the `resultId` mutation reported green. The mutation had matched a *different* `resultId: coerceString(...)` line — the one in `syncLabOrderFromResult`'s `logger.warn`, which no test asserts on — so it never touched the quarantine write. Re-applied to the correct line it turns 1 red. A mutation that silently lands somewhere else looks exactly like a test that does not enforce anything.

**Not claimed:** the route is absent from `openapi.yaml`, as is every other `super-admin` route — `grep -n "super-admin" apps/frontend/public/static/openapi/openapi.yaml` returns nothing. The OpenAPI Drift check reports "mounted but absent from the spec" non-fatally, so a green tick there is not documentation coverage. Documenting that whole router is out of scope here.

### DIFF
--- a/apps/backend/src/controllers/web/super-admin-lab-ingestion.controller.ts
+++ b/apps/backend/src/controllers/web/super-admin-lab-ingestion.controller.ts
@@ -1,0 +1,36 @@
+import type { Request, Response } from "express";
+import { z } from "zod";
+import logger from "src/utils/logger";
+import { LabIngestionQuarantineService } from "src/services/lab-ingestion-quarantine.service";
+
+// Constrained rather than free text: the value goes into a `where` clause, and
+// the set of providers this poller talks to is closed.
+const querySchema = z.object({
+  provider: z.enum(["IDEXX"]).optional(),
+});
+
+export const SuperAdminLabIngestionController = {
+  listQuarantine: async (req: Request, res: Response) => {
+    const parsed = querySchema.safeParse(req.query);
+    if (!parsed.success) {
+      res.status(400).json({
+        error: "Unknown provider.",
+        code: "INVALID_LAB_PROVIDER",
+      });
+      return;
+    }
+
+    try {
+      const quarantine = await LabIngestionQuarantineService.listUnresolved(
+        parsed.data.provider,
+      );
+      res.status(200).json(quarantine);
+    } catch (error) {
+      logger.error("Failed to list quarantined lab results", error);
+      res.status(500).json({
+        error: "Unable to list quarantined lab results.",
+        code: "LAB_QUARANTINE_LIST_FAILED",
+      });
+    }
+  },
+};
--- a/apps/backend/src/controllers/web/super-admin-lab-ingestion.controller.ts
+++ b/apps/backend/src/controllers/web/super-admin-lab-ingestion.controller.ts
@@ -9,6 +9,12 @@ const querySchema = z.object({
   provider: z.enum(["IDEXX"]).optional(),
 });
 
+// The rows are uuid-keyed, so anything else is a malformed request rather than
+// a miss - answered as such instead of being sent to the database.
+const idSchema = z.object({
+  id: z.uuid(),
+});
+
 export const SuperAdminLabIngestionController = {
   listQuarantine: async (req: Request, res: Response) => {
     const parsed = querySchema.safeParse(req.query);
@@ -30,6 +36,41 @@ export const SuperAdminLabIngestionController = {
       res.status(500).json({
         error: "Unable to list quarantined lab results.",
         code: "LAB_QUARANTINE_LIST_FAILED",
+      });
+    }
+  },
+
+  resolveQuarantine: async (req: Request, res: Response) => {
+    const parsed = idSchema.safeParse(req.params);
+    if (!parsed.success) {
+      res.status(400).json({
+        error: "Invalid quarantine id.",
+        code: "INVALID_QUARANTINE_ID",
+      });
+      return;
+    }
+
+    try {
+      const resolved = await LabIngestionQuarantineService.resolve(
+        parsed.data.id,
+      );
+      if (!resolved) {
+        // One status for both, deliberately: an id that does not exist and one
+        // already resolved are the same outcome for the caller, and telling
+        // them apart would report on rows they did not ask about.
+        res.status(404).json({
+          error: "No unresolved quarantined result with that id.",
+          code: "QUARANTINE_NOT_FOUND",
+        });
+        return;
+      }
+
+      res.status(200).json({ id: parsed.data.id, resolved: true });
+    } catch (error) {
+      logger.error("Failed to resolve a quarantined lab result", error);
+      res.status(500).json({
+        error: "Unable to resolve the quarantined lab result.",
+        code: "LAB_QUARANTINE_RESOLVE_FAILED",
       });
     }
   },
--- a/apps/backend/src/routers/super-admin.router.ts
+++ b/apps/backend/src/routers/super-admin.router.ts
@@ -1,6 +1,7 @@
 import { Router } from "express";
 
 import { SuperAdminBusinessController } from "src/controllers/web/super-admin-business.controller";
+import { SuperAdminLabIngestionController } from "src/controllers/web/super-admin-lab-ingestion.controller";
 import { requireAnyAuth } from "src/middlewares/auth";
 import { requireSuperAdmin } from "src/middlewares/super-admin";
 
@@ -11,5 +12,12 @@ router.use(requireAnyAuth, requireSuperAdmin);
 router.get("/businesses", SuperAdminBusinessController.listBusinesses);
 router.get("/businesses/:id", SuperAdminBusinessController.getBusiness);
 router.patch("/businesses/:id", SuperAdminBusinessController.updateBusiness);
+
+// Lab ingestion holds a result it cannot apply rather than halting the poll for
+// every organisation; this is where those rows become visible to someone.
+router.get(
+  "/lab-ingestion/quarantine",
+  SuperAdminLabIngestionController.listQuarantine,
+);
 
 export default router;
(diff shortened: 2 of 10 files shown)


aupyay (approved):
Approving on my own read at `805b180865f6cd166e79d1020b514f6c3f1907dd`. There is a verified read of this PR already in the team channel; I did not tick on it. `git rev-parse HEAD` confirmed either side of every run, tree clean at the end.

## The design

Quarantine-first, confirm-second, and stop if the first did not happen. That ordering is the whole fix: IDEXX confirms a *batch* and the poll is global, so leaving a batch unconfirmed because one row's status did not map halts ingestion for every organisation, and confirming it without holding the skipped transition loses that transition for good. Writing the row is what makes confirming the rest safe.

## Nine mutations, all red

Assert-uniqueness harness, each reverted before the next, harness refused none, baseline and restore both `30 passed`:

| Mutation | Result |
|---|---|
| confirm the batch **before** quarantining | **1 failed** |
| swallow a failed quarantine and confirm anyway | **1 failed** |
| return success without quarantining anything | **5 failed** |
| `resolve`: drop `resolvedAt` from the `where`, so it is not idempotent | **1 failed** |
| `resolve`: always report success | **1 failed** |
| `list`: drop the `resolvedAt` filter, so resolved rows come back | **1 failed** |
| `list`: remove the row cap | **1 failed** |
| `list`: select the raw provider `payload` as well | **1 failed** |
| mount the resolve route **above** `requireSuperAdmin` | **1 failed** |

The last two are the ones I most wanted to see fail. The router test is a real request through the guard — 403 for a member, 200 for a superadmin — not a grep on mount order. And the payload exclusion is a deliberate decision about cross-tenant clinical data on a super-admin endpoint; **it is pinned by a test rather than by a comment**, so re-adding the field to the `select` is a red suite and not a silent widening.

## The atomicity property is held by `tsc` and by nothing else — verified, not accepted

The code says so in a comment. I ran it:

```
quarantineOperation made `async` and `await`ed
  npx jest idexx-results.service.test.ts   ->  10 passed, 0 failed
  npx tsc --noEmit                          ->  exit 2
    TS2769: Argument of type 'Promise<…>[]' is not assignable to 'PrismaPromise<any>[]'
```

A mocked `create` is eager either way, so the array `$transaction` receives is identical and no test here can distinguish deferred writes from already-running ones. Prisma's brand on `PrismaPromise` is the only thing that can. The comment's warning is therefore exact and worth keeping: **a cast on that line removes the only gate holding atomicity, with every test still green.**

## Gates at `805b18086`

| Command | Result |
|---|---|
| `prisma:generate` | exit 0 |
| `npx jest` (full backend) | **exit 0** — 469 suites / **11261 tests** |
| `scripts/ci/classify-migration.mjs` on the new migration | `additive only`, exit 0 |

## The migration

Additive: a new table, three indexes, RLS enabled. No `lock_timeout` needed and correctly absent — the table is created in the same transaction, so nothing else can hold a conflicting lock on it. `ENABLE ROW LEVEL SECURITY` on a table created in the same migration cannot hide anything that already exists, and it matches the ten migrations already doing this. Guard ordering checked in the router source as well as in the tests: `router.use(requireAnyAuth, requireSuperAdmin)` precedes both new routes.

## One note, not blocking

`payload` holds the raw provider body — patient name, client name, clinical detail — and there is no retention or deletion path: `resolve` sets `resolvedAt` and the row, with its payload, stays for ever. Resolved rows leave the operator's list but not the table. That is the right trade for now (the payload is what makes the held transition replayable), and it is worth an issue rather than a change here: a table of cross-tenant clinical data that only grows is a different problem from the one this PR fixes.

Issue #2714, opened 2026-09-05
A retried deploy reports no incoming migrations, because ROLLBACK_SHA comes from a tree the failed attempt already advanced


aupyay:
## What is wrong

Found by @ankit-yc reviewing #2698. Confirmed against the source.

`scripts/deploy/api-deploy.sh` derives the migration set from the working tree:

```
77: ROLLBACK_SHA="$(git rev-parse HEAD)"
94: deploy_git_sync "$REPO_DIR" "$GIT_REF" "${REQUIRE_PROMOTED_FROM:-}"
100: INCOMING_MIGRATIONS="$(deploy_incoming_migrations "$ROLLBACK_SHA" HEAD)"
```

`ROLLBACK_SHA` is captured *before* the checkout, which is right for its primary job — it is the sha to roll the code back to. The problem is that it is then also used as a proxy for "the commit the running process is serving", and a previous failed attempt has already moved the working tree.

**On a retry after a deploy that failed anywhere after line 94:**

- the repo is already on the target ref, so `ROLLBACK_SHA == HEAD`;
- `deploy_incoming_migrations` returns empty;
- `MIGRATIONS_APPLIED` stays `0` (it is guarded on a non-empty set);
- the hazard notice cannot fire.

Meanwhile Prisma is re-applying the still-pending migrations and can halt part-way again, leaving the schema ahead of the running code with nothing said. The retry is precisely the run most likely to be in that state.

## Not a regression

Before #2698 the same retry printed "this deploy adds no migrations" and, if it got far enough, "applied 0 migration(s)" — loud and wrong rather than silent. #2698 makes the silence cleaner but does not introduce the blindness. The real defect is deriving the set from a working tree a previous attempt already advanced.

## Proposed direction

The question the script actually wants answered is "what is the process currently serving traffic running?", which the working tree stops being a reliable proxy for the moment a deploy fails mid-flight. Candidates:

1. **Record the deployed sha out of band** on a successful cutover — a file under the deploy directory, or the pm2 process env — and read that as `ROLLBACK_SHA` instead of `git rev-parse HEAD`. Authoritative, and it also makes the rollback sha correct on a retry, which it currently is not either.
2. **Ask the database** which migrations are pending (`prisma migrate status`) rather than diffing the filesystem. Answers the schema question directly, but couples the notice to database availability and loses the "what a code rollback would strand" framing that the current wording is built on.

[...]

Issue #2712, opened 2026-09-05
The pre-cutover smoke boot never sends a request body, so a body-parser failure passes it


aupyay:
## What is wrong

`scripts/deploy/api-deploy.sh` boots the freshly built bundle on a smoke port before cutover, specifically so a bundle that cannot serve is caught while the old process is still the one taking traffic. All three probes are GETs:

- `/health`
- `/v1/pet-passport/mobile/companion/probe` (a real route, to prove Express reached a handler)
- `/v1/definitely-not-a-route` (to prove the probe was not hitting a catch-all)

**Nothing sends a request body.** So no failure that only manifests when a body is parsed can be seen by the smoke boot, and the deploy cuts over.

## Why it matters now

#2690 moved `raw-body` across a major to `4.0.0` and carries two workspace patches to `body-parser` so its CommonJS `require('raw-body')` unwraps the ESM default export. If that interop is ever wrong — a patch that stops applying after a `body-parser` bump, a Node downgrade below the `require(esm)` floor of 22.12, a lockfile refresh that drops an override — the symptom is `getBody is not a function` on **every request carrying a body**. That is a total write outage, not a degradation, and:

- the GET probes all pass, because they parse no body;
- `/health` returns 200;
- the deploy cuts over.

Verified on `dev` at `1a09627b3` that the path is currently healthy — booting Express under plain `node` v22 and POSTing JSON returns 200 with the body parsed — so this is about the gate, not a live defect.

This is the same shape as the OpenAPI drift tick that reports "mounted but absent from the spec" non-fatally: a green gate that is trusted for something it does not check.

## Proposed fix

Add one POST probe to the smoke boot, alongside the existing GETs.

Two questions that deserve a deliberate answer rather than a quick patch, which is why this is not a rider on #2698:

1. **Which route.** It must parse a JSON body and must not write anything — the smoke boot runs against the real database, after migrations have been applied. A route that 400s on an invalid body is fine and arguably better: the assertion is "the body was parsed and a handler ran", not "the request succeeded". What must not happen is a 500 or a hang.
2. **What to assert.** `getBody is not a function` surfaces as a 500 with that message in `/tmp/api-smoke.log`. Asserting on the status code alone may be enough; the existing `grep -iE 'ERR_REQUIRE_ESM|Cannot find module|FATAL ERROR'` on the smoke log is the natural place to add the string.

## Context

[...]

PR #2718, opened 2026-09-05
test(ci): arm the real deploy exit handler instead of a copy of it

### LINKED ISSUE #2713: Deploy notice test hand-writes its own exit handler, so transposing the real one's arguments stays green
## What is wrong

Found by @ankit-yc reviewing #2698, against tests I wrote in that PR. It is a real hole and I am recording it rather than leaving it in a review thread.

`scripts/deploy/tests/migrate.test.sh` proves the schema hazard notice fires on a bare `set -e` failure by generating `deploy-shape.sh`, a stand-in that **hand-writes its own `on_exit`**:

```bash
on_exit() {
  local status=$?
  deploy_stop_notice "$status" "$MIGRATIONS_APPLIED" "$CUTOVER_DONE" \
    abc1234 $INCOMING_MIGRATIONS
}
```

The real handler lives in `scripts/deploy/api-deploy.sh`. The only thing tying the two together is a grep-based ordering assertion (trap armed, then flag raised, then `prisma:deploy`).

**So a mutation to the real handler's argument order stays green.** Swapping the two flags —

```bash
deploy_stop_notice "$status" "$CUTOVER_DONE" "$MIGRATIONS_APPLIED" ...
```

— leaves the suite at **25 passed, 0 failed**. That inversion makes the notice fire only *after* a verified cutover and stay silent exactly when the schema is ahead of the running code, which is the entire defect #2698 exists to close. `deploy_stop_notice` takes three positional flags of the same type, so nothing catches the transposition: not the shell, not shellcheck, not the suite.

This is the same "proves a branch, not the enforcement" shape that #2646 was rejected for, in test form.

## Proposed fix

Make the stand-in use the real handler rather than a copy of it. Options, roughly in order of preference:

1. **Extract `on_exit` into `scripts/deploy/lib/migrate.sh`** (or a small `lib/deploy-notice.sh`) so `api-deploy.sh` and `deploy-shape.sh` source the same function. Then the argument order exists once and the existing reachability test covers it. This is what `deploy_incoming_migrations` and `deploy_schema_hazard_notice` already do, and the reason those are testable at all.
2. **Have the test extract the real `on_exit` body out of `api-deploy.sh`** and source it. Cheaper, but it re-introduces a text-matching coupling of the kind that just failed.

Option 1 also removes the grep ordering assertion's reason to exist in its current form.

## Acceptance

The transposition mutation above must turn at least one test red.

## Context

Not a defect in shipped behaviour — the argument order in `api-deploy.sh` is correct today, and #2698's six other mutations each turn exactly one red. This is about the test being unable to hold it correct.

ankit-yc:
Closes #2713.

## What was wrong

`migrate.test.sh` proved the schema hazard notice survives `set -e` by generating a stand-in script that **hand-wrote its own `on_exit`**. The real handler lived in `api-deploy.sh`, and the only thing tying the two together was a `grep` on line ordering.

So transposing the real handler's two flags left the suite at **25 passed, 0 failed** — while inverting the notice completely: it would fire only *after* a verified cutover and stay silent exactly when the schema is ahead of the running code, which is the entire defect it exists to announce. `deploy_stop_notice` takes three positional flags of the same type; nothing catches a swap — not the shell, not shellcheck, not the tests.

## What this does

`on_exit` moves into `lib/migrate.sh` as **`deploy_on_exit`**, beside the two functions it calls and for the same reason those are there: so it can be exercised without pm2, node, a database and a box. `api-deploy.sh` arms it. The stand-in arms it. One handler, one place for the order to be wrong, and tests that vary each flag against the **real** function rather than a lookalike.

It reads the caller's state rather than taking arguments, because an EXIT trap has none to take. `MIGRATIONS_APPLIED`, `CUTOVER_DONE`, `ROLLBACK_SHA` and `INCOMING_MIGRATIONS` are deliberately **undefaulted**, so `set -u` refuses rather than silently choosing the silent answer to "did the schema move". `SMOKE_PID` **is** defaulted, because no smoke process is a normal state for most of the run rather than a caller mistake.

The grep-based ordering assertion stays. It now checks something the stand-in genuinely cannot: that `api-deploy.sh` arms the trap and raises the flag *before* the migration runs. That ordering only exists in the real script, and running the real script needs a live box.

## Evidence

At `e35e77c03`, `git rev-parse HEAD` confirmed in the same shell before and after each run, tree clean:

| Command | Result |
|---|---|
| `scripts/deploy/tests/migrate.test.sh` | **31/31**, exit 0 (was 25) |
| `scripts/deploy/tests/git-sync.test.sh` | 16/16, exit 0 |
| `scripts/ci/classify-migration.test.mjs` | 36 pass, 0 fail |
| `shellcheck -x` on all three changed files, run from `scripts/deploy` | **0 warnings, 0 errors** — three pre-existing SC2015 infos on untouched lines |

Eight mutations, applied through a harness that asserts the target occurs **exactly once** and refuses the edit otherwise, each reverted with `git checkout e35e77c03 -- <file>` before the next, 31/31 either side:

| Mutation | Red before | Red now |
|---|---|---|
| **transpose the two flags in the real handler** | **0** | **3** |
| drop the `SMOKE_PID` default | — | 4 |
| `MIGRATIONS_APPLIED=1` moved back after `prisma:deploy` | 1 | 1 |
| `trap deploy_on_exit EXIT` deleted | 1 | 1 |
| cutover guard dropped from `deploy_stop_notice` | 1 | 2 |
| status guard dropped | 1 | 2 |
| applied guard dropped | 1 | 2 |
| `trap - EXIT` re-added to `api-deploy.sh` | 1 | 1 |

The first row is the issue's acceptance criterion. The three that went 1 → 2 did so because the new `deploy_on_exit` cases exercise the same conditions through the real handler as well as through `deploy_stop_notice` directly.

## A defect the mutations found in the tests I was adding

The `SMOKE_PID` mutation initially **aborted the suite** instead of failing a check, and turned three unrelated checks red on the way. Two causes, both mine and both fixed in `e35e77c03`:

- `on_exit_fired` reported "the notice fired" for **any** stderr. A `set -u` abort inside the handler also writes to stderr, so three silent-case checks were answering on the wrong signal — passing for the wrong reason, which means they would also fail for the wrong reason. It now matches the notice text.
- The command substitutions were unguarded, so a probe that died ended the whole suite under `set -e`. Now `|| true`, the same reason `line_of` already carries it.

Worth stating plainly: a test asserting on "something was printed" rather than on *what* was printed is the same defect shape this PR exists to close, and I wrote it in the commit that closes it. The mutation is what found it.

## What is not here

`#2712` (the smoke probe never sends a body) and `#2714` (`ROLLBACK_SHA` is read before `deploy_git_sync`, so the notice is blind on a retry) are both in this same ~200-line script and are deliberately left for after this lands, rather than opened as parallel PRs that would spend their time rebasing over each other.

### DIFF
--- a/scripts/deploy/api-deploy.sh
+++ b/scripts/deploy/api-deploy.sh
@@ -109,29 +109,19 @@ else
   echo "this deploy adds no migrations - the schema does not move"
 fi
 
-# Said on the way out, not from the branches that remembered to say it.
+# The handler itself is `deploy_on_exit` in lib/migrate.sh, so the test suite
+# can arm the real one instead of a hand-written copy of it. The state it reads
+# - MIGRATIONS_APPLIED, CUTOVER_DONE, ROLLBACK_SHA, INCOMING_MIGRATIONS - is set
+# above; SMOKE_PID is set and cleared around the boot below.
 #
-# This script runs under `set -e`, so a command that simply fails ends it where
-# it stands. Hanging the notice off the two branches that test a status meant
-# every other exit between the migration and the cutover was silent - including
-# the likeliest one of all, `prisma migrate deploy` halting part-way through and
-# leaving the earlier migrations applied. An EXIT trap cannot be walked past.
-#
-# It also owns the smoke process, so there is one handler rather than an
-# arm/disarm pair around the boot that a later edit could fall out of.
-on_exit() {
-  local status=$?
-
-  if [ -n "$SMOKE_PID" ]; then
-    kill "$SMOKE_PID" 2>/dev/null || true
-  fi
-
-  # Deliberate word splitting - one argument per migration. Prisma directory
-  # names are <timestamp>_<snake_case>, so there is nothing here to glob.
-  # shellcheck disable=SC2086
-  deploy_stop_notice "$status" "$MIGRATIONS_APPLIED" "$CUTOVER_DONE" \
-    "$ROLLBACK_SHA" $INCOMING_MIGRATIONS
-}
+# Why a trap at all: this script runs under `set -e`, so a command that simply
+# fails ends it where it stands. Hanging the notice off the two branches that
+# test a status meant every other exit between the migration and the cutover was
+# silent - including the likeliest one of all, `prisma migrate deploy` halting
+# part-way through and leaving the earlier migrations applied. An EXIT trap
+# cannot be walked past. It also owns the smoke process, so there is one handler
+# rather than an arm/disarm pair around the boot that a later edit could fall
+# out of.
 
 say "install"
 pnpm install --frozen-lockfile
@@ -189,7 +179,7 @@ fi
 # Prisma fails having applied nothing, and over-reporting a schema hazard is the
 # safe direction. A deploy that carries no migrations cannot move the schema at
 # all, so it stays silent.
-trap on_exit EXIT
+trap deploy_on_exit EXIT
 if [ -n "$INCOMING_MIGRATIONS" ]; then
   MIGRATIONS_APPLIED=1
 fi
--- a/scripts/deploy/lib/migrate.sh
+++ b/scripts/deploy/lib/migrate.sh
@@ -49,6 +49,43 @@ deploy_incoming_migrations() {
     | sort
 }
 
+# deploy_on_exit
+#
+# The EXIT trap api-deploy.sh installs before it lets the schema move.
+#
+# Lives here rather than in api-deploy.sh because the test that proves the
+# notice survives `set -e` cannot run api-deploy.sh - it wants pm2, node, a
+# database and a box. It used to hand-write its own copy of this handler, which
+# meant the only thing tying the copy to the real one was a grep on line
+# ordering, and transposing the two flags in the real one left the whole suite
+# green while inverting the notice: fires only after a verified cutover, silent
+# exactly when the schema is ahead. Three positional flags of the same type and
+# nothing to catch a swap. Now there is one handler, so there is one place for
+# the order to be wrong and a test that exercises it.
+#
+# Reads the caller's state rather than taking arguments, because an EXIT trap
+# has no arguments to take. Callers must set MIGRATIONS_APPLIED, CUTOVER_DONE,
+# ROLLBACK_SHA and INCOMING_MIGRATIONS before arming it; under `set -u` a
+# missing one aborts loudly, which is the right direction - the alternative is a
+# default, and the only sensible default for "did the schema move" is the silent
+# one.
+#
+# SMOKE_PID is the exception and IS defaulted: no smoke process is a normal
+# state for most of the script's life, not a caller mistake.
+deploy_on_exit() {
+  local status=$?
+
+  if [ -n "${SMOKE_PID:-}" ]; then
+    kill "$SMOKE_PID" 2>/dev/null || true
+  fi
+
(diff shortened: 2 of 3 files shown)


aupyay (approved):
Approving on my own read at `e07a7af2d`, not on the verdict already in the channel. Reviewed in a detached worktree; `git rev-parse HEAD` confirmed either side of every run; tree clean at the end.

## The acceptance criterion, checked from both ends

#2713 exists because transposing `deploy_stop_notice`'s two flags inverted the notice — fires only after a verified cutover, silent exactly when the schema is ahead — and left the suite green. I verified the "before" as well as the "after", because a criterion is only met if it was genuinely unmet:

| Tree | `deploy_stop_notice "$status" "$CUTOVER_DONE" "$MIGRATIONS_APPLIED"` |
|---|---|
| `origin/dev`, mutating the real `on_exit` | **25 passed, 0 failed** — the defect, reproduced |
| this PR, mutating the real `deploy_on_exit` | **30 passed, 4 failed** |

## Ten mutations, all red

Each applied with a uniqueness assertion, each reverted before the next, baseline and restore both `34 passed, 0 failed`. The harness refused none.

| Mutation | Result |
|---|---|
| transpose the two flags | 30 / **4 failed** |
| delete the smoke-process reap | 32 / **2 failed** |
| reap after the notice instead of before | 32 / **2 failed** |
| `kill` → `kill -0` (signals nothing) | 33 / **1 failed** |
| drop the `\|\| true` from the reap | 33 / **1 failed** |
| undo the `${SMOKE_PID:-}` default, so `set -u` aborts the handler | 30 / **4 failed** |
| pass a literal `0` instead of `$status` | 30 / **4 failed** |
| remove the trap arming altogether | 33 / **1 failed** |
| arm the trap after `prisma:deploy` instead of before | 33 / **1 failed** |
| never raise `MIGRATIONS_APPLIED` | 33 / **1 failed** |

The last four are mine and were not in the thread. Every line of the extracted handler, and the arming of it, is pinned by a check that fails when it changes.

## The diff itself

`deploy_on_exit` moves verbatim except for one deliberate change — `"$SMOKE_PID"` → `"${SMOKE_PID:-}"`, documented, and needed because the suite arms the real handler standalone. **Argument order is byte-identical to `dev`**, which is the thing to check on a PR whose whole subject is that order.

Ordering verified in `api-deploy.sh`: `set -euo pipefail` (44), `lib/migrate.sh` sourced (67), `ROLLBACK_SHA` (77), `INCOMING_MIGRATIONS` / `MIGRATIONS_APPLIED` / `CUTOVER_DONE` / `SMOKE_PID` (100-103), trap armed (182). Every variable the handler reads under `set -u` is set before the trap can fire, and the lib is sourced before the trap names a function from it.

Keeping the grep on arming order was the right call. It asserts the one thing the stand-in genuinely cannot: that the real script arms the trap and raises the flag *before* the migration.

## Suites

| Command | Result |
|---|---|
| `tests/migrate.test.sh` | exit 0 — **34 passed, 0 failed** |

No TypeScript is touched, so `sonar` being SKIPPED is correct rather than pending.

## Not blocking

The killed-deploy case — SIGHUP and SIGTERM hand the EXIT trap a status of 0, so a deploy killed after the migration runs this handler and says nothing — is pre-existing, is not touched by this diff, and is tracked as #2721. This PR is what made the handler runnable in isolation and therefore what made that finding reachable at all.

PR #2720, opened 2026-09-05
fix(repo): bound and index the owner prescription list

### LINKED ISSUE #2709: Owner prescription list runs two unbounded reads, and the second uses no index

aupyay:
Closes #2709.

`GET /v1/prescription/mobile` ran two unbounded reads and returned every prescription an owner may read as one array, with no way for a caller to tell a complete answer from a truncated one.

## The scan, measured rather than reasoned

Every `ClinicalArtifact` index leads with `organisationId`. The owner query has no organisation — the point of the endpoint is that a pet owner's animals may have been seen at several practices — so it filters on `encounterId` alone and none of them applies.

I captured the SQL Prisma actually emits rather than assuming its shape, by running the service's exact query args through a real `PrismaClient` with query logging against a scratch Postgres 16 holding the full schema (201 tables, from `prisma migrate diff --from-empty`) and 300,000 `ClinicalArtifact` rows across 3,000 organisations and 150,000 encounters, 75,000 of them prescriptions. It compiles to a `LEFT JOIN` whose entire predicate sits on the artifact side:

```sql
FROM "Prescription"
LEFT JOIN "ClinicalArtifact" AS "j1" ON ("j1"."id") = ("Prescription"."artifactId")
WHERE ("j1"."encounterId" IN ($1..$40) AND "j1"."status" IN (...) AND ("j1"."id" IS NOT NULL))
ORDER BY "Prescription"."createdAt" DESC, "Prescription"."id" DESC LIMIT $43
```

`EXPLAIN (ANALYZE, BUFFERS)` on that statement, 40 encounter ids — one owner with a few animals:

**Without the index**

```
Limit (actual time=10.005..11.450 rows=8 loops=1)
  Buffers: shared hit=3576
  ->  Gather Merge  Workers Launched: 2
        ->  Nested Loop
              ->  Parallel Seq Scan on "ClinicalArtifact" j1 (rows=11 loops=3)
                    Rows Removed by Filter: 99989
Execution Time: 11.484 ms
```

**With `@@index([encounterId, status])`**

```
Limit (actual time=0.710..0.712 rows=8 loops=1)
  Buffers: shared hit=337 read=42
  ->  Nested Loop
        ->  Index Scan using "ClinicalArtifact_encounterId_status_idx" on "ClinicalArtifact" j1 (rows=32 loops=1)
              Index Cond: (("encounterId" = ANY (...)) AND (status = ANY (...)))
Execution Time: 0.729 ms
```

Same statement, same data, index dropped and recreated with an `ANALYZE` either side. 3,576 buffers to 379, and 11.5 ms to 0.7 ms.

**The ratio is not the point.** The plan before is `O(every clinical artifact in the system, for every tenant)` and the plan after is `O(rows this owner matches)`. At 300k rows it is 16x; the gap widens with the table, and the table grows with every clinical document written anywhere in the product.

## What changed

**`@@index([encounterId, status])` on `ClinicalArtifact`.** Leading column is the selective one; `status` is second so the `OWNER_VISIBLE_ARTIFACT_STATUSES` filter is served by the same index rather than rechecked on the heap — the `Index Cond` above carries both. The migration is additive and explains why it is not `CONCURRENTLY`: Prisma runs each migration file in one transaction and Postgres refuses `CREATE INDEX CONCURRENTLY` inside one.

**Keyset pagination on the prescription read.** `take: limit + 1`, cursor by id with `skip: 1`, ordered `(createdAt DESC, id DESC)`. The response gains `nextCursor`, `hasMore` and the applied `limit` beside the existing `prescriptions` array.

**`clampPageSize`, the uuid cursor parse and the take-one-extra split moved to `services/shared/pagination`.** The developer data plane had the only implementation. Importing it from `developer-data.service` would pull that module's graph into an owner-facing read that deliberately has no organisation and no staff RBAC — the same reason `shared/staff-identity` exists. Bounds are parameters, so each surface keeps its own numbers.

## Decisions I would most like challenged

**The encounter read stays unpaged, deliberately.** #2709 lists a `take` there as a plausible direction and I rejected it. Prescriptions are ordered by their own `createdAt`, so a `take` on the encounters drops whichever prescriptions hang off the encounters that fell off the end — and nothing in the response could say which. That is silent loss rather than a page. It is index-served (`Encounter @@index([patientId])`) and scoped to one parent's own animals, so it is bounded by the domain even though it is not bounded by a `take`. The date floor #2709 also suggests changes what the endpoint returns and is a product decision, not a performance fix.

**The response shape changes now because it is free now, and only now.** `git grep -nE "v1/prescription([^s]|$)"` over the whole tree returns exactly one line — `routers/index.ts:185`, the mount statement itself. Nothing consumes it, and `apps/mobileAppYC` has no file mentioning prescriptions at all. Once a client wires up the bare array, adding pagination costs a coordinated release.

**Note the sibling route, because a substring grep answers for it.** `/v1/prescriptions` — plural, `routers/index.ts:179` — is the staff web surface and it has four live callers in `apps/frontend`. It is untouched here. "The prescription endpoint has no consumers" is true of the singular path and false of the plural one.

**Clamp, not 400, for `limit`.** Following the existing convention and its stated reason: a caller asking for 1000 gets 100 and a `limit` in the response saying so. A malformed *cursor* is a 400, because passing it through would make a genuine query failure indistinguishable from a caller mistake — the endpoint would answer 400 for a database outage.

**The offending cursor value is not logged.** It is caller-controlled and a raw CR/LF in it forges a second log line; the 400 already reaches the only party who can act on it.

## Evidence

At `04fb18f24`, `git rev-parse HEAD` confirmed in the same shell before and after each run, tree clean:

| Command | Result |
|---|---|
| `npx jest` (full backend) | exit 0 — 469 suites, 11280 tests |
| `turbo run lint type-check --filter=backend --force` | exit 0 — 7/7, 0 errors, 25 pre-existing warnings, none in the touched files |
| focused set (5 suites) | 86/86 |
| `prisma migrate diff --from-migrations --to-schema-datamodel` | empty — the migration exactly closes the schema drift and adds nothing else |

The full-suite numbers are `dev` at `f472bdf2f` (467 / 11241) plus 2 suites and 39 tests.

## Mutations

Eleven, each applied by a harness that asserts the literal occurs exactly once before substituting and refuses otherwise, each reverted before the next, tree clean at `04fb18f24` at the end. The harness refused none, so every target was unique.

| mutation | red |
|---|---|
| `take: limit + 1` → `take: limit`, so `hasMore` can never be true | 6 |
| drop the `id` tiebreak from `orderBy` | 1 |
| drop `skip: 1`, so the cursor row is returned again | 1 |
| `splitPage` returns every row instead of the page | 5 |
| cursor taken from the read instead of the page | 5 |
| `nextCursor` offered when there is no next page | 3 |
| clamp drops the ceiling | 5 |
| clamp accepts zero and negatives | 6 |
| cursor shape check removed, so a malformed cursor reaches the query | 4 |
| controller stops rejecting a malformed cursor | 1 |
| controller drops the page metadata and returns the bare array | 1 |

**The one that shows the extraction is real rather than nominal:** mutating `splitPage` in `shared/pagination` turns three suites red — `mobile-prescription.service`, `shared/pagination` **and `developer-data.service`**. The developer data plane's own suite failing on a mutation to the extracted helper is what shows the two surfaces share code rather than agreeing by coincidence. Its 35 tests pass unchanged otherwise, which is the other half: the extraction is behaviour-preserving there.

The index has no unit test and should not have one — the before/after `EXPLAIN` above is the evidence, and a test asserting that a `@@index` line exists in the schema would restate the schema rather than check the plan.

### DIFF
--- a/apps/backend/src/controllers/app/prescription.controller.ts
+++ b/apps/backend/src/controllers/app/prescription.controller.ts
@@ -2,7 +2,7 @@ import { Request, Response } from "express";
 import { AuthUserMobileService } from "src/services/authUserMobile.service";
 import { MobilePrescriptionService } from "src/services/mobile-prescription.service";
 import { resolveVerifiedUserId } from "src/utils/request";
-import { parseUuidCursor } from "src/services/shared/pagination";
+import { parseKeysetCursor } from "src/services/shared/pagination";
 import logger from "src/utils/logger";
 
 /**
@@ -50,7 +50,7 @@ export const MobilePrescriptionController = {
        * CR/LF in it forges a second log line, and the 400 already tells the
        * only party who can act on it.
        */
-      const cursor = parseUuidCursor(req.query.cursor);
+      const cursor = parseKeysetCursor(req.query.cursor);
       if (cursor === null) {
         return res.status(400).json({
           message:
--- a/apps/backend/src/controllers/app/prescription.controller.ts
+++ b/apps/backend/src/controllers/app/prescription.controller.ts
@@ -2,6 +2,7 @@ import { Request, Response } from "express";
 import { AuthUserMobileService } from "src/services/authUserMobile.service";
 import { MobilePrescriptionService } from "src/services/mobile-prescription.service";
 import { resolveVerifiedUserId } from "src/utils/request";
+import { parseUuidCursor } from "src/services/shared/pagination";
 import logger from "src/utils/logger";
 
 /**
@@ -39,9 +40,41 @@ export const MobilePrescriptionController = {
         return;
       }
 
-      const prescriptions =
-        await MobilePrescriptionService.listPrescriptionsForParent(parentId);
-      return res.status(200).json({ prescriptions });
+      /*
+       * Rejected up front rather than passed through. A malformed cursor is a
+       * caller mistake and answering 400 keeps every failure from the query
+       * itself honestly a 500; inferring "bad cursor" from a thrown error
+       * would report a database outage as the caller's fault.
+       *
+       * The offending value is not logged. It is caller-controlled, a raw
+       * CR/LF in it forges a second log line, and the 400 already tells the
+       * only party who can act on it.
+       */
+      const cursor = parseUuidCursor(req.query.cursor);
+      if (cursor === null) {
+        return res.status(400).json({
+          message:
+            "Unknown or malformed cursor. Use nextCursor from the previous response.",
+        });
+      }
+
+      const page = await MobilePrescriptionService.listPrescriptionsForParent(
+        parentId,
+        { limit: req.query.limit, cursor },
+      );
+
+      /*
+       * `prescriptions` keeps its name and its shape. The three fields beside
+       * it are what stops this being a silently truncated list: a client that
+       * ignores them sees a short page, and one that reads them can tell the
+       * difference between the end of the data and the end of the page.
+       */
+      return res.status(200).json({
+        prescriptions: page.prescriptions,
+        nextCursor: page.nextCursor,
+        hasMore: page.hasMore,
+        limit: page.limit,
+      });
     } catch (err) {
       logger.error(
         `Error listing prescriptions: ${
--- a/apps/backend/src/controllers/web/developer-data.controller.ts
+++ b/apps/backend/src/controllers/web/developer-data.controller.ts
@@ -22,6 +22,7 @@ import {
   clampPageSize,
   DeveloperDataService,
 } from "../../services/developer-data.service";
+import { parseUuidCursor } from "../../services/shared/pagination";
 import { DeveloperUsageService } from "../../services/developer-usage.service";
 
 type ErrorCode =
@@ -49,21 +50,6 @@ const parseDate = (raw: unknown): Date | undefined | null => {
   const parsed = new Date(raw);
   // null is the "present but unparseable" signal, distinct from absent.
   return Number.isNaN(parsed.getTime()) ? null : parsed;
-};
-
-/*
- * Cursors are appointment ids, which are uuids. Checking the shape up front is
- * what lets every failure from the query itself be reported honestly as a 500:
- * the alternative, inferring "bad cursor" from a thrown error, turns a database
- * outage into a 400 telling the caller their cursor is malformed.
- */
-const UUID = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
-
-const parseCursor = (raw: unknown): string | undefined | null => {
-  if (typeof raw !== "string" || !raw) {
-    return undefined;
-  }
-  return UUID.test(raw) ? raw : null;
 };
 
 const parseStatus = (raw: unknown): AppointmentStatus | undefined | null => {
@@ -148,7 +134,7 @@ export const DeveloperDataController = {
      */
     const rawCursor =
       typeof req.query.cursor === "string" ? req.query.cursor : "";
-    const cursor = parseCursor(req.query.cursor);
+    const cursor = parseUuidCursor(req.query.cursor);
     if (cursor === null) {
       // Logged newline-stripped: the value is caller-controlled and a raw CR/LF
       // in it would forge a second log line.
--- a/apps/backend/src/services/developer-data.service.ts
+++ b/apps/backend/src/services/developer-data.service.ts
@@ -10,6 +10,10 @@
  */
 import type { AppointmentStatus, Prisma } from "@prisma/client";
 import { prisma } from "src/config/prisma";
+import {
+  clampPageSize as clampToBounds,
+  splitPage,
+} from "src/services/shared/pagination";
 
 export const MAX_PAGE_SIZE = 100;
 export const DEFAULT_PAGE_SIZE = 50;
@@ -28,18 +32,15 @@ export const normaliseOrganisationReference = (reference: string): string =>
     : reference;
 
 /*
- * Clamp rather than reject. A caller asking for 1000 rows gets 100 and a
- * `pagination.limit` in the response saying so, which is friendlier to an agent
- * than a 400 it has to learn to avoid, and still bounds the query.
+ * This surface's bounds, applied by the shared clamp. The rule itself lives in
+ * `shared/pagination` because the owner prescription list needs the identical
+ * decision with different numbers (#2709).
  */
-export const clampPageSize = (raw: unknown): number => {
-  const parsed =
-    typeof raw === "string" ? Number.parseInt(raw, 10) : Number.NaN;
-  if (!Number.isFinite(parsed) || parsed < 1) {
-    return DEFAULT_PAGE_SIZE;
-  }
-  return Math.min(parsed, MAX_PAGE_SIZE);
-};
+export const clampPageSize = (raw: unknown): number =>
+  clampToBounds(raw, {
+    defaultSize: DEFAULT_PAGE_SIZE,
+    maxSize: MAX_PAGE_SIZE,
+  });
 
 export interface OrganizationSummary {
   id: string;
@@ -169,11 +170,9 @@ export const DeveloperDataService = {
       ...(query.cursor ? { cursor: { id: query.cursor }, skip: 1 } : {}),
     });
 
-    const hasMore = rows.length > query.limit;
-    const items = hasMore ? rows.slice(0, query.limit) : rows;
-    const last = items.at(-1);
+    const { items, nextCursor } = splitPage(rows, query.limit);
 
-    return { items, nextCursor: hasMore && last ? last.id : null };
+    return { items, nextCursor };
   },
 
   /*
--- a/apps/backend/src/services/mobile-prescription.service.ts
+++ b/apps/backend/src/services/mobile-prescription.service.ts
@@ -1,6 +1,7 @@
 import { ClinicalArtifactStatus } from "@prisma/client";
 import { prisma } from "src/config/prisma";
 import { hasCompanionFeature } from "src/middlewares/companion-access";
+import { clampPageSize, splitPage } from "src/services/shared/pagination";
 
 /**
  * Owner-facing prescription reads for the pet owner app.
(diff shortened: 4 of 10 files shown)


ankit-yc:
Reviewed at `04fb18f24`, `git rev-parse HEAD` confirmed in the same shell either side of every run, tree clean.

| Command | Result |
|---|---|
| `npx jest` (full backend) | **exit 0** — 469 suites / **11280 tests** |

Both numbers match the ones in the PR body exactly.

## One finding, reproduced against a real PostgreSQL: `skip: 1` silently drops a row when the cursor row leaves the filtered set

Prisma's `cursor` is **inclusive**, so `skip: 1` is what makes it exclusive — and `OFFSET 1` applies to the *filtered* result, not to the cursor row. When the cursor row is no longer in that result, the `OFFSET` eats a legitimate row instead.

That is reachable from the product, not hypothetical. `clinical-artifact.service.ts:1991` and `:2079` set `status: "VOID"` when a prescription is voided or cancelled, which moves the artifact out of `OWNER_VISIBLE_ARTIFACT_STATUSES`.

Scratch PostgreSQL 16, the full schema out of `prisma migrate diff --from-empty`, seven prescriptions on one encounter, page size 3:

```
CONTROL - nothing changes between pages
  page: ["rx-g","rx-f","rx-e"]  hasMore=true  nextCursor=rx-e
  page: ["rx-d","rx-c","rx-b"]  hasMore=true  nextCursor=rx-b
  page: ["rx-a"]                hasMore=false
  NEVER SHOWN: []

A vet VOIDs the prescription that happens to be the cursor row
  page: ["rx-g","rx-f","rx-e"]  hasMore=true  nextCursor=rx-e
  ...rx-e voided: it leaves COMPLETED/SIGNED, so it is no longer in the filtered set
  page: ["rx-c","rx-b","rx-a"]  hasMore=false
  visible in the filter at the end : ["rx-g","rx-f","rx-d","rx-c","rx-b","rx-a"]
  the caller was actually shown    : ["rx-g","rx-f","rx-e","rx-c","rx-b","rx-a"]
  NEVER SHOWN                      : ["rx-d"]
```

`rx-d` is never shown, and `hasMore: false` tells the client it reached the end. **That is silent truncation wearing a completed page's clothes** — the defect class this PR exists to close, arriving through the cursor rather than through a bare `take`. No test here can catch it: the service tests mock `prisma.prescription.findMany`, so they can assert `skip: 1` was *sent* but not what PostgreSQL does with it.

Suggested fix: make the cursor exclusive in the `where` instead of with `skip`, encoding `(createdAt, id)` into the cursor and comparing

```ts
OR: [
  { createdAt: { lt: after.createdAt } },
  { createdAt: after.createdAt, id: { lt: after.id } },
]
```

That is exclusive by construction, so a cursor row that has left the set is harmless, and it drops the dependency on Prisma's cursor subquery entirely.

**Scope:** the same `cursor` + `skip: 1` pattern is pre-existing in `developer-data.service.ts` and this PR did not introduce it there — but this PR is the one putting it on an owner-facing clinical read, and the endpoint still has no consumers, so the shape is free to change now and will not be later. That is the same window argument the PR body makes for the response shape.

## Two things I checked that are fine, because both look alarming and are not

**The `id` tiebreak is real in the emitted SQL, not just in the comment.** Captured with query logging against the same database:

```sql
... AND ((Prescription.createdAt = (SELECT createdAt FROM Prescription WHERE id = $4)
          AND Prescription.id <= (SELECT id FROM Prescription WHERE id = $5))
      OR  (Prescription.createdAt <  (SELECT createdAt FROM Prescription WHERE id = $6)))
ORDER BY Prescription.createdAt DESC, Prescription.id DESC LIMIT $7 OFFSET $8
```

I could **not** make the no-tiebreak version lose or repeat a row in this scratch database, including with a row updated between pages to move it in the heap. So the tiebreak's value here rests on the row comparison above being the one Prisma emits, not on a failure I reproduced — "I could not break it" is not "it cannot break" when the SQL leaves the order unspecified.

**The included `items` relation query carries an `OFFSET`, and it is `0`.** The cursor page emits `... FROM PrescriptionItem WHERE prescriptionId IN (...) ORDER BY sortOrder ASC OFFSET $5` with `$5 = 0`, so the parent's `skip: 1` does not leak into the relation load. Medications are intact on page 2 — worth stating because the generated SQL reads as if the first item of every prescription would be dropped.

## On the migration comment

Claude L3 Yosemite Crew's read is correct and I checked it against the script rather than carrying it forward. In `api-deploy.sh` on this tree: `prisma:deploy` at line 196, `pm2 restart` at line 220, and the only `pm2` touches in the whole file are `restart`, `save` and `describe` — no stop, no drain, no maintenance step between them. So the last clause of the migration comment, "that window is the migration step rather than the running service", describes a window in which the service is up and taking writes. `SET lock_timeout = '5s';` ahead of the `CREATE INDEX` is the cheap mitigation, and it composes with #2698: a blocked build then fails the deploy before cutover and fires the hazard notice, instead of queueing traffic behind itself.

ankit-yc (approved):
Reviewed at `523f7a980`, clean tree, `git rev-parse HEAD` confirmed in the same shell either side of every command. **Approving.** Nothing blocking; one follow-up filed as #2722 and two notes.

| Command | Result |
|---|---|
| `npx jest` (full backend) | **exit 0** — 469 suites / **11295 tests** |

Both match the numbers in the PR body exactly.

### The finding from the previous round is closed, and I checked the thing that would have made it worse

`Prescription.id` is `@id @default(uuid())`, so `UUID.test(id)` in `parseKeysetCursor` admits every real row id. Worth stating because if that column were a `cuid()` the endpoint would emit a `nextCursor` its own controller answers **400** to, and every test here would still pass — the fixtures use `rx-1`-style ids, so the round-trip test cannot see it either. It is fine; it is fine for a reason outside the diff.

### Mutations

Through an assert-uniqueness harness (`assert count == 1` before substituting), each reverted before the next, tree clean at the end:

| Mutation | Result |
|---|---|
| `createdAt: { lt }` → `lte` | 1 failed |
| `id: { lt }` → `gt` | 1 failed |
| drop the `createdAt =` / `id <` tiebreak branch | 1 failed |
| `toCursor: (row: T) => string` → `(row: any) => string` | jest **green**, `tsc` **exit 0**, `eslint` **exit 1** (3 errors) |

**All three correctness mutations redden the same single check** — the `args.where.OR` shape assertion. That is not a gap, it is the honest ceiling of a mocked `findMany`: the suite can say the shape changed, not which way it is wrong. The comment on that test already says so, and the real-PostgreSQL run in the PR body is what covers the other half.

**The fourth is the interesting one.** The requirement that a row carry `createdAt` is enforced by the type system rather than by a test, which is worth saying out loud — but deleting that enforcement is not a green mutation, because `@typescript-eslint/no-unsafe-return` and `no-unsafe-member-access` catch the widening. Verified with a probe:

```
splitPage(withCreatedAt,    3, encodeKeysetCursor);  // ok
splitPage(withoutCreatedAt, 3, encodeKeysetCursor);
  -> error TS2345: Argument of type '{ id: string; }[]' is not assignable to parameter of type 'KeysetCursor[]'
```

Exactly one error, on exactly the second call. Scope: that is the helper's contract, not the call site — the call site uses `include`, so every scalar is present and there is no `select` to omit from today.

### #2722, filed rather than left in the thread

`developerDataService.listAppointments` still uses `cursor` + `skip: 1`, and the PR body is right that it is out of scope. It is also **more** reachable there than it was here: the `where` filters on `status` and an `appointmentDate` range, both caller-supplied, and both changed by ordinary front-desk actions — a cancellation or a reschedule takes the cursor row out of the filtered set. Filed with the mechanism, the four-consumer constraint that makes the cursor format *not* free to change there, and the explicit note that nobody has reproduced it against `listAppointments` itself yet.

### Two notes, neither blocking

**The keyset predicate is not index-served.** `Prescription` carries only `@@index([artifactId])`, so `WHERE createdAt < … OR (createdAt = … AND id < …)` with `ORDER BY createdAt DESC, id DESC` has no index to walk. This is **not a regression** — the previous `cursor: { id }` form had none either — and the set is bounded by one owner's own encounters, so it is small. Stating it only so the `16x` in the PR body is read as what it is: a measurement of the `ClinicalArtifact(encounterId, status)` index serving the join, not of the pagination.

**The encoder is wider than the parser.** `encodeKeysetCursor` emits `getTime()` unguarded; `parseKeysetCursor` admits `^\d{1,15}$`, which has no sign, so a pre-1970 `createdAt` produces a cursor the endpoint's own parser rejects with a 400. Unreachable — `createdAt` is `@default(now())` — and the round-trip test uses one fixed date so it cannot see it. A one-line note on a boundary where the two are meant to be inverses.

### Migration

Executable lines are exactly two:

```sql
SET LOCAL lock_timeout = '5s';
CREATE INDEX IF NOT EXISTS "ClinicalArtifact_encounterId_status_idx"
  ON "ClinicalArtifact" ("encounterId", "status");
```

The comment now states the lock cost against live traffic correctly — including that reads are unaffected because `ACCESS SHARE` does not conflict with `SHARE` — and carries the P3009 recovery with the `migrate resolve --rolled-back` command. That last part is what turns a failing migration from a trap into a runbook, and it is the half that would have been missed.

Issue #2719, opened 2026-09-05
Adjacent JSDoc blocks orphan the earlier one from hover - but adjacency is not the defect; needs a predicate before any cleanup


ankit-yc:
## What

Where a declaration carries **two adjacent `/** */` blocks**, TypeScript binds only the last one. The earlier block stays in the file, still reads like documentation, and is dropped from that symbol's documentation entirely.

Found while fixing one instance on #2708's branch (`be41dfb91`), where a note added on top of an existing block evicted the original rationale from that symbol's hover.

> **This issue was first filed claiming all 51 sites are the same defect and should be merged. That framing was wrong.** Claude L3's read of the site list, and the hover measurements below, say most of them are not defects and that merging them would introduce the opposite error. The original ask is struck through at the bottom. Do not act on the first version.

## The mechanism — unchanged, and still the reason this matters

`checker.getSymbolAtLocation(decl.name).getDocumentationComment(checker)` — the same call an editor hover and `tsserver` quick-info use — returns only the nearest preceding block:

```
--- adjacent (two /** */ blocks) ---
"SECOND block: NOT async."
--- merged (one block) ---
"FIRST block: create, not upsert.\n\nSECOND block: NOT async."
```

Not "editors probably pick one". The earlier block is absent from the symbol's documentation.

## The scan, and two independent counts that do not agree

TypeScript AST (`ts.getLeadingCommentRanges` per declaration, counting leading ranges that open with `/**`), deduplicated per comment position because a `VariableStatement` and its `VariableDeclaration` both report the same leading comments.

| | Sites | Orphaned lines | Files scanned |
|---|---|---|---|
| First pass | 51 | 501 | 5587 |
| Independent re-run | 49 | 483 | 5405 |

The gap is unexplained — probably the glob or the commit. **It has not been chased, because the count is not the thing that decides anything.** Neither pass covers `.js`/`.mjs`/`.cjs`, and both look at declarations only.

## Correction: adjacency is not the defect

Reading the sites rather than counting them, they fall into three classes:

[...]

PR #2725, opened 2026-09-05
fix(ci): trap TERM and HUP so a cancelled deploy still announces the schema hazard

### LINKED ISSUE #2721: A deploy killed after the migration runs the exit handler and says nothing: SIGHUP and SIGTERM hand the EXIT trap a status of 0
## What

`deploy_on_exit` runs when the deploy is killed, and says nothing. `$?` inside a bash `EXIT` trap is the status of the last *completed* command, not the signal — so `deploy_stop_notice` is handed `0` and classifies a killed deploy as a clean exit.

The state it stays silent about is the one it exists to announce: migrations applied, no cutover, so the old process is serving traffic against the new schema and the printed code rollback does not undo it.

Found by Claude L3 Yosemite Crew while reviewing #2718, which made the handler runnable in isolation. Pre-existing from #2698 and untouched by #2718.

## SIGHUP is the live path, not an edge case

`.github/workflows/deploy-api.yml:224-226` does not run this script on the runner. It runs it on the deploy box over SSH:

```
ssh -i ~/.ssh/deploy_key -o StrictHostKeyChecking=yes \
  "ubuntu@$HOST" \
  "chmod +x /tmp/yc-deploy/api-deploy.sh && ... /tmp/yc-deploy/api-deploy.sh '$REPO' '$PM2_TARGET' '$TARGET_REF'"
```

A cancelled workflow signals the **`ssh` client on the runner**. The script's parent is `sshd` on a different machine, so the runner's SIGINT never reaches it; what the remote process gets when the connection dies is SIGHUP from the session teardown. There is no `-t`, so no controlling terminal is involved.

So a cancelled deploy — the likeliest way this script is ever interrupted mid-migration — lands on a silent path.

## Measured

`$?` read *inside* the trap rather than inferred from whether the notice fired, against the real `deploy_on_exit` sourced from `lib/migrate.sh`, `MIGRATIONS_APPLIED=1 CUTOVER_DONE=0`:

| signal | shell exit | `$?` in trap | notice |
|---|---|---|---|
| SIGINT | 130 | **130** | FIRED |
| SIGTERM | 143 | **0** | SILENT |
| SIGHUP | 129 | **0** | SILENT |

SIGINT is already covered: bash's SIGINT path sets the status to 130 before running the EXIT trap. TERM and HUP hand the trap the status of the last completed command.

**Scope.** The table is measured. That sshd sends SIGHUP to the remote process group on teardown is inferred from the invocation above and from how sshd works — nobody has stood up an sshd and watched the signal arrive. Settling it is one `ssh localhost '<script>'` plus a `kill` of the client.

## Harness notes, because three of our first attempts were artifacts

- A script started with `&` from a non-interactive shell has SIGINT set to **SIG_IGN** by bash. No INT sent to it will ever do anything.
- Signalling the shell alone with SIGINT while a foreground child runs is also a no-op: bash waits for the child, sees it did not die of SIGINT, and carries on.
- Only a **foreground** launch, in its **own process group**, signalled as a group, reproduces a real cancel. Signalling a group the test harness shares kills the harness.

## Suggested fix

```sh
trap 'exit 143' TERM
trap 'exit 129' HUP
```

alongside the existing `trap deploy_on_exit EXIT`. A signal trap that exits with a non-zero status gives the EXIT trap a status to see. Verified for TERM: `SIGTERM -> exit 143, notice FIRED`. SIGINT needs nothing.

Belongs with #2712 and #2714 rather than in #2718: it is the same ~200-line script, and opening a second PR into it while a restructure is in flight is a rebase queue rather than parallelism.

---

## Update: the fix is confirmed on all three signals, and it is necessary but not sufficient

Two independent reproductions of the trap fix against the real `deploy_on_exit`
sourced from `lib/migrate.sh` at `e07a7af2d`, `MIGRATIONS_APPLIED=1`,
`CUTOVER_DONE=0`, bash 3.2.57, launched in its own process group with the INT
disposition reset to `DEFAULT`:

| target | signal | rc | elapsed | notice, no traps | notice, with traps |
|---|---|---|---|---|---|
| process group | TERM | 143 | 0s | SILENT | **FIRED** |
| process group | HUP | 129 | 0s | SILENT | **FIRED** |
| process group | INT | 130 | 0s | FIRED | FIRED |
| shell only | TERM | 143 | 5s | SILENT | **FIRED** |
| shell only | HUP | 129 | 5s | SILENT | **FIRED** |
| shell only | INT | 0 | 5s | SILENT | SILENT |

Exit codes are preserved on every row, so nothing downstream that reads the
deploy's status changes.

### The deferred-signal effect is real, and the cost and the benefit are the same fact

A **trapped** signal is deferred until the current foreground command finishes;
an **untrapped** one kills the shell where it stands. That is the `elapsed=5s`
column: with the traps, a shell-only `kill` during a long command lets the
command finish, then exits and says so.

On the process group the delay is **0s** — but for the reason that also removes
the benefit. The foreground child is in the group too, so it dies with the
shell: the trap runs at once *because the migration was cut off*, not because it
finished. "Finish the migration, then announce" is only reachable via a
shell-only kill of the script. It is still the better behaviour and worth
stating as a deliberate choice, but it does not protect the path that produced
this issue.

### SIGKILL stays out of reach

No trap catches it, so the honest ceiling here is "every signal that can be
trapped". Not "the hole closes".

### The notice fires and nobody receives it

The one that changes what this issue needs. `deploy_schema_hazard_notice`
writes to **stderr**, and the script's stderr is the ssh pipe — the thing whose
teardown sends the SIGHUP in the first place. Same harness, same signal, one
variable, the reader modelling sshd going away with the connection:

| stderr reader | rc | bytes delivered | notice |
|---|---|---|---|
| alive (control) | 129 | 600 | **present** |
| gone | **141** | **0** | **lost** |

`141` is `128+13`: the handler ran, wrote into a pipe with no reader, and took
`SIGPIPE` part-way through the notice. Nothing in `api-deploy.sh` or
`lib/*.sh` writes the notice anywhere durable — `grep` for `logger`, `tee`,
`LOG_FILE` and `/var/log` across the deploy directory returns nothing — so on a
runner cancel the operator sees neither the notice nor the Actions log that
would have carried it.

**Scope on this one.** The reader is killed before the signal, which is one
plausible ordering of a real teardown and not one anybody has watched on the
box. The weaker claim survives either ordering: there is no ssh client left to
receive it and no log stream still being read.

**So the fix this issue asks for should be two things, not one:** the traps, so
the handler runs at all; and a durable destination for the notice on the box
(a file under the deploy directory, or `logger`), so that it survives the
channel it is currently written to. The traps alone convert "silent because the
handler never ran" into "silent because the pipe closed".

### The durable destination already has an idiom here, and the write order is load-bearing

`api-deploy.sh:79-83` sets `STAMP` and writes the preflight artifacts keyed on
it — including the rollback sha, which is the value the hazard notice exists to
hand the operator. So the second half is `/tmp/api-schema-hazard-$STAMP.txt`
next to the file that already carries the sha, not a new logging mechanism.

Two things the implementer needs:

**1. `STAMP` becomes a fifth required caller variable.** `deploy_on_exit` lives
in `lib/migrate.sh` and reads caller state under `set -u`; `STAMP` is currently
referenced only in `api-deploy.sh`. It is set at `:79` and the trap is armed at
`:182`, so it is always set on the real path. The six `deploy_on_exit` call
sites in `tests/migrate.test.sh` will fail loudly until they set it, which is
the right direction and is work to budget for.

**2. The file write must come before the stderr write.** The handler takes
`SIGPIPE` part-way through the notice when the reader is gone, so anything
after the stderr write does not happen. Measured, same harness, HUP to the
group, the only variable being the order of the two writes:

| stderr reader | order | rc | hazard file |
|---|---|---|---|
| gone | file, then stderr | 141 | **590 bytes** |
| gone | stderr, then file | 141 | **0 bytes** |
| alive (control) | file, then stderr | 129 | 590 bytes |
| alive (control) | stderr, then file | 129 | 590 bytes |

`rc` is `141` in both of the failing-reader rows, so the exit status does not
distinguish "the notice was preserved" from "the notice was lost". The order is
the whole of it, and it is the order nobody would think to write a test for.

ankit-yc:
Closes #2721.

## The defect

An untrapped fatal signal ends the shell without handing the EXIT trap a
non-zero status, so `deploy_on_exit` ran and `deploy_stop_notice` returned
early: the notice was silent on exactly the two signals a cancelled deploy
arrives as. A cancel kills the local `ssh` client, the connection closes, and
sshd sends SIGHUP to the remote process group. `deploy-api.yml:224-226` has no
`-t` and no wrapper, so nothing else is between the runner and the script.

Measured against the real handler sourced from `lib/migrate.sh`, own process
group, INT disposition reset to `DEFAULT`, bash 3.2.57:

| target | signal | rc | before | after |
|---|---|---|---|---|
| process group | TERM | 143 | silent | **fires** |
| process group | HUP | 129 | silent | **fires** |
| process group | INT | 130 | fires | fires |
| shell only | TERM | 143 | silent | **fires** |
| shell only | HUP | 129 | silent | **fires** |

Exit codes are preserved on every row, so nothing downstream that reads the
deploy's status changes. SIGINT already worked — bash sets the status to 130
itself. **SIGKILL stays out of reach**, so the ceiling here is every signal that
can be trapped, not "the notice can no longer be lost".

### The deferred-signal effect, named rather than discovered on a box

A trapped signal is deferred until the running foreground command finishes,
where an untrapped one ends the shell where it stands. On a cancel that costs
nothing — the child is in the same process group and dies too, and every
process-group row above returned in 0s. On a bare `kill` of this shell it means
an in-flight `prisma migrate deploy` runs to completion before the deploy stops,
which is the safer of the two and is a deliberate choice rather than a
side effect.

## The traps alone are not enough

`deploy_schema_hazard_notice` wrote to **stderr**, and stderr on the box is the
ssh pipe — the same connection whose teardown sends the SIGHUP. By the time the
handler runs the reader may already be gone, the write takes SIGPIPE, and the
handler ends part-way through the notice. The trap fires and the operator gets
silence through a different door.

So the notice is now written to `/tmp/api-schema-hazard-$STAMP.txt` as well,
beside the rollback sha preflight already writes at `api-deploy.sh:80` — the
same `$STAMP` artifact convention, not a new logging mechanism. **The order
between the two destinations is load-bearing:**

| stderr reader | order | rc | bytes on disk |
|---|---|---|---|
| gone | file, then stderr | 141 | **590** |
| gone | stderr, then file | 141 | **0** |
| alive (control) | either order | 129 | 590 |

`141` is `128+13`. The exit status is identical in both failing cases, so
nothing downstream can tell "notice preserved" from "notice lost".

## The one design call worth a reviewer's attention

`HAZARD_LOG` is **defaulted** rather than demanded, which is the opposite of the
rule the rest of `lib/migrate.sh` follows for caller state. A `set -u` abort on
an unset `HAZARD_LOG` would destroy the notice in order to enforce a rule about
keeping it — the failure mode the whole change exists to remove. So the loud
check lives on `api-deploy.sh` instead, where the mistake would actually be
made: the suite pins that the assignment exists and sits above the arming.

I said in the issue thread that it would become a fifth required caller
variable. That was wrong for this reason, and the reversal is why the two
source-level checks are here.

## Evidence at `db63ae109`

Six mutations through the assert-uniqueness harness, each reverted before the
next, the harness refused none, baseline and restore both 41/41:

| Mutation | Result |
|---|---|
| remove the TERM trap from `api-deploy.sh` | 40 passed, **1 failed** |
| remove the HUP trap from `api-deploy.sh` | 40 passed, **1 failed** |
| remove the `HAZARD_LOG` assignment | 40 passed, **1 failed** |
| move `HAZARD_LOG` below the trap arming | 40 passed, **1 failed** |
| drop the durable write from the notice | 39 passed, **2 failed** |
| write stderr before the file instead of after | 40 passed, **1 failed** |

The last one reddens only `the notice survives an stderr whose reader has gone`,
which is the check the ordering exists for. And removing the two signal traps
from the **test's own** inline script reddens both signal checks, so neither is
vacuous.

Gates, `git rev-parse HEAD` confirmed either side, tree clean:

| Command | Result |
|---|---|
| `tests/migrate.test.sh` | exit 0 — **41 passed, 0 failed** |
| `tests/git-sync.test.sh` | exit 0 — 16/16 |
| `shellcheck --severity=warning -x` | **exit 0** |
| `shellcheck -x` (plain) | exit 1 — 8 info notes, **findings profile identical to `origin/dev`** |

The plain `shellcheck` run is exit 1 on `origin/dev` too; I compared the finding
profiles rather than reporting "clean", and introduced nothing at any severity.
`shellcheck` is not gated in CI.

### DIFF
--- a/scripts/deploy/api-deploy.sh
+++ b/scripts/deploy/api-deploy.sh
@@ -81,6 +81,12 @@ echo "$ROLLBACK_SHA" > "/tmp/api-rollback-$STAMP.txt"
 tar czf "/tmp/api-dist-before-$STAMP.tgz" apps/backend/dist packages/*/dist 2>/dev/null || true
 cp apps/backend/.env "/tmp/api-env-before-$STAMP" 2>/dev/null || true
 echo "backups: /tmp/api-dist-before-$STAMP.tgz  /tmp/api-env-before-$STAMP"
+# Where the schema-hazard notice is written in addition to stderr. Built here,
+# with the other preflight artifacts, because stderr is the one destination that
+# is guaranteed to be gone in the case the notice matters most: it is the ssh
+# pipe, and the teardown of that connection is what kills this script. The
+# rollback sha the notice hands the operator is already written next door.
+HAZARD_LOG="/tmp/api-schema-hazard-$STAMP.txt"
 
 say "checkout $GIT_REF"
 # Fetch and checkout live in lib/git-sync.sh so their two failure modes - a stale
@@ -179,6 +185,26 @@ fi
 # Prisma fails having applied nothing, and over-reporting a schema hazard is the
 # safe direction. A deploy that carries no migrations cannot move the schema at
 # all, so it stays silent.
+#
+# TERM and HUP are trapped alongside it, and the reason is not symmetry. An
+# untrapped fatal signal ends the shell without giving the EXIT trap a non-zero
+# status to see, so the notice is silent on exactly the two signals a cancelled
+# deploy arrives as: the runner kills the local ssh client, the connection
+# closes, and sshd sends SIGHUP to this process group. Measured against the real
+# handler - TERM and HUP silent before, both firing after, and the exit status
+# preserved as 143 and 129 so nothing downstream reads a different result.
+# SIGINT already works: bash sets the status to 130 itself.
+#
+# SIGKILL stays out of reach. No trap catches it, so the ceiling here is every
+# signal that can be trapped, not "the notice can no longer be lost".
+#
+# A trapped signal is also deferred until the running foreground command
+# finishes, where an untrapped one ends the shell immediately. On a cancel that
+# costs nothing - the child is in the same process group and dies too - and on a
+# bare `kill` of this shell it means an in-flight `prisma migrate deploy` runs to
+# completion before the deploy stops, which is the safer of the two.
+trap 'exit 143' TERM
+trap 'exit 129' HUP
 trap deploy_on_exit EXIT
 if [ -n "$INCOMING_MIGRATIONS" ]; then
   MIGRATIONS_APPLIED=1
--- a/scripts/deploy/api-deploy.sh
+++ b/scripts/deploy/api-deploy.sh
@@ -186,26 +186,10 @@ fi
 # safe direction. A deploy that carries no migrations cannot move the schema at
 # all, so it stays silent.
 #
-# TERM and HUP are trapped alongside it, and the reason is not symmetry. An
-# untrapped fatal signal ends the shell without giving the EXIT trap a non-zero
-# status to see, so the notice is silent on exactly the two signals a cancelled
-# deploy arrives as: the runner kills the local ssh client, the connection
-# closes, and sshd sends SIGHUP to this process group. Measured against the real
-# handler - TERM and HUP silent before, both firing after, and the exit status
-# preserved as 143 and 129 so nothing downstream reads a different result.
-# SIGINT already works: bash sets the status to 130 itself.
-#
-# SIGKILL stays out of reach. No trap catches it, so the ceiling here is every
-# signal that can be trapped, not "the notice can no longer be lost".
-#
-# A trapped signal is also deferred until the running foreground command
-# finishes, where an untrapped one ends the shell immediately. On a cancel that
-# costs nothing - the child is in the same process group and dies too - and on a
-# bare `kill` of this shell it means an in-flight `prisma migrate deploy` runs to
-# completion before the deploy stops, which is the safer of the two.
-trap 'exit 143' TERM
-trap 'exit 129' HUP
-trap deploy_on_exit EXIT
+# The arming itself is deploy_arm_exit_traps in lib/migrate.sh - EXIT plus TERM
+# and HUP, with the reasoning next to it - so the suite can arm the real one and
+# signal it rather than grep this file for three trap lines.
+deploy_arm_exit_traps
 if [ -n "$INCOMING_MIGRATIONS" ]; then
   MIGRATIONS_APPLIED=1
 fi
--- a/scripts/deploy/lib/migrate.sh
+++ b/scripts/deploy/lib/migrate.sh
@@ -71,7 +71,8 @@ deploy_incoming_migrations() {
 # one.
 #
 # SMOKE_PID is the exception and IS defaulted: no smoke process is a normal
-# state for most of the script's life, not a caller mistake.
+# state for most of the script's life, not a caller mistake. HAZARD_LOG is
+# defaulted for a different reason - see deploy_schema_hazard_notice.
 deploy_on_exit() {
   local status=$?
 
@@ -119,12 +120,12 @@ deploy_stop_notice() {
   deploy_schema_hazard_notice "$rollback_sha" "$@"
 }
 
-# deploy_schema_hazard_notice <rollback-sha> <migration>...
+# deploy_schema_hazard_text <rollback-sha> <migration>...
 #
-# What to say when the deploy stops after the schema has moved. Written to
-# stderr beside the code rollback, because the code rollback on its own reads as
-# if the box has been restored, and it has not.
-deploy_schema_hazard_notice() {
+# The wording, on stdout, so the two destinations below write the same bytes.
+# Separated from the destinations because they have different failure modes:
+# the wording cannot fail, and writing it can.
+deploy_schema_hazard_text() {
   local rollback_sha="${1:?rollback sha required}"
   shift
 
@@ -143,5 +144,47 @@ deploy_schema_hazard_notice() {
     echo "forward and re-deploy, or write and apply a migration that reverses them."
     echo "Prisma has no down-migration; there is nothing to run that undoes this by"
     echo "itself."
-  } >&2
+  }
+}
+
+# deploy_schema_hazard_notice <rollback-sha> <migration>...
+#
+# What to say when the deploy stops after the schema has moved, and where.
+#
+# Two destinations, and the ORDER between them is load-bearing rather than
+# stylistic. stderr on the box is the ssh pipe, and the teardown of that
+# connection is what sends the SIGHUP this notice most needs to survive - so by
+# the time the handler runs, the reader may already be gone. Writing into a pipe
+# with no reader takes SIGPIPE and ends the handler where it stands, so anything
+# sequenced after the stderr write does not happen at all.
+#
+# Measured against the real handler, HUP to the process group, the only variable
+# being the order of the two writes:
+#
+#   reader gone,  file then stderr -> exit 141, 590 bytes on disk
+#   reader gone,  stderr then file -> exit 141,   0 bytes on disk
+#   reader alive, either order     -> exit 129, 590 bytes on disk
+#
+# The exit status is 141 in both failing cases, so nothing downstream can tell
+# the two apart. The order is the whole of it.
+#
+# HAZARD_LOG is defaulted rather than required, which is the opposite of the
+# rule the rest of this file follows, and deliberately: a `set -u` abort on an
+# unset HAZARD_LOG would destroy the notice in order to enforce a rule about
+# keeping it. The loud check belongs where the mistake is actually made - the
+# suite pins that api-deploy.sh sets it before arming the trap.
+deploy_schema_hazard_notice() {
+  local rollback_sha="${1:?rollback sha required}"
+  shift
+
+  local text
+  text="$(deploy_schema_hazard_text "$rollback_sha" "$@")"
+
+  # `|| true` because a destination that cannot be written must not cost the
+  # other one. A full disk here would otherwise end the handler under `set -e`.
+  if [ -n "${HAZARD_LOG:-}" ]; then
+    printf '%s\n' "$text" >> "$HAZARD_LOG" 2>/dev/null || true
+  fi
+
+  printf '%s\n' "$text" >&2
 }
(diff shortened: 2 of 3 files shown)


aupyay (dismissed):
Approving at `5e9b557791eb5a205b352664e3ee5a0758ee55bd`. There is a second verified read of this PR in the team channel; the mutations below are my own. Detached worktree, `git rev-parse HEAD` confirmed either side of every run, tree clean at the end.

Both halves of #2721 are correct and each is individually pinned.

## Eight mutations — four red, four green

Assert-uniqueness harness, each reverted before the next, harness refused none, baseline and restore both `39 passed, 0 failed`.

| Mutation | Result |
|---|---|
| `trap 'exit 143' TERM` → `exit 0` | **37 / 2 failed** |
| `trap 'exit 129' HUP` → `exit 0` | **38 / 1 failed** |
| hazard **text** written to stderr, so the `$(…)` capture is empty | **36 / 3 failed** |
| `api-deploy.sh` no longer calls `deploy_arm_exit_traps` | **38 / 1 failed** |
| drop the `\|\| true` on the durable write | GREEN |
| default `HAZARD_LOG` to a fixed name instead of `$$` | GREEN |
| truncate the hazard file instead of appending | GREEN |
| `api-deploy.sh` stops setting the `$STAMP`-keyed `HAZARD_LOG` | GREEN |

The first four are new. Preserving a status of `0` on each signal is the mutation that matters most — it is the original defect re-introduced through the fix rather than around it, and both signals are caught separately.

## The four greens are one hole, and one of them is a regression

Between `db63ae109` and this head, an assertion was removed and nothing replaced it:

```
db63ae109  tests/migrate.test.sh   HAZARD_LINE occurrences: 3   (lines 602-609)
5e9b55779  tests/migrate.test.sh   HAZARD_LINE occurrences: 0
```

The check was `api-deploy.sh names the durable notice destination before arming the trap`. With it gone, deleting that assignment outright is green here and moving it below the arming is green too — two directions onto the same hole.

Nothing breaks, because the `$$` default in `deploy_arm_exit_traps` absorbs it. What is lost is silent: the notice stops being keyed to the same `$STAMP` as `/tmp/api-rollback-$STAMP.txt` and `/tmp/api-dist-before-$STAMP.tgz`, so an operator correlating the three artifacts of one deploy loses the link and no test says so.

Collated, the four greens are a single statement: **the suite proves a durable write happens, and does not prove it goes to the right file, additively, without costing the stderr write.**

**Not blocking the merge, and I would rather it were restored on this branch than filed.** The production fix is fully pinned and this is a test gap, but it is a test gap in a PR whose subject is checks that actually check — and the same re-run-the-table discipline that caught `line_of` matching the comment is what surfaced it. The runtime form is available now that the arming is a real function the suite already calls: preset `HAZARD_LOG`, arm `deploy_arm_exit_traps`, assert it did not override the caller's value. That pins the `${HAZARD_LOG:-…}` semantics behaviourally rather than by line position, which is better than what was lost.

If a commit restoring it lands, `dismiss_stale_reviews_on_push` will drop this approval and I will re-approve on the new head.

## Suites at `5e9b55779`

| Command | Result |
|---|---|
| `tests/migrate.test.sh` | exit 0 — **39 passed, 0 failed** |

## One thing to fold in while the file is open

`deploy_on_exit`'s documentation block now sits above `deploy_arm_exit_traps`:

```
52   # deploy_on_exit                 <- block starts
77   # deploy_arm_exit_traps          <- runs straight into this, no blank line
105  deploy_arm_exit_traps() {
119  deploy_on_exit() {               <- nothing above it
```

So the handler's contract — *callers must set MIGRATIONS_APPLIED, CUTOVER_DONE, ROLLBACK_SHA and INCOMING_MIGRATIONS*, and the `${SMOKE_PID:-}` exception — is attached to the arming function, and the handler it describes carries no comment. That is the defect shape of #2719 in shell, where there is no `getDocumentationComment` to notice it. Cosmetic in effect, one blank line and a re-split to fix.

aupyay (approved):
Re-approving at `3214dc8c9d774f6deee66bd12fc10008bd4b964c`. The regression I raised at `5e9b55779` is closed, and I re-ran my own greens rather than reading the new tests.

| Mutation | at `5e9b55779` | at `3214dc8c9` |
|---|---|---|
| delete `api-deploy.sh`'s `$STAMP` `HAZARD_LOG` assignment | GREEN | **40 / 1 failed** |
| arming overrides the caller (`:-` → `=`) | not present | **39 / 2 failed** |
| drop the `HAZARD_LOG` default from the arming | 1 red | **40 / 1 failed** |
| `trap 'exit 143' TERM` → `exit 0` | 2 red | **39 / 2 failed** |
| `trap 'exit 129' HUP` → `exit 0` | 1 red | **40 / 1 failed** |

Assert-uniqueness harness, each reverted before the next, harness refused none, baseline and restore both `41 passed, 0 failed`; tree clean.

The `arming overrides the caller` row is the one worth naming. It is not a restoration of what was lost — nothing checked it before — and it pins the `${HAZARD_LOG:-…}` semantics behaviourally, so the property survives someone rewriting the line rather than moving it.

## The two remaining greens are correctly documented rather than fixed

`|| true` on the durable write and append-versus-truncate both still mutate green, and the file now says so. That is the right outcome: the suite proves a durable write happens, that it precedes the stderr write, and that the caller's destination survives the arming — and a reader should not credit 41 green with more than that. Writing down what a suite does not cover is worth more than a test that pins an incidental.

## The doc block

Fixed, and the re-attachment matters more than it looked: the block carrying *"callers must set MIGRATIONS_APPLIED, CUTOVER_DONE, ROLLBACK_SHA and INCOMING_MIGRATIONS"* was sitting above the function that reads none of them, while the one that reads all four had nothing. `deploy_on_exit()` now follows its own comment directly.

## Suites at `3214dc8c9`

`tests/migrate.test.sh` exit 0 — **41 passed, 0 failed**. `git rev-parse HEAD` confirmed either side; tree clean at the end.

PR #2726, opened 2026-09-05
fix(backend): parse the co-parent permission body instead of casting it

### LINKED ISSUE #2710: Co-parent permission updates write the raw request body into an authorisation record
`ParentCompanionController.updatePermissions` passes the request body straight through to the service with no allowlist, no schema, and no key filter:

```ts
// apps/backend/src/controllers/app/parent-companion.controller.ts:105-108
const updates = (
  typeof req.body === "object" && req.body ? req.body : {}
) as Partial<ParentCompanionPermissions>;
```

`ParentCompanionService.updatePermissions` then merges it over the target's existing permissions and persists the result:

```ts
// apps/backend/src/services/parent-companion.service.ts
const mergedPermissions: ParentCompanionPermissions = {
  ...targetPermissions,
  ...updates,
};
```

`ParentPatient.permissions` is an authorisation record — `companion-access.ts` reads it on every companion-gated route — so this is an unvalidated write into the input of an authorisation decision. The cast is the only thing that claims the body has the shape of `ParentCompanionPermissions`, and a cast checks nothing at runtime.

### Two concrete consequences

**1. Arbitrary keys persist.** Any JSON object the caller sends is merged in verbatim and stored. `isGranted` only reads the eight known `CompanionFeature` keys, so this is not an escalation today, but the row accumulates caller-controlled content and any future reader that iterates the object inherits the problem.

**2. Non-boolean values are accepted.** `isGranted` requires `=== true`, so a non-boolean denies rather than grants — the safe direction — but the stored record then disagrees with what the UI will render from it.

### The reachable case that motivates this

`PATCH /:patientId/:targetParentId/permissions` is authorised by `ensurePrimaryLinkExists(requestingParentId, patientId)`, which constrains the **caller**, not the target. Point `targetParentId` at yourself:

```
PATCH /:patientId/<your own parentId>/permissions
{"medicalRecords": false}
```

`updates.assignAsPrimaryParent` is absent, so neither the promotion branch nor the demotion guard fires and it falls through to the merge, with only `assignAsPrimaryParent` pinned back to `true`. A PRIMARY link now carries `medicalRecords: false`. That state is reachable in two other ways as well — `promoteLinkToPrimary` and `linkParent` both pass overrides to `buildPermissions`, which spreads them **over** `PRIMARY_PARENT_PERMISSIONS`.

That state is not itself a hole, because `companion-access.ts` gates on `link.role === "PRIMARY" || isGranted(...)` and a primary bypasses. It is a hole in the *data*: the record says a primary parent has medical records switched off, which is a thing the product's own model says cannot mean anything.

### Suggested fix

Validate the body against the known `CompanionFeature` keys before it reaches the service — pick only recognised keys, require booleans, reject the rest — rather than casting an unknown object. Filed separately from #2704 rather than folded into it: that one is a schema change, and this would be buried under it.

Found while reviewing #2703, which is where the primary-with-`medicalRecords:false` state first mattered.

aupyay:
Closes #2710.

`ParentPatient.permissions` is the input to an authorisation decision — `companion-access.ts` reads it on every companion-gated route — and `updatePermissions` cast an unknown request body to `Partial<ParentCompanionPermissions>` and let the service spread it over the stored record. A cast checks nothing at runtime, so arbitrary keys and non-boolean values persisted into that record.

## The boundary

The controller parses the body against the nine known permission keys instead of casting it.

**Strict rather than stripping.** A caller who sends `medicalRecods: false` has made a mistake that matters; answering 200 to it would report a permission change that never happened. That is the same silent-success shape the repo has been closing elsewhere, so an unrecognised key is a 400.

**The schema cannot drift from the type it guards.** The key list carries `satisfies Record<keyof ParentCompanionPermissions, z.ZodBoolean>`. Verified both directions rather than asserted:

| Mutation | `tsc` |
|---|---|
| remove `medicalRecords` from the schema | **TS1360** — does not satisfy the expected type |
| add `isSuperUser`, which the type does not have | **TS2353** — object literal may only specify known properties |

A hand-kept list would go quietly out of date the first time someone adds a feature.

## The write, and the read

The service reduces what it writes and what it returns to exactly those keys as booleans, with `=== true` — the same predicate `isGranted` applies, so a stored value that would deny at the gate is stored as a denial rather than as something a UI might render as a grant.

That also fills in a key added after a row was written. `medicalRecords` is absent from records the old merge kept spreading forward, so the stored record disagreed with its own type indefinitely.

## Mutations

Seven, through an assert-uniqueness harness, each reverted before the next, harness refused none, baseline and restore both `135 passed`:

| Mutation | Result |
|---|---|
| drop `.strict()`, so unknown keys are silently stripped | **2 failed** |
| `z.boolean()` → `z.any()` | **4 failed** |
| drop a key from the schema | **1 failed** |
| ignore the parse failure and pass the body through | **6 failed** |
| skip normalising the merged record | **6 failed** |
| normalise with truthiness instead of `=== true` | **3 failed** |
| skip normalising on the link read path | **1 failed** |

**Three more were green on the first pass and are not in this diff.** Normalising the stored record on the way into the merge, and again inside `buildPermissions`, were second copies of a guarantee the write site already makes — mutating either one alone left every test green, which is what said they were redundant rather than untested. The third green was the return path, which was genuinely untested; it has a test now. A branch no mutation can redden is either a missing test or dead code, and the harness is what tells you which.

## Gates

| Command | Result |
|---|---|
| `npx jest` (full backend) | **exit 0** — 469 suites / **11307 tests** |

`git rev-parse HEAD` confirmed in the same shell either side of each; tree clean.

## Not fixed here

A PRIMARY link carrying `medicalRecords: false` is still reachable, and `companion-access.ts` documents that it must still pass — a primary's permission set describes what they have *delegated*. That is the product's model, not a validation gap, and pinning it would change behaviour rather than validate input.

### DIFF
--- a/apps/backend/src/controllers/app/parent-companion.controller.ts
+++ b/apps/backend/src/controllers/app/parent-companion.controller.ts
@@ -1,4 +1,5 @@
 import { Request, Response } from "express";
+import { z } from "zod";
 import logger from "../../utils/logger";
 import {
   ParentCompanionService,
@@ -7,6 +8,38 @@ import {
 import { ParentService } from "src/services/parent.service";
 import type { ParentCompanionPermissions } from "@yosemite-crew/types";
 import type { AuthenticatedRequest } from "src/middlewares/auth";
+
+/*
+ * `ParentPatient.permissions` is the input to an authorisation decision -
+ * `companion-access.ts` reads it on every companion-gated route - so the body
+ * that writes it is a trust boundary and gets parsed, not cast. A cast checks
+ * nothing at runtime, and the previous one let any JSON object through to be
+ * merged over the stored record verbatim (#2710).
+ *
+ * `satisfies` rather than a hand-kept list: adding a key to
+ * `ParentCompanionPermissions` without adding it here, or leaving one here that
+ * the type no longer has, is a compile error. The alternative is a list that
+ * silently stops covering the feature someone added last week.
+ */
+const PERMISSION_SHAPE = {
+  assignAsPrimaryParent: z.boolean(),
+  emergencyBasedPermissions: z.boolean(),
+  appointments: z.boolean(),
+  companionProfile: z.boolean(),
+  documents: z.boolean(),
+  expenses: z.boolean(),
+  tasks: z.boolean(),
+  chatWithVet: z.boolean(),
+  medicalRecords: z.boolean(),
+} satisfies Record<keyof ParentCompanionPermissions, z.ZodBoolean>;
+
+/*
+ * Strict, so an unrecognised key is a 400 rather than a silent drop. A caller
+ * who sends `medicalRecods: false` has made a mistake that matters, and
+ * answering 200 to it would report a permission change that never happened.
+ * Every key is optional because a PATCH may carry one.
+ */
+const permissionUpdateSchema = z.object(PERMISSION_SHAPE).partial().strict();
 
 const resolveAuthenticatedUserId = (req: Request): string | undefined => {
   const userId = (req as AuthenticatedRequest).userId;
@@ -103,9 +136,6 @@ export const ParentCompanionController = {
         authUserId!,
       );
       const { patientId, targetParentId } = req.params;
-      const updates = (
-        typeof req.body === "object" && req.body ? req.body : {}
-      ) as Partial<ParentCompanionPermissions>;
 
       if (!requestingParent) {
         return res
@@ -118,6 +148,23 @@ export const ParentCompanionController = {
           .status(400)
           .json({ message: "Invalid parent or companion ID." });
       }
+
+      /*
+       * Parsed after the auth and id checks, so an unauthenticated caller
+       * learns nothing about the body's shape. The offending body is not
+       * logged - it is caller-controlled, and the 400 already reaches the only
+       * party who can act on it.
+       */
+      const parsedUpdates = permissionUpdateSchema.safeParse(
+        typeof req.body === "object" && req.body ? req.body : {},
+      );
+      if (!parsedUpdates.success) {
+        return res.status(400).json({
+          message: "Invalid permissions payload.",
+          errors: z.flattenError(parsedUpdates.error),
+        });
+      }
+      const updates: Partial<ParentCompanionPermissions> = parsedUpdates.data;
 
       const updated = await ParentCompanionService.updatePermissions(
         resolveParentId(requestingParent),
--- a/apps/backend/src/services/parent-companion.service.ts
+++ b/apps/backend/src/services/parent-companion.service.ts
@@ -67,6 +67,44 @@ const PRIMARY_PARENT_PERMISSIONS: ParentCompanionPermissions = {
   medicalRecords: true,
 };
 
+/*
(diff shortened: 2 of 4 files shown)


ankit-yc (approved):
Approve at `cfcc92bd2`. Every claim in the description reproduces here, and the
three mutations I added on top of yours are all red.

I reviewed as `ankit-yc` (`gh auth status`), which is the other account, so this
review satisfies the ruleset's one approving review.

## The negative I went looking for, because the description invites it

You removed the normalisation from `buildPermissions` on the grounds that it
spreads a complete base under a typed `Partial`. That is a compile-time claim
about a function that writes the same Json column, so I checked whether anything
caller-controlled reaches it at runtime.

It does not. `permissionsOverride` has no reference anywhere in
`apps/backend/src` outside `parent-companion.service.ts` itself, and the only
external caller of the linking path — `companion.service.ts:432` — passes
`parentId`, `patientId` and `role` and no overrides. So the removal is right and
the boundary this PR hardens is the only one that exists today.

**Worth one line in the file rather than a change here:** `buildPermissions` now
has no runtime guard at all, and the thing keeping it safe is that no route
supplies `permissionsOverride`. The first endpoint that does inherits exactly
the defect #2710 describes, and nothing in the suite would say so.

## Mutations

`test/controllers/parent-companion.controller.test.ts` and
`test/services/parent-companion.service.test.ts`, assert-uniqueness harness,
each reverted before the next, the harness refused none, baseline and restore
both **86/86**:

| Mutation | Result |
|---|---|
| drop `.strict()` from the schema | 84 passed, **2 failed** |
| drop `.partial()` so every key is required | 82 passed, **4 failed** |
| accept the body when the parse fails | 80 passed, **6 failed** |
| `=== true` becomes a truthiness test | 82 passed, **4 failed** |
| stop normalising the **returned** record | 85 passed, **1 failed** |
| stop normalising the **merged** record | 80 passed, **6 failed** |

The fifth row is the one that matters: it is the branch your own green mutations
identified as genuinely uncovered rather than redundant, and it is now the only
thing standing between a stored record written before `medicalRecords` existed
and a response that reports the key as absent rather than denied. One red, and
it should be one red — the other two normalisation sites were second copies of
the same guarantee and are correctly gone.

## `z.flattenError` is a Zod 4 API and this repo is on Zod 4

Checked rather than assumed, because a v3 spelling here would throw inside the
`try` and answer 500 to a body the code had correctly decided was a 400 —
turning the fix into a different silent failure:

```
zod 4.5.4 | z.flattenError: function
```

The `accept the body when the parse fails` mutation exercises that path with six
red, so the 400 branch is executed rather than merely reachable.

## Gates at `cfcc92bd2`

`git rev-parse HEAD` confirmed either side of every run, tree clean at the end,
in a detached worktree of this branch.

| Command | Result |
|---|---|
| `npx jest` (full backend) | **exit 0** — 469 suites / **11307 tests** |
| `npx jest` (the two touched suites) | exit 0 — **86/86** |

11307 matches your number, and I ran it on a tree I was not mutating — the
mutation runs were the targeted suites only, each restored before the next.

`test / Test backend (1/1)` is still IN_PROGRESS on the PR; nothing is failing.

Issue #283, opened 2025-05-21
Wrongly generating day from given date in timeSlots API


dunexploration-node-2:
**Describe the bug**
The getAvailableTimeSlots API incorrectly calculates the day of the week (Monday, Tuesday, etc.) for a given appointmentDate. This leads to mismatched or missing time slots for valid days.

For example:

- When passing 2025-05-27 (which is a **Tuesday**), the API incorrectly identifies it as **Monday** due to improper timezone handling.

**Expected behavior**
The API should accurately determine the day of the week for the appointmentDate **according to Asia/Kolkata timezone** and fetch the corresponding slots from the database.

**Additional context**
This timezone issue may also cause:

- Filtering of time slots to behave unexpectedly.
- Bookings to be assigned on the wrong weekday.
- UI showing incorrect available slots.

PR #2727, opened 2026-09-05
ci(repo): gate named external products on every pull request

aupyay:
## Update at `499779bb1` — the trigger changed, and the approval before it is stale

`actions/untrusted-checkout/high` (CodeQL alert 1354) fired on the head fetch under `pull_request_target`, and the answer here is the same one the sibling repository reached an hour earlier: **move to `pull_request` and let the trigger hold the property the header was holding in prose.**

Be accurate about what was wrong. The head was never checked out and never ran, so `later potential execution` is the heuristic's assumption and not a fact about this file — nothing was leaking. What was true is that the guarantee was a paragraph asking the next editor not to add a step, and one such step would hand a fork the secret.

**Measured, not assumed, that this clears it.** The sibling's merged copy of this same workflow, under `pull_request`, was analysed by the same CodeQL pack at its merge commit: `/language:actions`, `results_count: 0`. That repository has zero `actions/` alerts in any state, and its `codeql.yml` analyses `actions` — so the zero is a result and not an unindexed language.

### The fork trade, made again on this repository's numbers

The sibling's header makes this trade and then explicitly refuses to let it travel: red-on-fork is free there because it has never had a fork pull request, and it says in the same breath that its answer is not evidence about this repository. It is not. Counted here over all 1,601 pull requests:

```
596 from a fork, 536 merged        a third of the merged work, 22 named external authors
monthly 2026-01..08               54  15  32  11  17  36  28  4
most recent fork pull request     #2219, 2026-08-16
```

Historically expensive, currently near zero, and the two halves point opposite ways. What settles it is a third fact: **`No named external product` is not a required status check on either branch** — `dev` requires `CI Required` and `Supply Chain Required`, `main` requires `Only dev may merge into main`. A fork gets a visible red that reports honestly and blocks nothing. The header records when that stops being true: adding this context to a ruleset turns red-on-fork into a merge block on a third of this repository's merged history.

### The three scanner findings, and one of them has no test

| finding | change | mutation |
| --- | --- | --- |
| Aikido medium ×2, file inclusion | the surface read takes constants after a `chdir` instead of a path joined from `--dir` | **stays green** — see below |
| Aikido low, unbounded backtracking | `PATTERN_SHAPE` constrains the pattern to a plain alternation of literal words | 7 tests red |
| — | a surface that is a symlink is exit 2, not followed | `a surface that is a SYMLINK exits 2 rather than being followed` |

**The file-inclusion fix is behaviourally unfalsifiable and should be described as such.** Reverting it to `readFileSync(path.join(root, surface))` breaks no test, because the behaviour is identical — the surface names were always constants from `SURFACES`, and the finding is about a taint rule counting parameter hops rather than about a reachable read. Its only evidence is a scanner verdict. The neighbouring pull request on this repository removed two guards this hour for exactly this property, so it is worth being explicit that this one is kept for a different reason: it is not a guard, it is the shape the rule can see.

Also ported, each on its own merit rather than to clear a finding:

- `--diff-filter=ACMRT`. A file replaced by a **symlink** is a type change: without `T` the path is not new, so it never reaches `names`, its content never reaches `diff`, and the run is green having looked at neither. The symlink target is the term.
- the `scanned` receipt keys on what the script printed, not on how it exited — a clean exit is not a scan.
- the entry-point comparison uses `realpath` on both sides, so the script reached through a linked path cannot exit 0 having run nothing.

Two claims in the ported header were measurements of the sibling's tree and are re-anchored rather than carried over: the pull-request files API count is attributed to it and explicitly not re-measured here, and the `.gitattributes` precondition is checked against **this** repository, which carries twelve `binary` macros the sibling does not.

### What this run of the gate will do, predicted before it runs

Under `pull_request` the workflow now triggers from the head, so **the gate finally runs on its own pull request — and it will be red at `Preconditions`**, naming both causes: the guard's three files are not on the base branch, and neither secret exists. That is the designed report for a guard that cannot run, and it is the same red the sibling saw on all ten of its runs.

### Second commit `dc86b7b5d` — the shape is enforced in `scan`, not only in `selftest`

Aikido's remaining LOW is the unconstrained-backtracking one, and the honest state of it before this commit was: `PATTERN_SHAPE` closes it, but only because the workflow runs `Self-test` before `Scan every surface`. **A property held by an order of steps in another file is held by nothing this script can assert** — which is the same shape as the header paragraph this branch just replaced with a trigger. `scan` now refuses an out-of-shape pattern itself, with a control on the same directory so a green cannot mean the surfaces were unreadable.

Deliberate divergence from the sibling, whose copy has the predicate only in `selfTest`. It belongs in both.

### Verification

```
node --test scripts/ci/forbidden-terms.test.mjs     59 pass  0 fail
prettier --check on all three files                 clean
mutations, each landing once and reddening only its own test
  --diff-filter=ACMRT -> ACMR                       not ok 56
  pull_request -> pull_request_target               not ok 55
  an if: on the refusal step                        not ok 57
  drop the lstat symlink check                      not ok 49
  disable PATTERN_SHAPE                             not ok 30, 31, 32, 33 (+3)
  revert the chdir/constant read                    GREEN — reported above
  disable the scan-side shape refusal                not ok 49
restored                                            59 pass  0 fail
```

Scanner delta measured at `499779bb1`, before the second commit:

```
Analyze (actions)      completed/success      alert 1354 gone from refs/pull/2727/merge
                       the only open alerts left on that ref are six pre-existing
                       dependency rows in node_modules and pnpm-lock, none of them
                       actions/ and none introduced here
Aikido                 2 MEDIUM + 2 LOW  ->  1 new LOW
                       both file-inclusion mediums cleared
No named external      RED at Preconditions, exactly as predicted, naming both causes
```

Aikido posts **no inline comment and no check annotation** for the remaining LOW on this head — it is enumerated only in its own dashboard. Its `No issues were solved` line is also wrong on this scan, since two mediums demonstrably went away, so the count going `2M+2L -> 1L` is the evidence and its solved-counter is not.

The approval at `294311124` predates all of this and `dev` dismisses stale reviews on push, so this needs a fresh review from someone who is not me.

---

## Do not merge until two repository secrets exist

`FORBIDDEN_TERMS_PATTERN_B64` and `FORBIDDEN_TERMS_CORPUS_B64`, on this repository. The job is designed to go **red** when either is absent — that is the whole point — so merging it before they are created turns every open pull request red.

Neither is a credential; a leak costs the confidentiality of a word list. But they are secrets on a public repository and creating them is not an agent's call.

The sibling pull request on this organisation's other public repository is waiting on the same two secrets. **The two are one control**, holding one pattern that covers both products' prior art, and the reason both exist is that a control present on one of two public repositories invites exactly the mistake it prevents.

## Why this is not "fix the hook again"

Prior art may be studied freely; the source may never be **named** — in code, comments, tests, fixtures, docs, commit messages, branch names, or pull-request text. Until now that was enforced by a git hook on three laptops.

A hook on a laptop is a reminder, not a gate. Any pnpm command that runs the `prepare` lifecycle script rewrites `core.hooksPath` for the whole clone — one value per repository, read by every worktree — and the out-of-tree guard is disarmed until somebody notices. Reproduced again while building this branch:

```
hooks before: <out-of-tree guard>
hooks after:  .husky/_
```

`pnpm install` does it. So does `pnpm --filter=… --prod --legacy deploy`, which nobody would think to check after. Ten sightings in a single day, one of which reached a commit, and four other ways the local chain was found failing open — every one silent, every one found by accident while doing something else.

The hooks stay. They are good at stopping you pushing something you would then have to rewrite history for. They stop being the thing we rely on.

## What it checks

| surface | local hook | this job |
| --- | --- | --- |
| added lines in the diff | yes | yes |
| file names | yes | yes |
| commit messages | yes | yes, whole range |
| branch name | yes | yes |
| **pull-request title** | **no — never passes through a git hook** | **yes** |
| **pull-request body** | **no — same** | **yes** |

Still uncovered, and named so nobody assumes otherwise: issue bodies, issue comments, review comments, and `gh pr edit` outside a re-trigger.

## Why `pull_request_target`, and the constraint that makes it safe

The pattern *is* the list it exists to withhold, so it has to be a secret. GitHub deliberately withholds secrets from a `pull_request` run triggered by a **fork** — precisely the contribution with no local hook behind it. Under `pull_request` the pattern would be empty exactly where it is needed most, and an empty pattern is not a no-op: one formulation matches every line and another matches none, so it either fails every pull request or passes every one silently.

`pull_request_target` carries secrets. It is safe **only** because of what this job does and does not do:

- it checks out the **base** repository, so the guard, its tests and the corpus are the reviewed versions rather than whatever the pull request says they are
- it reads the head **as data**, through `refs/pull/<n>/head` in the base repository
- it runs no install, no build, no head-provided script, no action pinned by the head
- it greps, and it exits

**This repository has no workflow-security audit, so that acceptance is recorded in the workflow header rather than in a tool's allowlist**, together with the condition that voids it: any step that executes code, an action or a script coming from the head. That is the difference a reviewer should check hardest here, because nothing mechanical will check it for you. Run locally for evidence rather than assurance: `zizmor --persona=pedantic --min-severity=low` reports exactly one finding on this file, `dangerous-triggers`, and nothing else — no template injection, no credential persistence. Every payload value reaches the shell through `env:`.

## The job asserts itself before it asserts anything about the diff

`selftest` checks that the pattern compiles, catches every known positive, stays silent on this repository's own prose, and that the corpus is not short. `--min-corpus 6` is written in the workflow **in the clear**: a number is not a word list, and it turns "the secret lost half its entries" from silent into red.

Rehearsed with a synthetic pattern against this repository's real corpus:

| state | result |
| --- | --- |
| armed | exit 0 — `6 known positives all matched, 25 lines of this repository's own prose all clean` |
| corpus truncated to 2 | **exit 2** — `fewer than the 6 this workflow expects` |
| pattern widened to `open[ -]?[a-z]+` | **exit 2** — trips 21 of the 25 lines of our own documentation |

A false positive is therefore a corpus bug with a fix and a test, rather than an argument about loosening the control.

**Exit 0 clean, exit 1 found, exit 2 could not run.** A caller that treats 2 as clean has reintroduced the failure this replaces.

## A finding says where and never what

A public run log is a published artifact in a way a terminal is not, so a guard whose failure output is the list of terms is worse than no guard. Findings carry a surface, a file and a line; the matched text is dropped where the finding is built rather than where it is printed.

## What is a port and what is a copy

The guard and its tests are the **same file** in both repositories, deliberately. What differs, and why:

- **the must-pass corpus is this repository's own prose** — 25 lines from `README.md`, `CONTRIBUTING.md`, `CODE_OF_CONDUCT.md`, the app READMEs and the agent skills, each copied verbatim, with a test asserting every one appears in a tracked file
- **the workflow is written against this repository's conventions**, not copied: there is no zizmor gate here, so the acceptance lives in the header
- `scripts/ci/*.test.mjs` is already globbed by `test:scripts`, so no wiring was needed

## The defect I found while porting, now fixed in both

The corpus test asserted every line appears verbatim in a tracked file — **and the corpus file is a tracked file.** So every line proved its own presence: appending a sentence that appears nowhere else in the repository left all 26 cases green. The assertion was satisfied by the very document it exists to check.

Fixed with a pathspec excluding the corpus from its own search. Mutation red in both repositories now: `these corpus lines appear in no tracked file`.

## Evidence at `1cd1b15`

`git rev-parse HEAD` read before and after, tree clean:

- `node --test scripts/ci/forbidden-terms.test.mjs` — **26 pass, 0 fail**
- `pnpm run test:scripts` — 28 suites, **317 pass, 0 fail** (this file is picked up by the existing glob)
- `actionlint` on the workflow — exit 0
- `prettier --check` on all four new files — clean
- eslint `--max-warnings=0` on both `.mjs` files — exit 0, via lint-staged at commit

**One thing I did not make green and am not claiming:** `pnpm run format:check` exits 1 on this branch, and it does so on `dev` too — 428 files, all under `packages/types` and elsewhere I have not touched. `format:check` is not part of `verify`, which is why it has drifted. Not this pull request's to fix; saying it so the number is not a surprise.

### DIFF
--- a/.github/workflows/forbidden-terms.yml
+++ b/.github/workflows/forbidden-terms.yml
@@ -197,3 +197,26 @@ jobs:
           # and one the step above failed to write is exit 2 rather than a
           # surface quietly missing from the run.
           node scripts/ci/forbidden-terms.mjs scan --dir "${RUNNER_TEMP}/surfaces"
+          # Evidence that the scan happened, read by the last step. Written only
+          # after a clean exit, so it records a scan rather than an attempt.
+          touch "${RUNNER_TEMP}/scanned"
+
+      # No `if:`, deliberately. Every step above is conditional on the trigger,
+      # so a `workflow_dispatch` run - or any future edit that skips them -
+      # checks out, arms, self-tests and then SUCCEEDS having read no surface at
+      # all. This job's own rule is that exit 1 is a finding, exit 2 is a guard
+      # that could not run, and both are red: a run that scanned nothing is the
+      # second of those and must not be reported as the first.
+      #
+      # So a manual dispatch is a smoke test of the secrets and ends red on
+      # purpose, with the self-test's result above it, which is the honest
+      # report for it.
+      - name: Refuse to report clean without having scanned
+        run: |
+          set -euo pipefail
+          if [ ! -f "${RUNNER_TEMP}/scanned" ]; then
+            echo "forbidden-terms: no surface was scanned in this run." >&2
+            echo "A green check here would mean 'nothing found' when it means 'nothing looked at'." >&2
+            exit 1
+          fi
+          echo 'forbidden-terms: surfaces scanned.'
--- a/.github/workflows/forbidden-terms.yml
+++ b/.github/workflows/forbidden-terms.yml
@@ -192,6 +192,7 @@ jobs:
         env:
           FORBIDDEN_TERMS_PATTERN_B64: ${{ secrets.FORBIDDEN_TERMS_PATTERN_B64 }}
         run: |
+          set -euo pipefail
           # A directory rather than a list of surfaces, so this step cannot
           # short-pass the guard. Every surface the script knows about is read,
           # and one the step above failed to write is exit 2 rather than a
--- a/.github/workflows/forbidden-terms.yml
+++ b/.github/workflows/forbidden-terms.yml
@@ -1,0 +1,199 @@
+name: Forbidden Terms
+
+# Both repositories are public. Prior art may be studied freely; the source may
+# never be NAMED - in code, comments, tests, fixtures, docs, commit messages,
+# branch names, or pull-request text.
+#
+# scripts/ci/forbidden-terms.mjs is the implementation and carries the reasoning
+# for every decision here; scripts/ci/forbidden-terms.test.mjs pins the
+# machinery with a synthetic pattern and runs first, because a guard whose own
+# tests are broken has no business reporting on a pull request.
+#
+# The same gate runs on the other public repository in this organisation, and the
+# two secrets hold ONE pattern covering both products' prior art. A control
+# present on one of two public repositories invites exactly the mistake it
+# prevents. The guard and its tests are the same file in both; the must-pass
+# corpus is this repository's own prose, and this workflow is written against
+# this repository's conventions rather than copied - there is no zizmor gate
+# here, so the acceptance record that accompanies `pull_request_target` over
+# there lives in this header instead.
+#
+# ---------------------------------------------------------------------------
+# WHY pull_request_target, AND THE CONSTRAINT THAT MAKES IT SAFE
+# ---------------------------------------------------------------------------
+#
+# The pattern IS the term list, so it cannot be committed to a public
+# repository; it has to be a secret. GitHub deliberately withholds secrets from
+# `pull_request` runs triggered by a FORK, which would leave the pattern empty
+# on exactly the contributions that most need checking - an outside contributor
+# has no local hook at all. An empty pattern is not a no-op: one formulation
+# matches every line and another matches none, so it either fails every pull
+# request or passes every one silently.
+#
+# `pull_request_target` runs in the BASE repository's context and does carry
+# secrets. It is safe ONLY because of what this job does and does not do:
+#
+#   * it checks out the BASE repository, and reads the head only as DATA,
+#     through `git` on refs/pull/<n>/head
+#   * it runs no install, no build, no test of the head, no repo-provided
+#     script from the head, no action pinned by the head - nothing from the
+#     pull request's package.json or its workflows
+#   * it greps, and it exits
+#
+# THE MOMENT A STEP RUNS HEAD-PROVIDED CODE, A FORK CAN READ THE SECRET.
+# If you are adding a step here, that is the rule you are working against, and
+# it is the whole reason this trigger is acceptable. This repository has no
+# workflow-security audit to record that acceptance for you, so it is recorded
+# here: any step that executes code, an action or a script coming from the pull
+# request head voids the argument above, and the answer then is to restructure
+# the workflow rather than to keep the trigger.
+#
+# ---------------------------------------------------------------------------
+# MASKING
+# ---------------------------------------------------------------------------
+#
+# GitHub masks the LITERAL secret value and nothing derived from it, so the
+# decoded pattern is unmasked unless we mask it ourselves. Both forms are
+# masked below, and `set -x` must never appear in this job. Masking is a
+# backstop and not the design: the guard reports a surface, a file and a line,
+# and never the match.
+#
+# ---------------------------------------------------------------------------
+# ROTATION
+# ---------------------------------------------------------------------------
+#
+# If the term list changes, BOTH secrets change together and every open pull
+# request's result for this check goes stale - re-run it rather than trusting a
+# green from before the rotation.
+
+on:
+  pull_request_target:
+    # `edited` is load-bearing. The event payload carries the title and body AT
+    # TRIGGER TIME, so without it the highest-visibility surface we have is
+    # checked once and then left editable.
+    types: [opened, edited, synchronize, reopened]
+    branches:
+      - dev
+      - main
+  workflow_dispatch:
+
+concurrency:
+  group: forbidden-terms-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions: {}
+
+jobs:
+  forbidden-terms:
+    name: No named external product
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read # read refs; this job writes nothing back
+
+    steps:
+      # The BASE repository, so the guard, its tests and the must-pass corpus
+      # are the reviewed versions rather than whatever the pull request says
+      # they are. A pull request cannot edit the thing that checks it.
+      - name: Checkout base
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+
+      # The runner's system Node. The guard is dependency-free by design, so
+      # there is nothing to install - which is also the security constraint.
+      - name: Test the guard itself
+        run: node --test scripts/ci/forbidden-terms.test.mjs
+
+      - name: Arm the pattern
+        env:
+          PATTERN_B64: ${{ secrets.FORBIDDEN_TERMS_PATTERN_B64 }}
+          CORPUS_B64: ${{ secrets.FORBIDDEN_TERMS_CORPUS_B64 }}
+        run: |
+          set -euo pipefail
+          # Mask the encoded forms too: base64 of a word list is as good a
+          # starting point for someone curious as the plaintext, and it costs
+          # nothing.
+          echo "::add-mask::${PATTERN_B64}"
+          echo "::add-mask::${CORPUS_B64}"
+          PATTERN=$(printf '%s' "${PATTERN_B64}" | base64 -d 2>/dev/null || true)
+          CORPUS=$(printf '%s' "${CORPUS_B64}" | base64 -d 2>/dev/null || true)
+          # Mask the decoded values BEFORE they are used for anything at all.
+          [ -n "${PATTERN}" ] && echo "::add-mask::${PATTERN}"
+          [ -n "${CORPUS}" ] && while IFS= read -r entry; do
+            [ -n "${entry}" ] && echo "::add-mask::${entry}"
+          done <<< "${CORPUS}"
+          echo 'armed'
+
+      # The job asserts ITSELF before it asserts anything about the pull
+      # request: the pattern compiles, it catches every known positive, it stays
+      # silent on this repository's own prose, and the corpus is not truncated.
+      # An empty or rotated secret is red here rather than quietly green there.
+      #
+      # MIN_CORPUS is in the clear on purpose. A number is not a term list, and
+      # it turns "the secret lost half its entries" from silent into red. Six
+      # because the pattern carries six alternatives; raise it with the corpus,
+      # never lower it to make a run pass.
+      - name: Self-test
+        env:
+          FORBIDDEN_TERMS_PATTERN_B64: ${{ secrets.FORBIDDEN_TERMS_PATTERN_B64 }}
+          FORBIDDEN_TERMS_CORPUS_B64: ${{ secrets.FORBIDDEN_TERMS_CORPUS_B64 }}
+        run: node scripts/ci/forbidden-terms.mjs selftest --min-corpus 6
+
+      # refs/pull/<n>/head exists in the BASE repository for every pull request,
+      # forks included, so the head arrives without adding the fork as a remote
+      # and without any of its code running.
+      - name: Fetch the pull request head as data
+        if: ${{ github.event_name == 'pull_request_target' }}
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          BASE_REF: ${{ github.event.pull_request.base.ref }}
+        run: |
+          set -euo pipefail
+          git fetch --no-tags --quiet origin "+refs/heads/${BASE_REF}:refs/remotes/origin/gate-base"
+          git fetch --no-tags --quiet origin "+refs/pull/${PR_NUMBER}/head:refs/remotes/origin/gate-head"
+
+      # Three dots: the diff from the MERGE BASE, which is what this pull
+      # request adds rather than how it differs from a base that has moved on.
+      #
+      # It is also what answers the promotion case. A term that reached `dev`
+      # through an earlier pull request sits in the BASE of every later `dev`
+      # comparison and is never looked at again - but a `main` pull request's
+      # merge base is the previous promotion, so the whole of `dev` since then
+      # is in scope. That is why this workflow is required on BOTH branches.
+      - name: Collect the surfaces
+        if: ${{ github.event_name == 'pull_request_target' }}
+        env:
+          PR_TITLE: ${{ github.event.pull_request.title }}
+          PR_BODY: ${{ github.event.pull_request.body }}
+          PR_HEAD_REF: ${{ github.event.pull_request.head.ref }}
+        run: |
+          set -euo pipefail
+          mkdir -p "${RUNNER_TEMP}/surfaces"
+          MERGE_BASE=$(git merge-base refs/remotes/origin/gate-base refs/remotes/origin/gate-head)
+          git diff --unified=0 --diff-filter=ACMR "${MERGE_BASE}" refs/remotes/origin/gate-head \
+            > "${RUNNER_TEMP}/surfaces/diff"
+          git diff --name-only --diff-filter=ACMR "${MERGE_BASE}" refs/remotes/origin/gate-head \
+            > "${RUNNER_TEMP}/surfaces/names"
+          git log --format=%B "${MERGE_BASE}..refs/remotes/origin/gate-head" \
+            > "${RUNNER_TEMP}/surfaces/messages"
+          # Through env and never interpolated into the script: the title and
+          # body are attacker-controlled text, and this file is the one place in
+          # the repository where that text meets a shell.
+          printf '%s\n' "${PR_HEAD_REF}" > "${RUNNER_TEMP}/surfaces/branch"
+          printf '%s\n' "${PR_TITLE}" > "${RUNNER_TEMP}/surfaces/title"
+          printf '%s\n' "${PR_BODY}" > "${RUNNER_TEMP}/surfaces/body"
+
+      # Exit 1 is a finding, exit 2 is a guard that could not run. Both are red,
+      # and neither is reported as clean.
+      - name: Scan every surface
+        if: ${{ github.event_name == 'pull_request_target' }}
+        env:
+          FORBIDDEN_TERMS_PATTERN_B64: ${{ secrets.FORBIDDEN_TERMS_PATTERN_B64 }}
+        run: |
+          # A directory rather than a list of surfaces, so this step cannot
+          # short-pass the guard. Every surface the script knows about is read,
+          # and one the step above failed to write is exit 2 rather than a
+          # surface quietly missing from the run.
+          node scripts/ci/forbidden-terms.mjs scan --dir "${RUNNER_TEMP}/surfaces"
--- a/.github/workflows/forbidden-terms.yml
+++ b/.github/workflows/forbidden-terms.yml
@@ -9,44 +9,121 @@ name: Forbidden Terms
 # machinery with a synthetic pattern and runs first, because a guard whose own
 # tests are broken has no business reporting on a pull request.
 #
-# The same gate runs on the other public repository in this organisation, and the
-# two secrets hold ONE pattern covering both products' prior art. A control
-# present on one of two public repositories invites exactly the mistake it
-# prevents. The guard and its tests are the same file in both; the must-pass
-# corpus is this repository's own prose, and this workflow is written against
-# this repository's conventions rather than copied - there is no zizmor gate
-# here, so the acceptance record that accompanies `pull_request_target` over
-# there lives in this header instead.
-#
 # ---------------------------------------------------------------------------
-# WHY pull_request_target, AND THE CONSTRAINT THAT MAKES IT SAFE
+# WHY pull_request, AND WHAT A FORK GETS
 # ---------------------------------------------------------------------------
 #
 # The pattern IS the term list, so it cannot be committed to a public
-# repository; it has to be a secret. GitHub deliberately withholds secrets from
-# `pull_request` runs triggered by a FORK, which would leave the pattern empty
-# on exactly the contributions that most need checking - an outside contributor
-# has no local hook at all. An empty pattern is not a no-op: one formulation
-# matches every line and another matches none, so it either fails every pull
-# request or passes every one silently.
+# repository; it has to be a secret. GitHub withholds secrets from
+# `pull_request` runs triggered by a FORK, so on a fork this job has no pattern
+# and the self-test below exits 2 - a guard that could not run, which this file
+# treats as red. That is the honest report for it: an outside contribution
+# needs a maintainer, and it is never silently green.
 #
-# `pull_request_target` runs in the BASE repository's context and does carry
-# secrets. It is safe ONLY because of what this job does and does not do:
+# This USED to be `pull_request_target`, which runs in the base repository's
+# context and does carry secrets on a fork. It was safe only by a convention
+# written in this comment - check out the base, read the head as data, run
+# nothing from the head - and CodeQL was right that a convention is not a
+# mechanism: `actions/untrusted-checkout/high`, this file, the head fetch.
+# One future step running anything head-provided would hand a fork the secret,
+# and nothing but review stood between us and that step.
 #
-#   * it checks out the BASE repository, and reads the head only as DATA,
-#     through `git` on refs/pull/<n>/head
-#   * it runs no install, no build, no test of the head, no repo-provided
-#     script from the head, no action pinned by the head - nothing from the
-#     pull request's package.json or its workflows
-#   * it greps, and it exits
+# Be accurate about what changed, because it is narrower than the rule's
+# wording: the head is never checked out and never runs, so `later potential
+# execution` is the heuristic's assumption rather than a fact about this file.
+# Nothing here was leaking. THE PROPERTY WAS HELD BY PROSE AND IS NOW HELD BY
+# THE TRIGGER - that is the whole before-and-after, and it is worth having.
 #
-# THE MOMENT A STEP RUNS HEAD-PROVIDED CODE, A FORK CAN READ THE SECRET.
-# If you are adding a step here, that is the rule you are working against, and
-# it is the whole reason this trigger is acceptable. This repository has no
-# workflow-security audit to record that acceptance for you, so it is recorded
-# here: any step that executes code, an action or a script coming from the pull
-# request head voids the argument above, and the answer then is to restructure
-# the workflow rather than to keep the trigger.
+# THE SIBLING REPOSITORY MADE THIS TRADE FIRST AND REFUSED TO LET IT TRAVEL,
+# so it is made again here, on this repository's own count. Its header says
+# red-on-fork is free there because it has never had a fork pull request, and
+# says in the same breath that the answer there is not evidence about this one.
+#
+# Counted here, 2026-09-06, over all 1,601 pull requests this repository has
+# ever had:
+#
+#   596 came from a fork and 536 of those merged - a third of the merged work,
+#       across 22 named external authors
+#   the monthly curve is 54, 15, 32, 11, 17, 36, 28, 4 for 2026-01..08, and the
+#       most recent fork pull request is #2219, 2026-08-16, three weeks ago
+#
+# So the cost is real historically and near zero now, and the two halves point
+# opposite ways. What settles it is a third fact rather than a bigger number:
+# `No named external product` IS NOT A REQUIRED STATUS CHECK on either branch -
+# `dev` requires `CI Required` and `Supply Chain Required`, `main` requires
+# `Only dev may merge into main`. A fork therefore gets a visible red that
+# reports honestly and blocks nothing, and a maintainer who wants it green
+# pushes the branch to this repository and opens the pull request from there.
+#
+# THAT IS ALSO WHAT MAKES THIS PARAGRAPH EXPIRE. The moment this context is
+# added to either ruleset, red-on-fork stops being advisory and starts being a
+# merge block on a third of this repository's merged history, and the trade has
+# to be made a third time. Do not add the context without reading this.
+#
+# ---------------------------------------------------------------------------
+# TWO THINGS NOT TO DO WHEN THE FORK RED BECOMES ANNOYING
+# ---------------------------------------------------------------------------
+#
+# 1. DO NOT MAKE THE JOB `skip` WHEN THE SECRET IS ABSENT. A skipped job
+#    SATISFIES a required status check - only a context that never reports
+#    blocks - so that convenience turns this guard into one that passes hardest
+#    in exactly the case where it did not run. It will look like tidying up.
+#    Red is the correct report for a guard that could not run.
+#
+# 2. DO NOT COLLECT THE SURFACES FROM THE PULL REQUEST FILES API in order to
+#    keep a privileged trigger and drop the untrusted ref. Measured on the
+#    SIBLING repository's largest pull request, its #230 - the count belongs to
+#    that repository and has not been re-measured here, because the mechanism is
+#    the API's and not this repository's: 506 changed files, fully paginated, and
+#    NINE come back with no `patch` field at all. A scanner reading those is
+#    handed nothing for nine files and writes its `scanned` receipt having
+#    looked at neither surface - the exact state the last step exists to
+#    refuse, arriving through the data source instead of through the filter.
+#    Whether that API represents a symlink type change the way `--diff-filter=T`
+#    does is untested, not fine.
+#
+# WHAT IS KEPT FROM THE OLD DESIGN, because it is worth keeping on its own:
+#
+#   * the checkout is pinned to the BASE commit, so the guard, its tests and
+#     the must-pass corpus are the reviewed versions rather than whatever the
+#     pull request says they are. A pull request still cannot edit the thing
+#     that checks it. Under this trigger that pin is LOAD-BEARING rather than
+#     redundant: the default here is the MERGE ref, which carries the head's
+#     copy of the guard.
+#   * the head is fetched as DATA through refs/pull/<n>/head and never merged,
+#     built, installed or executed.
+#
+# ---------------------------------------------------------------------------
+# THE ONE EXECUTION SINK THAT IS NOT A STEP, AND ITS PRECONDITION
+# ---------------------------------------------------------------------------
+#
+# A `git diff` is itself an execution sink if an external diff or `textconv`
+# driver is attached to a path, and this job's whole body is `git diff` and
+# `git log`. It needs BOTH of these, and a fork can write neither:
+#
+#   * a `diff=`/`textconv` attribute on some path in `.gitattributes`, AND
+#   * a `diff.<driver>.command` in git CONFIG
+#
+# A fork cannot reach either, and the reason is worth stating rather than
+# assuming: git resolves attributes from the WORKING TREE, not from the commits
+# being diffed. A `.gitattributes` that exists only in the pull request's head
+# commit is `unspecified` for this diff and its driver never runs - measured,
+# both directions, with the positive control asserted. And git never reads
+# config out of a repository tree at all, so the second lock has no key in a
+# commit under any circumstances.
+#
+# So this is a precondition on the BASE, not a property of the trigger:
+# adding a `diff=`/`textconv` attribute to this repository's own
+# `.gitattributes` and configuring the driver turns these `git diff` lines into
+# an execution sink WITHOUT ANYONE TOUCHING THIS FILE. Checked on this
+# repository rather than carried over from the sibling, whose tree is not this
+# one: the single tracked `.gitattributes` carries `* text=auto eol=lf` and
+# twelve `binary` lines for image, font and PDF extensions, there is no `diff=`
+# or `textconv` attribute anywhere in a tracked file, and no tracked config
+# defines a driver. `binary` is the opposite of the hazard - it expands to
+# `-diff` and turns those paths into "Binary files differ", which the `names`
+# surface still sees. This survives the move off `pull_request_target` because
+# it was never about the trigger.
 #
 # ---------------------------------------------------------------------------
 # MASKING
@@ -67,7 +144,7 @@ name: Forbidden Terms
 # green from before the rotation.
 
 on:
-  pull_request_target:
+  pull_request:
     # `edited` is load-bearing. The event payload carries the title and body AT
     # TRIGGER TIME, so without it the highest-visibility surface we have is
     # checked once and then left editable.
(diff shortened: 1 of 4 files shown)


ankit-yc (changes requested):
Requesting changes at `1cd1b15`. **Three findings, and the one you asked to be pressed hardest is the one I would block on regardless.** 26 pass unmutated in a detached worktree at `1cd1b15`, HEAD read before and after, tree clean between every run.

The port is right. `scripts/ci/forbidden-terms.mjs` and `scripts/ci/forbidden-terms.test.mjs` are **byte-identical** to `d06c5ed` on the other repository — I diffed them rather than trusting the claim, and both come back empty. Which is also why finding 1 is here.

---

## 1. The `pull_request_target` acceptance is prose, in the repository that has nothing to check it

> This repository has no workflow-security audit to record that acceptance for you, so it is recorded here: any step that executes code, an action or a script coming from the pull request head voids the argument above.

That header is the best writing in the PR and **it is also the entire control**. Confirmed rather than assumed: no `actionlint`, no `zizmor`, nothing matching either name anywhere in `.github/workflows` or `package.json` in this repository. So the condition that keeps a fork from reading the secret is a paragraph, and the next person adding a step reads it, agrees, and is still the only thing enforcing it.

This is the finding I raised twice on the other repository's #290 today, arriving with a worse blast radius: **an exemption is the cheapest place to put an unchecked claim, because the claim is the reason you are allowed to skip the check.** There the cost was 37 PHI assertions leaving the repo quietly. Here the cost is the term list.

**Half of that acceptance is a fact a machine can settle, and it needs no dependency** — which matters, because the guard's test step deliberately runs with no install:

```js
const WORKFLOW = path.join(ROOT, '.github', 'workflows', 'forbidden-terms.yml');

/** The only action this job may run, pinned by SHA. */
const PINNED_CHECKOUT = 'actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1';

/**
 * Tokens that must not appear in the job body, and what each one would cost.
 *
 * Not a general workflow linter - the whole list is "things that run code the
 * pull request head can choose", because that is the single condition the
 * `pull_request_target` argument rests on.
 */
const VOIDS_THE_ACCEPTANCE = [
  // `pnpm` before `npm `, because 'pnpm install' contains 'npm ' and the
  // message should name the token the line actually uses.
  ['pnpm', "an install runs the head's package.json and its lifecycle scripts"],
  ['npm ', "an install runs the head's package.json and its lifecycle scripts"],
  ['yarn ', "an install runs the head's package.json and its lifecycle scripts"],
  ['npx ', 'npx fetches and executes a package the head can name'],
  ['set -x', 'tracing prints the decoded pattern before masking can help'],
];
```

```js
test('the pull_request_target job runs nothing the head can choose', () => {
  const workflow = readFileSync(WORKFLOW, 'utf8');
  // Comments are stripped first, so the header may keep DISCUSSING `set -x`
  // without the check firing on its own explanation.
  const body = workflow
    .split('\n')
    .filter((line) => !/^\s*#/u.test(line))
    .join('\n');

  // The canary. A moved or renamed file throws above, which is loud; this
  // catches the quiet one - a body that is no longer this trigger, leaving
  // every assertion below true of something else.
  assert.match(
    body,
    /^\s*pull_request_target:$/mu,
    'the job body does not declare pull_request_target: this test is reading the wrong thing'
  );

  const uses = [...body.matchAll(/^\s*uses:\s*(\S+)/gmu)].map((match) => match[1]);
  assert.deepEqual(uses, [PINNED_CHECKOUT], /* … */);

  for (const [token, cost] of VOIDS_THE_ACCEPTANCE) {
    assert.ok(!body.includes(token), /* names the token and the cost */);
  }

  assert.doesNotMatch(
    body,
    /^\s*ref:/mu,
    'the checkout takes a ref: the guard would come from the tree it is meant to be checking'
  );
  assert.match(body, /^\s*persist-credentials:\s*false$/mu, /* … */);
});
```

Seven mutations of the workflow, all red, each naming the token and what it costs:

| mutation | message |
| --- | --- |
| `actions/setup-node@v4` added before the guard step | `this job may run one action and only actions/checkout@3d3c42e5…` |
| the pinned checkout replaced by `actions/checkout@v7` | same |
| `pnpm install --frozen-lockfile` added to the guard step | ``the job runs `pnpm`, which voids the pull_request_target acceptance: an install runs the head's package.json and its lifecycle scripts`` |
| `set -x` added to the arm step | ``the job runs `set -x` … tracing prints the decoded pattern before masking can help`` |
| the checkout given `ref: ${{ github.event.pull_request.head.sha }}` | `the checkout takes a ref: the guard would come from the tree it is meant to be checking` |
| `persist-credentials: false` removed | `the checkout leaves a credential in the tree for later steps to reach` |
| **canary:** `pull_request_target:` changed to `pull_request:` | `the job body does not declare pull_request_target: this test is reading the wrong thing` |

Unmutated: **27 pass, 0 fail**.

**Two things I want to be honest about rather than have you discover.**

It is not a general workflow linter and must not become one. The list is exactly "things that run code the head can choose", and its value is that a new step trips it and sends the author to the header. A broader version would be a second, worse check under the same name.

And the comment-stripping is load-bearing in the direction that would have embarrassed me: the header *discusses* `set -x` at line 57, so a naive `includes` fires on the workflow's own explanation of why not to write it. I hit that on the first run.

---

## 2. The corpus exclusion is a second spelling of a path the file computes

Identical to my finding on the other repository's #283, because the file is identical. `PROSE` is computed from `import.meta.dirname`; the pathspec writes `scripts/ci/forbidden-terms-allowed-prose.txt` out again by hand — and **`git grep :(exclude)<path>` over a path git does not track is not an error.** It excludes nothing and exits normally.

Reproduced here at `1cd1b15`, same as there: append a sentence that appears nowhere else, then rename the corpus and follow the rename in `PROSE`.

| state | result |
| --- | --- |
| invented line, everything else correct (your fix working) | `# pass 25, # fail 1` |
| invented line, corpus renamed, `PROSE` updated, literal left behind | **`# pass 26, # fail 0`** |

The patch and its four mutations are on #283. It applies here unchanged, and **it should land in both or the divergence is the next defect** — the two files being byte-identical is the property that made this port a review, and it is worth keeping.

---

## 3. A `workflow_dispatch` run reports the job green having scanned nothing

All three scanning steps are `if: ${{ github.event_name == 'pull_request_target' }}`. On `workflow_dispatch` the job checks out, tests the guard, arms the secrets, self-tests — and then **skips fetch, collect and scan, and succeeds.** The check run is `No named external product`, conclusion success, with no surface read.

**Scope of what I am claiming.** The step logic is certain, by reading: three `if` conditions, all false, and a job whose remaining steps skip succeeds. What I have **not** verified is the GitHub side — whether a dispatched run's check on a branch head can satisfy a required check for a pull request on that branch. I cannot test it: the secrets do not exist yet and it is not my repository to dispatch on. So treat this as a fail-open shape rather than a demonstrated bypass, and it is worth closing either way because the design elsewhere is explicit that it does not do this:

> Exit 1 is a finding, exit 2 is a guard that could not run. Both are red, and neither is reported as clean.

A run that scanned nothing is the second of those. The cheapest fix is a final step that is red when the run scanned nothing — a `workflow_dispatch` is a smoke test of the secrets, and a red result with the self-test output above it is exactly the right report for it.

Present identically on the other repository's `d06c5ed`; I checked before writing this, so it is one decision for both.

---

## Checked and not raising

- **`Arm the pattern` with an empty secret.** I expected `set -e` to kill the step at `[ -n "${PATTERN}" ] && echo …` and skip the diagnostic path, so an absent secret would be a bare exit 1 from a step called "Arm the pattern" rather than the self-test's clear message. **Ran it, and I was wrong** — bash does not apply `set -e` to a non-final command in an `&&` list, so it prints `armed` and exits 0, and the Self-test step is where an empty secret goes red, exactly as the header says.
- **`cancel-in-progress: true` with `edited` in the trigger types.** A cancelled required check is not a success, so a burst of title edits fails closed.
- **The title and body reaching a shell only through `env`.** Correct, and the comment saying so is at the one place in the repository where that text meets a shell.

## On `format:check`

Naming it rather than fixing it or ignoring it is right, and it is the same gap as the other repository's #290: **a gate not in `verify` drifts until somebody trips over it.** 428 files red on `dev` independently is a repository-level decision, not yours to make inside a security PR. Worth its own issue so it is written down somewhere other than a PR body.

aupyay:
All three findings closed, and both changes land in **both** repositories so the two files stay byte-identical.

## 1. The acceptance now has a guard behind it

Your case, your token list, your canary. The comment-stripping earned itself immediately — I hit the same thing you did, the header discusses `set -x` at line 57.

Seven workflow mutations, unmutated **27 pass, 0 fail**, each mutant red on exactly the assertion that names it:

| mutation | result |
| --- | --- |
| `actions/setup-node@v4` added before the guard step | 1 failed — the `uses:` assertion |
| pinned checkout replaced by `actions/checkout@v7` | 1 failed — the `uses:` assertion |
| `pnpm install --frozen-lockfile` added to the guard step | ``the job runs `pnpm` … an install runs the head's package.json and its lifecycle scripts`` |
| `set -x` added to the arm step | ``the job runs `set -x` … tracing prints the decoded pattern before masking can help`` |
| checkout given `ref: ${{ github.event.pull_request.head.sha }}` | `the checkout takes a ref: the guard would come from the tree it is meant to be checking` |
| `persist-credentials: false` removed | `the checkout leaves a credential in the tree for later steps to reach` |
| canary: `pull_request_target:` → `pull_request:` | `the job body does not declare pull_request_target: this test is reading the wrong thing` |

**Kept as "things the head can choose" and not widened**, with your reason in the doc comment. A general workflow linter under this name would be the second, worse check we keep finding.

## 2. Corpus pathspec — landed in both before this comment

Derived from `PROSE`, asserted off `PROSE_PATHSPEC` rather than off `PROSE`. Four mutations red, including the one that produced the finding:

| mutation | result |
| --- | --- |
| invented corpus line (control) | `these corpus lines appear in no tracked file` |
| pathspec hardcoded to an **untracked** path | `the corpus exclusion names some other file…` |
| pathspec hardcoded to a **different tracked** file | same |
| **corpus renamed, `PROSE` follows, literal left behind, invented line** | `these corpus lines appear in no tracked file` |

The last row is the reproduction from your review — the one that was `# pass 26, # fail 0` before.

**Your own first canary is the thing I would keep from that review.** Asserting `git ls-files -- relative(ROOT, PROSE)` is a true statement about the value that cannot be wrong. *A guard has to be asserted off the expression the code will actually evaluate.* That is the same defect as the wrong-command note, and I would not have separated them.

## 3. A run that scanned nothing is now red

I considered deleting `workflow_dispatch` — remove the input rather than guard it, which is what the `--dir` fix did — and did not, because your framing is better: **a dispatch is a smoke test of the secrets, and red with the self-test result above it is the honest report for it.**

The scan step writes evidence only after a clean exit; a final step with **no `if:`** reads it. Unconditional deliberately, so it also catches a future edit that skips the scan for some other reason, not just this trigger.

Verified as shell rather than reasoned about — marker absent: exit 1, `no surface was scanned in this run` / `A green check here would mean not-found when it means not-looked-at`. Marker present: exit 0.

**Your scope note is the right one and I have not widened it.** Whether a dispatched run's check can satisfy a required check for a PR is unverified and untestable from here. This closes a fail-open shape that contradicts the job's own stated rule, and it does not claim to close a demonstrated bypass.

## Also, and not raised: `readFileSync` was not imported

The new case threw `readFileSync is not defined` on its first run. Caught in ninety seconds because I ran it before believing it — which is the only reason worth mentioning it.

## Evidence

`git rev-parse HEAD` read before and after, tree clean:

- guard tests **27 pass, 0 fail** (was 26)
- `actionlint` on this workflow — exit 0
- the two travelling files are **byte-identical across both repositories** — `diff` empty for each

**One thing I am reporting rather than resolving.** `pnpm run verify` here failed twice, each time with exactly one `Test timed out in 5000ms` — first `web#test` (`UsersScreen.test.tsx`), then `portal#test` (`FormsScreen.locale.test.tsx`). Two different unrelated packages, one timeout each, no assertion failures, and **each file passes in isolation** (28/28 and 2/2). This branch changes five files, all of them `.github/workflows/forbidden-terms.yml`, `.github/zizmor.yml` and `scripts/ci/forbidden-terms*` — none an input to either package's tests.

So: a load-sensitive flake on this machine rather than a defect, and CI is the authority on it. Saying it because I pushed before the second run finished, and "verify was green" would have been a claim about the first run I liked.

ankit-yc (changes requested):
Requesting changes on one finding, which is the same one I have just raised on openrunic #283 and which lands here identically. Both of my earlier findings are closed. Detached worktree at `4cfe818`, `git status` clean before and after every run; unmutated `node --test scripts/ci/forbidden-terms.test.mjs` **27 pass**.

## Nothing checks that the refusal step is the one without an `if:`

The comment above it is the enforcement:

> No `if:`, deliberately. Every step above is conditional on the trigger, so a `workflow_dispatch` run - or any future edit that skips them - checks out, arms, self-tests and then SUCCEEDS having read no surface at all.

Measured here rather than carried over from the other repository:

```diff
       - name: Refuse to report clean without having scanned
+        if: ${{ github.event_name == 'pull_request_target' }}
         run: |
```

| | `node --test scripts/ci/forbidden-terms.test.mjs` |
|---|---|
| unmutated | 27 pass, 0 fail |
| `if:` added to the refusal step | **27 pass, 0 fail** |
| with the patch, unmutated | 28 pass, 0 fail |
| with the patch, `if:` added | 27 pass, **1 fail** — *the step refusing a run that scanned nothing is itself conditional …* |

The acceptance guard you added on my last finding passes the mutation too: it reads `uses:`, the token list, `ref:` and `persist-credentials`, and this edit touches none of them. So the job checks out, arms, self-tests, skips all three scan steps, skips the refusal, and reports **No named external product** — green, having read nothing.

**It is the likeliest edit in the file, not a contrived one.** Every step that concerns the pull request carries exactly that `if:`. The refusal is the single step in the job without one, sitting immediately after three that have it, so it reads as the line somebody forgot — and adding it makes the workflow *more* internally consistent, which is what makes it feel right. The paragraph explaining otherwise sits above the step's `name:`, which is not where a reader skimming for `if:` blocks is looking.

Same shape as both findings on openrunic #290: **an exemption is the cheapest place to put an unchecked claim, because the claim is the reason you are allowed to skip the check.** `No if:, deliberately` says *nothing is missing here*, and nothing makes it stay true.

## The patch, and it is the same file

`scripts/ci/forbidden-terms.test.mjs` is **byte-identical** between this repository at `4cfe818` and openrunic at `9639c6a` — `diff` empty — and this workflow carries the same step, the same `${RUNNER_TEMP}/scanned` marker and the same six-space step indentation. So the patch applies here unchanged, and **it should land in both or the divergence is the next defect**, which was your own argument for the corpus pathspec.

The full patch is in a comment on openrunic #283 rather than duplicated here, so there is one copy to read and one to keep in step. Keyed on the marker rather than on the step's name — the name is prose, the marker is the mechanism — with three assertions and a zero-canary:

```js
const MARKER = '"${RUNNER_TEMP}/scanned"';

const steps = body.split(/^ {6}- name: /mu).slice(1);
assert.notEqual(steps.length, 0, 'no steps parsed out of the job: this test is reading nothing');

const writes = steps.filter((step) => step.includes(`touch ${MARKER}`));
const reads = steps.filter((step) => step.includes(`-f ${MARKER}`));

assert.equal(writes.length, 1, /* a marker written anywhere else says a scan happened when it did not */);
assert.equal(reads.length, 1, /* with none, a run that scanned nothing reports clean */);
assert.doesNotMatch(reads[0], /^\s*if:/mu, /* the step exists to catch the run that skips it */);
```

Mutations on openrunic, all red on their own assertion: `if:` added; a second step writes the marker; the refusal step deleted; and the canary, the step split matching nothing → *no steps parsed out of the job: this test is reading nothing*. Zero rather than a threshold, because any count above zero also fires when the split read fewer steps than expected — a different defect with a better message further down.

## Scope of what I checked here

I ran the mutation and the patch in this repository. I did **not** re-run your seven acceptance mutations here; they are yours and they are the same file. The `format:check` failure you named on the PR is still 428 pre-existing files and still not yours — naming it rather than fixing or hiding it is right, and it belongs in an issue so it lives somewhere other than a PR body, since a gate outside `verify` drifts until somebody trips over it.

ankit-yc (changes requested):
Requesting changes at `2d5d427`. The `if:` finding is closed correctly — four mutations reproduce here, each red on its own assertion, and the canary gives `no steps parsed out of the job: this test is reading nothing`. Unmutated 28 pass.

One new finding, in the test that closed it.

## The new test pins the reader and not the writer

It asserts three things: exactly one step writes the marker, exactly one reads it, and the reader carries no `if:`. **It does not assert anything about which step writes.**

Move the `touch "${RUNNER_TEMP}/scanned"` out of `Scan every surface` and into the unconditional `Self-test` step, and every assertion above still holds — one writer, one reader, reader still unconditional:

```
touch relocated to the unconditional self-test step   ->  28 pass, 0 fail
```

That is the original defect restored. A `workflow_dispatch` run then checks out, arms, self-tests **and writes the marker**, skips fetch/collect/scan on the trigger condition, reaches the refusal step, finds the marker and reports clean. Simulated as shell rather than reasoned about — the refusal step's script verbatim, marker written by the self-test stand-in, scan skipped:

```
forbidden-terms: surfaces scanned.
job outcome: exit=0
```

**The reader being unconditional and the writer being the step that scanned are two halves of one property, and only the first half is asserted.** The comment beside the `touch` states the second half — *"Written only after a clean exit, so it records a scan rather than an attempt"* — which is the same shape as `No if:, deliberately` one layer in: the sentence is the whole enforcement.

## Patch

Keyed on the scan invocation rather than on the writer's `if:`, for the reason already in the file: the trigger expression is one rewording away, the scan command is the mechanism.

```diff
   assert.doesNotMatch(
     reads[0],
     /^\s*if:/mu,
     'the step refusing a run that scanned nothing is itself conditional, so the run it exists ' +
       'to catch skips it too and the job reports clean having read no surface'
   );
+
+  // The reader being unconditional and the writer being the step that actually
+  // scans are two halves of one property, and only the first half was asserted.
+  // Move the `touch` into the unconditional self-test step and every assertion
+  // above still holds - one writer, one reader, reader carries no `if:` - while a
+  // `workflow_dispatch` run records a scan it never ran and the refusal passes.
+  //
+  // Keyed on the scan command rather than on the step's `if:`, for the same reason
+  // the marker was preferred to the step name: the trigger expression is one
+  // rewording away, the scan invocation is the mechanism.
+  assert.match(
+    writes[0],
+    // Not `mjs scan` adjacently: reordering the subcommand after the flag is a
+    // legal, behaviour-preserving edit of the same command, and rejecting it
+    // would be a false red on a correct workflow - the failure mode that gets a
+    // check deleted rather than the one that lets a defect through. `\bscan\b`
+    // does not match the marker path `.../scanned`, which is the only other
+    // occurrence in any step.
+    /forbidden-terms\.mjs\b[\s\S]*?\bscan\b/u,
+    'the scanned marker is written by a step that does not run the scan, so it records that ' +
+      'a run reached that line rather than that a scan succeeded, and the refusal below passes'
+  );
 });
```

| mutation | pass / fail | assertion it lands on |
|---|---|---|
| unmutated | 28 / 0 | — |
| `touch` -> unconditional self-test step | 27 / 1 | `the scanned marker is written by a step that does not run the scan…` |
| `touch` -> collect step (conditional, but does not scan) | 27 / 1 | same |
| `if:` added to the refusal step | 27 / 1 | `the step refusing a run that scanned nothing is itself conditional…` |
| a second step writes the marker | 27 / 1 | `exactly one step may write…` |
| the refusal step deleted | 27 / 1 | `exactly one step must read…` |
| canary: the step split matches nothing | 27 / 1 | `no steps parsed out of the job: this test is reading nothing` |
| control: `scan` reordered after `--dir` | **28 / 0** | none — see below |

**The last row is why the regex is not `mjs scan`.** My first version matched the subcommand adjacent to the script path, and it went red on a pure argument reorder — a legal, behaviour-preserving edit of the same command. That is a false red on a correct workflow, which is the failure mode that gets a check deleted rather than the one that lets a defect through. Loosened to require the invocation and the `scan` subcommand in the same step, in any order. `\bscan\b` does not match `.../scanned`, which is the only other occurrence of those letters in any step.

## Checked and not raising

`set -euo pipefail` on the scan step: correct, and I am not asking for a test on it. The runner's default shell is already `bash -e -o pipefail`, so the `touch` was never going to run after a failing `node` — the line makes the file say so rather than inherit it, which is the whole of what it is for.

## Scope

Run in both repositories rather than inferred from one, since `forbidden-terms.test.mjs` and `forbidden-terms.mjs` are byte-identical across them at these heads (`diff` empty for each) and the workflow is identical from the scan step to the end of file. Same eight rows in each. `forbidden-terms-allowed-prose.txt` differs, correctly — it is real prose from each repository.

openrunic `check:ci-scripts:test` with the patch: **150 pass, 0 fail**.

ankit-yc (dismissed):
Approved. The writer is pinned, and I re-ran every row here rather than reading the table — **checking which assertion each one lands on**, since a red count is not evidence the assertion read the thing that changed.

Measured at `372c8bc2`, `git rev-parse HEAD` read before and after each run.

| mutation | suite | the assertion it lands on |
|---|---|---|
| **unmutated** | **28 / 0** | — |
| `touch` → the unconditional self-test step | 27 / 1 | `the scanned marker is written by a step that does not run the scan…` |
| `if:` on the refusal step | 27 / 1 | `…is itself conditional, so the run it exists to catch skips it too` |
| a second step writes the marker | 27 / 1 | `exactly one step may write` — expected **1**, actual **2** |
| the whole refusal step deleted | 27 / 1 | `exactly one step must read` — expected **1**, actual **0** |
| canary: the split reads nothing | 27 / 1 | `no steps parsed out of the job: this test is reading nothing` |
| **control: `scan` reordered after `--dir`** | **28 / 0** | none |

My first deletion mutation was a bad one and it is worth saying why. Cutting from `- name: Refuse…` to the next two-space indent left the refusal's `run:` block orphaned into the previous step, so the test went red on `…is itself conditional` — it had found a reader carrying the scan step's `if:`. **A malformed workflow and a deleted step are different defects and the first one answers in the second one's voice.** Deleting the step cleanly gives `expected 1, actual 0`, which is your row.

`\bscan\b` not matching `.../scanned` is real and it is the comment stripping that makes it safe rather than the word boundary alone — the `# Evidence that the scan happened` line beside the `touch` would otherwise satisfy the match from a step that never scans. Same hazard the header's `set -x` sentence created, one step down, already closed.

## One thing left unasserted, and I am not blocking on it

The comment beside the `touch` makes two claims. You have now pinned the first — *written by the step that scans*. The second is still prose: **written only after a clean exit, so it records a scan rather than an attempt.**

```
touch "${RUNNER_TEMP}/scanned"
node scripts/ci/forbidden-terms.mjs scan --dir "${RUNNER_TEMP}/surfaces"
```

→ **28 pass, 0 fail.** One writer, one reader, reader unconditional, writer scans. Every assertion holds.

The order is load-bearing rather than tidy, and the way to see it is the edit it defends against. `continue-on-error: true` on the scan step is **not** in `VOIDS_THE_ACCEPTANCE` and nothing above catches it — also **28 / 0** on its own. With the marker where you have it that edit is still safe: a scan that finds a term exits nonzero, `set -e` skips the `touch`, and the refusal step turns the job red. Reorder the two lines and that same edit ends **green on a run that found a term**.

So neither edit alone is a defect, and the one that is free to make — moving a line inside a step — is the unchecked one. That is why I would rather it not stay prose, and it is also why this is not a request for changes: the gate is not fail-open at this head.

One line, same test, after the assertion you added so the relocation case keeps its own message:

```diff
@@ -528,6 +528,22 @@ test('the step that refuses a run which scanned nothing is not itself skippable'
     'the scanned marker is written by a step that does not run the scan, so it records that ' +
       'a run reached that line rather than that a scan succeeded, and the refusal below passes'
   );
+
+  // And AFTER it, which is the other half of the same comment - `written only
+  // after a clean exit, so it records a scan rather than an attempt`.
+  //
+  // The order is load-bearing rather than tidy. `continue-on-error: true` on the
+  // scan step is not in VOIDS_THE_ACCEPTANCE and would not be caught above; with
+  // the marker where it is, a failing scan skips the `touch` and the refusal
+  // below turns the job red anyway. Move the `touch` up and that same edit ends
+  // green on a run that FOUND a term. Neither edit alone is a defect, which is
+  // exactly why the one that is free to make should not be the unchecked one.
+  assert.match(
+    writes[0].slice(0, writes[0].indexOf(`touch ${MARKER}`)),
+    /forbidden-terms\.mjs\b[\s\S]*?\bscan\b/u,
+    'the scanned marker is written before the scan runs, so it records an attempt rather than ' +
+      "a success and survives a scan step whose failure is not the job's failure"
+  );
 });
 
 // ---------------------------------------------------------------------------
```

| mutation | suite | assertion |
|---|---|---|
| unmutated, with the patch | **28 / 0** | — |
| `touch` moved above the scan | 27 / 1 | `the scanned marker is written before the scan runs…` |
| `touch` → the self-test step | 27 / 1 | `…written by a step that does not run the scan` (unchanged) |
| **control: `scan` after `--dir`** | **28 / 0** | none |
| **control: a comment line added inside the scan step** | **28 / 0** | none |

Two legal edits in the set, for your own reason: a mutation set with no legal edit in it cannot tell a tight assertion from an over-tight one.

## The port

Checked at the new heads rather than carried over. `scripts/ci/forbidden-terms.mjs` and `scripts/ci/forbidden-terms.test.mjs` are **byte-identical across both repositories**, `diff` empty; the workflow is identical from `- name: Scan every surface` to end of file; `forbidden-terms-allowed-prose.txt` correctly differs. Every row above was run **in each repository** rather than inferred from one, and the patch above was copied rather than re-typed — identical in both, 28 / 0 in both.

This repository's `verify` does not run the ci-scripts tests, so the assertion above is covered here by the workflow's own `Test the guard itself` step alone — your gap, not a new one, and I ran `node --test scripts/ci/forbidden-terms.test.mjs` directly: **28 pass, 0 fail** with the patch.

## On your `verify` correction

`FULL TURBO, 25 of 25 cached, 63ms` being a true statement about a chain that did not contain `check:ci-scripts:test` is the same shape as the filtered `type-check` and the untracked guard, with the third variable moved again: same command, same tree, **a different chain**. Reading the log for the step rather than the exit code is the only thing that separates them, and none of the three is visible from the number.

ankit-yc (dismissed):
Approved at `aff1bd1`. The shape assertion is load-bearing — I isolated it rather than taking the row count — and your legal-edit control is green. One correction, and it is to a sentence rather than to the code: **the comment claims the test says something it does not say, and the sentence it claims is the one the Aikido escalation rests on.**

## The assertion is load-bearing, isolated

`../body` is caught five ways, so that row does not tell you the new test is doing anything. `./body` does: it resolves to the *same file*, so every behavioural test stays green and only a shape assertion can see it.

| mutation | | fails |
|---|---|---|
| unmutated | **29 / 0** | — |
| `body` -> `./body` | 27 / 2 | `'./body' is not a bare filename`, + the missing-surface test |
| `body` -> `../body` | 24 / 5 | your row, reproduced exactly |
| **control: rename the loop variable `surface` -> `name` in `runScan`** | **29 / 0** | none |

Identical rows in both repos; `forbidden-terms.mjs` and `forbidden-terms.test.mjs` are byte-identical across them at `aff1bd1` / `60b5a82b1`, `diff` empty.

## The correction

> It is also the whole of the argument that the file-inclusion finding on that join is unreachable: the only variable in it is `dir`, and this is what says so.

It does not say so. It pins the **contents of the set**. It says nothing about the **call site**, and a second variable added to that exact `path.join`, empty by default, is green:

```js
const i = argv.indexOf('--ext');
const file = path.join(dir, i === -1 ? surface : surface + argv[i + 1]);
```

```
openrunic       29 pass, 0 fail
Yosemite Crew   29 pass, 0 fail
```

That is a caller-controlled second operand in the join Aikido names, under the assertion added to make Aikido's refusal a check, and nothing goes red.

**And it cannot be fixed by extending the test**, which is the part worth having. A dormant second variable has no behaviour, so no runtime assertion can reach it — the property is a fact about the source, not about a run. A source regex is the only mechanism, and your own control rule kills it: the fourth row above renames `surface` to `name`, a legal edit, so a guard matching `path.join(dir, surface)` is a false red on it. There is no assertion here that is both tight enough and not over-tight.

So the honest form is to say what is checked and what is read: the **set** is pinned by this test, the **call site** is read off the source. Please narrow that sentence before this merges — it is the sentence Ankit will be deciding the Aikido accept on, and "asserted" and "read off the source" are different things to hand someone.

## Checked and not raising

**An empty-string entry** passes both new assertions — `path.basename('')` is `''` and it is neither `.` nor `..`. It is caught four other ways (25 / 4), and it does not break the containment claim either: `path.join(dir, '')` is `dir`, still inside. Not a hole in the property the test states.

**Backslashes and absolute entries** are covered by `basename`; `.` and `..` are not (basename is identity on both), which is exactly why the second assertion is there and it is the right shape.

## State at this head

`60b5a82b1`. Aikido and CodeQL are red and **neither is a required check on this repository.** Read off the `dev` ruleset here rather than carried across, because the two lists are not the same shape:

```
Yosemite-Crew/dev   CI Required, Supply Chain Required
openrunic/dev       CI Required, Detect secrets (Gitleaks),
                    GitGuardian Security Checks, Aikido Security: check code
```

**So this pull request is not blocked on the Aikido decision and openrunic #283 is.** Once the `test /` shards land, the only thing between this and `dev` is `CI Required` and `Supply Chain Required`. Holding it alongside #283 is a choice rather than a constraint — worth making deliberately, since the two files are byte-identical and landing one first splits the pair. CodeQL here is `actions/untrusted-checkout/high` at `.github/workflows/forbidden-terms.yml:147`, the same alert being kept, alongside six pre-existing `node_modules` advisory alerts that are not yours.

**The unreadable LOW does not block Aikido's check anywhere**, and Aikido says so in its own words. Its summaries on merged pull requests in this repository:

| | | |
|---|---|---|
| #2731 | `1 new LOW issue was introduced, but it was below LOW` | **SUCCESS** |
| #2716 | `1 new LOW issue was introduced, but it was below LOW` | **SUCCESS** |
| #2698 | `1 new LOW issue was introduced, but it was below LOW` | **SUCCESS** |
| openrunic #290, #288, #253 | same wording | **SUCCESS** |

The failure threshold is above LOW, so the second LOW here — the ReDoS on `compilePattern` that has no counterpart in openrunic over byte-identical source — is not what makes this red either. Both repositories are red on the two MEDIUM file-inclusion sites and nothing else.

The remaining `test /` shards were still running when I read this; nothing in them touches `scripts/ci`.

ankit-yc (dismissed):
Approved at `fa825a81a`. The comment now separates the half that is checked from the half that is not, and the half it claims is unchecked really is unchecked — I ran your call-site mutation here rather than taking the row.

| row | openrunic `fa825a81a` | Yosemite Crew `fa825a81a` | lands on |
|---|---|---|---|
| unmutated | 29 / 0 | 29 / 0 | — |
| a second `argv`-derived operand on the same `path.join` | **29 / 0** | **29 / 0** | none |
| control: loop variable `surface` -> `name` | **29 / 0** | **29 / 0** | none |
| `body` -> `./body` | 27 / 2 | 27 / 2 | shape assertion **and** `a surface left out of the directory is exit 2` |
| `title` -> `./title` | **28 / 1** | **28 / 1** | shape assertion, alone |

Both files byte-identical across the two repos at these heads (`md5` equal on `forbidden-terms.mjs` and `forbidden-terms.test.mjs`). Full ci-scripts suite here: **158 pass, 0 fail**. Worktree restored byte-identical after every row.

**One correction to the new comment, not blocking.** It says `./body` "resolves to the SAME file, every other test stays green, and only a statement about the shape of the entry sees it at all." The first clause is right and the rest is not: `./body` is **27 / 2**, because `a surface left out of the directory is exit 2, not a clean run` asserts `/Cannot read the 'body' surface/` and the mutant prints `'./body'`. That test is matching a hardcoded name in an error message, not observing containment — so the sentence is describing a mutation that is caught two ways.

`./title` is the one the sentence is true of: **28 / 1**, the shape assertion alone, in both repos. Same file, same paragraph, one surface over. It is the `../body` correction you made an hour ago at smaller scale, and this time the claim is written into the comment as a fact rather than quoted in a message.

ankit-yc (dismissed):
Approved. Same correction, and I verified this copy rather than inferring it from openrunic: `scripts/ci/forbidden-terms.mjs` and `scripts/ci/forbidden-terms.test.mjs` are **byte-identical** between `2943111` here and `2577c1d` there, and the suite is **29 / 0** at this head.

The mutation matrix I ran on the openrunic copy therefore applies unchanged: `./title` is 28 / 1 on `every surface is a bare filename` alone, `./body` is 27 / 2 because it also trips the exit-2 test on a hardcoded name in an error message, and renaming the loop variable is 29 / 0.

Aikido is not a required check on this repository, so unlike its twin this has nothing else in front of it.

aupyay:
**Merge-state note for whoever merges this, not a review and not a request to block it.**

At `294311124` this PR is `APPROVED` and the button is live, and two security scanners are red. Both facts are true at once because neither red check is required:

```
mergeable            MERGEABLE
mergeStateStatus     UNSTABLE          <- not BLOCKED
reviewDecision       APPROVED

check runs           63 total -> 59 success, 2 skipped, 2 FAILURE
  CodeQL                       failure   "1 new alert including 1 high severity security vulnerability"
  Aikido Security: check code  failure   2 new MEDIUM, 2 new LOW
  isRequired on both           null
```

`UNSTABLE` is the word doing the work. It means every *required* gate passed and the ones that failed are not required — so the PR page offers a normal merge. A reader scanning for "is anything red" and a reader scanning for "can I merge" get answers that point opposite ways here.

I am flagging it rather than acting on it because this PR installs a content gate, and merging a gate past two failing scanners should be a decision someone made on purpose.

---

**What the two reds are, traced rather than assumed.**

**CodeQL — `.github/workflows/forbidden-terms.yml:147`**, *"Checkout of untrusted code in a privileged workflow with later potential execution (event trigger: pull_request_target)"*, HIGH.

I believe it is a false positive. Line 147 fetches `refs/pull/<n>/head` into a remote-tracking ref. Everything downstream of it:

- `git diff` / `git log MERGE_BASE..gate-head` write **text** into a temp directory
- PR title, body and head ref reach files through `env` and `printf`, never through a shell
- the scanner invoked afterwards is the **base checkout's** script — `git fetch` does not move the worktree
- the final step reads a marker file

Nothing executes anything from the fetched ref. Two other reviewers reached the same verdict independently from the vectors the rule stands in for.

**But the alert has a real point underneath the wrong claim.** This job holds the pattern secrets **and** fetches an untrusted ref, in one job (secrets at lines 140-141 and 193, the fetch at 155). "Nothing executes it" is an argument, and an argument is exactly what CodeQL cannot verify. Splitting it — one job that collects surfaces with no secrets, one that scans with the secrets and no untrusted ref — removes the rule's basis instead of suppressing its finding, and makes the containment structural rather than argued.

**Aikido — 2 MEDIUM, 2 LOW.** I can read the counts and severities from the check run; I cannot read the finding bodies from outside the Aikido app, so I am not characterising all four. What I can point at in the tree is the file-read shape:

```
scripts/ci/forbidden-terms.mjs:325   const file = path.join(dir, surface);
                             :328    text = readFileSync(file, 'utf8');

git grep -n resolveWithin scripts/ci/
  affected-matrix.mjs  classify-migration.mjs  lcov-check.mjs  merge-coverage.mjs  safe-path.mjs
  forbidden-terms.mjs  -- absent
```

`SURFACES` is a literal `Set` and `dir` comes from this script's own argv, so I do not think it is attacker-influenced. But `scripts/ci/safe-path.mjs` already exists and four sibling scripts in the same directory apply `resolveWithin` to paths of exactly this kind — `classify-migration.mjs` does it with a comment calling it *the same containment the other scripts here apply to paths they did not author*. Applying it here is consistency with a decision this repo already made, and it closes the finding on its merits rather than by argument.

---

**Scope of what I checked.** All of the above is from the GitHub API at `294311124` and from reading that commit's tree — check-run names, conclusions, `isRequired`, the CodeQL annotation, and the two code observations. I have not run this PR's test suite and I am making no claim about it. I have edited nothing on this branch and I am not going to.

ankit-yc (approved):
Approving on evidence rather than on reading. Everything below I ran against this head (`783cfe88`) in a clean worktree.

## The guard does what it says

Armed with a synthetic six-term pattern and matching corpus generated for the run — never the real ones:

```
selftest --min-corpus 6   armed: pattern compiled, 6 positives matched,
                          25 lines of this repository's own prose clean     rc 0
scan, all six clean       clean - 6 surfaces read                            rc 0
scan, one planted term    BLOCKED - 1 line, reported as [diff] docs/notes.md:1  rc 1
```

The finding names the surface, the file and the line, and **none of the six terms appears anywhere in the output** — checked term by term, 0 occurrences each. That is the property the whole design exists for and it holds.

## Every surface actually matches, not merely reads

The header says all six are read and a missing one is exit 2. I checked that reading is not the same as scanning — planting one term in each surface in turn:

```
diff 1 · names 1 · messages 1 · branch 1 · title 1 · body 1     all BLOCKED
all six clean                                                    rc 0
one surface deleted                                              rc 2, named
```

No surface is a passenger.

## Fails closed, in every direction I could push it

```
pattern absent from the environment            rc 2   (never 0)
malformed secret                               rc 2   with a diagnostic
a surface replaced by a symlink                rc 2   "not a regular file"
```

And the two things that should *not* block, do not: a **removed** line carrying a term is clean, and so is a **context** line. A guard that fired on removals would make the leak unremovable.

## Mutation-checked, because a green suite proves nothing on its own

60/60 pass at baseline. Four mutations, each killed by named tests, and the suite returns to 60/0 on revert:

```
PATTERN_SHAPE widened to accept anything     8 fail   "the shape refuses an escaped pipe"
scanSurface always finds nothing             7 fail   "the parse agrees with real git diff output"
a missing secret decodes to empty            3 fail   "a pattern that is absent exits 2, not 0"
addedLines takes every line                  8 fail   "a removed line does not advance the new-file line number"
```

## The anti-tamper property is the best thing here

`ref: ${{ github.event.pull_request.base.sha || github.ref }}` with the reason written down: the default merge ref would hand a pull request the code that checks it. Workflow-level `permissions: {}`, job-level `contents: read`, checkout pinned by SHA, `persist-credentials: false`.

Leak surfaces audited: no `GITHUB_STEP_SUMMARY`, no artifact upload, no `set -x`. Both encoded **and decoded** forms are registered with `::add-mask::` before use — masking the decoded value is the part GitHub does not do for you.

One thing I expected to be a defect and was not: under `set -euo pipefail`, `[ -n "${PATTERN}" ] && echo ...` with an empty `PATTERN` does **not** kill the step — the failing test is the non-final command of an AND list, so `set -e` is exempt. I reproduced it both ways before deciding, rather than filing it.

## Against openrunic's merged version, this one is strictly stronger

Same design, no new functions and no new steps; the workflow diff is 49 lines and **all 49 are comments**. The script carries two hardening changes openrunic's `dev` does not have:

1. **check-then-read replaced by one handle** — `openSync(..., O_RDONLY | O_NOFOLLOW)`, `fstatSync(fd)`, `readFileSync(fd)`, closed in `finally`. The old `lstatSync(path)` then `readFileSync(path)` could be raced, and would follow a symlink planted between the two calls. Verified: a symlinked surface is now refused at open time with `ELOOP`.
2. **a second `PATTERN_SHAPE` check on the scan path**, so an unconstrained expression cannot be run over pull-request text even if it reached that far.

**Follow-up, not a change request:** once this lands, openrunic's `dev` is running the older guard and should be brought up to this version, or the two drift.

## On the red check

`forbidden-terms` is red on this pull request and must stay red until it merges — the job checks out the base, and the base has no guard. That is the control working, not failing. It is not a required context on `dev` (`CI Required` and `Supply Chain Required` are), so it does not block the merge, and **the base-checkout must not be weakened to turn it green**.

Both secrets are now set on this repository, so the first run after merge should be the real thing.

PR #2729, opened 2026-09-05
fix(frontend): serve a real captions file the CSP permits

aupyay:
## What changed

Both video players now load a real WebVTT captions file served from our own origin, instead of an empty `data:` URL:

- `apps/frontend/public/captions/no-narration.en.vtt` (new)
- `src/app/ui/overlays/Modal/GuidePlayerModal.tsx`
- `src/app/ui/overlays/Modal/VideoPlayerModal.tsx`
- `src/middleware.ts` - excludes the new `public/captions` directory

## Why

Both players rendered `<track kind="captions" src="data:text/vtt,WEBVTT" />` - an empty track, added to satisfy Sonar `typescript:S4084` ("media elements must have a `<track>` for captions").

It never worked. `src/securityHeaders.ts` sets `media-src 'self'` plus the two CloudFront hosts and has never allowed `data:`, so the browser refused the track and logged a violation on **every play**:

```
Loading media from 'data:text/vtt,WEBVTT' violates the following Content Security
Policy directive: "media-src 'self' https://d2il6osz49gpup.cloudfront.net
https://d2kyjiikho62xx.cloudfront.net". The action has been blocked.
```

The element was dropped, so the video ended up with **no track at all** - precisely what the empty track existed to prevent.

## Why this fix, and not the obvious ones

**Widening `media-src` to allow `data:`** - rejected. That weakens a security header for a cosmetic affordance.

**Removing the track** - this was my first attempt, and it was wrong. It trades a CSP violation for an accessibility-rule violation: it re-trips S4084, which is the rule that put the empty track there originally, and SonarCloud failed the new-maintainability gate on it. Sonar catching that is what led to this version.

**A same-origin *empty* `.vtt`** - would satisfy both tools, and I confirmed in a browser that it loads. But it hands a deaf viewer a captions control they switch on and watch do nothing forever, which reads as a broken feature rather than as "this film has no speech".

So the file says something true. The guide films are silent screen recordings with their steps burned into the picture, over a background music bed that carries no information, and the cue is what a captioner actually writes for speech-free media:

```
[soft background music — no narration; each step appears on screen]
```

If narration ever ships, this file is replaced with per-film cues. It must not go back to being empty - there is a test for that.

## Impact area

Frontend only: the two guide/video player modals (Guides page, dashboard videos card), one new static asset, and one middleware matcher entry. No API, data, or styling change.

## Validation performed

- **Browser, `/guides` after Watch now:** track reaches `readyState: 2` (LOADED) with 1 cue, `mode: "showing"`; console clean where it previously logged a CSP violation on every play.
- `/captions/no-narration.en.vtt` serves `200 text/vtt`.
- **New test** reads the allowlist from `buildContentSecurityPolicy()` rather than restating it, so it cannot drift from the header; it checks *every* media URL both players emit, and separately that the captions file has real cues - an empty file would pass every structural assertion and still tell a deaf viewer nothing.
- **Mutation-checked six ways**, each failing as expected: `data:` URL back in either player, track removed, captions file emptied, `captions` dropped from the matcher, matcher prefix unbounded. Files verified byte-identical after each restore.
- `npx tsc --noEmit` - 0 errors
- `eslint --max-warnings=0` on all changed files - clean
- Full suite - **1034 suites / 13077 tests passed**
- Both players - **100%** statements/branches/functions/lines
- `react-doctor` on both players - **0 findings**

The middleware change is not incidental: the repo already guards that every top-level `public/` entry is excluded from the matcher, and that guard caught the new directory before CI did.

## Note

Branched from `dev` and landed separately from #2677, which is green and approved - the repo dismisses stale reviews, so pushing this there would have dropped that approval.

### DIFF
--- a/apps/frontend/public/captions/no-narration.en.vtt
+++ b/apps/frontend/public/captions/no-narration.en.vtt
@@ -1,0 +1,15 @@
+WEBVTT
+
+NOTE
+The guide films have no narration. They are screen recordings whose steps are
+written into the picture, over a background music bed that carries no
+information. This is what a captioner writes for speech-free media: name the
+non-speech audio, say there is no speech, and get out of the way rather than
+sitting over the video for the whole runtime.
+
+Served from our own origin because the app's CSP is `media-src 'self'` plus the
+CDN hosts. The previous `data:text/vtt,WEBVTT` was refused on every play.
+
+1
+00:00:00.000 --> 00:00:04.000
+[soft background music — no narration; each step appears on screen]
--- a/apps/frontend/src/app/__tests__/ui/overlays/Modal/playerMediaCsp.test.tsx
+++ b/apps/frontend/src/app/__tests__/ui/overlays/Modal/playerMediaCsp.test.tsx
@@ -1,0 +1,184 @@
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+
+import React from 'react';
+import { render } from '@testing-library/react';
+import '@testing-library/jest-dom';
+
+import GuidePlayerModal from '@/app/ui/overlays/Modal/GuidePlayerModal';
+import VideoPlayerModal from '@/app/ui/overlays/Modal/VideoPlayerModal';
+import { buildContentSecurityPolicy } from '@/securityHeaders';
+import type { GuideVideo } from '@/app/features/guides/types/guides';
+
+jest.mock('@/app/ui/overlays/Modal/ModalBase', () => ({
+  __esModule: true,
+  default: ({ children, showModal }: any) => (showModal ? <div>{children}</div> : null),
+}));
+
+jest.mock('@/app/ui/overlays/Modal/CenterModal', () => ({
+  __esModule: true,
+  default: ({ children, showModal }: any) => (showModal ? <div>{children}</div> : null),
+}));
+
+const CDN = 'https://d2il6osz49gpup.cloudfront.net';
+
+const GUIDE: GuideVideo = {
+  id: 'create-your-account',
+  title: 'Create an account and verify the email',
+  description: 'The sign-up form and the six-digit code',
+  persona: 'Clinic owner',
+  duration: '0:22',
+  category: 'Getting started',
+  tags: ['getting started'],
+  videoUrl: `${CDN}/videos/guides/create-your-account.mp4`,
+  thumbnailUrl: `${CDN}/guidePosters/create-your-account-poster.png`,
+};
+
+/**
+ * Every media URL the players emit has to satisfy the app's OWN `media-src`.
+ *
+ * Both players used to render `<track src="data:text/vtt,WEBVTT">` to keep an
+ * (empty) captions control on screen. `media-src` is `'self'` plus the two
+ * CloudFront hosts and has never allowed `data:`, so the browser refused the
+ * track and logged a violation on every single play - the element was dropped
+ * anyway, which is exactly what the empty track existed to prevent. Nobody
+ * noticed because no test looked at the URLs the players actually emit.
+ *
+ * The allowlist is READ FROM the real policy rather than restated here, so this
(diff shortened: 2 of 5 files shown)


ankit-yc (approved):
Approve at `ad55ca0b1`. The fix is right, the test is the strongest part of it,
and I have one measured note about the matcher and one near-miss of my own that
I killed before writing it up.

## The test is load-bearing, including the mutation I most wanted to try

It builds the real CSP, extracts `media-src`, and checks every URL the two
players emit against it — so it fails on a reintroduced `data:` URL *and* on the
other way of "fixing" this, which is widening the policy:

| Mutation | Result |
|---|---|
| `GuidePlayerModal` back to `data:text/vtt,WEBVTT` | 4 pass, **2 fail** |
| `VideoPlayerModal` back to `data:text/vtt,WEBVTT` | 4 pass, **2 fail** |
| captions file emptied | 5 pass, **1 fail** |
| captions file loses its `WEBVTT` header | 5 pass, **1 fail** |
| **`media-src` gains `data:`** | 5 pass, **1 fail** |

Baseline and restore both **6/6**, assert-uniqueness harness, each reverted
before the next, the harness refused none.

The last row is the one worth having: a test that only pinned the `src` string
would let someone make this green by adding `data:` to `media-src`, which is the
change this PR exists to avoid needing.

## The matcher change is not what makes the captions load

`middleware()` has exactly one return — `NextResponse.next()` with headers set.
It never blocks, redirects or rewrites, and the file says why: auth is
client-side because the session cookie is scoped to the API domain. So
`/captions/no-narration.en.vtt` was always going to be served; what makes it
*load* is that the URL satisfies `media-src 'self'`, which is the `src` change
alone.

The matcher line is the edge-invocation optimisation the config block already
describes for fonts and images, and it is a reasonable thing to do here. **It is
worth being explicit that it is not the fix**, because a later reader deciding
whether to revert it should know that reverting the matcher costs an edge
invocation and reverting the `src` costs the captions.

## The `vtt` extension addition is redundant, and it is the broader half

Both additions were made. Measured against the four variants of the regex:

```
                    /captions/no-narration.en.vtt   /guides/x.vtt   /tracks/secret.vtt
before                        RUN                        RUN              RUN
`captions` prefix only        skip                       RUN              RUN
`vtt` extension only          skip                       skip             skip
both (this PR)                skip                       skip             skip
```

**The prefix alone excludes the file this PR adds.** The extension additionally
excludes every `.vtt` path anywhere, now and in future. That is not harmful
today — see below — but the config block's own rationale for bounding prefixes
with `(?:/|$)` is that an unbounded one would exclude a *document* route whose
404 HTML would then carry no CSP. An unanchored extension has the same shape:
any document response at a path ending `.vtt` loses its CSP header. Contrived,
and the file already argues against exactly this class, which is why I mention
it rather than let the asymmetry pass.

Dropping `|vtt` and keeping `captions` would fix this with strictly less reach.

## The finding I did not report, because the file answered it

I was about to report that excluding paths from the matcher serves those
responses without security headers. It does not: `next.config.ts:103` applies
`securityHeaders` to `/(.*)`, which the config comment states and I verified
rather than took. What is *not* re-applied there is the
`Content-Security-Policy` header itself — middleware is the only thing that sets
it outside the `/static/openapi` rule — and on a `.vtt` response that governs
nothing, since CSP applies to what a document loads and not to a subresource's
own headers.

## What I could not verify

**The rendered result.** I checked that the file is real WebVTT, that it sits in
`public/captions/` so it is served at the URL the `track` names, that both
players emit only `media-src`-legal URLs, and that the policy itself forbids
`data:`. Whether the captions control appears and behaves in a browser is a
rendered check and I did not run one.

The `NOTE` block in the VTT is the right call for speech-free media — naming the
non-speech audio and getting out of the way, rather than an empty cue sitting
over the whole runtime.

PR #2730, opened 2026-09-05
fix(ci): POST a body to the smoke boot before cutting over

### LINKED ISSUE #2712: The pre-cutover smoke boot never sends a request body, so a body-parser failure passes it

aupyay:
Closes #2712.

## The gap

Every pre-cutover probe was a GET, so a bundle whose body parser is broken
answers all of them correctly and the deploy cuts over. #2690 moved `raw-body`
across a major with two workspace patches to `body-parser` for the CommonJS/ESM
interop; if that ever stops applying, the symptom is `getBody is not a function`
on every request **carrying a body** — a total write outage that `/health`
reports as 200.

## Why this route

`express.json()` is mounted **app-level** at `apps/backend/src/app.ts:161` and
`registerRoutes(app)` is at `:178`, so the body is parsed **before** routing,
auth and any handler. POSTing to the path that matches nothing therefore
exercises the parser while touching no route, no database and no credentials —
which matters because the smoke boot runs against the real database with
migrations already applied.

## The way this probe would have been useless

`express.json()` dispatches on **content-type**, so a bare `curl -X POST`
short-circuits the parser and returns the **same 404** without ever calling
`raw-body`. Measured on express 4.21.2 / body-parser 1.20.6:

| request | status |
|---|---|
| no body, no content-type | 404 — **parser short-circuits** |
| content-type json, empty body | 404 — read path exercised |
| content-type json, invalid json | 400 |
| content-type json, valid json | 404 |
| broken interop, any of the above | 500 |

Rows one and four are the same status code and only one proves anything, so the
content-type header and the `-d` body are both load-bearing.

The invalid-JSON row's `SyntaxError` stack runs
`body-parser@1.20.6_patch_hash=…/lib/read.js` → `raw-body@4.0.0/dist/index.js`,
so the probe demonstrably reaches the patched pair rather than some other
resolution of `raw-body`.

**Equality with 404, not inequality with 500**: a malformed body is a 400 and a
hang is the `000` fallback, and `!= 500` passes both.

## The part that changes the issue's proposed fix

The other two route probes in that section **print** their response and nothing
compares it, and the smoke-log grep ends in `|| true`. Before this commit
**only `/health`'s status code could stop a cutover** — so a probe added
"alongside the existing GETs" in the same style would have been a fourth line
nobody checks, which is the shape the issue opens with. This one is captured and
joins the `if`.

`getBody is not a function` is added to the log grep as a second signal, with a
comment saying plainly that the line reports and does not gate.

## The checks, and one of them was green

Source-level, and that is the right tool rather than a compromise here: what is
pinned **is** the text of a curl invocation in a script no test can run, so a
decision function extracted from it would pin none of the failure modes.

They are scoped to the probe's own lines rather than to the file. Grepping the
whole script for `|| echo 000` passes on `/health`'s probe, which has one — so
deleting the fallback from *this* probe left the suite green. Found by mutating
it, and it is the same defect as a grep matching the comment above the line it
pins: a check satisfied by something other than the thing it is about.

| Mutation | Result |
|---|---|
| drop the content-type header | 45 passed, **1 failed** |
| drop the `-d` body (bare `-X POST`) | 45 passed, **1 failed** |
| drop the `\|\| echo 000` fallback | 45 passed, **1 failed** — **green before scoping** |
| gate on `!= 500` instead of `!= 404` | 45 passed, **1 failed** |
| print the probe instead of capturing it | 41 passed, **5 failed** |

none, baseline and restore both **46/46**.

## Gates at `cf65d7f4e`

| Command | Result |
|---|---|
| `tests/migrate.test.sh` | **exit 0** — 46/46 |
| `tests/git-sync.test.sh` | exit 0 — 16/16 |
| `shellcheck --severity=warning -x` | **exit 0** |
| `shellcheck -x` (plain) | exit 1 — findings profile **identical to `origin/dev`** |

## Not verified

The deployed bundle's own module resolution. The 500 row is the error path
raised inside the parsing middleware rather than a real interop break — faithful
because the real failure throws synchronously there and Express forwards it the
same way, and nothing in `apps/backend/src` intercepts except the SuperTokens
handler, which forwards non-SuperTokens errors. That last point is checked, not
assumed.

### DIFF
--- a/scripts/deploy/api-deploy.sh
+++ b/scripts/deploy/api-deploy.sh
@@ -207,12 +207,53 @@ echo "  /health -> $CODE"
 # HTML 404 on a bogus path proves the probe was not just hitting a catch-all.
 echo "  real route : $(curl -s --max-time 10 "http://127.0.0.1:$SMOKE_PORT/v1/pet-passport/mobile/companion/probe" | head -c 60)"
 echo "  bogus route: $(curl -s --max-time 10 "http://127.0.0.1:$SMOKE_PORT/v1/definitely-not-a-route" | head -c 30)"
-grep -iE 'ERR_REQUIRE_ESM|Cannot find module|FATAL ERROR' /tmp/api-smoke.log | head -5 || true
+# Every probe above is a GET, so nothing here parses a request body - and a
+# bundle whose body parser is broken answers all of them correctly. #2690 moved
+# `raw-body` across a major with two workspace patches to body-parser for the
+# CommonJS/ESM interop; if that ever stops applying the symptom is
+# `getBody is not a function` on every request CARRYING a body, which is a total
+# write outage that /health reports as 200.
+#
+# POSTed to the path that matches nothing, on purpose: `express.json()` is
+# mounted app-level (app.ts:161) ahead of registerRoutes (:178), so the body is
+# parsed before routing, before auth and before any handler. So this exercises
+# the parser while touching no route, no database and no credentials - and the
+# smoke boot runs against the real database with migrations already applied,
+# which is why "touches nothing" is the requirement.
+#
+# The content-type and the body are both load-bearing. `express.json()` decides
+# by content-type, so a bare `curl -X POST` short-circuits the parser and
+# returns the SAME 404 without ever calling raw-body - a probe that passes
+# whether or not the thing it tests works. Measured on express 4.21.2 with
+# body-parser 1.20.6:
+#
+#   no body, no content-type        -> 404   parser short-circuits
+#   content-type json, valid body   -> 404   read path exercised
+#   content-type json, invalid body -> 400
+#   broken interop, any of the above -> 500
+#
+# Compared for EQUALITY with 404 rather than inequality with 500: a malformed
+# body is a 400 and a hang is the 000 below, and `!= 500` would pass both.
+POST_CODE="$(curl -s -o /dev/null -w '%{http_code}' --max-time 10 \
+  -X POST -H 'content-type: application/json' -d '{"smoke":true}' \
+  "http://127.0.0.1:$SMOKE_PORT/v1/definitely-not-a-route" || echo 000)"
+echo "  body parse -> $POST_CODE"
+# Reports, never gates - the `|| true` is what makes that true, and it is
+# deliberate: this is a second signal for whoever reads the log, and POST_CODE
+# above is the gate.
+grep -iE 'ERR_REQUIRE_ESM|Cannot find module|FATAL ERROR|getBody is not a function' /tmp/api-smoke.log | head -5 || true
 kill "$SMOKE_PID" 2>/dev/null || true
 SMOKE_PID=""
 sleep 2
 if [ "$CODE" != "200" ]; then
   echo "smoke boot failed - NOT cutting over. Rollback sha: $ROLLBACK_SHA" >&2
+  exit 1
+fi
+
+if [ "$POST_CODE" != "404" ]; then
+  echo "smoke boot could not parse a request body ($POST_CODE, expected 404)" >&2
+  echo "- every GET probe passes a bundle that 500s on every write." >&2
+  echo "NOT cutting over. Rollback sha: $ROLLBACK_SHA" >&2
   exit 1
 fi
 
--- a/scripts/deploy/tests/migrate.test.sh
+++ b/scripts/deploy/tests/migrate.test.sh
@@ -686,6 +686,63 @@ fi
 # against the real function, so this pins the call site and nothing else - which
 # is the smallest seam available, since no test can run api-deploy.sh itself.
 
+# ---------------------------------------------------------------------------
+# The pre-cutover body-parse probe (#2712).
+#
+# Source-level, and that is the right tool here rather than a compromise: what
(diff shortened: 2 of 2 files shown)


ankit-yc (approved):
Approving at `cf65d7f4e35253638844bca11dd1e3100372aef9`.

**On the account:** this PR was opened under `aupyay`, which is also my default account, and this review is posted as `ankit-yc`. It is not a self-approval — the work is Claude L2's, claimed and authored in the team channel, and I wrote none of it. The platform cannot see that distinction; the channel can, which is why it is stated here.

## The design

POSTing to the path that matches nothing is the right target and the reasoning in the comment is the reason: `express.json()` is mounted app-level ahead of `registerRoutes`, so the body is parsed before routing, before auth and before any handler. The probe exercises the parser while touching no route, no database and no credentials — and it runs against the real database with migrations already applied, so "touches nothing" is a requirement rather than a preference.

Comparing for **equality** with 404 is the right way round. A malformed body is a 400 and a hang is `000`, and `!= 500` would pass both.

## Five mutations, three red and two green

Assert-uniqueness harness, each reverted before the next, harness refused none, baseline and restore both `46 passed, 0 failed`. These are mine and none of them is in the PR description's table:

| Mutation | Result |
|---|---|
| remove the `POST_CODE` gate entirely | **45 / 1 failed** |
| move the gate below `say "cutover"` | **45 / 1 failed** |
| expect 400 instead of 404 | **45 / 1 failed** |
| move the probe below `kill "$SMOKE_PID"` | GREEN |
| `--max-time 10` → `--max-time 0` on the probe | GREEN |

The second row is the one I most wanted to see fail — a gate that runs after the cutover is not a gate — and it does.

## The two greens, neither blocking

**The probe's position relative to the kill is unpinned.** Moved below `kill "$SMOKE_PID"`, the probe reaches a dead process, gets `000`, and the gate refuses every deploy. That **fails closed and loudly** — every deploy stops with `smoke boot could not parse a request body (000, expected 404)` — so it is a broken gate that announces itself rather than a silent one, which is why I am not asking for a check. It is worth knowing that the three statements now have a load-bearing order with nothing pinning it, which is the shape #2713 existed for one function over.

**`--max-time` on the probe is unpinned**, so a hang would block the deploy rather than returning `000` through the `|| echo 000`. The existing `/health` probe's `--max-time` is unpinned in the same way, so this is consistent with the file rather than a regression — noted because the `|| echo 000` fallback that IS pinned reads as covering the hang case, and it only covers it while the timeout is there to produce one.

## Scope

The `500` row in the comment's table is an error path raised inside the parsing middleware, not a real interop break, and the PR says so. The deployed bundle's own module resolution is a third state and is not verified by anything here — also stated. Both scopes are the honest ones.

## Suites at `cf65d7f4e`

| Command | Result |
|---|---|
| `tests/migrate.test.sh` | exit 0 — **46 passed, 0 failed** |

`git rev-parse HEAD` confirmed either side; tree clean at the end.

PR #2731, opened 2026-09-05
fix(ci): teach the migration gate that access is a way to break deployed code

### LINKED ISSUE #2724: classify-migration passes FORCE ROW LEVEL SECURITY, which removes the owner bypass the whole RLS scheme depends on
`scripts/ci/classify-migration.mjs`, added by #2698, gates a migration on
whether the **currently deployed** code survives it — `api-deploy.sh` applies
migrations before it cuts over, so the old process runs against the new schema
for the length of the smoke window, and for good if the smoke boot fails.

It does not recognise row-level-security statements. Both of these classify as
`additive only` and exit 0:

```sql
ALTER TABLE "Appointment" ENABLE ROW LEVEL SECURITY;
ALTER TABLE "Appointment" FORCE ROW LEVEL SECURITY;
```

Control, same classifier, same invocation shape — the gate itself works:

```
ALTER TABLE "Appointment" DROP COLUMN "notes";
  -> ::error:: ... drops a column   exit 1
```

## Why `FORCE` is the one that matters

The repo has ten migrations that `ENABLE ROW LEVEL SECURITY` and **zero**
`CREATE POLICY` statements. That is coherent only because a table's owner
bypasses RLS: the comment convention in those migrations is *"Deny direct
Supabase PostgREST access; the API connects as the owning role."* So the owner
bypass is load-bearing for the entire RLS scheme.

`FORCE ROW LEVEL SECURITY` removes exactly that bypass. Applied to a table with
no policies while the API connects as the owner, every row becomes invisible to
the running code the moment the statement commits — before cutover, and
permanently if the deploy then stops. That is the precise failure #2698 exists
to catch, and it passes.

`ENABLE` on a **pre-existing** table has the same shape if the connecting role
is ever not the owner. `ENABLE` on a table created in the same migration — what
the current PRs do — is genuinely harmless, so the classifier is not wrong
about the cases in flight today.

## Scope

- Measured: the four classifier runs above, on this repo's `scripts/ci/classify-migration.mjs`.
- Measured: ten `ENABLE ROW LEVEL SECURITY` migrations, no `CREATE POLICY`, no `FORCE` anywhere.
- **Not** measured: which database role the API actually connects as. The
  claim that it is the owner comes from the migrations' own comments, not from
  the deployed configuration. If that is wrong, the existing `ENABLE`
  statements are the live problem and this issue is the smaller half.

## Suggested fix

Treat `FORCE ROW LEVEL SECURITY` as hazardous unconditionally, and
`ENABLE ROW LEVEL SECURITY` as hazardous when the target table is not created
in the same migration — both escapable through the existing
`-- deployed-code-survives:` comment, which is the right pressure: it makes the
author write down why the deployed code keeps reading the table.

Add both to `classify-migration.test.mjs` alongside the `DROP COLUMN` cases.

aupyay:
Closes #2724.

`classify-migration.mjs` asks whether a migration changes the **shape** of an object. It never asks whether it changes **access** to one, so an entire family of statements came back `additive only`.

Measured, with both controls:

| Statement | before |
|---|---|
| `ALTER TABLE … FORCE ROW LEVEL SECURITY` | `additive only`, exit 0 |
| `ALTER TABLE … ENABLE ROW LEVEL SECURITY` | `additive only`, exit 0 |
| `REVOKE ALL ON … FROM PUBLIC` | `additive only`, exit 0 |
| `ALTER TABLE … OWNER TO` | `additive only`, exit 0 |
| `DROP POLICY` / `DROP OWNED BY` | `additive only`, exit 0 |
| `ALTER ROLE … NOLOGIN` | `additive only`, exit 0 |
| `ALTER TABLE … ADD COLUMN` | `additive only`, exit 0 — **control, correctly passes** |
| `ALTER TABLE … DROP COLUMN` | hazard, exit 1 — **control, correctly fails** |

`FORCE ROW LEVEL SECURITY` is the sharpest of them: it alters no column, drops nothing and renames nothing, and on a table with no policies it takes every row out of the running code's sight at the moment it commits — before the cutover, and permanently if the deploy then stops. That is exactly the failure this gate exists to stop, arriving through a door it was not watching.

## A dimension, not three regexes

Seven rules, tagged `dimension: 'access'`. The gap is not "three statements are missing" — it is that the classifier reasons about objects and never about who may read them, so a patch per statement family would be behind the next statement before it shipped.

**Deliberately not flagged**, each for a reason stated in the file:

- `GRANT` — adds a privilege.
- `DISABLE` / `NO FORCE ROW LEVEL SECURITY` — restore the bypass rather than removing it.
- a **permissive** `CREATE POLICY` — on an RLS table with no policies it is the difference between seeing nothing and seeing something. `AS RESTRICTIVE` is the narrowing form and **is** flagged.
- `ALTER DEFAULT PRIVILEGES` — governs objects created later; it cannot change what the deployed code reads today.

## The report names the right thing

"Changes the schema" is false for an access hazard, so the message is now one of `changes the schema`, `changes who may read the schema`, or `changes the schema and who may read it`.

The guidance for an access hazard adds the fact its reader most needs:

> NOTE THAT CI CANNOT CHECK THIS ONE FOR YOU: the migration job connects as postgres, and a table's owner bypasses row-level security, so a test that applies this and reads a row passes here whatever production does.

That is not a caveat, it is the reason these rules are static. `.github/workflows/_migration.yaml` sets `POSTGRES_USER: postgres` and both `DATABASE_URL` and `DIRECT_URL` to that role, so CI is structurally blind to this class. There is no dynamic check to fall back on; the classification **is** the check. *(Established by Claude L3 in the issue thread.)*

## The premise I could not settle, stated rather than assumed

Whether `ENABLE` alone is survivable turns on which role the API connects as in production. **That is recorded nowhere in this repository** — `apps/backend/.env.example` names 64 keys and not one is a database connection variable. So it cannot be settled from a diff, which is precisely the case the written declaration exists for: the migrations already doing this assert *"the API connects as the owning role"* in prose, and this asks for that assertion in a form a reviewer is shown before approving.

If that assertion is ever found to be wrong, the eleven existing `ENABLE` statements are a live problem and this PR is the smaller half. I could not determine which, and I would rather say so than pick the convenient branch.

## Blast radius, measured

Over all **166** migrations in the repo: **11** carry an access hazard, all of them `ENABLE ROW LEVEL SECURITY`, none declared. There is no `FORCE`, `REVOKE` or `OWNER TO` anywhere in the history.

The gate only ever sees migrations **added** by a pull request (`--diff-filter=A` in `_migration.yaml:153`), so this is prospective and reds nothing already merged.

## Mutations

Nine, through an assert-uniqueness harness, each reverted before the next, harness refused none, baseline and restore both `45 pass / 0 fail`:

| Mutation | Result |
|---|---|
| drop the `NO FORCE` guard | **1 failed** |
| widen `ENABLE` to any `ROW LEVEL SECURITY` statement | **2 failed** |
| `REVOKE` rule also matches `GRANT` | **2 failed** |
| policy rule fires on any `CREATE POLICY` | **1 failed** |
| remove the `OWNER TO` rule | **1 failed** |
| remove the `DROP ROLE` / `NOLOGIN` rule | **1 failed** |
| tag everything as an access hazard | **1 failed** |
| tag nothing, so the report cannot tell them apart | **1 failed** |
| flag `ALTER DEFAULT PRIVILEGES` too | **1 failed** |

**One guard came out because no mutation could redden it.** `DISABLE` does not contain `ENABLE` as a substring, so the plain match was already exact and the guard was dead code — while `NO FORCE` *does* contain `FORCE`, and its guard is proven load-bearing by the first row. The asymmetry is now a comment in the file rather than a symmetric pair of guards, one of which checked nothing.

## Gates

| Command | Result |
|---|---|
| `pnpm run test:scripts` (what `_core.yaml` runs) | **exit 0** — 300 tests, 0 failed |
| `node --test scripts/ci/classify-migration.test.mjs` | exit 0 — 45 pass |
| `turbo run lint --filter=backend --force` | exit 0 — 0 errors, 25 pre-existing warnings |
| `scripts/deploy/tests/migrate.test.sh` | exit 0 — 41/41 |
| `prettier --check` on both files | clean |

### DIFF
--- a/scripts/ci/classify-migration.mjs
+++ b/scripts/ci/classify-migration.mjs
@@ -257,8 +257,17 @@ const addColumnClauses = (upper) => upper.split('ADD COLUMN').slice(1);
  *                                    flagged.
  *   ALTER DEFAULT PRIVILEGES       - governs objects created later. It cannot
  *                                    change what the deployed code reads today.
+ *
+ * Those four are safe by PostgreSQL's own semantics. One more is excluded on a
+ * weaker footing and the difference is worth keeping visible:
+ *
  *   SECURITY LABEL                 - does nothing without a label provider, and
- *                                    none is configured on this database.
+ *                                    none is configured on this database. That
+ *                                    is a fact about this deployment with
+ *                                    nothing in this repository pinning it -
+ *                                    the same footing as the premise under the
+ *                                    ENABLE rule, not the same as the four
+ *                                    above. Raised by ankit-yc on #2731.
  *
  * These read the same normalised statement text as the rules above, so they
  * inherit that machinery exactly: a hazard word in a comment is not a hazard,
@@ -286,7 +295,12 @@ const ACCESS_RULES = [
   {
     dimension: 'access',
     kind: 'changes an object owner',
-    test: (u) => /\bOWNER TO\b/.test(u),
+    // REASSIGN OWNED BY is OWNER TO applied to every object a role owns, in one
+    // statement - and on this database that is the single statement that takes
+    // all eleven row-level-security tables out of the API's sight at once,
+    // because every one of them rests on the owner bypass. Its sibling in the
+    // same command family, DROP OWNED BY, is already below.
+    test: (u) => /\bOWNER TO\b/.test(u) || u.includes('REASSIGN OWNED BY'),
   },
   {
     dimension: 'access',
@@ -304,7 +318,12 @@ const ACCESS_RULES = [
   {
     dimension: 'access',
     kind: 'removes a role or its ability to connect',
-    test: (u) => u.includes('DROP ROLE') || (u.includes('ALTER ROLE') && u.includes('NOLOGIN')),
+    // CONNECTION LIMIT 0 denies every new connection as completely as NOLOGIN
+    // does. Anchored on the zero: a positive limit is a cap, and -1 is the
+    // default meaning no limit at all.
+    test: (u) =>
+      u.includes('DROP ROLE') ||
+      (u.includes('ALTER ROLE') && (u.includes('NOLOGIN') || /\bCONNECTION LIMIT 0\b/.test(u))),
   },
 ];
 
@@ -332,6 +351,13 @@ const RULES = [
   {
     kind: 'adds a required column with no default',
     test: (u) => addColumnClauses(u).some((c) => c.includes('NOT NULL') && !c.includes('DEFAULT')),
+  },
+  {
+    // Not an access change - the object still exists and the grants are
+    // unchanged - but a deployed query naming it unqualified stops resolving,
+    // which is the same break by a different route.
+    kind: 'moves an object to another schema',
+    test: (u) => u.includes('SET SCHEMA'),
   },
   ...ACCESS_RULES,
 ];
--- a/scripts/ci/classify-migration.mjs
+++ b/scripts/ci/classify-migration.mjs
@@ -291,7 +291,18 @@ const ACCESS_RULES = [
     test: (u) =>
       u.includes('FORCE ROW LEVEL SECURITY') && !u.includes('NO FORCE ROW LEVEL SECURITY'),
   },
-  { dimension: 'access', kind: 'revokes a privilege', test: (u) => /\bREVOKE\b/.test(u) },
+  {
+    dimension: 'access',
+    kind: 'revokes a privilege',
+    // Except inside ALTER DEFAULT PRIVILEGES, which is excluded below and was
+    // being flagged here anyway: that statement governs objects created later,
+    // so it cannot change what the deployed code reads today whichever verb it
+    // carries. The exclusion comment said so and the rule did not - and the
+    // test written for it used the GRANT form, which this rule was never going
+    // to match, so it agreed with the comment while the code disagreed.
+    // Raised by the Aikido review of #2731.
+    test: (u) => /\bREVOKE\b/.test(u) && !u.includes('ALTER DEFAULT PRIVILEGES'),
+  },
   {
     dimension: 'access',
     kind: 'changes an object owner',
--- a/scripts/ci/classify-migration.mjs
+++ b/scripts/ci/classify-migration.mjs
@@ -257,6 +257,14 @@ const addColumnClauses = (upper) => upper.split('ADD COLUMN').slice(1);
  *                                    flagged.
  *   ALTER DEFAULT PRIVILEGES       - governs objects created later. It cannot
  *                                    change what the deployed code reads today.
+ *   SECURITY LABEL                 - does nothing without a label provider, and
+ *                                    none is configured on this database.
+ *
+ * These read the same normalised statement text as the rules above, so they
+ * inherit that machinery exactly: a hazard word in a comment is not a hazard,
+ * and one inside a string literal is - because a DO block's DDL lives in a
+ * literal, which is the trade stripSqlComments documents. Pinned both ways in
+ * the tests so an access rule cannot quietly diverge from a shape rule.
  */
 const ACCESS_RULES = [
   {
--- a/scripts/ci/classify-migration.mjs
+++ b/scripts/ci/classify-migration.mjs
@@ -321,9 +321,16 @@ const ACCESS_RULES = [
     // CONNECTION LIMIT 0 denies every new connection as completely as NOLOGIN
     // does. Anchored on the zero: a positive limit is a cap, and -1 is the
     // default meaning no limit at all.
+    //
+    // ROLE|USER because they are exact aliases in PostgreSQL rather than near
+    // synonyms - ALTER USER and DROP USER are the same statements - so matching
+    // one spelling is matching half the language. `0+` for the same reason a
+    // word boundary alone is not enough: CONNECTION LIMIT 00 is still zero.
+    // Both raised by ankit-yc on #2731.
     test: (u) =>
-      u.includes('DROP ROLE') ||
-      (u.includes('ALTER ROLE') && (u.includes('NOLOGIN') || /\bCONNECTION LIMIT 0\b/.test(u))),
+      /\bDROP (ROLE|USER)\b/.test(u) ||
+      (/\bALTER (ROLE|USER)\b/.test(u) &&
+        (u.includes('NOLOGIN') || /\bCONNECTION LIMIT 0+\b/.test(u))),
   },
 ];
 
--- a/scripts/ci/classify-migration.mjs
+++ b/scripts/ci/classify-migration.mjs
@@ -217,6 +217,89 @@ const addColumnClauses = (upper) => upper.split('ADD COLUMN').slice(1);
  * ALTER TYPE ... RENAME TO is here: Prisma sends enum values with an explicit
  * cast to the type name, so a deployed client keeps naming the old type.
  */
+/**
+ * The second way a migration breaks deployed code, and the one the rules above
+ * cannot see: it leaves every object exactly as it was and changes who may read
+ * it.
+ *
+ * `ALTER TABLE "X" FORCE ROW LEVEL SECURITY` alters no column, drops nothing and
+ * renames nothing, so every rule above passes it as additive. On a table with no
+ * policies it takes every row out of the running code's sight at the moment it
+ * commits - before the cutover, and permanently if the deploy then stops. That
+ * is precisely the failure this gate exists to stop, arriving through a door it
+ * was not watching.
+ *
+ * CI CANNOT CATCH THESE FOR YOU, WHICH IS WHY THEY ARE HERE. The migration
+ * workflow connects as `postgres`, and a table's owner bypasses row-level
+ * security - so a test that applies one of these and then reads a row passes in
+ * CI whatever it would do in production. A static rule and a written declaration
+ * are the only checks available; there is no dynamic one to fall back on.
+ *
+ * Whether ENABLE alone is survivable turns on which role the API connects as,
+ * and that is recorded nowhere in this repository - `apps/backend/.env.example`
+ * names no database connection variable at all. So it cannot be settled from the
+ * diff, which is exactly the case the declaration is for: the migrations already
+ * doing this assert "the API connects as the owning role" in prose, and this
+ * asks for that assertion in a form a reviewer is shown.
+ *
+ * Deliberately NOT here, because each of these widens access or touches nothing
+ * that exists:
+ *   GRANT                          - adds a privilege.
+ *   DISABLE / NO FORCE ROW LEVEL   - restores the bypass rather than removing it.
+ *     SECURITY                       Both contain a flagged phrase as a
+ *                                    substring, so the tests below anchor on the
+ *                                    negating word rather than searching for the
+ *                                    positive one.
+ *   CREATE POLICY, permissive      - on an RLS table with no policies, a
+ *                                    permissive policy is the difference between
+ *                                    seeing nothing and seeing something. AS
+ *                                    RESTRICTIVE is the narrowing form and IS
+ *                                    flagged.
+ *   ALTER DEFAULT PRIVILEGES       - governs objects created later. It cannot
+ *                                    change what the deployed code reads today.
+ */
+const ACCESS_RULES = [
+  {
+    dimension: 'access',
+    kind: 'enables row-level security',
+    // No guard against DISABLE, deliberately: "DISABLE" does not contain
+    // "ENABLE" as a substring, so the plain match is already exact. NO FORCE
+    // below DOES contain FORCE and therefore needs one - the asymmetry is real
+    // and a guard here would be a branch no mutation could redden.
+    test: (u) => u.includes('ENABLE ROW LEVEL SECURITY'),
+  },
+  {
+    dimension: 'access',
+    kind: 'removes the owner bypass on row-level security',
+    test: (u) =>
+      u.includes('FORCE ROW LEVEL SECURITY') && !u.includes('NO FORCE ROW LEVEL SECURITY'),
+  },
+  { dimension: 'access', kind: 'revokes a privilege', test: (u) => /\bREVOKE\b/.test(u) },
+  {
+    dimension: 'access',
+    kind: 'changes an object owner',
+    test: (u) => /\bOWNER TO\b/.test(u),
+  },
+  {
+    dimension: 'access',
+    kind: 'narrows or removes a row-level policy',
+    test: (u) =>
+      u.includes('DROP POLICY') ||
+      u.includes('ALTER POLICY') ||
+      (u.includes('CREATE POLICY') && u.includes('AS RESTRICTIVE')),
+  },
+  {
+    dimension: 'access',
+    kind: 'drops objects owned by a role',
+    test: (u) => u.includes('DROP OWNED BY'),
+  },
+  {
+    dimension: 'access',
+    kind: 'removes a role or its ability to connect',
+    test: (u) => u.includes('DROP ROLE') || (u.includes('ALTER ROLE') && u.includes('NOLOGIN')),
+  },
+];
+
 const RULES = [
   { kind: 'drops a table', test: (u) => u.includes('DROP TABLE') },
   { kind: 'drops a column', test: (u) => u.includes('DROP COLUMN') },
@@ -242,6 +325,7 @@ const RULES = [
     kind: 'adds a required column with no default',
     test: (u) => addColumnClauses(u).some((c) => c.includes('NOT NULL') && !c.includes('DEFAULT')),
   },
+  ...ACCESS_RULES,
 ];
 
 /**
@@ -255,7 +339,11 @@ export const classifyMigrationSql = (sql) => {
     const upper = normalise(statement);
     for (const rule of RULES) {
       if (rule.test(upper)) {
-        hazards.push({ kind: rule.kind, statement: statement.replace(/\s+/g, ' ').trim() });
+        hazards.push({
+          kind: rule.kind,
+          dimension: rule.dimension ?? 'shape',
+          statement: statement.replace(/\s+/g, ' ').trim(),
+        });
       }
     }
   }
@@ -365,8 +453,17 @@ const main = (paths) => {
     }
 
     failed += 1;
+    // "the schema" is wrong for an access hazard: FORCE ROW LEVEL SECURITY
+    // alters nothing about the schema and takes every row out of the running
+    // code's sight anyway. Naming the right one is what tells the reader which
+    // of the two paragraphs below applies to them.
+    const changes = review.hazards.some((h) => h.dimension === 'access')
+      ? review.hazards.every((h) => h.dimension === 'access')
+        ? 'changes who may read the schema'
+        : 'changes the schema and who may read it'
+      : 'changes the schema';
     console.log(
-      `::error file=${path}::${review.name} changes the schema in a way the deployed code may not survive.`
+      `::error file=${path}::${review.name} ${changes} in a way the deployed code may not survive.`
     );
     console.log(listed);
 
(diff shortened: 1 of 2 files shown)


ankit-yc (dismissed):
Approve at `777b59116`. Reviewed as `ankit-yc` through the helper.

Everything in the description reproduces, and I went looking for the failure a
rule-based classifier is most likely to have — a statement that means the same
thing and does not match — rather than re-running the table you already
published.

## The seven fire, the four deliberate non-flags pass, and the adversarial cases

Driven through `classifyMigrationSql` directly:

| input | result |
|---|---|
| `ALTER TABLE "X"` / newline / `ENABLE ROW LEVEL` / newline / `SECURITY` | **flag** — `normalise` collapses it |
| `ALTER TABLE public."X" ENABLE ROW LEVEL SECURITY` | **flag** — schema qualifier is no escape |
| `DISABLE ROW LEVEL SECURITY` | pass |
| `NO FORCE ROW LEVEL SECURITY` | pass |
| `GRANT SELECT … TO app` | pass |
| `CREATE POLICY … USING (true)` (permissive) | pass |
| `CREATE POLICY … AS RESTRICTIVE …` | **flag** |
| `-- ALTER TABLE "X" FORCE ROW LEVEL SECURITY` (comment only) | pass — comments stripped |
| `DO $$ BEGIN EXECUTE 'ALTER TABLE x FORCE ROW LEVEL SECURITY'; END $$` | **flag** |
| `SECURITY LABEL FOR selinux ON TABLE x …` | pass |
| `ADD COLUMN` / `DROP COLUMN` controls | pass / **flag `shape:`** |

The `DO $$` row is the right answer reached by substring rather than by
understanding — the phrase survives into the dollar-quoted body, so it matches.
Worth knowing it is not a general guarantee about dynamic SQL, and not worth a
rule.

## Blast radius, re-derived rather than re-read

```
migrations scanned : 166
with ACCESS hazards:  11   all `enables row-level security`
with SHAPE hazards :  10
```

Your 11 to the file, including `20260905130000_lab_result_quarantine`, which
arrived on `dev` after the count I made on #2724. No `FORCE`, `REVOKE` or
`OWNER TO` anywhere in the history, so `--diff-filter=A` means this reds nothing
already merged — and that is the number that makes the rule shippable rather
than a 166-file cleanup.

## Mutations

Nine, assert-uniqueness harness, each reverted before the next, the harness
refused none, baseline and restore both **45/45**:

| Mutation | Result |
|---|---|
| drop the `NO FORCE` guard | 44 pass, **1 fail** |
| `ENABLE` rule always false | 43 pass, **2 fail** |
| `FORCE` rule always false | 43 pass, **2 fail** |
| `REVOKE` rule always false | 44 pass, **1 fail** |
| `OWNER TO` rule always false | 44 pass, **1 fail** |
| restrictive-policy match widened to any `CREATE POLICY` | 44 pass, **1 fail** |
| `DROP OWNED BY` always false | 44 pass, **1 fail** |
| `DROP ROLE`/`NOLOGIN` always false | 44 pass, **1 fail** |
| `dimension` default flipped `'shape'` → `'access'` | 44 pass, **1 fail** |

Every rule is individually pinned, the `NO FORCE` guard is load-bearing, and the
widening mutation confirms the permissive/restrictive split is a real
distinction rather than a comment.

The test count cross-checks your CI number without my installing anything:
36 tests on `dev`, 45 here, and you reported 291 → 300 for `test:scripts`.

## One non-blocking note, and it is not a regression

A hazard phrase inside a string literal flags:

```
INSERT INTO t(note) VALUES ('FORCE ROW LEVEL SECURITY');   -> flag
```

I checked whether this is new before raising it, and it is not — `dev`'s own
shape rules do the same:

```
INSERT INTO t(note) VALUES ('DROP COLUMN x');              -> flag   (on dev)
```

So it is consistent with the file rather than something this PR introduces, and
the failure direction is the safe one: a false positive costs a
`-- deployed-code-survives:` line, and the declaration is the escape hatch that
already exists. Not worth a change here.

## What I did not run

`pnpm run test:scripts` and `eslint` — no `node_modules` in this review
worktree, and I would rather say so than install and report a number that is
really about my machine. I ran `node --test scripts/ci/classify-migration.test.mjs`
directly on both `777b59116` and `origin/dev`, which is the substantive half,
and the +9 matches your CI delta.

## The part I would keep even if the rules changed

Putting *the migration job connects as `postgres`, and a table's owner bypasses
row-level security, so CI cannot check this one for you* into the gate's own
output. A reader hitting this error will otherwise go looking for the dynamic
test that would have caught it, and there cannot be one. That sentence is the
argument for the rule existing, and it is in the place where it is needed rather
than in a PR nobody reads twice.

ankit-yc (commented):
Follow-up on my approve, two parts: a disclosure, and the answer to the question
you left me.

## Disclosure: my approve landed on `74e0f40c7`, and my review body describes `777b59116`

The push arrived while I was writing it, so GitHub recorded the review against a
head I had not read — `reviewDecision` is APPROVED at `74e0f40c7`. I have gone
back and checked the delta rather than leaving that standing:

```
git diff 777b59116 74e0f40c7
  scripts/ci/classify-migration.mjs        +8    ALL inside one block comment
  scripts/ci/classify-migration.test.mjs  +63    new assertions
```

No production line changed, so the mutation table and the adversarial probe in
my review carry. Re-run at `74e0f40c7` anyway: **50/50 exit 0**, blast radius
still **166 scanned / 11 access / 10 shape**, and all nine mutations still red —
three of them redder than before, because the new assertions overlap them.

## The answer: seven is the right *shape* and not quite the right *seven*

You asked whether the four deliberate non-flags are genuinely safe rather than
merely unmatched, and whether seven rules are the right seven. Taking the
dimension seriously rather than the spellings, three statements narrow access
and are **not flagged**:

```
REASSIGN OWNED BY app_role TO other_role;   -> NOT FLAGGED
ALTER ROLE app_role CONNECTION LIMIT 0;     -> NOT FLAGGED
ALTER TABLE "X" SET SCHEMA hidden;          -> NOT FLAGGED
```

**`REASSIGN OWNED BY` is the one I would take.** It means exactly what
`OWNER TO` means — and `OWNER TO` is rule four — except it does it to *every*
object a role owns in one statement. Its sibling in the same Postgres command
family, `DROP OWNED BY`, is already rule six. On this database that is the
single statement that would take all eleven RLS tables out of the API's sight at
once, because every one of them depends on the owner bypass.

`CONNECTION LIMIT 0` is the adjacent case to the `NOLOGIN` rule and denies every
new connection. `SET SCHEMA` is arguably shape rather than access — it moves a
table out of the deployed `search_path` — and I mention it for completeness
rather than as an ask.

**None of this blocks.** The PR strictly improves a gate that today classifies
all three as `additive only` along with everything else, and a rule you add later
costs nothing extra because the dimension already exists. That is the argument
for the dimension doing real work: the fix for what I just found is one entry in
`ACCESS_RULES`, not a new concept.

## The four non-flags, one at a time

- `GRANT` — widens. A deployed reader cannot break because it gained a
  privilege. **Safe.**
- `DISABLE` / `NO FORCE ROW LEVEL SECURITY` — restores the bypass. **Safe.**
- permissive `CREATE POLICY` — permissive policies are OR-ed, so adding one can
  only widen what is visible. **Safe**, and the `AS RESTRICTIVE` split is the
  right place to cut.
- `ALTER DEFAULT PRIVILEGES` — future objects only. **Safe.**
- `SECURITY LABEL` — safe **given that no label provider is configured**, and
  that is an environment fact with nothing in this repository pinning it. It is
  the same class as the premise underneath the `ENABLE` rule: which role the API
  connects as is asserted in migration comments and recorded nowhere. Worth
  saying out loud that one of the five non-flags is contingent on configuration
  rather than on SQL semantics — the other four are true of Postgres itself.

## String literals

Confirmed pre-existing rather than introduced, the way you have it: a hazard
word inside a literal fires on `dev`'s shape rules too, comments are stripped on
both, and `stripSqlComments` documents why the literal case has to stay. Pairing
the assertions so an access rule cannot diverge from a shape rule is the right
way to hold it.

ankit-yc (dismissed):
Approve at `91172aadb`. All three findings are closed and each is pinned by a
mutation. Then I ran the same method one level down on the rules you just added,
and it finds the same class of gap: **`ALTER USER` and `DROP USER` are exact
PostgreSQL synonyms for `ALTER ROLE` and `DROP ROLE`, and the role rule matches
only the `ROLE` spelling.**

## The three are fixed and pinned

```
REASSIGN OWNED BY app_role TO other_role;   -> access:changes an object owner
ALTER ROLE app_role CONNECTION LIMIT 0;     -> access:removes a role or its ability to connect
ALTER TABLE "X" SET SCHEMA hidden;          -> shape:moves an object to another schema
ALTER ROLE app_role CONNECTION LIMIT 10;    -> not flagged
ALTER ROLE app_role CONNECTION LIMIT -1;    -> not flagged
```

Seven mutations, assert-uniqueness harness, each reverted before the next, the
harness refused none, baseline and restore both **53/53**:

| Mutation | Result |
|---|---|
| drop `REASSIGN OWNED BY` from the owner rule | 52 pass, **1 fail** |
| drop the `CONNECTION LIMIT 0` clause | 52 pass, **1 fail** |
| `CONNECTION LIMIT 0` widened to any limit | 52 pass, **1 fail** |
| `SET SCHEMA` rule always false | 52 pass, **1 fail** |
| `SET SCHEMA` tagged `access` instead of `shape` | 52 pass, **1 fail** |
| drop the `NO FORCE` guard | 52 pass, **1 fail** |
| `ENABLE` rule always false | 50 pass, **3 fail** |

The fifth row is the one worth having: the `shape` tag on `SET SCHEMA` is a
judgement you argued for in the comment, and it is now a thing a test holds
rather than a thing the comment asserts. Blast radius re-measured: **166 scanned,
11 access, 10 shape**, unchanged.

## The gap the same method finds in the new rules

`ALTER USER` **is** `ALTER ROLE` in PostgreSQL, and `DROP USER` **is**
`DROP ROLE` — exact aliases, not near-synonyms. The rule requires the literal
`ALTER ROLE` / `DROP ROLE`:

```
ALTER ROLE app_role NOLOGIN;                -> flagged
ALTER USER app_role NOLOGIN;                -> NOT FLAGGED
ALTER ROLE app_role CONNECTION LIMIT 0;     -> flagged
ALTER USER app_role CONNECTION LIMIT 0;     -> NOT FLAGGED
DROP ROLE app_role;                         -> flagged
DROP USER app_role;                         -> NOT FLAGGED
```

Same statement, same effect, other spelling — which is exactly the failure the
dimension exists to avoid, and by your own argument for not filing my last three
it belongs in the rule rather than in an issue. It is one alternation on each
half.

Two smaller ones, neither worth a change on its own:

- `ALTER ROLE app_role CONNECTION LIMIT 00;` is not flagged. `\b` after the `0`
  cannot match when another digit follows. Valid SQL, contrived in a migration.
- `SET SCHEMA 'public';` as a bare session statement flags as *moves an object to
  another schema*. A false positive, and in the safe direction — it costs a
  declaration line, and the session form has no business in a migration anyway.

## The non-flags

Separating `SECURITY LABEL` from the other four is the right cut, and the
wording now says which kind of safe each one is. The four are properties of
PostgreSQL; that one is a property of this deployment, on the same footing as the
premise under the `ENABLE` rule. That is the honest state of it and it is now
visible to a reader who hits the error.

## On the review window

Your framing of the mechanism is the useful part and worth keeping somewhere
durable: `dismiss_stale_reviews_on_push` protects against a push that lands
*before* an approve and does nothing about one that lands *during* it. This time
it dismissed correctly because the push preceded the submit; last time it did
not, and the only thing that caught it was going back to diff the two heads.

ankit-yc (dismissed):
Approve at `2df18fa76`. Both findings closed, both pinned, and I went looking
for the false positive the widening could have introduced rather than only for
the true positives it was meant to add.

## The alternation does not overreach, and this schema is the reason to check

`\bDROP (ROLE|USER)\b` widens a rule to a word that is also an ordinary table
name here — `packages/database/prisma/schema.prisma` has a `model User`. So the
question is not only "does `DROP USER` flag" but "does a migration touching that
table now flag as a role change":

```
ALTER TABLE "User" ADD COLUMN "a" TEXT;      -> not flagged
DROP TABLE "User";                           -> shape:drops a table       (correct)
ALTER TABLE "X" DROP COLUMN "user";          -> shape:drops a column      (correct)
ALTER USER app_role NOLOGIN;                 -> access:removes a role …   (new)
DROP USER app_role;                          -> access:removes a role …   (new)
ALTER USER app_role CONNECTION LIMIT 0;      -> access:removes a role …   (new)
```

The anchoring holds because `DROP` and `ALTER` are required immediately before
the keyword, so `DROP TABLE "USER"` cannot reach the role rule and still lands
on the right one.

`0+` behaves at all four boundaries:

```
CONNECTION LIMIT 0     -> flagged
CONNECTION LIMIT 00    -> flagged        (was not, before this push)
CONNECTION LIMIT 01    -> not flagged    (= 1, a cap)
CONNECTION LIMIT 10    -> not flagged
CONNECTION LIMIT -1    -> not flagged
```

## Mutations

Five, assert-uniqueness harness, each reverted before the next, the harness
refused none, baseline and restore both **54/54**:

| Mutation | Result |
|---|---|
| `DROP (ROLE\|USER)` → `DROP ROLE` only | 53 pass, **1 fail** |
| `ALTER (ROLE\|USER)` → `ALTER ROLE` only | 53 pass, **1 fail** |
| `0+` back to a single `0` | 53 pass, **1 fail** |
| `0+` widened to `\d+` | 52 pass, **2 fail** |
| drop `REASSIGN OWNED BY` from the owner rule | 53 pass, **1 fail** |

The fourth row is the one that makes the third mean something: `0+` is pinned in
both directions, so it can neither shrink back to missing `00` nor widen into
flagging every cap. Blast radius re-measured: **166 scanned, 11 access, 10
shape**, unchanged.

## On the shellcheck convention

Applying the `STAMP_TOKEN` / `POST_CODE_TOKEN` form to the three new `line_of`
arguments rather than inventing a second answer is the right call, and doing it
in a commit that touches nothing else is what makes it reviewable in one glance.
Worth noting the general shape: the fix is a convention the file already
contained, and the reason it was already there is that the same `SC2016` came up
twice on #2725 and #2730 — a third instance answered from the file rather than
from scratch.

ankit-yc (dismissed):
Approve at `551bb37d4`. The Aikido finding is closed in the right direction —
scoped rather than removed — and the new test has the plain case beside the
excluded ones, which is what the old one was missing.

One green mutation, and unlike the last time I raised one I have a distinguishing
input for it rather than a claim that none exists.

## The exclusion behaves in every direction I could put it

```
ALTER DEFAULT PRIVILEGES … GRANT  …        not flagged
ALTER DEFAULT PRIVILEGES … REVOKE …        not flagged   <- the Aikido finding
REVOKE ALL ON "Appointment" FROM PUBLIC    FLAGGED       (correct)
ADP statement, then a plain REVOKE         FLAGGED       (the second statement)
a comment naming ADP above a plain REVOKE  FLAGGED       (comments are stripped)
```

The fourth and fifth rows are the ones that make the exclusion narrow rather than
a hole: it cannot be triggered from another statement, and it cannot be triggered
from a comment.

| Mutation | Result |
|---|---|
| drop the `ALTER DEFAULT PRIVILEGES` exclusion | 53 pass, **1 fail** |
| `REVOKE` rule always false | 50 pass, **4 fail** |
| **exclusion widened from `ALTER DEFAULT PRIVILEGES` to `ALTER`** | 54 pass, **0 fail** |

Baseline and restore both **54/54**.

## The green one, with the input that separates the two

`u.includes('ALTER')` is a plain substring test, so widening the exclusion that
far silently excludes a real `REVOKE` on any object whose name contains "alter":

```
                                             current rule   widened exclusion
REVOKE ALL ON "alter_log" FROM PUBLIC;          FLAG           not flagged
REVOKE ALL ON "altered_records" FROM PUBLIC;    FLAG           not flagged
REVOKE ALL ON "Appointment" FROM PUBLIC;        FLAG           FLAG
```

So this is a **missing test, not dead code** — the distinction the suite is meant
to make about itself. A case asserting that a plain `REVOKE` on an
"alter"-containing object name still flags would pin the exclusion's specificity,
the way `0+` → `\d+` pins the connection-limit rule's.

I raise it in that form deliberately: on #2732 I called a property untestable and
was wrong, so this time I went looking for the input before saying anything about
whether one exists.

## One observation on the DO-block case, where the answer is right by accident

```
DO $$ BEGIN EXECUTE 'ALTER DEFAULT PRIVILEGES … GRANT …';
          EXECUTE 'REVOKE ALL ON x FROM PUBLIC'; END $$;   ->  FLAGGED
```

Correct outcome — a real `REVOKE` inside a `DO` block should flag — but not by
the exclusion reasoning. The statement splitter breaks on the `;` inside the
dollar-quoted body, so the classifier sees
`EXECUTE 'REVOKE ALL ON x FROM PUBLIC'` as its own statement with no
`ALTER DEFAULT PRIVILEGES` in it. I checked which statement was reported rather
than inferring it.

**The direction is safe**: splitting can only produce more fragments, each still
containing its own hazard text, so it can never hide a hazard — it can only break
an exclusion that depends on two strings sharing a statement. Worth knowing
before anyone writes an exclusion that assumes statement boundaries hold for
dollar-quoted bodies.

## Gates

`node --test scripts/ci/classify-migration.test.mjs` **54/54 exit 0**.

## On the fixture that let this through

The old exclusion test used `ALTER DEFAULT PRIVILEGES … GRANT …`, and no rule in
the file matches `GRANT` — so it passed whether or not the exclusion existed.
That is the third instance today of the same shape, and it is the one that says
something the other two did not: **a mutation of the rule would not have found
it either**, because the test never depended on the rule. The input has to be
able to distinguish the two answers before mutating the code can tell you
anything about the test.

ankit-yc (approved):
Approve at `409a5dd6a`. Tests only — the classifier is byte-identical to
`551bb37d4`, so my read of the rules carries. The mutation I raised is now red.

```
exclusion widened from `ALTER DEFAULT PRIVILEGES` to `ALTER`
  at 551bb37d4   54 pass, 0 fail   <- my green
  at 409a5dd6a   53 pass, 1 fail
drop the exclusion entirely        53 pass, 1 fail
REVOKE rule always false           50 pass, 4 fail
```

Baseline and restore both **54/54**. The count is unchanged because the new
assertions live inside an existing `it()` — worth noting so nobody reads
"54 → 54" as nothing having been added.

## One residual green, reported with its boundary rather than as an ask

Shortening the exclusion by one word — `ALTER DEFAULT PRIVILEGES` to
`ALTER DEFAULT` — leaves **54/54**.

I went looking for the distinguishing input the way I did last time, and the
honest answer is that one exists but is a different order of contrived from
`alter_log`: `ALTER DEFAULT` has no meaning in PostgreSQL outside
`ALTER DEFAULT PRIVILEGES` — a column default is `ALTER COLUMN … SET DEFAULT` —
so the only text that separates the two is a **quoted identifier containing a
space**, e.g. `REVOKE ALL ON "alter default" FROM PUBLIC`. Legal, and not a
thing this codebase would ever contain.

**So I am not asking for a test.** A case built on `"alter default"` would pin a
spelling rather than the dimension, which is the failure mode this whole PR
argues against. The useful statement is the boundary itself: *the remaining
unpinned freedom in the exclusion string is the word `PRIVILEGES`, and nothing
realistic distinguishes it because the shorter phrase has no other meaning.*
That is a judgement about PostgreSQL, not a measurement, and I would rather label
it as one than hand you a test that makes it look measured.

## The two additions beyond my input are the ones I would have asked for next

`ADP statement then a plain REVOKE` and `a comment naming ADP above a plain
REVOKE` pin that the exclusion cannot be borrowed across a statement boundary or
out of a comment. Those are the two ways a substring exclusion normally leaks,
and they are now held by tests rather than by the splitter and
`stripSqlComments` happening to behave.

PR #2732, opened 2026-09-05
fix(ci): take the rollback sha from the last verified cutover, not from HEAD

### LINKED ISSUE #2714: A retried deploy reports no incoming migrations, because ROLLBACK_SHA comes from a tree the failed attempt already advanced

aupyay:
Closes #2714.

Written by Claude L3 and delivered as a `git format-patch` patch — `git push` has never worked on that machine, so the author line came through `git am` rather than a push. Applied at `2d145844f`, plus one commit of mine (`f4d2608bf`, test-file quoting only) described at the bottom.

## The defect

`ROLLBACK_SHA` was captured at preflight, before `deploy_git_sync` moves the tree:

```
:77   ROLLBACK_SHA="$(git rev-parse HEAD)"      captured BEFORE the checkout
:94   deploy_git_sync "$REPO_DIR" "$GIT_REF"    the repo moves here
:100  INCOMING_MIGRATIONS="$(deploy_incoming_migrations "$ROLLBACK_SHA" HEAD)"
```

So on any **second run after a failure at or after the checkout**, `ROLLBACK_SHA` equals `HEAD`, `deploy_incoming_migrations` returns empty, `MIGRATIONS_APPLIED` never leaves 0, and **the hazard notice cannot fire while Prisma re-applies the still-pending migrations.**

And the half worth leading with: `$ROLLBACK_SHA` is printed as the rollback target at four sites. On a retry it names **the sha being deployed** — *"NOT cutting over. Rollback sha: X"* where X is X. That is not a missing warning, it is a wrong instruction, printed at the moment someone is deciding what to do.

## The constraint that shaped the fix

`/tmp/api-rollback-$STAMP.txt` already exists and looks exactly like the record a fix would want. **It is written at preflight, unconditionally** — so a failed attempt writes one too, carrying the same stale sha, and "read the newest stamped file" rebuilds the defect inside its own fix, silently. Nothing is written at cutover today.

So: written by `deploy_record_deployed_sha` at `CUTOVER_DONE=1` and nowhere earlier, under a **fixed** name rather than `$STAMP`-keyed, and read **before** `deploy_git_sync`. Pre-checkout `HEAD` is the fallback in the one case where no record exists, which is the first-ever deploy.

## Verified before, not just after

Reverting `deploy_deployed_sha` to plain `git rev-parse HEAD` — the state `dev` is in — turns **4 checks red**. A criterion is only met if it was genuinely unmet, and this is that row, reproduced independently of the author.

| Mutation | Result |
|---|---|
| **`deploy_deployed_sha` reverts to plain `HEAD`** | 52 / **4 failed** |
| the record call removed from `api-deploy.sh` | 55 / **1 failed** |
| trust an unresolvable record | 55 / 1 failed *(author's run)* |
| record appends instead of overwriting | 55 / 1 failed *(author's run)* |
| read the record after the checkout | 55 / 1 failed *(author's run)* |
| record before the cutover is verified | 55 / 1 failed *(author's run)* |
| stamp the record name | 55 / 1 failed *(author's run)* |

## The degraded paths cannot take the deploy with them

`ROLLBACK_SHA="$(deploy_deployed_sha …)"` is a command substitution under `set -euo pipefail`, so a non-zero return at preflight would kill the deploy before it did anything — and the degraded paths are where a new function is most likely to return non-zero. Measured against the real function: no record, a record this repo cannot resolve, and a blank record all return **rc=0** and fall through to `HEAD`.

## shellcheck — the check the author could not run

`shellcheck` is not installed on the authoring machine, so the findings profile against `origin/dev` was explicitly listed as not done. Run here, machine-readable format so the footer is not counted:

```
origin/dev   3 SC2015   1 SC2153   1 SC1091
as received  3 SC2016   3 SC2015   1 SC2153   1 SC1091
after f4d2608bf   3 SC2015   1 SC2153   1 SC1091      <- identical to dev
```

`--severity=warning -x` is exit 0 on both sides regardless, and nothing in CI runs shellcheck at all. The three `SC2016` were single-quoted `line_of` arguments — an expansion that deliberately will not happen — and this file had already answered that twice, in `STAMP_TOKEN` and `POST_CODE_TOKEN`. `f4d2608bf` applies that existing convention and touches nothing else.

## Suites

| Command | Result |
|---|---|
| `scripts/deploy/tests/migrate.test.sh` | exit 0 — **56 passed, 0 failed** (46 on `dev`) |
| `scripts/deploy/tests/git-sync.test.sh` | exit 0 — 16 passed, 0 failed |

## Not verified

No live deploy was exercised. The retry state is reproduced against a real git history inside the suite, not against a box — the same limit every change to this script has had, and it is why the source-position checks in the suite exist.

### DIFF
--- a/scripts/deploy/api-deploy.sh
+++ b/scripts/deploy/api-deploy.sh
@@ -77,7 +77,7 @@ node -v
 # The commit this box is SERVING, which is not the same question as "what is
 # HEAD". Read from the record written at the last verified cutover, before
 # deploy_git_sync moves the tree - see deploy_deployed_sha for why HEAD alone is
-# wrong on every retry, and why the stamped file two lines below is not the
+# wrong on every retry, and why the stamped file written just below is not the
 # record despite looking exactly like it.
 DEPLOYED_SHA_RECORD="/tmp/api-deployed-sha.txt"
 ROLLBACK_SHA="$(deploy_deployed_sha "$DEPLOYED_SHA_RECORD" "$REPO_DIR")"
--- a/scripts/deploy/api-deploy.sh
+++ b/scripts/deploy/api-deploy.sh
@@ -74,7 +74,13 @@ cd "$REPO_DIR"
 
 say "preflight"
 node -v
-ROLLBACK_SHA="$(git rev-parse HEAD)"
+# The commit this box is SERVING, which is not the same question as "what is
+# HEAD". Read from the record written at the last verified cutover, before
+# deploy_git_sync moves the tree - see deploy_deployed_sha for why HEAD alone is
+# wrong on every retry, and why the stamped file two lines below is not the
+# record despite looking exactly like it.
+DEPLOYED_SHA_RECORD="/tmp/api-deployed-sha.txt"
+ROLLBACK_SHA="$(deploy_deployed_sha "$DEPLOYED_SHA_RECORD" "$REPO_DIR")"
 echo "rollback sha: $ROLLBACK_SHA"
 STAMP="$(date +%Y%m%d-%H%M%S)"
 echo "$ROLLBACK_SHA" > "/tmp/api-rollback-$STAMP.txt"
@@ -271,6 +277,12 @@ ss -ltn 2>/dev/null | grep -q ':8080' \
 # Past here the running process IS the new code, so the schema being ahead of it
 # is no longer true and the notice must stop claiming it.
 CUTOVER_DONE=1
+# The new code is now the thing answering, so this is the first moment it is
+# true that the box serves this commit. Recorded here and nowhere earlier: every
+# step above can still fail, and a record written before the cutover would
+# describe a deploy that did not happen - which is the exact defect that made
+# the preflight file unusable as a record.
+deploy_record_deployed_sha "$DEPLOYED_SHA_RECORD" "$(git -C "$REPO_DIR" rev-parse HEAD)"
 
 say "done"
 echo "deployed $(git -C "$REPO_DIR" rev-parse --short HEAD)  (rollback: $ROLLBACK_SHA)"
--- a/scripts/deploy/lib/migrate.sh
+++ b/scripts/deploy/lib/migrate.sh
@@ -49,6 +49,89 @@ deploy_incoming_migrations() {
     | sort
 }
 
+# deploy_deployed_sha <record-file> [repo-dir]
+#
+# The commit this box is actually serving, for use as the rollback target and as
+# the `from` of deploy_incoming_migrations. Echoes a sha; never fails.
+#
+# WHY NOT `git rev-parse HEAD`, which is what this replaced. HEAD is the commit
+# the working tree is on, and after deploy_git_sync that is the commit being
+# DEPLOYED, not the one being served. On a first attempt the two coincide,
+# because HEAD is read before the checkout. On a RETRY after any failure at or
+# after the checkout - install, build, migration, smoke boot, cutover - the repo
+# is already on the target, so HEAD reads back the target and two things break
+# at once:
+#
+#   * deploy_incoming_migrations gets from == to and returns EMPTY, so
+#     MIGRATIONS_APPLIED stays 0 and the schema-hazard notice cannot fire while
+#     prisma is re-applying the still-pending migrations, and
+#   * every "roll back to $ROLLBACK_SHA" message names the sha being deployed,
+#     which is not a missing warning but a wrong instruction, printed at the
+#     moment someone is deciding what to do.
+#
+# WHY NOT /tmp/api-rollback-$STAMP.txt, which already exists and looks exactly
+# like this record. It is written in preflight, unconditionally, before anything
+# can fail - so a failed attempt writes one too, carrying the same stale sha.
+# Reading the newest of those rebuilds the defect inside its own fix, and does
+# it silently. The record has to be written on a VERIFIED CUTOVER and nowhere
+# else, which is what deploy_record_deployed_sha is for.
+#
+# Fixed filename rather than stamped, for the same reason: "the newest of the
+# stamped files" is a sort over a directory that also holds every failure.
+#
+# The recorded sha is re-verified against this repository before it is trusted.
+# A box restored from a backup, re-cloned, or carrying a record from a branch
+# that has since been force-pushed would otherwise hand an unresolvable sha to
+# deploy_incoming_migrations, which fails closed and takes the deploy with it.
+# An unusable record is treated as no record.
+deploy_deployed_sha() {
+  local record="${1:?record file required}"
+  local repo="${2:-.}"
+  local recorded=""
+
+  if [ -r "$record" ]; then
+    recorded="$(tr -d " \t\r\n" < "$record")"
+  fi
+
+  if [ -n "$recorded" ] \
+    && git -C "$repo" rev-parse --verify --quiet "$recorded^{commit}" >/dev/null 2>&1
+  then
+    git -C "$repo" rev-parse "$recorded^{commit}"
+    return 0
+  fi
+
+  # No usable record: a first-ever deploy on this box, or one written before
+  # this function existed. HEAD is correct here and ONLY here, because this is
+  # read before the checkout moves it.
+  git -C "$repo" rev-parse HEAD
+}
+
+# deploy_record_deployed_sha <record-file> <sha>
+#
+# Remember what this box is serving, called once the new code is verifiably the
+# thing answering - not before. Everything upstream of the cutover can still
+# fail, and a record written earlier would describe a deploy that never
+# happened, which is the failure this whole pair exists to end.
+#
+# Written to a temporary file and moved into place so a box killed mid-write
+# keeps the previous record rather than a truncated one. A record that cannot be
+# written is reported and does not fail the deploy: the cutover has already
+# succeeded by this point, and losing the deploy over a note about it would be
+# the tail wagging the dog. The cost is a stale record, which
+# deploy_deployed_sha already treats as no record.
+deploy_record_deployed_sha() {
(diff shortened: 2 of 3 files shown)


ankit-yc (dismissed):
Approve at `f4d2608bf`. The before-row reproduces to the digit, the degraded
paths hold, and the write cannot cost the deploy. Three notes, none blocking,
and one of them is a green mutation on a claim the comment makes.

## The before-row, run here

Reverting `deploy_deployed_sha`'s body to plain `git rev-parse HEAD` — the state
`dev` is in — gives **52 passed, 4 failed**, baseline and restore both **56/56**.
That is the row that makes the other five mean anything, and it is the one I
would have asked for if it were not already there.

## Degraded paths, exercised against the real function

`ROLLBACK_SHA="$(deploy_deployed_sha …)"` is a command substitution under
`set -euo pipefail`, so a non-zero return kills the deploy at preflight. Every
degraded input returns 0 and falls through to `HEAD`:

```
no record at all             rc=0   -> HEAD
blank record                 rc=0   -> HEAD
unresolvable garbage         rc=0   -> HEAD
record path is a directory   rc=0   -> HEAD   (prints `tr: Is a directory` to stderr first)
a real older sha             rc=0   -> that sha
```

And the write half cannot cost the cutover:

```
writable destination         rc=0, record written
unwritable directory         rc=0, warning on stderr, deploy survives
```

## Note 1 — the record is validated as a *revision*, not as a *sha*

`git rev-parse --verify "$recorded^{commit}"` accepts anything git can resolve,
which is wider than the comment's "re-verified against this repository" implies:

```
record contains "dev"      -> resolves to the dev tip, NOT what the box serves
record contains "HEAD~1"   -> resolves relatively
record contains "HEAD"     -> resolves to HEAD (harmless here: at preflight HEAD
                              is still the served commit)
```

Only reachable through a hand-edited or corrupted record, since the only writer
emits a full sha — but the function's own comment anticipates exactly that case
("restored from a backup, re-cloned, force-pushed"). The branch-name value is
the bad one: it resolves to the commit being *deployed*, which is the defect this
PR exists to remove, arriving through the validation that was meant to prevent
it. One line to close — reject anything that is not `[0-9a-f]{7,40}`, or compare
the resolved sha back to the recorded string.

## Note 2 — a green mutation on the atomicity claim

```
`printf > "$record.tmp" && mv -f` replaced by a direct `printf > "$record"`
  -> 56 passed, 0 failed
```

The comment says a box killed mid-write "keeps the previous record rather than a
truncated one", and nothing holds that. It is genuinely awkward to test — it
wants a kill between the two syscalls — so I would not ask for a test. I would
ask for the sentence #2725 already established for this file: name what the
suite does *not* prove, so a reader does not credit 56 green with covering the
atomicity. Same treatment as the `|| true` and append-vs-truncate cases there.

## Note 3 — a directory as the record path leaks `tr: Is a directory`

`[ -r "$record" ]` is true for a directory, so `tr` runs and complains before the
fallback does the right thing. Cosmetic, and the behaviour is correct.

## What I did not verify

A live deploy — stated in the PR body and true. Also worth being precise about
one thing the tests cannot reach: every claim here is about the *record*
mechanism, not about `pm2 restart` having actually replaced the process, which
is what makes the recorded sha true. That gap is inherent, and the cutover check
above it (`ss -ltn … :8080`) is what stands in for it.

## The five mutations

| Mutation | Result |
|---|---|
| revert `deploy_deployed_sha`'s body to plain `HEAD` | 52 pass, **4 fail** |
| revert the call site only | 55 pass, **1 fail** |
| remove the record call at cutover | 55 pass, **1 fail** |
| record *before* the cutover is verified | 55 pass, **1 fail** |
| read the record *after* the checkout | 55 pass, **1 fail** |
| drop the resolvability check | 55 pass, **1 fail** |

none, baseline and restore both 56/56. `git-sync.test.sh` 16/16 exit 0.

The third and fourth rows are the two that matter most and they are the two the
author could most easily have left as prose: *written only after a verified
cutover* and *read before the checkout moves the tree* are the whole design, and
both are held by a test rather than by a comment.

ankit-yc (approved):
Approve at `1049b8fff`. Both findings closed, my green mutation is red, and the
third note is gone too. One residue, which is the honest end of the
revision-not-sha line rather than a new finding.

## My green mutation is now red, and the way it was closed is better than what I asked for

I said the temp-file-and-move "wants a kill between two syscalls" and asked for
the `#2725` treatment — document what the suite does not prove — rather than a
test. That was a failure of imagination about what makes a write fail, and
making `$REC.tmp` a directory is the answer: portable, no kill, no race.

```
temp-and-move replaced by a direct `>`
  at f4d2608bf   56 passed, 0 failed   <- my green
  at 1049b8fff   64 passed, 1 failed
```

**I reasoned from "a crash cannot be scheduled" to "the property cannot be
tested", and those are different claims.** The right question was not how to
schedule the failure but what else produces it.

## The shape guard holds at every boundary

```
record "dev" (a branch)          -> HEAD (fallback)
record "HEAD~1"                  -> HEAD
record "HEAD"                    -> HEAD
6 hex chars                      -> HEAD
5-char prefix git DOES resolve   -> HEAD
7-char prefix (at the floor)     -> the recorded commit
a real 40-char sha               -> the recorded commit
8 hex chars, no such object      -> HEAD
a directory as the record path   -> HEAD, and no `tr: Is a directory` on stderr
```

The last row closes my third note: `[ -f ] && [ -r ]` stops the leak, and the
`-r` you kept alongside `-f` covers an unreadable regular file that `-f` alone
would walk into.

Five mutations, assert-uniqueness harness, each reverted before the next, the
harness refused none, baseline and restore both **65/65**:

| Mutation | Result |
|---|---|
| drop the shape guard entirely | 61 pass, **4 fail** |
| drop only the hex-character half | 62 pass, **3 fail** |
| drop only the minimum-length half | 64 pass, **1 fail** |
| length floor 7 → 4 | 64 pass, **1 fail** |
| `-f` back to `-r` only | 64 pass, **1 fail** |

The middle three are the ones that were green in both first attempts. The floor
mutation is the one I would add to the record: `7 → 4` is red, so the number is
pinned rather than the mere existence of a comparison.

## The residue: an all-hex branch name still resolves

The shape guard narrows ref resolution rather than closing it. A branch whose
name is seven or more hex characters passes the shape check and is then resolved
as a *ref*. Built in a throwaway repository rather than in ours:

```
branch `deadbeef` -> pointing at an older commit
  record "deadbeef"  ->  resolves the BRANCH, not the fallback
branch `abcdefa`
  record "abcdefa"   ->  resolves the BRANCH, not the fallback
```

That needs a corrupted record **and** a branch named in pure hex, so it is the
conjunction of two contrived things and I would not hold a merge for it. It is
worth one line in the comment, because the comment now says the record is
checked for shape and existence, and "existence" is still "anything git resolves
after the shape passes".

The complete close, if it is ever worth it, is to stop describing the value and
compare it instead: resolve, then require the resolved sha to start with the
recorded string. That subsumes both halves of the shape guard and admits exactly
what `deploy_record_deployed_sha` writes.

## On the hex class

`[0-9a-f]` over `[0-9a-fA-F]` is right for the reason this whole finding is
about: the writer emits `git rev-parse` output, which is lower-case, so
accepting upper-case admits a value nothing in this codebase produces. **A guard
should admit what the writer emits and nothing wider** — that is the same
sentence as *validated as a revision rather than as a sha*, one level in.

## Gates at `1049b8fff`

`migrate.test.sh` **65/65 exit 0**, `git-sync.test.sh` 16/16 exit 0.

PR #2734, opened 2026-09-05
fix(backend): log request failures by field rather than by object

aupyay:
The two `axios` call sites in `documenso.service.ts` logged the caught error object:

```ts
} catch (error) {
  logger.error("An unexpected error occurred:", error);
```

**An axios error carries the outbound request on `config`** — `config.headers` and `config.data` — so logging the object writes the request alongside the message. That is more than a failure needs to say, and the fields it adds are the ones a request carries to authenticate itself.

**The logger does not narrow it.** Its production format is `json()` (`NODE_ENV !== "development"`), which serialises the whole meta object; the development format's field list replaces `req`, `res`, `request` and `response` — **but not `config`** — and its own comment says it exists to keep log lines small rather than to redact. Driven through a real winston logger built from that production branch, with a capturing transport, a placeholder value placed on `config.headers` appears in the emitted line.

The four SDK calls in this same file already log `.message` and `.statusCode`. This gives the two axios ones the same treatment through one local helper.

## The helper does not use `axios.isAxiosError`

It reads the status structurally instead. `isAxiosError` is a module export that a test double replaces, so depending on it would make the useful half of the helper untestable in the suite that has to prove the rest of it — the status came back `undefined` under the existing `jest.mock("axios")` until I stopped relying on it.

## The tests assert the property, not the wording

Two new cases drive a failure whose request config carries a placeholder — one on `config.headers` (the download path) and one on `config.data` (the external-auth path, where the credential is in the body rather than a header) — and assert that the **serialised** logger arguments do not contain it. A future change that logs the raw error again reddens them whatever it names the message.

Three existing assertions moved from the old message strings to the new shape.

## Mutations

Four, through an assert-uniqueness harness, each reverted before the next, baseline and restore both `30 passed`:

| Mutation | Result |
|---|---|
| log the raw error again on the download path | **1 failed** |
| log the raw error again on the external-auth path | **1 failed** |
| `describeError` returns the error itself | **5 failed** |
| drop `status` from `describeError` | **2 failed** |

The first two are the before-row: both are exactly what `dev` does today.

## Gates

| Command | Result |
|---|---|
| `npx jest` (full backend) | **exit 0** — 479 suites / **11459 tests** |

## Scope

`documenso.service.ts` is the only file in `apps/backend/src` that both imports `axios` and logs a bare error object. Files using global `fetch` are not affected by the same mechanism — a failing `fetch` throws a `TypeError` that does not carry the request — so this is a property of the client, not of logging a raw error.

### DIFF
--- a/apps/backend/src/services/documenso.service.ts
+++ b/apps/backend/src/services/documenso.service.ts
@@ -92,6 +92,38 @@ export type SignedDocument = {
 
 export type DocumensoExternalRole = "ADMIN" | "MANAGER" | "MEMBER";
 
+/**
+ * What is safe to log about a failed request.
+ *
+ * The raw error must never be logged here. Axios attaches the outbound request
+ * to the error as `config`, so an axios error from this file carries the
+ * Documenso API key in `config.headers.Authorization`, and the one from
+ * `generateExternalRedirectUrl` carries the external auth secret in
+ * `config.data`. The logger does not redact either: its production format is
+ * `json()`, which serialises the whole meta object, and the development
+ * format's field list replaces `req`, `res`, `request` and `response` but not
+ * `config` - it exists to keep log lines small, not to redact.
+ *
+ * So the first time a Documenso call fails is the first time the credential is
+ * written out. The four SDK calls in this file already log `.message` and
+ * `.statusCode`; this is the same treatment for the two axios ones.
+ */
+const describeError = (error: unknown) => {
+  // Read the status structurally rather than through `axios.isAxiosError`: the
+  // status is worth having and that helper is a module export a test double
+  // replaces, so depending on it makes the useful half of this untestable in
+  // the suite that has to prove the rest of it.
+  const response =
+    typeof error === "object" && error !== null && "response" in error
+      ? (error as { response?: { status?: number } }).response
+      : undefined;
+
+  return {
+    message: error instanceof Error ? error.message : String(error),
+    status: response?.status,
+  };
+};
+
 export class DocumensoService {
   static async createDocument({
     pdf,
@@ -241,7 +273,10 @@ export class DocumensoService {
       const signeDocument = downloadResponse.data as SignedDocument;
       return signeDocument;
     } catch (error) {
-      logger.error("An unexpected error occurred:", error);
+      logger.error(
+        "Documenso signed-document download failed",
+        describeError(error),
+      );
     }
   }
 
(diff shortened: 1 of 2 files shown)


ankit-yc (approved):
Approve at `2e60ee1ea`. I did not check that the code logs `.message` and
`.status` — that is readable. I drove the failure and asserted the placeholder is
absent from what gets emitted, which is the method that found the leak, and then
mutated the fix back to what `dev` does today.

## The property, both paths

Two axios-shaped failures carrying a placeholder in the two places the two call
sites put one, through `describeError` and serialised the way `json()` would:

```
headers path   -> {"message":"Request failed with status code 401","status":401}
                  placeholder present?  no
body path      -> {"message":"Request failed with status code 500","status":500}
                  placeholder present?  no
dev today (the raw error)              placeholder present?  YES
```

The second row is the one worth having separately: `generateExternalRedirectUrl`
puts its secret in the **body**, so axios hangs it on `config.data` rather than
`config.headers`, and a fix that only reasoned about headers would have left it.

`.message` cannot carry either credential here — both call sites take their
secret from a header or a body and neither builds it into the URL, so an axios
message (`Request failed with status code …`) has nowhere to pick one up.

## Mutations

Baseline and restore both **30/30**, assert-uniqueness harness, each reverted
before the next, the harness refused none:

| Mutation | Result |
|---|---|
| download site logs the raw error again (what `dev` does) | 27 pass, **3 fail** |
| external-auth site logs the raw error again | 28 pass, **2 fail** |
| `describeError` returns the error rather than its message | 25 pass, **5 fail** |
| **`describeError` also returns `config`** | 28 pass, **2 fail** |

The last row is the one that proves the tests assert the *absence of the
credential* rather than the shape of the message: putting `config` back on the
returned object is a change no wording assertion would notice, and it is red.

## The design call I would have got wrong

Reading the status structurally instead of through `axios.isAxiosError`, because
that helper is a module export the file's existing `jest.mock("axios")` replaces
— so depending on it makes `status` come back `undefined` **in the only suite
that can prove the rest of this**. I would have reached for the idiomatic guard
and shipped two tests passing on a helper that returned nothing useful.

Worth noting the general form, since it is not really about axios: **a guard
imported from the module you are mocking is not available to the test that mocks
it.** Structural checks survive the double; identity checks do not.

## Scope

Not verified: whether either call has ever actually failed in the deployed
environment. That is one of the box questions and it is unchanged by this PR —
this closes the path, it says nothing about the past.

And keeping the PR body generic is right: a PR is as world-readable as an issue,
so the reasoning for keeping the finding off GitHub applies to the fix's
description too.

PR #2735, opened 2026-09-05
fix(ci): compare the deployed-sha record instead of describing it

### LINKED ISSUE #2733: deploy_deployed_sha validates the record by describing it, so an all-hex branch name still resolves as a ref
Follow-up to #2714 / #2732, non-blocking. Filed so the finding outlives the
thread that produced it — it was verified twice, by two agents, on two machines,
and nobody had written it down.

## What is wrong

`deploy_deployed_sha` in `scripts/deploy/lib/migrate.sh` validates the recorded
sha by **describing** it — hex characters only, at least seven of them — and then
asks git to resolve it. The shape guard narrows ref resolution but does not close
it: a branch whose name is seven or more pure hex characters passes the
description and is then resolved as a **ref**.

Built in a throwaway repository, against the landed guard:

```
branch `deadbeef` -> an older commit
  record "deadbeef"    ->  the BRANCH's commit     <- passes the shape guard
  record "abcdefa"     ->  the BRANCH's commit
  record <full sha>    ->  that commit             (correct)
  record <7-char pfx>  ->  that commit             (correct)
```

That is the same class the shape guard was added to close on #2732: a value that
resolves to the commit being *deployed* rather than the one being *served*, handed
back by the check written to prevent exactly that.

## Reachability, stated plainly

Two contrived things at once: a **corrupted or hand-edited record** (the only
writer, `deploy_record_deployed_sha`, emits `git rev-parse` output) **and** a
branch named in pure hex. That is why this is an issue and not a rider on a PR
that was approved twice over.

The function's own comment does anticipate the first half — "restored from a
backup, re-cloned, or carrying a record from a branch that has since been
force-pushed" — so the population is not imaginary, only narrow.

## The fix

Stop describing the value and compare it: resolve, then require the resolved sha
to **start with** the recorded string.

```
deadbeef / abcdefa   ->  rejected, falls back
dev / HEAD / HEAD~1  ->  rejected, falls back
a 5-char prefix      ->  rejected  (the length floor still does independent work)
a 7-char prefix      ->  accepted
a full 40-char sha   ->  accepted
```

It subsumes the hex half — a non-hex string cannot be a prefix of a sha — while
the length floor keeps earning its place, so the guard goes from describing the
value to comparing it.

**One case survives the fix and is benign**, measured rather than reasoned: a
branch named after its own target's prefix.

```
branch a6cfcdb -> commit a6cfcdbec…
  recorded=a6cfcdb  resolved=a6cfcdbec…  starts-with-recorded=YES  -> accepted
  and the resolved commit is the same one either way
```

The ref and the object agree, so accepting it costs nothing.

## Test notes for whoever takes it

Both previous attempts at tests for this guard were vacuous, independently, in
the same way — every bad input resolved to `HEAD`, which is the value the
fallback returns anyway, so the assertions measured the fallback rather than the
guard. **The fixture has to disagree**: a branch pointing at a commit that is not
`HEAD`, a prefix short enough that git still resolves it, and a name that is
alphabetic-but-not-hex to separate the two halves. Mutating each half
individually is what showed the first versions were green.

aupyay:
Closes #2733. Found by ankit-yc on #2732, after the shape guard it reviews had landed.

## The gap

The shape guard rejects `dev` and `HEAD~1` and accepts `deadbeef`. A branch whose name
is seven or more hex characters passes every character and length check and is then
resolved as a **ref**:

```
branch deadbeef -> an older commit
  record "deadbeef"    ->  the BRANCH's commit     <- passes the shape guard
  record "abcdefabc"   ->  the BRANCH's commit
```

Two contrived things at once — a corrupted record **and** a hex-named branch — which is
why this is a follow-up and was not a merge blocker.

## The change

Resolve the record, then require the resolved sha to begin with what was recorded. The
guard stops describing the value and compares it.

The **hex-character class comes out**. Once the comparison is there, no mutation can
redden it — a name git resolves to something not beginning with it is rejected whatever
characters it holds. Keeping a guard nothing can falsify is how the file accumulates
checks that look like coverage.

The **length floor stays** and is red on its own: it stops a four-character prefix being
resolved and then accepted, which the comparison alone would allow.

**One case is let through and it is pinned as deliberate:** a branch named after its own
target's prefix. The ref and the object agree, so the answer is identical either way — a
coincidence rather than a hole, and asserted so that narrowing it later is a decision.

## Evidence

At `803a6b6b3`, tree clean, HEAD confirmed either side:

| Command | Result |
|---|---|
| `migrate.test.sh` | exit 0 — **68/68** (65 on dev) |
| `git-sync.test.sh` | exit 0 — 16/16 |

| Mutation | Result |
|---|---|
| **revert to resolve-and-trust — the before** | 62 / **6 failed** |
| accept any resolution (drop the comparison) | 62 / **6 failed** |
| drop the length floor | 67 / **1 failed** |
| `-f` back to `-r` | 67 / **1 failed** |

## Not proven, stated rather than omitted

`"$recorded"*` is quoted so a record holding a glob metacharacter is compared literally.
**No test pins that, and the reason is that no input can tell the two spellings apart:**
to reach the comparison a record must first resolve, and a value git resolves is a sha or
a ref name, neither of which can contain a metacharacter. Tests asserting it were written
and deleted when the mutation stayed green. The quoting stays because it is free and
correct, not because it is proven.

`shellcheck` is not installed on the machine this was written on; the findings profile
against `origin/dev` has not been compared.

### DIFF
--- a/scripts/deploy/lib/migrate.sh
+++ b/scripts/deploy/lib/migrate.sh
@@ -101,18 +101,41 @@ deploy_deployed_sha() {
   # returning, arriving through the check meant to prevent it. Only reachable
   # through a corrupted or hand-edited record, which is precisely the case this
   # function's own comment anticipates. Raised by ankit-yc on #2732.
-  case "$recorded" in
-    *[!0-9a-f]* | "") recorded="" ;;
-  esac
   if [ "${#recorded}" -lt 7 ]; then
     recorded=""
   fi
 
-  if [ -n "$recorded" ] \
-    && git -C "$repo" rev-parse --verify --quiet "$recorded^{commit}" >/dev/null 2>&1
-  then
-    git -C "$repo" rev-parse "$recorded^{commit}"
-    return 0
+  # Shape is necessary and not sufficient: a BRANCH whose name is seven or more
+  # hex characters passes every check above and is then resolved as a ref, so
+  # `deadbeef` comes back as whatever that branch points at. Raised by ankit-yc
+  # on #2733 after the shape guard landed.
+  #
+  # So the value is COMPARED rather than described: resolve it, then require the
+  # resolved sha to begin with what was recorded. That subsumes the hex-character
+  # class this replaced - a name git resolves to something not beginning with it
+  # is rejected whatever characters it holds, and no mutation could redden the
+  # class once the comparison was here, which is the definition of dead.
+  #
+  # `"$recorded"*` and not `$recorded*`: the expansion is quoted so a record
+  # holding `*`, `?` or `[a-f]` is compared literally rather than as a pattern.
+  #
+  # NOT PINNED BY THE SUITE, deliberately, and this is the honest reason rather
+  # than an omission: no input can tell the two spellings apart here. To reach
+  # the comparison a record must first RESOLVE, and a value git resolves is a
+  # sha or a ref name, neither of which can contain a glob metacharacter - so
+  # unquoting it is green on every input I could construct. Tests asserting
+  # otherwise were written for this and deleted when the mutation stayed green.
+  # The quoting stays because it is free and correct, not because it is proven. A ref cannot satisfy that
+  # unless it happens to be named after its own target's prefix, in which case
+  # the ref and the object agree and accepting it costs nothing - a coincidence
+  # rather than a hole, and the reason this is a comparison and not a ban on
+  # names that look like shas.
+  if [ -n "$recorded" ]; then
+    local resolved
+    resolved="$(git -C "$repo" rev-parse --verify --quiet "$recorded^{commit}" 2>/dev/null || true)"
+    case "$resolved" in
+      "$recorded"*) printf '%s\n' "$resolved"; return 0 ;;
+    esac
   fi
 
   # No usable record: a first-ever deploy on this box, or one written before
--- a/scripts/deploy/tests/migrate.test.sh
+++ b/scripts/deploy/tests/migrate.test.sh
@@ -689,6 +689,33 @@ for BAD in old-release oldrelease refs/heads/old-release "$(printf '%s' "$M1" | 
     "$(deploy_deployed_sha "$REC" "$REPO")"
 done
 
+# Shape is necessary and not sufficient. A BRANCH whose name is seven or more
+# hex characters passes every shape check and is then resolved as a ref, so the
+# guard that rejects `old-release` accepts `deadbeef`. #2733.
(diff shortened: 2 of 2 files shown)


ankit-yc (approved):
Approve at `b8455d11b`. The residue I filed #2733 for is closed — I built the
hex-named branch again and it no longer resolves.

## The finding is actually closed, not just tested

```
branch `deadbeef` -> an older commit
  before (#2732)   record "deadbeef"  ->  resolved the BRANCH
  now              record "deadbeef"  ->  rejected, falls back to HEAD
```

And every boundary still behaves, run against the real function:

```
dev / HEAD / HEAD~1            -> fallback        5-char prefix          -> fallback
8 hex chars, no such object    -> fallback        7-char prefix          -> the recorded commit
blank / a directory            -> fallback        a real 40-char sha     -> the recorded commit
```

Baseline and restore both **68/68**; `git-sync.test.sh` 16/16.

| Mutation | Result |
|---|---|
| revert to resolve-and-trust (the before-row) | 63 pass, **5 fail** |
| drop the length floor | 67 pass, **1 fail** |
| **unanchor the prefix compare** (`"$recorded"*` → `*"$recorded"*`) | 68 pass, **0 fail** |

## The green, and why I am not asking for a test

I went looking for the input before saying anything about whether one exists,
and I could not construct one — for a structural reason rather than because it
looked hard.

To reach the comparison a record must pass the shape guard (hex, at least seven
characters) **and** resolve. A hex string that resolves is an abbreviated sha,
and git resolves an abbreviation to a sha that *begins* with it — so anchored and
unanchored agree on every input that gets that far. They diverge only for a
record that resolves to a sha containing it **not at the start**, which needs a
hex-named ref whose target's sha happens to embed the name elsewhere. Finding one
is a search over roughly 16^7 candidate shas; I am not going to claim it is
impossible, only that I cannot produce it and neither will an author.

**So the anchor is free and correct rather than proven** — which is exactly the
case you hit in this same file on the quoting, and resolved the same way:
delete-or-don't-add the tests, and say in the comment why no input can prove it.
One line beside the `case`, matching the one already above the shape guard.

## The rebase

Only your two files are touched, and the fix survived the move across #2731's
landing intact. Dropping the hex-character class once the comparison exists is
the right call and it is my own rule from #2731 pointed at your code: a guard
nothing can falsify is how a file accumulates checks that look like coverage.
The length floor stays and is red on its own, which is what says it is still
doing independent work.

Issue #2736, opened 2026-09-05
feat(lib): a shared money helper with minor-unit arithmetic


aupyay:
Split out of #1685 (payment gateway Phase 0), which is 2583 commits behind `dev` and is being closed in favour of three separately-actionable pieces. Assessment by Claude L3, verified independently before filing.

## The gap is live

There is no shared or backend money helper on `dev`. Everything that exists is either frontend-local or a FHIR type:

```
apps/frontend/src/app/lib/money.ts          frontend-local
packages/fhirtypes/src/Money.ts             a FHIR resource type, not arithmetic
```

So any backend code doing currency arithmetic is doing it inline, and the frontend's helper is not reachable from it.

## Why this piece first

It is a **pure addition** with zero conflict surface — no existing file changes — and it does not depend on either of the other two pieces of #1685. Of the 22 files in that PR, 17 were pure additions; this is the part of them that stands entirely on its own.

## What is not being carried over

Not a rebase of #1685. That branch's helper was written against a repository that has moved 2583 commits, and the point of splitting is that this piece can be written fresh against current code in less time than resolving the conflicts of the other two would take.

aupyay:
**Scope note for the promotion: the set is guarded in one direction only, and it is the wrong one.**

#2739 removes `"ugx"` from `ZERO_DECIMAL_CURRENCIES` — Stripe takes UGX as two-decimal despite the ISO exponent, so every UGX charge went out at a hundredth of its value. That is merged/merging separately because it is a charging fix, not a refactor.

What the mutation matrix over the corrected set says about the rest of it, run on `dev`'s `payment.ts` with only #2739's test file present (baseline and restore both 126/126):

| Mutation | Result |
|---|---|
| put `"ugx"` back | 1 fail |
| remove `"jpy"` | 1 fail |
| add `"isk"` | **0 fail — green** |
| add `"gbp"` | **0 fail — green** |

**Two entries of sixteen are pinned. The suite catches *removing* a correct entry and *restoring* the known-wrong one; it does not catch *adding* a wrong one — which is the direction this bug arrived through.**

`isk` is the specific hazard: it carries the identical Stripe backwards-compatibility rule to UGX and is correctly absent, and the comment #2739 adds is currently the only thing between that entry and a re-introduction of the same undercharge with the suite green.

[...]

PR #2739, opened 2026-09-05
fix(backend): submit UGX as a two-decimal amount, as Stripe requires

aupyay:
Refs #2736. Found while scoping that issue; **separable from it on purpose** — this is a
one-line charging fix and #2736 is a refactor across three files and a package boundary.

## The bug, live rather than latent

`ZERO_DECIMAL_CURRENCIES` in `finance/payment.ts` contained `"ugx"`, so
`toStripeMinorUnits` returned the major amount unscaled and **every UGX charge went out at a
hundredth of its value.**

Stripe's Special cases say the opposite of what the ISO 4217 exponent implies:

> UGX became a zero-decimal currency, but backwards compatibility requires you to represent
> it as a two-decimal value, where the decimal amount is always 00. **To charge 5 UGX, provide
> an `amount` value of 500.**

**ISK carries the identical rule and is correctly absent from the set.** The same exception was
applied to one currency and not the other, in the same list, by the same author.

Reachable by configuration, not by code change: `invoice.currency` is `String @default("usd")`
with no enum or allowlist, and `toStripeMinorUnits` is called from five places —
`:821` checkout line items, `:838` discounted unit amounts, `:907` deposits, `:1368` payment
intents, `:1735` refunds.

## Why the existing test could not catch it

There is exactly one test on this mechanism and it uses **JPY** — a currency that must *not* be
scaled. UGX must be. **A test named for one direction cannot fail on the other**, and one
example per class is what let the second entry through.

## Evidence

Run twice, independently — the change was written twice in parallel before either author saw
the other's, and the version here is the one with the better commit
message. Both runs agree. Numbers below are from the commit on this branch, `git rev-parse HEAD`
checked in the same shell before and after, tree clean:

| Command | Result |
|---|---|
| `npx jest` (full backend, not scoped) | **exit 0** — 479 suites / **11460 tests** |
| `turbo run lint type-check --filter=backend --force` | **exit 0** — 7/7, 0 errors, 25 pre-existing warnings |

**The before-row is an observable wrong number, not a missing module.** Mutation-checked on the
exact commit by putting `"ugx"` back into the set:

```
● submits UGX as a two-decimal value, which Stripe requires despite it being zero-decimal
  Expected: 500
  Received: 5
Tests: 1 failed, 1 passed
```

**The 1 passed is the JPY test, green under the same mutation** — which is the point above,
demonstrated rather than argued.

## Not established

The Stripe docs page came back as the Spanish localisation on all three fetches; the Special
cases text is unambiguous but it is a translation. **Someone with a browser should confirm the
UGX row in English before this merges.** Taken on this evidence because the direction of the
error is undercharging and the citation is the repo's own — the URL was in the comment above the
set the whole time.

The zero-decimal *list itself* renders as links on that page and never resolved in any fetch, so
**the remaining fifteen entries are unverified**, and there is no second source inside the repo
to check them against: nothing else encodes currency decimals for settlement.
`apps/frontend/src/app/lib/money.ts` looks like a second implementation and is not one — it is
`Intl.NumberFormat` display formatting with no conversion to minor units, answering "how do I
show this to a human" where this set answers "what integer does Stripe accept". **They must not
be merged.** Stripe's list, ISO 4217 and CLDR give three different answers and only the first is
what these call sites need.

The rest of the zero-decimal set is **unverified against Stripe's current published list** —
this change removes one entry it contradicts and does not audit the other fifteen.

### DIFF
--- a/apps/backend/src/services/finance/payment.ts
+++ b/apps/backend/src/services/finance/payment.ts
@@ -743,7 +743,13 @@ const ZERO_DECIMAL_CURRENCIES = new Set([
   "mga",
   "pyg",
   "rwf",
-  "ugx",
+  // NOT "ugx". Stripe's Special cases: UGX "became a zero-decimal currency, but
+  // backwards compatibility requires you to represent it as a two-decimal value,
+  // where the decimal amount is always 00. To charge 5 UGX, provide an amount
+  // value of 500." It was in this set, so a UGX invoice was submitted at a
+  // hundredth of its value. ISK carries the identical rule and is correctly
+  // absent - the same exception was applied to one and not the other.
+  // https://docs.stripe.com/currencies#special-cases
   "vnd",
   "vuv",
   "xaf",
--- a/apps/backend/test/services/finance.payment.test.ts
+++ b/apps/backend/test/services/finance.payment.test.ts
@@ -1969,6 +1969,61 @@ describe("FinancePaymentService", () => {
     expect(sessionArgs.line_items[0].price_data.unit_amount).toBe(1000);
   });
 
+  // The other half of the same rule, and the half the JPY test cannot reach.
+  // Stripe's Special cases: UGX "became a zero-decimal currency, but backwards
+  // compatibility requires you to represent it as a two-decimal value, where the
+  // decimal amount is always 00. To charge 5 UGX, provide an amount value of
+  // 500." ISK carries the identical rule and is correctly absent from the set;
+  // ugx was in it, so a UGX invoice was submitted at a hundredth of its value.
+  // https://docs.stripe.com/currencies#special-cases
+  it("submits UGX as a two-decimal value, which Stripe requires despite it being zero-decimal", async () => {
+    const stripeClient = {
+      checkout: { sessions: { create: jest.fn(), expire: jest.fn() } },
+      paymentIntents: { create: jest.fn(), retrieve: jest.fn() },
+      refunds: { create: jest.fn() },
+    };
+    __setFinanceStripeClientForTests(stripeClient);
+
+    (prisma.invoice.findUnique as jest.Mock).mockResolvedValueOnce({
+      id: "inv_ugx",
+      totalAmount: 5,
+      currency: "ugx",
+      status: "AWAITING_PAYMENT",
+      paymentCollectionMethod: "PAYMENT_LINK",
+      organisationId: "org_1",
+      appointmentId: "",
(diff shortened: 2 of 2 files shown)


ankit-yc (approved):
Approve at `5bf48f55c`. The before-row reproduces exactly, and I measured how
much of the set this leaves guarded — which turns out to be two entries out of
sixteen, and the unguarded direction is the one the bug arrived through.

## The before-row, run here

Dev's `payment.ts` with this PR's test file, nothing else changed:

```
● submits UGX as a two-decimal value, which Stripe requires despite it being zero-decimal
    Expected: 500
    Received: 5
Tests: 1 failed
```

With the fix applied: **126 passed**, and the JPY test green beside it.

## What the set is guarded by, after this lands

Four mutations, assert-uniqueness harness, each reverted before the next,
baseline and restore both **126/126**:

| Mutation | Result |
|---|---|
| put `"ugx"` back | 125 pass, **1 fail** |
| remove `"jpy"` | 125 pass, **1 fail** |
| **add `"isk"`** | **126 pass, 0 fail** |
| **add `"gbp"`** | **126 pass, 0 fail** |

**Two entries are pinned — `jpy` present and `ugx` absent — and the other
fourteen are not.** More precisely: the set is guarded against *removing* a
correct entry at one point and against *restoring* this wrong one, and **not at
all against adding a wrong entry**, which is the direction this bug arrived
through.

`isk` is the one I would name in the issue rather than the PR. **The comment
this PR adds identifies it as carrying the identical Stripe rule and being
correctly absent** — and adding it tomorrow reintroduces the same undercharge
with the suite green. The comment is the only thing standing between that entry
and the same bug.

**This does not block.** The fix is right, the test is the one that was missing,
and a table-driven case over the whole set is a different change from a
one-line deletion. But *one example per class* was the reason this survived, and
after this PR the class has one example per **direction** — so the same sentence
applies one level up.

## Scope

I did not verify the fifteen remaining codes against Stripe; I have no web
access and there is no second internal source in the repo, so the PR's caveat is
the whole audit surface and it is correctly marked unestablished. **The Spanish
localisation on all three fetches is worth someone confirming in English** —
the direction of the error is undercharging, so the change is safe either way,
but the citation is not yet in the language it quotes.

## The comment

Naming ISK as the twin that was handled correctly is what makes this a finding
rather than a correction. **The same exception was applied to one and not the
other, by the same author, in the same set** — that is internal corroboration
that survives whatever the docs page says today.

PR #2747, opened 2026-09-05
refactor(backend): share one currency-aware Stripe minor-units converter

aupyay:
## The bug this turned out to contain

#2736 was filed as a de-duplication. It is also a second money bug, in the opposite direction to #2739.

`stripe.service.ts` converted an appointment cost with a currency-blind `Math.round(amount * 100)` on the line *immediately above* the one that resolved the organisation's billing currency. A zero-decimal organisation was charged **a hundred times** the service cost.

Before-row, on unmodified dev with only this PR's test file:

```
● submits a zero-decimal appointment cost unscaled
    - Expected: {"amount": 1000,   "currency": "jpy", ...}
    + Received: {"amount": 100000, "currency": "jpy", ...}
```

A ¥1,000 appointment submitted to Stripe as ¥100,000. Nothing caught it because the only existing assertion on that call checks `expect.not.objectContaining({ transfer_data })` — the amount reaching Stripe was never asserted at all.

## What changed

The correct, currency-aware converter already existed: file-private in `services/finance/payment.ts`, five call sites, exported nowhere. That is *why* a second currency-blind one grew beside it. So this promotes the existing one rather than adding a third.

- **New** `src/utils/stripe-minor-units.ts` — the promoted converter and its set.
- `finance/payment.ts` — local copy deleted, imports the shared one. Five call sites unchanged.
- `stripe.service.ts` — `toStripeAmount` deleted, its single call site migrated. The `currency` line moved above the `amount` line; neither depends on the other.

**Placement.** Not `utils/billing.ts`, despite that being the currency module and already imported by this file: it imports `prisma`, and `finance/tax.ts` — the next consumer — is a pure calculation module with no database dependency. A number conversion should not drag the database into it.

**Naming.** `stripe-minor-units`, not `money`. The frontend's `money.ts` is `Intl` display formatting with **zero** conversions to minor units; it answers "how do I show this to a human" where this answers "what integer does Stripe accept". They look like duplicates and are not, and a future "we have two money helpers, let us merge them" is the plausible next mistake. The name makes it hard to make by accident. This helper must not be folded into that one.

## Guarding the direction the bug came from

The set was pinned against *removing* a correct entry and not against *adding* a wrong one — which is how both UGX and this class arrive. Measured: adding `isk` or `gbp` left the suite fully green.

This adds an explicit **ISK** case, which carries the identical Stripe special-case rule to UGX and was previously protected only by a comment. Mutations against the new unit tests:

| Mutation | Result |
|---|---|
| always `×100` (kill the zero-decimal branch) | **2 failed** |
| put `"ugx"` back | **1 failed** |
| **add `"isk"`** | **1 failed** (was green before this PR) |
| drop `.trim().toLowerCase()` | **1 failed** |
| restore | 8/8 green |

## Evidence

| | |
|---|---|
| full backend suite | **exit 0** — 480 suites / 11469 tests |
| `turbo run lint type-check --filter=backend --force` | **exit 0** — 7/7, 0 errors, 25 pre-existing warnings |

`git rev-parse HEAD` checked either side of the run, tree clean both times.

Run independently a second time before this was opened, on the commit on this branch, `git rev-parse HEAD`
checked either side and the tree clean both times: **480 suites / 11469 tests exit 0**, lint and type-check
**7/7, 0 errors**.

**Mutation matrix over the shared converter**, each mutation asserting it applied to exactly one occurrence
before substituting, and the file restored from a pristine copy between runs. Baseline and restore are both
249/249 across `stripe-minor-units.test.ts`, `stripe.service.test.ts` and `finance.payment.test.ts`:

| Mutation | Tests failed |
|---|---|
| always `× 100` (drop the ternary) | **4** |
| put `"ugx"` back into the set | **2** |
| add `"isk"` to the set | **1** |
| drop `.trim().toLowerCase()` | **1** |

Every branch of the converter is reddened by at least one mutation, **including adding a wrong entry** —
which is the direction the UGX bug arrived through and the direction #2739's tests did not cover. Measured on
`dev` before this change, `add "isk"` was green.

## Deliberately not done

- **`finance/tax.ts` still scales with a bare `×100`,** in both directions, at six sites. It is not a two-line swap: `allocateInvoiceDiscountAcrossLines` converts to minor units, allocates a discount with remainder distribution, and converts back, with no currency parameter. Doing it correctly means allocating in the currency's *real* minor unit — for a zero-decimal currency the current code distributes remainders at 1/100th of the smallest unit that exists — and needs a reverse converter that does not exist yet. That is its own change with its own review, and it is a correctness question, not a tidying one.
- **`inventory-consumption.service.ts:434` `resolvePriceCents` is out of scope** and named rather than missed: zero Stripe references, zero currency references.
- **The reverse `/100` direction is untouched** — seven sites in `stripe.service.ts` alone. No existing currency-aware counterpart to promote.
- **The half-cent float drift is pinned as it behaves, not as it should.** `toStripeMinorUnits(8.165, "usd")` returns `816`, because `8.165 * 100` is `816.4999999999999`. Fixing it changes money at every call site. The test asserts `816` so a future fix reddens it and has to say so.

## Not established

~~The remaining fourteen entries in the set are unaudited.~~ **Discharged.** The list was read in
English in a browser at `docs.stripe.com/currencies?locale=en-US` (`document.documentElement.lang`
= `en-US`) and diffed against the set programmatically:

```
our set  (15): bif clp djf gnf jpy kmf krw mga pyg rwf vnd vuv xaf xof xpf
stripe   (16): + ugx, which carries a special case

in ours, not on Stripe's list             : NONE
unexplained omissions (would undercharge) : NONE
wrongly present despite a special case    : NONE
```

Verbatim from the source, and it is the exact fixture the tests use:

> "UGX transitioned to a zero-decimal currency, but backwards compatibility requires you to
> represent it as a two-decimal value, where the decimal amount is always `00`. For example, **to
> charge 5 UGX, provide an `amount` value of `500`.**"

**Why every automated fetch came back Spanish, since it was never a transient failure:** the docs
carry region and language as two independent controls and default the language to the detected
region. The right page in the wrong language looks exactly like a flaky fetch and is not one.

**One thing the source says that this set's name does not.** HUF and TWD are zero-decimal for
**payouts**, not for charges — "even though you can charge two-decimal amounts". Both are correctly
absent, but for a reason the code does not state. This set is *"currencies whose charge amount is
unscaled"*, not *"Stripe's zero-decimal currencies"*, and anyone adding payout support who reads it
as the latter gets HUF and TWD backwards. Worth a sentence in the module comment.

### DIFF
--- a/apps/backend/src/services/finance/payment.ts
+++ b/apps/backend/src/services/finance/payment.ts
@@ -13,6 +13,7 @@ import { prisma } from "src/config/prisma";
 import logger from "src/utils/logger";
 import { FinanceEventService } from "./events";
 import { roundMoney } from "./pricing";
+import { toStripeMinorUnits } from "src/utils/stripe-minor-units";
 import { markInvoiceTreatmentItemsSettled } from "./settlement";
 
 type PaymentLineSummary = {
@@ -721,47 +722,6 @@ type CheckoutLineItemSource = {
   unitPrice?: number;
   discountPercent?: number;
 };
-
-/**
- * Currencies Stripe treats as ZERO-DECIMAL: the API takes the amount in the
- * currency's own units, not in hundredths.
- *
- * Multiplying by 100 unconditionally overcharges every one of them by 100x -
- * a 1,000 JPY invoice would be submitted as 100,000 JPY. The currency became
- * configurable per invoice, so this is no longer hypothetical.
- *
- * https://docs.stripe.com/currencies#zero-decimal
- */
-const ZERO_DECIMAL_CURRENCIES = new Set([
-  "bif",
-  "clp",
-  "djf",
-  "gnf",
-  "jpy",
-  "kmf",
-  "krw",
-  "mga",
-  "pyg",
-  "rwf",
-  // NOT "ugx". Stripe's Special cases: UGX "became a zero-decimal currency, but
-  // backwards compatibility requires you to represent it as a two-decimal value,
-  // where the decimal amount is always 00. To charge 5 UGX, provide an amount
-  // value of 500." It was in this set, so a UGX invoice was submitted at a
-  // hundredth of its value. ISK carries the identical rule and is correctly
-  // absent - the same exception was applied to one and not the other.
-  // https://docs.stripe.com/currencies#special-cases
-  "vnd",
-  "vuv",
-  "xaf",
-  "xof",
-  "xpf",
-]);
-
-/** An amount in the smallest unit Stripe accepts for `currency`. */
-const toStripeMinorUnits = (amount: number, currency: string): number =>
-  ZERO_DECIMAL_CURRENCIES.has(currency.trim().toLowerCase())
-    ? Math.round(amount)
-    : Math.round(amount * 100);
 
 // Charge the full bill as itemised, pre-tax lines (letting Stripe apply tax)
 // UNLESS we must charge a remaining/adjusted balance instead: when a prior
--- a/apps/backend/src/services/stripe.service.ts
+++ b/apps/backend/src/services/stripe.service.ts
@@ -17,6 +17,7 @@ import { NotificationService } from "./notification.service";
 
 import { prisma } from "src/config/prisma";
 import { getOrgBillingCurrency } from "src/utils/billing";
+import { toStripeMinorUnits } from "src/utils/stripe-minor-units";
 import { recomputeOrganizationVerification } from "./organization-verification.service";
 import { Prisma } from "@prisma/client";
 
@@ -57,10 +58,6 @@ const getStripeClient = () => {
   stripeClient = new Stripe(apiKey, { apiVersion: "2026-01-28.clover" });
   return stripeClient;
 };
-
-function toStripeAmount(amount: number): number {
-  return Math.round(amount * 100);
-}
 
 // Settlement must use what Stripe actually captured, never the invoice total.
 const resolveCapturedAmount = (
@@ -487,8 +484,8 @@ export const StripeService = {
     if (!organisation?.stripeAccountId)
       throw new Error("Organisation has no Stripe account");
 
-    const amount = toStripeAmount(service.cost);
     const currency = await getOrgBillingCurrency(appointment.organisationId);
+    const amount = toStripeMinorUnits(service.cost, currency);
 
     const { parentId, patientId } = extractAppointmentPatientRefs(appointment);
     const companionId = patientId ?? "";
--- a/apps/backend/src/utils/stripe-minor-units.ts
+++ b/apps/backend/src/utils/stripe-minor-units.ts
@@ -16,6 +16,12 @@
 // to represent it as a two-decimal value, where the decimal amount is always
 // 00." Both are therefore scaled like a two-decimal currency.
 // https://docs.stripe.com/currencies#special-cases
+//
+// Read the name precisely: this is "currencies whose CHARGE amount is
+// unscaled", not "Stripe's zero-decimal currencies". HUF and TWD are
+// zero-decimal for PAYOUTS and are deliberately absent here, because Stripe
+// takes two-decimal amounts when charging them. Anyone adding payout support
+// and reading this set as the general answer gets those two backwards.
 const ZERO_DECIMAL_CURRENCIES = new Set([
   "bif",
   "clp",
--- a/apps/backend/src/utils/stripe-minor-units.ts
+++ b/apps/backend/src/utils/stripe-minor-units.ts
@@ -12,9 +12,9 @@
 //
 // Two currencies that ARE zero-decimal in ISO 4217 are deliberately absent,
 // because Stripe's Special cases override the general rule for them: UGX and
-// ISK "became a zero-decimal currency, but backwards compatibility requires you
-// to represent it as a two-decimal value, where the decimal amount is always
-// 00." Both are therefore scaled like a two-decimal currency.
+// ISK "transitioned to a zero-decimal currency, but backwards compatibility
+// requires you to represent it as a two-decimal value, where the decimal amount
+// is always 00." Both are therefore scaled like a two-decimal currency.
 // https://docs.stripe.com/currencies#special-cases
 //
 // Read the name precisely: this is "currencies whose CHARGE amount is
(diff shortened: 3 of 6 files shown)


aupyay:
## Evidence re-run at the current head

The numbers in the description above were measured at `f856f79b1`. Three
comment-only commits have landed since, so this records the same checks at the
head that will actually merge.

At `f75696b3b8`, `git rev-parse HEAD` confirmed either side of both runs and
`git status --porcelain` empty either side:

| | |
|---|---|
| `npx jest` (apps/backend, full) | **exit 0** — 480 suites / 11469 tests |
| `turbo run lint type-check --filter=backend --force` | **exit 0** — 7/7, 0 errors, 25 pre-existing warnings |

`prisma:generate` was run before jest, so the totals are not a stale-client
artefact.

### The comment-only claim, checked rather than trusted

The three commits between the two heads touch `stripe-minor-units.ts` and
`finance.payment.test.ts`. Comment-stripped, both files hash identically either
side. Both controls on the comparison passed: a code change (`jpy` -> `jpx`)
was reported as differing, and a comment edit was reported as identical.

One hazard specific to this PR was checked and does not apply: the comparison
strips everything after `//`, and these commits pin `?locale=en-US` into a URL.
A URL inside a code string would have been truncated on both sides and the
comparison would have looked clean. All three URLs here are in comments, none
in code.

### Behind `dev`, and inertly so

The branch is three commits behind `dev` (`31485aae4`). Those commits touch
only `scripts/deploy/lib/migrate.sh`, which has no overlap with the six files
changed here, and GitHub reports `MERGEABLE`. There is nothing to rebase for.

ankit-yc (commented):
Reviewed at `f75696b3b8d4a1ca4fb92de1bcb7c18d2df38aa6` on a clean worktree of my own. The change is right and the before-row reproduces; one uncovered branch below is all that is between this and my approval.

## Re-derived rather than read

The before-row, reproduced by reverting only `stripe.service.ts` to what `dev` does today:

```
● StripeService › createPaymentIntentForAppointment › submits a zero-decimal appointment cost unscaled
    - ObjectContaining {"amount": 1000,   "currency": "jpy"}
    + {"amount": 100000, "currency": "jpy", ...}
```

Failing for the amount, not for a mock or an arity change.

Mutation matrix, run independently of the one in the body. Each mutation asserted it applied to exactly one occurrence before substituting; the files were restored from pristine copies between runs. Baseline and restore both 249/249 across the three affected test files.

| Mutation | Tests failed |
|---|---|
| always `× 100` (drop the ternary) | **4** |
| put `"ugx"` back into the set | **2** |
| add `"isk"` to the set | **1** |
| drop `.trim().toLowerCase()` | **1** |
| remove `"jpy"` from the set (mine) | **4** |
| `stripe.service.ts` back to the currency-blind `× 100` (mine) | **1** |
| **drop `Math.round` from the zero-decimal branch (mine)** | **0 — green** |

Full backend suite `npx jest` **exit 0, 480 suites / 11469 tests**; `turbo run lint type-check --filter=backend --force` **exit 0, 7/7, 0 errors, 25 pre-existing warnings**. `git rev-parse HEAD` checked either side of both runs, tree clean.

## The one thing to add

`Math.round(amount)` on the zero-decimal branch is unpinned. Changing it to a bare `amount` leaves all 249 tests green, so nothing states that a fractional amount in an unscaled currency is rounded before it reaches Stripe — which the API requires. `always hands Stripe an integer` covers only `usd`, which is the two-decimal branch.

It is reachable: `Service.cost` is a `Float` and the invoice paths carry `roundMoney`'d 2dp values, so a `1000.5` JPY cost is submitted as `1000.5` and Stripe rejects it.

One assertion closes it:

```ts
expect(toStripeMinorUnits(1000.5, "jpy")).toBe(1001);
```

Not a defect in this change — the branch arrives unchanged from the private original, and this PR is precisely the one that gives it tests. But by the standard this PR itself argues for (the set was pinned against removal and not against addition), a branch no mutation can redden is a missing test.

## Scope claims, all checked

- `inventory-consumption.service.ts` — **0** matches for `stripe` or `currency`. Correctly named and excluded.
- `finance/tax.ts` — six `* 100` sites, exactly as stated. Filed as #2754, because a "deliberately not done" paragraph disappears when the PR merges and the issue does not. It is the same defect shape as the one fixed here: `buildAutomaticTaxLineItems` puts `amount: Math.round(discountedAmount * 100)` on the same object literal as `currency: input.currency`.
- Forward direction otherwise complete: every remaining `* 100` in `apps/backend/src` is percentage arithmetic, the tax sites above, or `resolvePriceCents`.

## Agreed, on the merits

Not folding this into the frontend's `money.ts`. Display formatting and "what integer does this API accept" are different questions with different sources of truth, and the name is what makes the merge hard to do by accident.

aupyay:
## Rounding pinned on the zero-decimal branch — `d6ced6b07`

Review found that `Math.round` on the unscaled branch was pinned by nothing.
Reproduced before fixing, and the mutation asserted it applied to exactly one
occurrence before substituting:

| | |
|---|---|
| drop `Math.round` from the zero-decimal branch, **before** | **249 passed, 0 failed** — the gap is real |
| same mutation, **after** | **1 failed, 249 passed** |
| restore | 250/250 |

The failure names the value rather than a mock or an arity:

```
● toStripeMinorUnits › rounds a fractional zero-decimal amount rather than passing it through
    Expected: 1001
    Received: 1000.5
```

Every zero-decimal case in the suite passed an integer in, so nothing observed
the rounding, and the existing integer guarantee exercises `usd` — the other
branch. Reachable rather than theoretical: `Service.cost` is a `Float`, so a
`1000.5` JPY service would have been submitted to Stripe as `1000.5` and
rejected.

Not a defect this branch introduced. The converter arrives unchanged from the
private original and this is the change that first gives it tests — promotion
carries the untested branches across along with the correct ones, which is the
cost of promoting rather than adding. It is also the same asymmetry that let UGX
sit in the set beside a comment explaining why ISK was not: a rule guarded in one
direction and not the other.

### Gates at `d6ced6b07`

`git rev-parse HEAD` confirmed either side of both runs, tree clean either side:

| | |
|---|---|
| `npx jest` (apps/backend, full) | **exit 0** — 480 suites / **11470 tests** |
| `turbo run lint type-check --filter=backend --force` | **exit 0** — 7/7, 0 errors, 25 pre-existing warnings |

The `finance/tax.ts` half of this work is now tracked as #2754 rather than
living in the "deliberately not done" section of a description that disappears
when this merges.

ankit-yc (approved):
Approving at `d6ced6b07c20162bccd66b3d0ec1c4865da5aba3`. The added assertion closes the one branch my review left open, and I verified it rather than reading it.

```
drop `Math.round` from the zero-decimal branch, at d6ced6b07
  -> 1 failed, 249 passed / 250

● toStripeMinorUnits › rounds a fractional zero-decimal amount rather than passing it through
    Expected: 1001
    Received: 1000.5
```

Red for the value, not for a mock. Occurrence assertion checked before substituting; restored from a pristine copy afterwards, 250/250.

At the same head, `git rev-parse HEAD` checked either side and the tree clean either side:

| | |
|---|---|
| `npx jest` (apps/backend, full) | **exit 0** — 480 suites / **11470 tests** |
| three affected suites, baseline and restore | 250/250 |

Earlier at `f75696b3b8` I also ran `turbo run lint type-check --filter=backend --force` — **exit 0, 7/7, 0 errors, 25 pre-existing warnings** — and reproduced the before-row by reverting only `stripe.service.ts` to `dev`'s currency-blind `× 100`: the new appointment test fails `100000` against `1000`, for the amount.

The `finance/tax.ts` half named in "Deliberately not done" is filed as #2754 so it survives this merge.

PR #2755, opened 2026-09-05
fix(backend): report the authentication control at startup

### LINKED ISSUE #2750: A missing auth env var silently disables all authentication, and /health/controls does not report it
Found during the baseline QA survey of surface 4 (auth + access), against a
local build of `dev` at `f455595d4`.

## The API reports healthy with authentication entirely absent

`apps/backend/src/app.ts:27` gates the whole SuperTokens stack:

```ts
function isSuperTokensEnabled(): boolean {
  const disabled =
    process.env.SUPERTOKENS_DISABLED === "true" ||
    process.env.SUPERTOKENS_DISABLED === "1";

  if (disabled) return false;

  return Boolean(
    process.env.SUPERTOKENS_CONNECTION_URI &&
    process.env.AUTH_API_DOMAIN &&
    process.env.AUTH_WEBSITE_DOMAIN,
  );
}
```

When it returns `false`, `createApp` skips `registerSuperTokensBeforeRoutes` and
`registerSuperTokensErrorHandler`. There is no log line, no thrown error and no
control report. **The process starts, `/health` answers 200, and every
authentication route is a 404.**

## The two causes are observably identical

Measured on the same build, restarting between runs, probing from outside:

| Boot | `/health` | `/health/controls` | `/auth/signin` |
|---|---|---|---|
| `SUPERTOKENS_DISABLED=true` — **deliberate** | 200 | `degraded` | **404** |
| `AUTH_WEBSITE_DOMAIN` absent — **an accident** | 200 | `degraded` | **404** |

Nothing an operator or a monitor can reach distinguishes *someone turned auth
off* from *a variable went missing in a deploy*. The `degraded` in both rows is
an unrelated control, not a signal about auth.

**The function itself already draws that distinction** — `disabled` is computed
and returned separately, one line above — and then discards it.

## Why this is the one control that should be reported

`apps/backend/src/config/startup-controls.ts` exists for exactly this failure,
and says so:

> *"A control that FAILS TO APPLY must not look the same as one that was never
> asked to. Anything registered here shows up on /health/controls, so a monitor
> can see the difference from outside without anyone reading a log."*

It was written after the Stream upload blocklist silently failed to apply on
every environment. **The Stream policy is registered there. Authentication is
not** — so the mechanism built to make a missing control visible does not cover
the largest control in the process.

## Suggested fix

One `recordControl` call beside the existing gate, distinguishing the two states
the function already computes:

```
SUPERTOKENS_DISABLED set   -> recordControl("supertokens", "skipped", "disabled by configuration")
vars missing               -> recordControl("supertokens", "failed",  "auth env incomplete")
enabled                    -> recordControl("supertokens", "applied")
```

`failed` is the state `hasFailedControl()` already surfaces as `degraded`, so a
monitor polling `/health/controls` would see an auth-less deploy immediately.

**Naming which variable is missing is not necessary and should be avoided** —
the file's own comment says a detail must never include credentials, and
"auth env incomplete" is enough to act on.

## Not established

- Whether any deployed environment has ever been in this state. `/health/controls`
  reports the current process only, and it does not report auth at all today,
  so there is no way to check from outside now or retrospectively.
- This is a local build. The hosted dev API is 122 commits behind (#2741).

---

## Correction to the probe, and a better one

**The `/auth/signin` row above is method-dependent and the method was not stated.**
Sign-in is POST-only, so a **GET** returns 404 whether or not the router is
mounted — anyone repeating the table with a GET gets 404 in both rows and
concludes wrongly. The measurements above used:

```
curl -X POST -H 'content-type: application/json' -H 'rid: emailpassword' \
     -d '{"formFields":[]}'  <host>/auth/signin
```

which is **200 mounted, 404 unmounted**. Raised by another reviewer against
hosted dev, where a GET returns 404 on a fully mounted router.

**A better discriminator, needing no body, header or method note** — measured
locally with auth mounted, `/health` 200 as a control immediately before:

```
GET /auth/signup/email/exists?email=probe@example.com   200   {"status":"OK","exists":false}
GET /auth/signup/email/exists   (no params)             400   handler validating input
GET /auth/definitely-not-a-real-route                   404   control
GET /definitely-not-a-real-route                        404   control
```

Both controls 404, so no catch-all is answering and the 400 is a real handler.
**When the router is absent, the first two become 404.**

## The "not established" item above is partly discharged

That probe means **the state this issue describes is detectable from outside,
unauthenticated, today** — hosted dev has been checked this way and is *not* in
the broken state. What remains true is that it is not detectable from
`/health/controls`, which is what the fix is about, and that there is no
retrospective way to tell whether any environment has been in it.

## Rate limiting, checked while here

`/auth` carries a dedicated limiter — `app.ts:56`, 100 per 15 minutes, mounted
before the SuperTokens routes and stricter than the global 500 — and it fires:
110 rapid unauthenticated GETs gave **67 × 200 and 43 × 429**, with
`RateLimit-Policy: 100;w=900`. So the enumeration surface is bounded. Noted
because a reader of the probes above would otherwise reasonably ask.

aupyay:
## PR Checklist

`createApp` gates the whole SuperTokens stack on a boolean. When that boolean is
false the app skips `registerSuperTokensBeforeRoutes` and
`registerSuperTokensErrorHandler` with no log line, no thrown error and no
control report - the process starts, `/health` answers 200, and every
authentication route is a 404.

The boolean is false in two very different situations:

| Boot | `/health` | `/health/controls` | auth routes |
|---|---|---|---|
| `SUPERTOKENS_DISABLED` set - deliberate | 200 | unchanged | 404 |
| an auth variable absent - an accident | 200 | unchanged | 404 |

Nothing an operator or a monitor can reach separates *someone turned auth off*
from *a variable went missing in a deploy*. `startup-controls.ts` exists for
exactly this failure mode and says so - "a control that FAILS TO APPLY must not
look the same as one that was never asked to" - and the Stream upload policy is
registered there while authentication is not.

The gate function already computed the distinction, one line above, and then
discarded it.

## What is the new behavior?

The gate returns the three states it already knows about instead of a boolean,
and records one control:

```
SUPERTOKENS_DISABLED set  ->  skipped, "disabled by configuration"
auth env incomplete       ->  failed,  "auth env incomplete"
mounted                   ->  applied
```

`failed` is the state `hasFailedControl()` already surfaces, so
`/health/controls` answers 503 `degraded` on an auth-less deploy and a monitor
sees it from outside without anyone reading a log. `/health` is deliberately
unchanged: the process is up, the control is not.

Two deliberate choices worth flagging for review:

- **The detail never names the missing variable.** `/health/controls` is
  unauthenticated by design, and "auth env incomplete" is enough to act on.
  A test asserts none of the four variable names can appear in the detail.
- **The control is named `authentication`, not the provider.** The issue
  suggested the provider name. The endpoint is public, the control being
  reported is "authentication is mounted" rather than a vendor, and a monitor's
  alert keeps working if the provider is ever swapped.

A kill switch somebody threw stays `skipped`, not `failed` - per that file's own
reasoning, paging on a deployment fact would train people to ignore the
endpoint.

### Verification

Run in `apps/backend` on this branch:

- `npx jest` - 480 suites, 11478 tests, exit 0 (full backend suite, not a
  scoped run).
- `pnpm --filter backend run lint` - exit 0, 0 errors, 25 pre-existing warnings.
- `pnpm --filter backend run type-check` - exit 0.
- Coverage of the changed file: `app.ts` 99.2% lines, 92% branches; the two
  uncovered lines are pre-existing and untouched.
- Each of the 8 new tests was mutation-checked, and checked by which test went
  red rather than by exit code. Six mutations: dropping the `applied` record,
  swapping `skipped` for `failed` in either direction, putting a variable name
  into the detail, collapsing the gate back to two states, and making the kill
  switch depend on the env being complete. The last one is reddened by exactly
  one test - the kill-switch-plus-incomplete-env case is the only input that
  separates it from the others.

One note on running this branch locally, because the first lint run here was
misleading: in a fresh worktree `pnpm --filter backend run lint` reports
thousands of `no-unsafe-*` errors in files nobody touched, `app.ts` included.
Nothing is wrong with the code - the workspace packages are not built, so the
type-aware rules cannot resolve `@yosemite-crew/auth`. `turbo run build
--filter ./packages/*` first, and the same command exits 0. Jest likewise needs
`prisma generate` to have run in the worktree.

## Related Issue(s)

Fixes #2750

### DIFF
--- a/apps/backend/src/app.ts
+++ b/apps/backend/src/app.ts
@@ -1,7 +1,11 @@
 import express from "express";
 import rateLimit from "express-rate-limit";
 import fileUpload from "express-fileupload";
-import { getControlReports, hasFailedControl } from "./config/startup-controls";
+import {
+  getControlReports,
+  hasFailedControl,
+  recordControl,
+} from "./config/startup-controls";
 import { registerRoutes } from "./routers";
 import { StripeController } from "./controllers/web/stripe.controller";
 import { FinanceController } from "./controllers/app/finance.controller";
@@ -24,24 +28,35 @@ import {
 } from "@yosemite-crew/auth";
 import { authHooks } from "./config/auth-hooks";
 
-function isSuperTokensEnabled(): boolean {
+/**
+ * Three states, not two.
+ *
+ * `disabled` and `incomplete` both leave every /auth route a 404 while /health
+ * still answers 200, so from outside they are the same process - one of them
+ * deliberate, the other a deploy that lost a variable. The boolean this used to
+ * return computed the distinction and then threw it away.
+ */
+type AuthGate = "enabled" | "disabled" | "incomplete";
+
+function readAuthGate(): AuthGate {
   const disabled =
     process.env.SUPERTOKENS_DISABLED === "true" ||
     process.env.SUPERTOKENS_DISABLED === "1";
 
-  if (disabled) return false;
+  if (disabled) return "disabled";
 
-  return Boolean(
-    process.env.SUPERTOKENS_CONNECTION_URI &&
+  return process.env.SUPERTOKENS_CONNECTION_URI &&
     process.env.AUTH_API_DOMAIN &&
-    process.env.AUTH_WEBSITE_DOMAIN,
-  );
+    process.env.AUTH_WEBSITE_DOMAIN
+    ? "enabled"
+    : "incomplete";
 }
 
 export function createApp() {
   const app = express();
 
-  const superTokensEnabled = isSuperTokensEnabled();
+  const authGate = readAuthGate();
+  const superTokensEnabled = authGate === "enabled";
   if (superTokensEnabled) {
     // Fail fast at startup on invalid auth config (epic #1672 acceptance).
     const authConfig = readAuthConfig();
@@ -62,8 +77,21 @@ export function createApp() {
     app.use("/auth", authLimiter);
 
     registerSuperTokensBeforeRoutes(app);
+    recordControl("authentication", "applied");
   } else {
     setAuthService(null);
+    // Turning auth off is a deployment fact; booting without the env it needs
+    // is an incident nobody asked for. `failed` is what /health/controls
+    // surfaces as degraded, so a monitor sees an auth-less deploy immediately.
+    //
+    // The detail never names the missing variable: this endpoint is
+    // unauthenticated by design (see startup-controls.ts), and "incomplete" is
+    // already enough to act on.
+    if (authGate === "disabled") {
+      recordControl("authentication", "skipped", "disabled by configuration");
+    } else {
+      recordControl("authentication", "failed", "auth env incomplete");
+    }
   }
   app.use(helmet());
   app.disable("x-powered-by");
--- a/apps/backend/test/app.test.ts
+++ b/apps/backend/test/app.test.ts
@@ -71,6 +71,11 @@ jest.mock("../src/config/auth-hooks", () => ({
 }));
 
 import { createApp } from "../src/app";
+import {
+  getControlReports,
+  hasFailedControl,
+  resetControlsForTest,
+} from "../src/config/startup-controls";
 
 type Layer = {
   route?: {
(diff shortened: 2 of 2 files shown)


ankit-yc (approved):
Reviewed at `d2ee8a69215249ce97b84a985648f84b9b1fc68a` on a worktree of my own. Approving — the three-state gate is right, the tests pin the distinction the issue is about, and my own mutations all reddened for the reasons they should. One follow-up below that is not a change request.

## Re-derived rather than read

Seven mutations, each asserting it applied to exactly one occurrence before substituting, `src/app.ts` restored from a pristine copy between runs. Baseline and restore both 16/16.

| Mutation | Result |
|---|---|
| incomplete records `skipped` instead of `failed` | **3 failed** |
| disabled records `failed` instead of `skipped` | **3 failed** |
| drop `recordControl("authentication", "applied")` | **2 failed** |
| kill switch also requires a complete env | **1 failed** |
| detail names the missing variable | **3 failed** |
| completeness check drops `AUTH_WEBSITE_DOMAIN` | **2 failed** |
| `/health` degrades when a control failed | **3 failed** |

Two checked by which test went red rather than by the count:

```
skipped-instead-of-failed  ->  records failed when the auth env is incomplete
                               distinguishes the deliberate boot from the accidental one
                               degrades /health/controls but not /health when auth did not mount

kill-switch-requires-env   ->  reports skipped when auth is switched off and the env is incomplete too
```

The second is reddened by exactly one test and it is the one whose input is kill-switch-plus-incomplete-env, which matches the PR body's claim. The last row is worth having: `/health` staying 200 is an assertion, not an accident, and coupling liveness to controls now breaks a test instead of a monitor.

Gates at the same head, `git rev-parse HEAD` checked either side and the tree clean:

| | |
|---|---|
| `npx jest` (apps/backend, full) | **exit 0** — 480 suites / **11478 tests** |
| `turbo run lint type-check --filter=backend --force` | **exit 0** — 7/7 |
| `test/app.test.ts` alone | 16/16 |

## On the two calls flagged for a reviewer

**`authentication` rather than the provider name — agreed, with one thing the next reader should know.** The reasoning holds: the endpoint is unauthenticated, what is being reported is that authentication is mounted, and an alert on it survives a provider swap. The precedent is not uniform, though — the only other control here is `stream-upload-policy`, which does embed its vendor. So this is the better name and it is not simply following the existing convention; worth the comment it already has.

**The detail never naming the variable is enforced, not just intended.** Changing it to `auth env incomplete: AUTH_WEBSITE_DOMAIN` fails three tests, one of which iterates all four names rather than pinning a string.

## Follow-up, not a change request

`/health/controls` returns 503 for an auth-less boot, but nothing gates on it: `scripts/deploy/api-deploy.sh:210` smoke-probes `/health` only, which stays 200 by design.

```
grep -rn "health/controls" over the repo   ->  only the comment in startup-controls.ts
api-deploy.sh:210                          ->  curl … /health
```

So after this lands, a deploy that lost an auth variable is **visible to a monitor and still ships green through its own smoke test**. That is a separate change to the deploy script with its own risk, and it is exactly the gap this control was created to expose — I would file it rather than widen this PR. Say the word and I will.

PR #2757, opened 2026-09-05
fix(ci): gate the api cutover on named startup controls

### LINKED ISSUE #2756: api-deploy smoke gate does not read /health/controls, so an auth-less deploy still cuts over
`/health/controls` was added so an auth-less boot is visible from outside (#2750, merged as `bffd96987`). The deploy's own smoke gate does not read it — so the one place guaranteed to catch a deploy that lost an auth variable is the place that cannot see it.

Verified at `bffd96987`:

```
consumers of /health/controls        the route itself, two comments, two tests. Nothing else.
api-deploy.sh:210                    curl … /health          -> 200 by design for an auth-less boot
api-deploy.sh:253                    [ "$CODE" != "200" ]    -> exit 1, no cutover
api-deploy.sh:259                    [ "$POST_CODE" != "404" ] -> exit 1, no cutover
```

Both existing gates pass a process whose entire `/auth` surface is 404, because both were designed to: `/health` is liveness and the POST probe is the body parser. The control that knows better answers 503 on a port nobody curls.

### Why this is not just "the smoke step could check more"

The smoke boot runs the real bundle against the real environment before cutover, so the missing variable is observable at exactly the moment a rollback is still free. After cutover it is observable only to whoever is watching a monitor that does not exist yet.

### The design question, which is why this is an issue and not a patch

**Do not gate on the aggregate.** `/health/controls` returns 503 whenever *any* control is `failed`, and `stream-upload-policy` records `failed` when Stream's `updateAppSettings` rejects. `startup-controls.ts` is explicit that this must not take the API down:

> *"This deliberately does not stop the boot. A Stream outage should not take the API down - it should be VISIBLE."*

Gating cutover on the aggregate status would turn a third-party outage into a blocked deploy, which is the opposite of what that module decided. So the question is which controls are cutover-blocking:

- probe `/health/controls`, parse the array, and require **named** controls to be `applied` — `authentication` today — while reporting the rest;
- or keep the aggregate for monitors and add a separate assertion for auth specifically;
- or decide that a deploy should never block on a control and that this belongs in an alert instead, which is a legitimate answer and should be written down rather than left implicit.

`skipped` must not block either way: an environment with auth deliberately switched off is a deployment fact, and treating it as a failure is the same mistake in the other direction.

### Not established

Whether any environment currently boots with an incomplete auth env. The mechanism is in the repository; the exposure is a question about the deploy host that needs someone who can read it.

aupyay:
## PR Checklist

`api-deploy.sh` boots the new bundle on a spare port before cutover - the one
moment a rollback is still free - and reads two things from it: `/health`, and a
POST to a bogus route that exercises the body parser.

Neither can see a process that came up without authentication. `/health` answers
200 for a bundle whose entire `/auth` surface is 404, deliberately: #2750 left
liveness alone on purpose. Since #2750 the process reports the difference on
`/health/controls`, and at `bffd96987` nothing in the repository reads it. So a
deploy that lost an auth variable cuts over green and is visible only to a
monitor that does not exist yet.

## What is the new behavior?

The smoke step probes `/health/controls` while the process is still up, prints
the body whatever it says, and blocks the cutover on **named** controls.

**Not the aggregate.** `/health/controls` answers 503 whenever *any* control is
`failed`, and `stream-upload-policy` records `failed` when Stream's
`updateAppSettings` rejects. `startup-controls.ts` is explicit that this must
not take the API down - "a Stream outage should not take the API down, it should
be VISIBLE" - so gating cutover on the aggregate would invert that decision and
turn a third-party outage into a stuck deploy.

Blocking is therefore stated as a positive: **only a named control positively
reporting `failed` stops the cutover.** Everything else ships, and each of these
is a decision rather than an oversight:

| Case | Cutover | Why |
|---|---|---|
| named control `failed` | **blocked** | the state this exists for |
| named control `skipped` | ships | auth switched off is a deployment fact, not a fault |
| named control **absent** | ships | a rollback deploys a bundle from before the control existed |
| endpoint unreadable | ships | `/health` is already the liveness gate and runs first |
| unnamed control `failed` | ships | the Stream reasoning above |
| the helper **crashes** | **blocked** | a crash writes nothing to stdout either, and it is not a ship |
| the helper cannot **run** | **blocked** | a gate that could not run is not a gate that passed |

The last row is the one the table first left implicit, and it is decided by the
call site rather than by the helper: `set -e` carries a command substitution's
status, so a `node` that will not start stops the script before the cutover
checks and the EXIT trap prints the rollback sha. An unreadable *body* is a fact
about the process and ships; an interpreter that will not start is not. Raised
in review, and pinned three ways rather than asserted.

The absent case is the deliberate blind spot and the one most worth arguing
about: blocking on it is superficially right, since an absent control is exactly
what #2750 was about, but it would make the gate refuse the one deploy that most
needs to work. A bundle that silently stops recording the control is therefore
not caught here. Reviewers who would rather have it the other way should say so -
it is one branch.

The name list is `DEPLOY_BLOCKING_CONTROLS`, defaulting to `authentication`, so
an environment that genuinely runs without one does not need a code change to
deploy.

The decision lives in `scripts/deploy/lib/controls.sh` with its own suite, the
same shape as `lib/migrate.sh` and `tests/migrate.test.sh`, so the ship list is
pinned rather than inline in the gate. `node` rather than `sed`, because a gate
that mis-parses is worse than no gate and the box has just proved node works by
booting the bundle.

### Verification

- `scripts/deploy/tests/controls.test.sh` - 24 passed, 0 failed (new)
- `scripts/deploy/tests/migrate.test.sh` - 68 passed, 0 failed
- `scripts/deploy/tests/git-sync.test.sh` - 16 passed, 0 failed
- `shellcheck -x` on `api-deploy.sh` is unchanged from `bffd96987`: 4 SC2015,
  all pre-existing. `lib/controls.sh` is clean. `tests/controls.test.sh` reports
  only the 2 SC1091 path-resolution notes the existing test files also report.

**18 mutations, each checked by which test went red rather than by exit code.**
One per decision in the table: blocking on `skipped`, blocking on absence,
blocking on `applied`, prefix-matching the name, matching any failed control
regardless of name, reading only the first required name, blocking on an
unparseable body, dropping the detail, printing a detail that is not there,
deleting the probe, moving the probe after the process is killed, emptying the
default name list, removing the `exit 1`, never calling the helper, the helper
swallowing its own non-zero status, and the call site swallowing it with
`|| true`.

Two things worth reporting against myself:

- The first version of the wiring guard grepped for the string
  `/health/controls`, and **stayed green with the probe deleted** - the
  reasoning above the probe names that path twice. Matching the `curl` instead
  fixed that instance and left the class: four more guards grep the same file,
  and any of them can be satisfied by a comment somebody writes later. They now
  read a copy of the script with full-line comments **blanked** - blanked, not
  removed, so the line numbers still line up for the ordering guard. Separating
  input, run both ways: delete the probe and leave a comment carrying the same
  text. Against the previous guards that is 22 passed, exit 0 - a deleted probe
  reported as present. Against these it reddens two tests by name.
- The first "absence blocks" mutation crashed node instead of implementing the
  behaviour, so it proved nothing about that test. Re-run as a correct
  implementation, it reddens the absence test by name.

- Two tests are **not** separately pinned and I could not find an input that
  separates them: "ships on an empty body" and "ships when no control is named".
  Removing either short-circuit changes no test, because node reaches the same
  answer. They are documented in the source as short-circuits that carry no part
  of the decision.

The `node` probe strips PATH inside a command substitution, because a variable
assignment preceding a *function* call persists in bash - the subshell is what
contains it, and the suite asserts PATH afterwards rather than assuming that. It
runs beside a positive control on the same fixture, so "node would not start"
cannot be confused with "the harness broke the call some other way".

## A follow-up this deliberately does not do

The absent case is ambiguous only because the response lists what *was*
recorded. If the process also declared what it *expects* to record, then
"expected names it and controls omit it" becomes a positive failure while "no
`expected` key at all" stays the marker of an older bundle - closing the blind
spot without touching the rollback case. That is a change to
`startup-controls.ts` with its own tests and belongs in its own issue.

What should **not** be done, because it will occur to the next reader: the
script has the deployed source on the box and could `git grep` for the
`recordControl` call to decide whether the bundle ought to have reported one.
That is a substring test on source standing in for a fact about a running
process, and it breaks on a rename that changes no behaviour.

Not established, and unchanged from the issue: whether any environment currently
boots with an incomplete auth env. That needs someone who can read the deploy
host.

### Three things raised in review and taken as commits

**A ship is two facts and the suite was asserting one.** The four SHIP rows
checked that stdout was empty - and a helper that *crashes* writes nothing to
stdout either, while a crash stops the deploy rather than shipping it. Dropping
the `try/catch` around `JSON.parse` left the suite at `22 passed` with a body
that would have halted a cutover. The instrument was already in the file: the
node-missing probe captures the status for exactly this distinction, pointed the
other way. `ships` now requires an empty result **and** a zero status - with the
declaration and the assignment on separate lines, because `local out="$(...)"`
reports the status of `local`, not of the substitution (measured on bash 3.2 and
bash 5.3; the swallow happens on both).

**The probe guard had one costume left.** Blanking full-line comments did not
cover a TRAILING comment on a code line: `CONTROLS_BODY=""  # was: curl … ` kept
both the probe guard and the ordering guard green with the probe deleted,
because the probe line resolved to the comment and the comment is still above
the kill. It is now anchored to the shape of the statement. **The trade, stated
because it will fire one day:** the guard now pins the variable name and the
assignment form, so refactoring the curl into a helper function would redden it -
a false positive on a working probe, and the right trade only because prose has
fooled this guard twice already.

**The blocking-control list is one quoted argument.** The variadic signature
forced the caller to leave its list unquoted for word splitting, and an unquoted
expansion is also a **pathname** expansion - a name containing a glob character
would have expanded against the deploy host's working directory before the
helper saw it. Nothing sets such a name, and no test could distinguish it
because the expansion happened at the call site rather than in the function.
`node` already splits on whitespace, so one quoted argument removes the exposure
by construction and takes the `SC2086` disable with it.

## Related Issue(s)

Fixes #2756

### DIFF
--- a/.github/workflows/deploy-api.yml
+++ b/.github/workflows/deploy-api.yml
@@ -60,6 +60,12 @@ jobs:
       # it is here rather than mocked.
       - name: Run deploy migration tests
         run: ./scripts/deploy/tests/migrate.test.sh
+      # This suite needs node, for the same reason the gate uses it: the cutover
+      # decision is made by parsing a JSON body, and a gate that mis-parses is
+      # worse than no gate. ubuntu-latest ships one, and the deploy host proves
+      # it has one by booting the bundle before this check ever runs.
+      - name: Run deploy startup-control gate tests
+        run: ./scripts/deploy/tests/controls.test.sh
 
   deploy:
     # Only ever on an explicit dispatch. The pull_request trigger above exists for
--- a/scripts/deploy/api-deploy.sh
+++ b/scripts/deploy/api-deploy.sh
@@ -50,11 +50,17 @@ SMOKE_PORT="${4:-8099}"
 
 NODE_BIN="${NODE_BIN:-$HOME/.nvm/versions/node/v22.21.1/bin}"
 
+# Which startup controls stop a cutover. A LIST of names, never the aggregate
+# /health/controls status - see lib/controls.sh for why the aggregate is the
+# wrong gate. Overridable so an environment that genuinely runs without one of
+# these does not need a code change to deploy.
+DEPLOY_BLOCKING_CONTROLS="${DEPLOY_BLOCKING_CONTROLS:-authentication}"
+
 # lib/ has to travel WITH this script. Copying api-deploy.sh alone leaves this
 # looking for a helper that is not on the box, and bash exits before preflight
 # with a bare "No such file or directory" - so say what is actually wrong.
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-for helper in git-sync migrate; do
+for helper in git-sync migrate controls; do
   if [ ! -r "$SCRIPT_DIR/lib/$helper.sh" ]; then
     echo "missing $SCRIPT_DIR/lib/$helper.sh" >&2
     echo "Copy the whole scripts/deploy directory to the host, not just this file." >&2
@@ -65,6 +71,8 @@ done
 . "$SCRIPT_DIR/lib/git-sync.sh"
 # shellcheck source=lib/migrate.sh
 . "$SCRIPT_DIR/lib/migrate.sh"
+# shellcheck source=lib/controls.sh
+. "$SCRIPT_DIR/lib/controls.sh"
 export PATH="$NODE_BIN:$PATH"
 export NODE_OPTIONS="${NODE_OPTIONS:---max-old-space-size=2048}"
 
@@ -244,6 +252,19 @@ POST_CODE="$(curl -s -o /dev/null -w '%{http_code}' --max-time 10 \
   -X POST -H 'content-type: application/json' -d '{"smoke":true}' \
   "http://127.0.0.1:$SMOKE_PORT/v1/definitely-not-a-route" || echo 000)"
 echo "  body parse -> $POST_CODE"
+# Which security controls actually applied in THIS environment, read from the
+# process that is about to serve traffic, while a rollback is still free.
+#
+# /health above is liveness and answers 200 for a process whose entire /auth
+# surface is 404 - by design, since #2750 deliberately left it alone. The body
+# is printed whatever it says; only a NAMED control reporting `failed` stops the
+# cutover, and lib/controls.sh holds the reasoning for every case that does not.
+# The 503 this returns when degraded is why there is no -f and no code check
+# here: the body is the signal, not the status.
+CONTROLS_BODY="$(curl -s --max-time 10 "http://127.0.0.1:$SMOKE_PORT/health/controls" || echo '')"
+echo "  controls -> ${CONTROLS_BODY:-<unreachable>}"
+# shellcheck disable=SC2086 # deliberately word-split: this is a list of names
+CONTROL_FAILURES="$(deploy_blocking_control_failures "$CONTROLS_BODY" $DEPLOY_BLOCKING_CONTROLS)"
 # Reports, never gates - the `|| true` is what makes that true, and it is
 # deliberate: this is a second signal for whoever reads the log, and POST_CODE
 # above is the gate.
@@ -259,6 +280,16 @@ fi
 if [ "$POST_CODE" != "404" ]; then
   echo "smoke boot could not parse a request body ($POST_CODE, expected 404)" >&2
   echo "- every GET probe passes a bundle that 500s on every write." >&2
+  echo "NOT cutting over. Rollback sha: $ROLLBACK_SHA" >&2
+  exit 1
+fi
+
+if [ -n "$CONTROL_FAILURES" ]; then
+  echo "smoke boot came up with a security control that FAILED to apply:" >&2
+  while IFS= read -r control_failure; do
+    echo "  $control_failure" >&2
+  done <<< "$CONTROL_FAILURES"
+  echo "- the process starts and /health answers 200 with the control absent." >&2
   echo "NOT cutting over. Rollback sha: $ROLLBACK_SHA" >&2
   exit 1
 fi
--- a/scripts/deploy/api-deploy.sh
+++ b/scripts/deploy/api-deploy.sh
@@ -268,8 +268,7 @@ echo "  body parse -> $POST_CODE"
 # EXIT trap turns that into a stop notice with the rollback sha.
 CONTROLS_BODY="$(curl -s --max-time 10 "http://127.0.0.1:$SMOKE_PORT/health/controls" || echo '')"
 echo "  controls -> ${CONTROLS_BODY:-<unreachable>}"
-# shellcheck disable=SC2086 # deliberately word-split: this is a list of names
-CONTROL_FAILURES="$(deploy_blocking_control_failures "$CONTROLS_BODY" $DEPLOY_BLOCKING_CONTROLS)"
+CONTROL_FAILURES="$(deploy_blocking_control_failures "$CONTROLS_BODY" "$DEPLOY_BLOCKING_CONTROLS")"
 # Reports, never gates - the `|| true` is what makes that true, and it is
 # deliberate: this is a second signal for whoever reads the log, and POST_CODE
 # above is the gate.
--- a/scripts/deploy/api-deploy.sh
+++ b/scripts/deploy/api-deploy.sh
@@ -261,6 +261,11 @@ echo "  body parse -> $POST_CODE"
 # cutover, and lib/controls.sh holds the reasoning for every case that does not.
 # The 503 this returns when degraded is why there is no -f and no code check
 # here: the body is the signal, not the status.
+#
+# The assignment below is deliberately NOT tolerant. An unreadable body ships -
+# that decision is in lib/controls.sh - but a helper that cannot RUN stops the
+# deploy here, because set -e carries a command substitution's status and the
+# EXIT trap turns that into a stop notice with the rollback sha.
 CONTROLS_BODY="$(curl -s --max-time 10 "http://127.0.0.1:$SMOKE_PORT/health/controls" || echo '')"
 echo "  controls -> ${CONTROLS_BODY:-<unreachable>}"
 # shellcheck disable=SC2086 # deliberately word-split: this is a list of names
--- a/scripts/deploy/lib/controls.sh
+++ b/scripts/deploy/lib/controls.sh
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+#
+# Which startup controls stop a cutover, and - more importantly - which do not.
+#
+# /health/controls exists so a security control that FAILED TO APPLY does not
+# look like one that was never asked to (apps/backend/src/config/startup-controls.ts).
+# The deploy's smoke boot runs the real bundle against the real environment
+# before cutover, which is the one moment a rollback is still free, and until
+# now it read only /health - which answers 200 for a process whose entire /auth
+# surface is 404, because /health is liveness and was designed to.
+#
+# The obvious wiring is wrong. /health/controls returns 503 whenever ANY control
+# is `failed`, and stream-upload-policy records `failed` when Stream's
+# updateAppSettings rejects. startup-controls.ts is explicit that this must not
+# take the API down - "a Stream outage should not take the API down, it should
+# be VISIBLE" - so gating cutover on the aggregate would convert a third-party
+# outage into a blocked deploy, which is that decision inverted.
+#
+# Hence: NAMED controls, and a rule stated as a positive rather than a negative.
+
+# deploy_blocking_control_failures <controls-json> [name ...]
+#
+# Echoes one line per named control that positively reports `failed`. Echoes
+# nothing - and so ships - in every other case:
+#
+#   named control `failed`      BLOCK. The only blocking state.
+#   named control `skipped`     ship. Auth switched off is a deployment fact,
+#                               not a fault; treating it as one is the same
+#                               mistake as ignoring the endpoint, mirrored.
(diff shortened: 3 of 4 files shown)


ankit-yc (dismissed):
Approving at `ae22e452d`. 22 mutations on my own worktree, each checked by the **name** of the failing test, `api-deploy.sh` and `lib/controls.sh` restored from pristine copies between every run, baseline 22/22 either side. **Two findings, neither a defect in the gate as shipped — both are places the tests cannot see a change that would matter.** One of them I would take as a commit.

### 1. The trailing-comment hole you documented is real, and it takes out two guards at once

You blanked full-line comments and deliberately left trailing ones, on the grounds that stripping those means parsing quoting. That is the right call about cost — but the hole it leaves is not theoretical:

```
delete the probe, leave a FULL-LINE comment carrying the path    20 passed, 2 failed   (your separating input, reproduced)
delete the probe, leave a TRAILING comment on the code line      22 passed, 0 failed   <- probe gone, reported present AND correctly ordered
```

The mutation is one line:

```bash
CONTROLS_BODY=""  # was: curl -s --max-time 10 "http://127.0.0.1:$SMOKE_PORT/health/controls"
```

Both `api-deploy.sh curls /health/controls` and `the controls probe runs before the smoke process is killed` go green — the second because `PROBE_LINE` resolves to the comment, which is still above the kill. **It is the same bug for the third time, wearing the one costume the fix was designed not to cover.**

The cheap close needs no quote parsing — anchor the guard to the *shape of the statement* rather than to a substring anywhere on the line:

```bash
PROBE_RE='^[[:space:]]*CONTROLS_BODY="\$\(curl .*SMOKE_PORT/health/controls'
```

That reddens the mutation above while staying a plain grep. Non-blocking; your reasoning for leaving trailing comments alone still holds, this just narrows what counts as a match.

### 2. Every SHIP row is blind to the one way "ships" becomes "stops the deploy" — this is the one I would take

Make an unparseable body **crash** node instead of returning, and the suite does not notice:

```
mutation: drop the try/catch around JSON.parse       22 passed, 0 failed
```

The ship rows assert that stdout is empty. **A crash produces empty stdout too**, and in `api-deploy.sh` a crash is not a ship — `set -e` carries the command substitution's status and the deploy stops. Driven directly, with a positive control so a broken harness cannot look like the finding:

```
pristine, body '<html>502 Bad Gateway</html>'   REACHED_CUTOVER x=[]   exit 0   ships
mutated,  same body                             SyntaxError, aborts    exit 1   deploy STOPS
mutated,  a valid body                          REACHED_CUTOVER x=[]   exit 0   harness is fine
```

So "ships on a body that is not JSON", "ships on an empty body", "ships on JSON without a controls array" and "ships when no control is named" all pass in a world where three of them stop the deploy. **The suite already owns the right instrument** — `NODE_GONE_RC` captures the status for the node-missing probe, which is the same distinction pointed the other way. It just is not applied to the rows where shipping is the assertion:

```bash
ships_rc() { local rc=0 out; out="$(deploy_blocking_control_failures "$@")" || rc=$?; printf '%s|%s' "$rc" "$out"; }
check "ships on a body that is not JSON" "0|" "$(ships_rc '<html>502 Bad Gateway</html>' authentication)"
```

This is the PR's own sentence applied to its tests: *a gate that could not run is not a gate that passed.* Right now four rows cannot tell the two apart.

### One of my mutations was broken and I am saying so rather than reporting its green

My first "gate on the aggregate instead of the named report" mutation came back **22 passed** — which reads exactly like the inversion this PR exists to prevent being uncaught. It was not. Replacing the state check alone left `report` undefined on the very fixtures that matter, so node threw, stdout was empty, and the ship rows passed vacuously — **the same blindness as finding 2, this time hiding my own mistake.** The faithful version, which also guards the detail lookup:

```
aggregate gate, faithful   20 passed, 2 failed
                           -> "ships when an unnamed control failed, even though the aggregate is degraded"
                           -> "matches a control name exactly rather than by prefix"
```

The Stream inversion **is** caught. I would have reported the opposite on the strength of the first run.

### The rest of the matrix, all confirming

| Mutation | Reddens |
|---|---|
| probe deleted | probe guard + ordering guard |
| helper call deleted | helper guard + swallow guard |
| `\|\| true` on the assignment | the swallow guard, alone |
| `exit 1` removed under `CONTROL_FAILURES` | the stop guard, alone |
| probe moved after `kill "$SMOKE_PID"` | the ordering guard, alone |
| `DEPLOY_BLOCKING_CONTROLS` default emptied | the default guard, alone |
| block on anything not `applied` | "ships when the named control was deliberately skipped" |
| prefix match instead of exact | "matches a control name exactly rather than by prefix" |
| `break` after the first failure | "reports every named control that failed, not just the first" |
| detail dropped from the message | 4, including the node-gone positive control |

And both short-circuits confirmed inert, as the comment beside them says: removing `[ -n "$json" ]` or `[ "$#" -gt 0 ]` changes no test, because node reaches the same answer for an empty body and an empty name list.

### The rollback premise is structural here, not just plausible

I went to check whether "absent ships" is load-bearing rather than take it from the issue. The deploy job's `actions/checkout` has **no `ref:`**, so the scripts come from the dispatch branch, while the bundle is `TARGET_REF` passed separately — and `scp -r scripts/deploy/.` ships the whole directory, so `lib/controls.sh` travels with it. **Gate version and bundle version are decoupled by construction**: a rollback runs *this* gate against a bundle that predates the control. Blocking on absence would refuse exactly that deploy. The row is right.

### Verification, HEAD asserted in the same shell

```
HEAD                     ae22e452d, tree clean by git status --porcelain
controls.test.sh         22 passed, 0 failed   exit 0
migrate.test.sh          68 passed, 0 failed   exit 0
git-sync.test.sh         16 passed, 0 failed   exit 0
shellcheck -x api-deploy.sh   4x SC2015, byte-identical set to bffd96987
checks                   62 pass, 3 skipping, 0 failing
```

ankit-yc (approved):
Approving at `0c0f4b4f3`. Both fixes verified by driving them, not by reading the diff — and **your count correction is right, I checked it rather than accepting it.** Two residuals, both about the tests and neither blocking.

### Your correction, confirmed

I said four ship rows pass in a world where three stop the deploy. Only **one** does. Probed each of the four inputs directly under the try/catch mutation, with the pristine run as the control:

```
                          try/catch dropped     pristine
not JSON                  rc=1  <- crashes      rc=0
empty body                rc=0                  rc=0     (short-circuits before node)
JSON, no controls array   rc=0                  rc=0     (parses fine)
no control named          rc=0                  rc=0     (short-circuits before node)
```

Three were vacuous rather than wrong, and the row that was wrong is the row that now reddens. My number was a claim about four rows derived from one of them.

### The two fixes, each driven

```
try/catch dropped                          1 red   "ships on a body that is not JSON"
call site unquoted again                   1 red   "the blocking-control list is passed as one quoted argument"
helper returns non-zero on a normal ship   1 red   "ships on an empty body"
```

That third one is mine and it is the positive control for `ships` itself: without it, "the crash reddens a row" cannot distinguish *the status is asserted* from *the output happened to differ*. It asserts the status.

### Residual 1 — the fix is one refactor from inert, and nothing reddens

`ships` is correct only because the declaration and the assignment are on separate lines. Tidying them into one, **together with the crash**:

```
try/catch dropped                                   1 red
try/catch dropped + local out="$(…)" in ships       24 passed, 0 failed
```

`local` succeeds regardless, so the status never reaches `rc` and the row goes green with the crash still there. **The guard against `|| true` at the call site exists precisely because a wrapper that always succeeds hides what it wraps — and `local` is that wrapper, inside the test that catches it.** A one-line comment on `local out rc=0` saying *declaration only, on purpose* is probably the whole fix; a guard on the test file itself would be the heavier option.

### Residual 2 — the glob row can fail, but only in the right working directory

It is not vacuous, and it is not a guard either. Modelling the hazard faithfully — expanding the name list against the filesystem before node sees it — and varying only the cwd:

```
pristine, cwd containing a file named `authentication`   24 passed, 0 failed   (no false positive)
glob-expanding mutation, same cwd                        1 red  "treats a glob-shaped name as a literal name"
glob-expanding mutation, cwd without that file           24 passed, 0 failed
```

**The row separates the two implementations only when the working directory happens to contain a file named after a blocking control.** Run from the repo root, which is how CI runs it, `*` expands to `apps`, `packages`, `scripts`… and the hazardous implementation is green. So as written the row documents the intention; the thing that actually protects against the exposure is the new quoting guard, which does redden. Seeding a temp cwd with a file named `authentication` for that one row would make it a guard — worth it or not is your call, and your own comment already says the expansion happens at the call site rather than in the function, which is the honest version of this.

### On the `local` trap you avoided by layout

Worth keeping the distinction you drew: you would have written it that way regardless, so the file being right is not evidence the hazard was considered. What makes it checked is a two-shell table, and it needed one because `local` swallowing a status is a property of the shell rather than of the file — confirmed identical on bash 3.2.57 and bash 5.3.15, with a succeeding-command control on each.

### Verification, HEAD asserted in the same shell

```
HEAD                       0c0f4b4f3, tree clean by git status --porcelain
controls.test.sh           24 passed, 0 failed   exit 0
migrate.test.sh            68 passed, 0 failed   exit 0
git-sync.test.sh           16 passed, 0 failed   exit 0
shellcheck api-deploy.sh   4x SC2015, same set as bffd96987 - and the SC2086 disable is gone
shellcheck lib/controls.sh clean
```

The variadic-to-one-argument change is the better fix than `set -f`: it removes the exposure by construction rather than suppressing it at one call site, and it deletes a `disable` directive instead of adding one.

PR #2759, opened 2026-09-06
feat(backend): declare the controls /health/controls expects to report

### LINKED ISSUE #2758: /health/controls should declare the controls it expects, so the deploy gate can tell an old bundle from a missing control
Follow-up to #2756 / PR #2757. The deploy gate blocks only on a named control that *positively* reports `failed`, and ships when a named control is **absent** — deliberately, because a rollback deploys an older bundle that never records it. `lib/controls.sh` says so in the table:

> *named control ABSENT — ship. … This is the deliberate weak spot: a bundle that silently stops recording is not caught here.*

That weak spot is a property of the response shape, not of the problem. Today `/health/controls` reports only what **was** recorded, so "too old to record this control" and "should have recorded it and did not" are the same bytes.

### Proposal

Have the process declare what it expects to register:

```jsonc
{
  "status": "ok",
  "controls": [ { "name": "authentication", "state": "applied", "at": "…" } ],
  "expected": ["authentication", "stream-upload-policy"]
}
```

The deploy rule stays a positive report of failure, applied one level up:

| response | decision |
|---|---|
| `expected` names it, `controls` omit it | **BLOCK** — the bundle says it registers this and did not |
| `expected` absent from the body entirely | ship — an older bundle, which is what a rollback deploys |
| everything else | unchanged from #2757 |

**The absence of the `expected` key is itself the version marker**, so this closes the blind spot without touching the rollback case that motivated it.

### Checked precondition, and the invariant it creates

`expected`-vs-`controls` is only meaningful if every declared control is recorded before the port answers. It is, today:

```
apps/backend/src/main.ts
  await initQueues();
  await configureStreamUploadPolicy();   <- records stream-upload-policy (applied | skipped | failed)
  const app = createApp();               <- records authentication
  app.listen(PORT, …)
```

Both controls are recorded before `listen`, so a declared control missing from a response is a real fault and not a race. **That becomes an invariant this change depends on**: a control registered after `listen` — or from a background task — would be legitimately absent when the smoke probe runs 25 seconds in, and would block a healthy deploy. Worth a test asserting it rather than a comment.

### Staleness, the obvious failure mode of a declaration

A name left in `expected` after its control is deleted blocks every deploy forever. Mitigation is a test in the backend suite: boot the app in each configuration and assert the recorded set covers the declared set. That keeps the declaration honest in the place that can see both.

### Explicitly not

Deciding absence by grepping the deployed source (`git grep -q 'recordControl("authentication"'`). The deploy host has the source, so it will occur to someone. It is a substring test on source standing in for a fact about a running process, and it breaks on a rename that changes no behaviour.

### Sequencing

Backend first: the gate must keep treating a missing `expected` key as "ship", or the first deploy of the new backend blocks on every host still running the old one. The `lib/controls.sh` change is one table row and its tests, after the key exists.

aupyay:
## PR Checklist

`/health/controls` reports only what **was** recorded. So "this bundle is too old
to record `authentication`" and "this bundle registers `authentication` and did
not" are the same bytes, and the deploy gate in #2757 has to ship on the first
and stop on the second. It ships on both - the blind spot that gate documents in
its own table.

## What is the new behavior?

The process also declares what it registers:

```jsonc
{
  "status": "ok",
  "controls": [ { "name": "authentication", "state": "applied", "at": "…" } ],
  "expected": ["authentication", "stream-upload-policy"]
}
```

**The absence of the key is the version marker**, so it is sent
unconditionally - on the degraded response, on an auth-less boot, in every
configuration. A gate can then treat "declared and not reported" as a positive
failure while "no `expected` key at all" stays a rollback to an older bundle.

**A manifest, not a `declare()` beside each `recordControl`.** Co-locating them
makes drift impossible, and also makes the declaration vanish exactly when the
code path is skipped - which is the case a deploy gate most needs to see. The
cost is staleness: a name left here after its control is deleted would block
every deploy forever.

That cost is paid by tests rather than by a comment:

- **Honesty, both directions.** The app is booted in three configurations - auth
  configured, auth switched off with no Stream credentials, auth env incomplete
  with Stream rejecting - and the recorded set must cover the declared set in
  each. A separate test asserts the reverse, that nothing is recorded
  undeclared, so the manifest cannot silently stop describing the process.
- **Ordering.** A control recorded after `listen` would be legitimately absent
  when the smoke probe runs 25 seconds into a boot, and would block a healthy
  deploy. The test drives the real bootstrap - `main.ts` imported with only the
  heavy externals mocked - and snapshots the recorded set at the moment `listen`
  is called. That snapshot is also asserted **equal** to the declared set, not
  merely to contain it: the two honesty tests above drive `createApp` and
  `configureStreamUploadPolicy` by hand and can only see what those two record,
  so a control registered anywhere else in the bootstrap would pass all of them.
  A `recordControl` added to `main.ts` and left out of the manifest reddens this
  assertion and nothing else. Asserted on **both** ordering boots - the
  configured one and the kill-switch one - since `SUPERTOKENS_DISABLED` is the
  configuration most likely to grow a branch of its own.

The ordering test deliberately does **not** read `main.ts`. A source check
would stay green for any refactor that kept the two lines in order while
changing when they actually resolve - which is exactly what the second mutation
below does.

**Not in this change:** the `lib/controls.sh` table row. The gate must keep
treating a missing `expected` key as *ship* until the hosts are running a bundle
that sends one, so that is a separate PR after this lands.

### Verification

- `npx jest` full backend suite - 481 suites, 11485 tests, exit 0 (at
  `0458c26a6`)
- `pnpm --filter backend run lint` - exit 0, 0 errors, 25 pre-existing warnings
- `pnpm --filter backend run type-check` - exit 0
- Coverage of the changed files: `startup-controls.ts` 100% lines and branches;
  `app.ts` 99.22% lines, 92.85% branches, the two uncovered lines pre-existing

**Six mutations, each checked by which test went red rather than by exit code:**

| Mutation | Reddens |
|---|---|
| `configureStreamUploadPolicy` moved after `listen` | both ordering tests |
| the same call left un-`await`ed | the ordering test whose control is genuinely async - the other stays green, correctly, because the credentials-absent path records before its first `await` |
| manifest names a control nothing records | 6 tests |
| manifest drops a control that is recorded | the reverse-direction honesty test |
| `expected` omitted from the response | 3 tests |
| `expected` sent only when healthy | 2, including the degraded one |

### Two known limits, stated rather than left to be found

`listen` in the ordering test is a mock, so what is proven is that the
recordings precede the **call** to `listen` - not that they precede the socket
accepting a connection. Those are the same thing here because `app.listen` is
the last statement in `startServer` and nothing after it records. Narrowed in
review: moving a declared control *into the listen callback* is also caught,
because the snapshot is taken before the mock invokes it, so what remains
uncovered is only code after `app.listen(...)` returns.

Second, and less obvious: `jest.isolateModules` is only active while its
callback runs, and `void startServer()` executes synchronously up to its first
`await`. A control registered by a module that `main.ts` resolves **lazily**,
during the async tail, would land in the outer module registry - a different
`startup-controls` singleton with a different Map - and be invisible to this
test rather than weakly covered. Nothing does that today; every recorder is
reached through a top-level import, which resolves inside the isolated
registry. This limit was found by a reviewer's third mutation form and is the
same mechanism as the defect in the first draft of this test, pointing the
other way: that one *read* the outer instance, this one would *write* it.

**One defect found in my own test and worth stating.** The ordering test first
read the control reports from the module imported at the top of the file, while
`jest.isolateModules` gives `main.ts` a *fresh* registry - so it was reading an
instance nothing had written to and reported `[]` for every boot. It passed
`expect(listen).toHaveBeenCalled()` and failed only because the array was empty;
had the assertion been `not.toContain`, it would have passed forever while
measuring nothing. It now reads the instance the bootstrap actually writes to.

**And one about hygiene rather than correctness.** Importing the real `app` and
`main` pulls modules that construct a bullmq `Queue` at import, whose redis
client outlives the suite and logs into whatever runs next. Measured across full
runs: 34-38 late-log warnings on the untouched baseline, 44 and 78 with this
suite added, 32 with `bullmq` mocked at the library boundary. Nothing failed in
any of those runs - `481 passed` each time - but the suite should not be adding
to it.

## Related Issue(s)

Fixes #2758

### DIFF
--- a/apps/backend/src/app.ts
+++ b/apps/backend/src/app.ts
@@ -3,6 +3,7 @@ import rateLimit from "express-rate-limit";
 import fileUpload from "express-fileupload";
 import {
   getControlReports,
+  getExpectedControls,
   hasFailedControl,
   recordControl,
 } from "./config/startup-controls";
@@ -241,6 +242,11 @@ export function createApp() {
     return res.status(degraded ? 503 : 200).json({
       status: degraded ? "degraded" : "ok",
       controls: getControlReports(),
+      // What this bundle registers, always sent. A reader comparing the two
+      // lists can tell a control that is missing from one that never existed
+      // in this version - the absence of this key is the version marker, so it
+      // must not depend on anything.
+      expected: getExpectedControls(),
     });
   });
 
--- a/apps/backend/src/config/startup-controls.ts
+++ b/apps/backend/src/config/startup-controls.ts
@@ -14,6 +14,32 @@
  * This deliberately does not stop the boot. A Stream outage should not take the
  * API down - it should be VISIBLE.
  */
+/**
+ * The controls this process registers before it answers on its port.
+ *
+ * Reported alongside the reports themselves, because a response that lists only
+ * what WAS recorded cannot distinguish "this bundle is too old to record X"
+ * from "this bundle registers X and did not". Those are the same bytes, and the
+ * deploy gate has to ship on the first and stop on the second: absence of this
+ * key is what marks an older bundle, so it must never be conditional.
+ *
+ * A MANIFEST rather than a declare() call beside each recordControl. Putting the
+ * declaration next to the recording makes them impossible to drift apart - and
+ * also makes the declaration vanish exactly when the code path is skipped,
+ * which is the case a deploy gate most needs to see. The cost is that this list
+ * can go stale, and a name left here after its control is deleted would block
+ * every deploy forever, so test/config/expected-controls.test.ts boots the app
+ * in each configuration and asserts the recorded set covers this one.
+ *
+ * Anything added here must be recorded BEFORE app.listen - see that same file.
+ */
+export const EXPECTED_CONTROLS = [
+  "authentication",
+  "stream-upload-policy",
+] as const;
+
+export const getExpectedControls = (): string[] => [...EXPECTED_CONTROLS];
+
 export type ControlState = "applied" | "failed" | "skipped";
 
 export type ControlReport = {
--- a/apps/backend/test/app.test.ts
+++ b/apps/backend/test/app.test.ts
@@ -72,6 +72,7 @@ jest.mock("../src/config/auth-hooks", () => ({
 
 import { createApp } from "../src/app";
 import {
+  EXPECTED_CONTROLS,
   getControlReports,
   hasFailedControl,
   resetControlsForTest,
@@ -463,12 +464,16 @@ describe("createApp reports the authentication control", () => {
     const body = JSON.parse(controls.body) as {
       status: string;
       controls: Array<{ name: string; state: string; detail?: string }>;
+      expected: string[];
     };
 
     // Liveness is unchanged on purpose: the process is up, the control is not.
     expect(liveness.statusCode).toBe(200);
     expect(controls.statusCode).toBe(503);
     expect(body.status).toBe("degraded");
+    // Sent on the degraded response too: a reader comparing the two lists needs
+    // it most exactly when something is wrong.
+    expect(body.expected).toEqual([...EXPECTED_CONTROLS]);
     expect(body.controls).toContainEqual(
       expect.objectContaining({
         name: "authentication",
@@ -487,6 +492,7 @@ describe("createApp reports the authentication control", () => {
     const body = JSON.parse(controls.body) as {
       status: string;
       controls: Array<{ name: string; state: string }>;
+      expected: string[];
     };
 
     expect(controls.statusCode).toBe(200);
@@ -494,5 +500,21 @@ describe("createApp reports the authentication control", () => {
     expect(body.controls).toContainEqual(
       expect.objectContaining({ name: "authentication", state: "applied" }),
     );
+    expect(body.expected).toEqual([...EXPECTED_CONTROLS]);
+  });
+
+  // #2758: the deploy gate ships when a named control is absent, because a
+  // rollback deploys a bundle that never recorded it. This key is what lets the
+  // gate tell that apart from a bundle that should have recorded it and did
+  // not - so its ABSENCE is the version marker, and it must not be conditional
+  // on state, on configuration, or on anything else.
+  it("declares the controls it registers on every boot, including an auth-less one", async () => {
+    const app = createApp();
+
+    const controls = await request(app, { path: "/health/controls" });
+    const body = JSON.parse(controls.body) as { expected: string[] };
+
+    expect(body.expected).toEqual([...EXPECTED_CONTROLS]);
+    expect(body.expected).toContain("authentication");
   });
 });
--- a/apps/backend/test/config/expected-controls.test.ts
+++ b/apps/backend/test/config/expected-controls.test.ts
@@ -277,5 +277,9 @@ describe("every declared control is recorded before the port answers", () => {
     expect(recordedAtListen.sort()).toEqual(
       expect.arrayContaining([...EXPECTED_CONTROLS]),
     );
+    // Equality on this path too, not only the configured one. The kill switch
+    // is the configuration most likely to grow a branch of its own, and an
+    // undeclared control recorded only there would otherwise be uncovered.
+    expect(recordedAtListen.sort()).toEqual(getExpectedControls().sort());
   });
 });
--- a/apps/backend/test/config/expected-controls.test.ts
+++ b/apps/backend/test/config/expected-controls.test.ts
@@ -1,0 +1,275 @@
+/**
+ * The two things `expected` on /health/controls silently depends on.
+ *
+ * The deploy gate (scripts/deploy/lib/controls.sh) blocks when the response
+ * declares a control and does not report it. That is only a fault signal if
+ * (1) the manifest never names a control nothing records, and (2) everything it
+ * names is recorded before the port answers - a control registered after
+ * `listen`, or from a background task, would be legitimately absent when the
+ * smoke probe runs 25 seconds into the boot, and would block a healthy deploy.
+ *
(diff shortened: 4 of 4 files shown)


ankit-yc (dismissed):
Approving at `0458c26a6`. Verified on my own worktree rather than read: every mutation in the PR body re-derived and checked by the **name** of the failing test, plus five of my own. Baseline 23/23 either side of every run, `src/main.ts`, `src/app.ts`, `startup-controls.ts` and the test file restored from pristine copies between each.

### Your six, re-derived

| Mutation | Reddens | Matches the table |
|---|---|---|
| `expected` removed from the response | 3, by name | yes |
| `expected` only when not degraded | 2, including the degraded one | yes |
| manifest names a control nothing records | 6 | yes |
| manifest drops a recorded control | 1, the reverse-direction test | yes |
| `configureStreamUploadPolicy` after `listen` | both ordering tests | yes |
| the same call left un-`await`ed | **exactly 1**, the Stream-configured one | yes |

### Why the mutation that "did not take effect" came back green

Not a broken mutation — the same wrong-singleton mechanism as the defect in the test itself, one level out. Three forms of *the same* statement, an undeclared control recorded during the real boot before `listen`:

```
A  top-level import, recorded before createApp     1 failed  <- "has recorded every declared
                                                                control by the time listen is called"
B  inline require(), AFTER the first await          23 passed 0 failed
C  inline require(), BEFORE the first await         1 failed  <- same test, by name
```

B and C differ only in which side of `await initQueues()` they sit on. `void startServer()` runs synchronously up to its first `await`, so anything resolved before it lands in the registry `jest.isolateModules` installed — the one the snapshot reads. Anything resolved after it hits the **outer** registry, a different `startup-controls` singleton with a different `Map`.

C is what makes A worth anything: A alone says the assertion catches *something*, A-vs-B says a mutation can vanish, and only C names the property that decides. Without it, "inline `require` does not take effect" is a plausible wrong rule.

### Residual limit worth one sentence in the file

B is not only a mutation artifact. A control registered by a module that `main.ts` resolves **lazily during the async tail** of `startServer` is invisible to this test — not weakly covered, invisible. Nothing does that today, and it is the same class as the `listen`-mock caveat already written in, which is why it belongs next to it rather than in a follow-up.

Non-blocking, one line if you touch the file again: the new equality assertion is only in the first ordering test. Mutation A reddens **1 of the 2**, so an undeclared control recorded only on the auth-off path stays uncovered, and `SUPERTOKENS_DISABLED` is the configuration most likely to grow a path of its own.

### The listen-mock caveat is narrower than it reads

Moving `configureStreamUploadPolicy` **into the `listen` callback** reddens both ordering tests — the snapshot is taken before `mockListen` invokes the callback. So the shape that most looks like "after listen" to a reader is caught, and what remains uncovered is only code after `app.listen(...)` returns, which for a declared control is already red. The caveat is honest; the gap under it is smaller than the sentence suggests.

### Two of my own, and one checked negative

```
ordering test reads the OUTER instance (the defect, restored)   both ordering tests red
configureStreamUploadPolicy moved INTO the listen callback      both ordering tests red
```

And a trap that is **not** there: `jest.doMock("src/app", ...)` is registered inside `isolateModules`, so I checked whether it leaks to a test appended after the ordering block. It does not — an appended test gets the real `createApp` with a real `.use`. Scoped to that: a test appended to this file, after that describe.

### Verification, HEAD asserted in the same shell as every run

```
HEAD                          0458c26a6, tree clean by git status --porcelain
npx jest, full backend        481 suites / 11485 tests   exit 0
pnpm --filter backend lint    exit 0     0 errors, 25 pre-existing warnings
type-check                    exit 0     0 errors
Sonar quality gate (PR 2759)  OK, new_coverage 100.0, 0 duplicated, hotspots 100% reviewed
checks                        36 pass, 3 skipping, 0 failing
```

One thing another reviewer should not misread as a regression here: `npx jest` on the two changed files together prints *"Jest did not exit one second after the test run has completed"* and the runner hangs. I ran `app.test.ts` with an unrelated second file as the control and it hangs identically — **pre-existing in this suite, not introduced by this PR**, and the full suite exits 0 regardless.

Design is right: the aggregate is never read, absence of the key is the version marker so it is sent unconditionally, and the manifest's staleness cost is paid by tests rather than by a comment.

ankit-yc (approved):
Re-approving at `52b23dc5c`. The earlier approval was dismissed on push (`dismiss_stale_reviews_on_push` is true on the `dev` ruleset), so this is a fresh read of the delta plus a re-run of the whole matrix, not a re-tick.

**Your added assertion does what it claims, verified rather than counted.** The undeclared-control mutation — recorded during the real boot, before `listen`, via a top-level import so it resolves inside the isolated registry — now reddens **2 of 2** ordering tests:

```
every declared control is recorded before the port answers
  > has recorded every declared control by the time listen is called          RED
  > records them before listen even when every control is in a non-applied state  RED
```

It was 1 of 2 at `0458c26a6`, which is the gap I raised. Closed.

**Verification, HEAD asserted in the same shell as every run, tree clean by `git status --porcelain`:**

```
HEAD                          52b23dc5c
npx jest, full backend        481 suites / 11485 tests   exit 0
pnpm --filter backend lint    exit 0   0 errors, 25 pre-existing warnings
type-check                    exit 0   0 errors
```

Everything from my `0458c26a6` review still holds: the six mutations in the PR body all reproduce by failing-test name, the un-`await`ed one reddens exactly the Stream-configured ordering test and correctly leaves the credentials-absent one green, and the residual limit you wrote into the body — a control registered by a module `main.ts` resolves lazily in the async tail lands in the outer registry and is invisible here — is stated accurately.

PR #2763, opened 2026-09-06
fix(ci): block the cutover when a bundle declares a control and never reports it

### LINKED ISSUE #2761: fix(ci): the deploy gate ignores the `expected` controls declaration it was given
## The declaration has no consumer

`#2759` added an `expected` key to `/health/controls`:

```jsonc
{ "status": "ok", "controls": [ ... ], "expected": ["authentication", "stream-upload-policy"] }
```

`apps/backend/src/config/startup-controls.ts` states what it is for, in its own doc comment:

> Reported alongside the reports themselves, because a response that lists only what WAS
> recorded cannot distinguish "this bundle is too old to record X" from "this bundle
> registers X and did not". Those are the same bytes, and the deploy gate has to ship on
> the first and stop on the second.

Nothing in `scripts/deploy` reads it.

```
$ git grep -n expected -- scripts/deploy/lib/controls.sh scripts/deploy/api-deploy.sh
scripts/deploy/api-deploy.sh:285:  echo "smoke boot could not parse a request body ($POST_CODE, expected 404)" >&2
```

So the deploy gate added in `#2757` still ships on `authentication` being absent for either
reason, and `lib/controls.sh` names that as its deliberate weak spot:

```
#   named control ABSENT        ship. ... This is the deliberate weak spot: a bundle
#                               that silently stops recording is not caught here.
```

The data needed to close it now arrives in the response. This issue is the consumer.

## Is the absent case actually reachable, or is this hypothetical?

Reachable, and currently the state of both API hosts. `/health/controls` is unauthenticated
by design, so it can be read from outside:

```
devapi  200  {"status":"ok","controls":[{"name":"stream-upload-policy","state":"applied","at":"2026-09-04T22:21:40.663Z"}]}
api     200  {"status":"ok","controls":[{"name":"stream-upload-policy","state":"applied","at":"2026-09-02T18:03:21.052Z"}]}
```

No `authentication`, no `expected`, on either. Both processes booted days before `#2755`
merged, and `recordControl("authentication", ...)` runs on all three paths in `app.ts`, so a
post-`#2755` bundle cannot omit it. Those bodies are what a rollback to any ref before
`#2755` produces, and the gate must keep shipping them.

## Proposed rule

Absence alone stays a ship. Absence *contradicted by the bundle's own declaration* blocks:

| response | decision |
|---|---|
| no `expected` key | ship — an older bundle, exactly as today |
| `expected` is not an array | ship — same reasoning as an unparseable body |
| `expected` omits the blocking name | ship — this bundle does not claim to record it |
| `expected` names it, `controls` omits it | **BLOCK** — a positive report of failure |
| `controls` reports it `failed` | **BLOCK** — unchanged |
| `controls` reports it `applied`/`skipped` | ship — unchanged |

The third row is the one that keeps this a version marker rather than a membership check,
and it is the row a test has to separate: a gate that blocks whenever `expected` is present
and the name is missing passes every other row.

## Explicitly not this

Do not decide absence by grepping the deployed source for `recordControl("authentication"`.
It is a substring test on source standing in for a fact about a running process, and it
breaks on a rename that changes no behaviour. Raised and rejected during `#2757` review;
recorded here so it is rejected once rather than re-proposed.

## Scope

`scripts/deploy/lib/controls.sh` and `scripts/deploy/tests/controls.test.sh` only. No backend
change — `#2759` already ships the key.

aupyay:
Closes #2761.

`#2759` added an `expected` list to `/health/controls` so a reader could tell *"this bundle is
too old to record X"* from *"this bundle registers X and did not"* — the same bytes without it.
`startup-controls.ts` says so in its own doc comment, and names the deploy gate as the consumer.
Nothing read the key. This is the consumer.

## The rule

Absence alone still ships. Absence **contradicted by the bundle's own declaration** blocks.

| response | before | after |
|---|---|---|
| no `expected` key, control absent | ship | ship |
| `expected` not an array, control absent | ship | ship |
| `expected` omits the name, control absent | ship | ship |
| **`expected` names it, `controls` omits it** | **ship** | **BLOCK** |
| `controls` reports it `failed` | BLOCK | BLOCK |
| `controls` reports `applied`/`skipped` | ship | ship |

Row 4 is the deliberate weak spot `lib/controls.sh` documented and this closes. Rows 1–3 are why
it is not simply "block when the name is missing".

## Membership, not presence — and that distinction is measured

The question asked of each name is *did this bundle claim it*, never *does this bundle have the
key*. Presence-based sounds equivalent and is not: a bundle that publishes `expected` without the
name has not claimed it, and blocking there refuses a rollback from the other side.

Raised in review as a trap — `parsed.expected || []` collapsing **absent** into
**present-and-empty**, the same shape as `local` swallowing a status. Under a membership rule it
is inert, and I ran it rather than arguing it:

```
mutation: `: []` in place of `: null` (absent collapsed into empty)   30 passed, 0 failed
```

Green, because both make `declared.includes(name)` false. Recorded in the source so nobody
"fixes" it into a presence check, where that same substitution would be a real defect.

`Array.isArray` rather than truthiness **is** load-bearing and is a separate question: a string
also has `.includes`, and `"authentication".includes("authentication")` is true, so a malformed
key would block every deploy.

## Mutations — 6, each checked by the failing test NAME

Tree restored from a pristine copy between runs; baseline 30/30 either side; the landed line
printed before each run.

| Mutation | Reddens |
|---|---|
| declared branch unreachable (pre-#2761 behaviour) | 2 — the block row and the mixed row |
| membership ignored, any absence blocks once the key is present | **1 — `expected` omits the name** |
| truthiness instead of `Array.isArray` | 1 — `expected` present but not an array |
| absence blocks with no declaration at all | 7, incl. the rollback row |
| declaration ignored entirely, every absence emits | 8 |
| absent collapsed into present-and-empty | **0 — inert, see above** |

Row 2 is the separating input. It is the only one that fails a gate reading the key as a
membership list rather than a version marker, and every other row above passes such a gate.

**One mutation of mine was broken and I am reporting that rather than its number.** A first pass
at the `absent blocks` and `absent collapsed` rows came back **19 failed** — which reads as a
suite catching everything. It was shell escaping: `\&\&` reached the file as literal
backslash-ampersands, node threw, and stdout was empty for every case. The faithful versions are
7 and 0. A mutation that dies and a mutation the code survives are both "not the number you
meant", and here they were on opposite sides of the same run.

## Is the blocking row reachable?

Not today, and that is stated rather than implied. `EXPECTED_CONTROLS` is static, `app.ts`
records `authentication` on all three paths before `listen`, and `#2759`'s
`expected-controls.test.ts` asserts the recorded set covers the manifest at boot. This gate is
what sees it if any of that stops being true on a bundle that still deploys — which is the case
the endpoint was built for in the first place.

## The absent case, on the other hand, is live

`/health/controls` is unauthenticated by design, so both hosts can be read from outside:

```
devapi  200  {"status":"ok","controls":[{"name":"stream-upload-policy","state":"applied","at":"2026-09-04T22:21:40.663Z"}]}
api     200  {"status":"ok","controls":[{"name":"stream-upload-policy","state":"applied","at":"2026-09-02T18:03:21.052Z"}]}
```

No `authentication`, no `expected`, on either — both processes booted before `#2755` merged.
Read twice from two machines on different networks, byte-identical including both differing `at`
stamps, `Cache-Control: no-store`, cache-buster query changes nothing.

Those exact bytes are a test fixture here, pasted rather than built by `body()`. **Stated
honestly: it is subsumed** — no mutation of `lib/controls.sh` reddens it without also reddening
the `body()`-built absent row, and I could not construct one. It earns its place as the only
input in the file that the endpoint actually produced; every other case asserts against a builder
that could drift from it. It exercises the *ship* branch, so it is the control in the pair, not
the subject — the blocking body is one no host currently emits and has to be constructed.

## Also in this diff

`api-deploy.sh`: one log line, `FAILED to apply` → `did not apply`, because the message now heads
a list that can contain a control which was never recorded rather than one that tried and failed.
The sentence under it already read correctly for both.

## Verification

```
controls.test.sh    30 passed, 0 failed   (was 24)
migrate.test.sh     68 passed, 0 failed
git-sync.test.sh    16 passed, 0 failed
shellcheck -x       lib/controls.sh clean, exit 0
                    api-deploy.sh identical set to origin/dev under the same command
                    (4x SC1091, 4x SC2015, 4x SC2034) — compared against a
                    baseline extraction, not remembered
hooks               ran: gitleaks, secretlint, lint-staged; no "hooks did not run" line
HEAD                asserted in the same shell as every run
```

**One finding against myself, caught by keeping the baseline.** A comment I added *inside* the
`node -e` script introduced a new `SC2016` that `origin/dev` does not have — backticks inside a
single-quoted shell string. The file's own comment says no apostrophe and no backtick belong in
there, because either one ends the quoting and the rest of the file becomes shell; I wrote that
line and then broke it four edits later. Removed, and the script now asserts both characters are
absent at patch time. A single `shellcheck` run reporting findings would not have told me which
were mine.

### DIFF
--- a/scripts/deploy/api-deploy.sh
+++ b/scripts/deploy/api-deploy.sh
@@ -289,7 +289,7 @@ if [ "$POST_CODE" != "404" ]; then
 fi
 
 if [ -n "$CONTROL_FAILURES" ]; then
-  echo "smoke boot came up with a security control that FAILED to apply:" >&2
+  echo "smoke boot came up with a security control that did not apply:" >&2
   while IFS= read -r control_failure; do
     echo "  $control_failure" >&2
   done <<< "$CONTROL_FAILURES"
--- a/scripts/deploy/lib/controls.sh
+++ b/scripts/deploy/lib/controls.sh
@@ -36,11 +36,26 @@
 #   named control `skipped`     ship. Auth switched off is a deployment fact,
 #                               not a fault; treating it as one is the same
 #                               mistake as ignoring the endpoint, mirrored.
-#   named control ABSENT        ship. A ROLLBACK deploys an older bundle that
-#                               never records the control, so blocking on
+#   named control ABSENT,       ship. A ROLLBACK deploys an older bundle that
+#   body has no `expected`      never records the control, so blocking on
 #                               absence would refuse the deploy most needed to
-#                               work. This is the deliberate weak spot: a bundle
-#                               that silently stops recording is not caught here.
+#                               work. Absence alone is not evidence of a fault,
+#                               and the absence of `expected` is what marks the
+#                               bundle as older than the declaration itself.
+#
+#   named control ABSENT,       ship. This bundle declares what it records and
+#   `expected` omits it         does not claim this one. Same reasoning as the
+#                               row above, arrived at positively rather than by
+#                               a missing key: what the deploy asks to block on
+#                               is not what this bundle knows how to report.
+#
+#   named control ABSENT,       BLOCK. The bundle said it would record this
+#   `expected` NAMES it         control before it answered on its port, and then
+#                               did not. That is a positive report of failure
+#                               reconstructed from two facts instead of one, and
+#                               it is why startup-controls.ts reports `expected`
+#                               alongside the reports (#2759, #2761). It used to
+#                               be indistinguishable from the two rows above.
 #   endpoint unreachable        ship. /health is already the liveness gate and
 #   or not JSON                 it runs first; a second liveness check that
 #                               guesses at a body is not one.
@@ -93,11 +108,43 @@ deploy_blocking_control_failures() {
       const reports =
         parsed && Array.isArray(parsed.controls) ? parsed.controls : [];
 
+      // What this bundle says it registers before answering on its port. A
+      // body without the key predates the declaration, and a key of the wrong
+      // shape is a body we cannot read - both ship, for the same reason an
+      // unparseable body does. Only a well-formed list can turn an absent
+      // report into a failure.
+      //
+      // The rule below is MEMBERSHIP, not presence, and that is deliberate:
+      // the question asked of each name is "did this bundle claim it", never
+      // "does this bundle have the key". Presence and an empty list therefore
+      // decide identically, and an empty-array default in place of the null
+      // below changes no answer - measured, not assumed. A presence-based
+      // rule would be broken
+      // by exactly that substitution, which is why it is not one.
+      //
+      // Array.isArray rather than truthiness is load-bearing and separate: a
+      // string also has .includes, and "authentication".includes("authentication")
+      // is true, so a malformed key would block every deploy.
+      const declared =
+        parsed && Array.isArray(parsed.expected) ? parsed.expected : null;
+
       for (const name of required) {
         const report = reports.find(
           (candidate) => candidate && candidate.name === name,
         );
-        if (!report || report.state !== "failed") continue;
+        if (!report) {
+          // Absence decides nothing by itself. It becomes a failure only when
+          // the declaration published by this same bundle contradicts it.
+          //
+          // No apostrophe and no backtick anywhere in this script: it is the
+          // argument to node -e inside a SINGLE-quoted bash string, so either
+          // one ends the quoting and the rest of the file becomes shell.
+          if (declared && declared.includes(name)) {
+            process.stdout.write(name + ": declared but never reported\n");
+          }
+          continue;
+        }
+        if (report.state !== "failed") continue;
         const detail =
           typeof report.detail === "string" && report.detail
             ? " (" + report.detail + ")"
--- a/scripts/deploy/lib/controls.sh
+++ b/scripts/deploy/lib/controls.sh
@@ -49,16 +49,21 @@
 #                               a missing key: what the deploy asks to block on
 #                               is not what this bundle knows how to report.
 #
-#   named control ABSENT,       BLOCK. The bundle said it would record this
-#   `expected` NAMES it         control before it answered on its port, and then
-#                               did not. That is a positive report of failure
+#   named control ABSENT from   BLOCK. The bundle said it would record this
+#   a READABLE controls list,   control before it answered on its port, and then
+#   `expected` NAMES it         #                               did not. That is a positive report of failure
 #                               reconstructed from two facts instead of one, and
 #                               it is why startup-controls.ts reports `expected`
 #                               alongside the reports (#2759, #2761). It used to
 #                               be indistinguishable from the two rows above.
-#   endpoint unreachable        ship. /health is already the liveness gate and
-#   or not JSON                 it runs first; a second liveness check that
-#                               guesses at a body is not one.
+#   endpoint unreachable,       ship. /health is already the liveness gate and
+#   not JSON, or JSON whose     it runs first; a second liveness check that
+#   `controls` is not an array  guesses at a body is not one. The last of those
+#                               matters because of the BLOCK row below: a
+#                               declaration can only contradict a list that was
+#                               actually read, so an unreadable `controls` makes
+#                               `expected` inert rather than damning. A readable
+#                               EMPTY list is not this case - see that row.
 #   unnamed control `failed`    ship. See the Stream reasoning above.
 #
 #   node itself cannot run      STOP THE DEPLOY, and this is the one row that is
@@ -105,8 +110,21 @@ deploy_blocking_control_failures() {
         return;
       }
 
-      const reports =
-        parsed && Array.isArray(parsed.controls) ? parsed.controls : [];
+      // Whether the list of what WAS recorded can be read at all. This used to
+      // be an inert defensive fallback - nothing distinguished an unreadable
+      // list from an empty one, because only a report positively saying
+      // failed could block and neither shape has one. Reading absence made it
+      // load-bearing, and in the wrong direction: a body with a good
+      // declaration and a controls key that is a string or missing would have
+      // blocked, which is a body we cannot read stopping the deploy. Raised in
+      // review on #2763.
+      //
+      // An EMPTY array is a different answer and still blocks: that is a list
+      // we read, saying nothing was recorded, against a bundle that said it
+      // would. Readable-and-empty is a contradiction; unreadable is not an
+      // answer at all.
+      const readable = Boolean(parsed) && Array.isArray(parsed.controls);
+      const reports = readable ? parsed.controls : [];
 
       // What this bundle says it registers before answering on its port. A
       // body without the key predates the declaration, and a key of the wrong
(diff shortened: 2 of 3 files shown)


aupyay:
## Review round 2 — `3cfbddf7d` → `738e8bf2c`

Two findings from review, both taken. The first is a defect I introduced and could not see.

### The `controls` fallback became load-bearing in the wrong direction

`reports` and `declared` are sibling ternaries with opposite policies, four lines apart:

```js
const reports  = parsed && Array.isArray(parsed.controls) ? parsed.controls : [];   // -> BLOCK
const declared = parsed && Array.isArray(parsed.expected) ? parsed.expected : null; // -> ship
```

The first was an inert defence for as long as only a report positively saying `failed` could
block. **Reading absence made it decide a block** — and the table four lines above says an
unreadable body ships. Driven against the shipped function at `3cfbddf7d`:

```
expected NAMES it, no `controls` key      out=[authentication: declared but never reported]
expected NAMES it, controls is "nope"     out=[authentication: declared but never reported]
expected NAMES it, controls is null       out=[authentication: declared but never reported]
```

A malformed `controls` blocked while a malformed `expected` shipped. **Same body, opposite
answers, and the suite decided it in neither direction.**

**Answer: ship.** A declaration can only contradict a list that was actually read.

**A readable but EMPTY list still blocks**, and that is the whole distinction: `[]` is a list we
read saying nothing was recorded, from a bundle that said it would record this. Unreadable is not
an answer; readable-and-empty is a contradiction.

### The fixture builder expanded its name list against the cwd

`body_declaring` used `for name in $expected` — unquoted, so a pathname expansion. The same
hazard #2757 deleted an `SC2086` disable to remove from the call site, this time in **the builder
every other fixture in this file asserts against**. Nothing passes a glob today, so nothing is
wrong; a builder whose output varies with the working directory is how each machine covers
something different while every self-test passes. Now `read -ra`.

## New rows and the mutations that reach them

```
controls.test.sh   37 passed, 0 failed   (was 30 on this branch, 24 on dev)
```

| Mutation | Reddens |
|---|---|
| `declared` reads `parsed` again, not `readable` (the finding, reverted) | **3** — the three unreadable-`controls` ship rows |
| readable-and-empty treated as unreadable (`.length > 0`) | **1** — the empty-`controls` block row |
| builder split unquoted again | **1** — the glob-name builder row |
| declared branch unreachable | 3 — both block rows and the empty-list one |
| membership ignored | 1 — `expected` omits the name |
| truthiness instead of `Array.isArray` | 1 |
| absence blocks with no declaration | 10 |
| absent collapsed into present-and-empty | 0 — still inert, still deliberate |

Rows 1 and 2 are a pair: the first says an unreadable list cannot damn a bundle, the second says
an empty one still can. Either alone permits the wrong half.

**The builder row needed a vacuity control and it earns its place.** With an empty working
directory an unquoted expansion *also* leaves `*` alone, so the test would pass on the broken
builder. The seeded directory is asserted by name before it is used:

```
builder unquoted, seed present   1 red   "the fixture builder treats a glob-shaped name as a literal name"
builder unquoted, seed REMOVED   1 red   "the glob fixture directory has something to expand against"
                                         — and the builder row goes GREEN
```

The second run is the point: without the seed the broken builder passes, and the only thing that
says so is the guard over the seed.

## Verification

```
controls 37/37   migrate 68/68   git-sync 16/16   exit 0
suite re-run from /tmp and from $HOME    37/37 each — a fixture builder was the subject
shellcheck   lib/controls.sh clean exit 0
             controls.test.sh identical set to origin/dev at the real path (2x SC1091)
             api-deploy.sh unchanged from round 1
temp dir     removed and asserted absent by name inside the suite
HEAD asserted in the same shell as every run; tree clean
```

**Two of my own errors this round, both caught by guards rather than by review.** The
patch script asserts no apostrophe and no backtick inside the single-quoted `node -e` string —
it refused my first attempt at the new comment, which is the same mistake round 1 shipped as an
`SC2016`. And the first version of the seed assertion used `ls | wc -l`, adding **2x SC2012**
that `origin/dev` does not have; replaced with a check that names the file rather than counting,
which is the better assertion regardless.

ankit-yc (approved):
Approved at `738e8bf2c`.

The pair is right, and it is a better answer than the single row I asked for. `controls: []`
blocking is the half I would have got wrong: unreadable is not an answer, readable-and-empty is
a contradiction.

**Eight mutations re-derived on my own tree, each by failing test NAME, both files restored from
pristine between runs:**

```
`declared` reads `parsed` again (the finding, reverted)   3 red   the three unreadable rows only
readable-and-empty treated as unreadable (.length > 0)    1 red   the empty-list block row only
truthiness instead of Array.isArray                       1 red
membership ignored                                        1 red
absent collapsed into empty (`: []`)                      0       still inert
builder reverted to `for name in $expected`               1 red   the builder row
seed line removed                                         1 red   the seed row, and the builder
                                                                  row goes GREEN - the vacuity
                                                                  the seed assertion exists to catch
pristine                                                  37 passed, 0 failed
```

The first two are genuinely a pair: neither reddens the other's rows, so either alone would
permit the wrong half.

**Suite is cwd-independent** — `37/37` from the repo root, from `/tmp` and from `$HOME`, which
matters this round because a fixture builder was the subject.

**Composition with #2760, built rather than predicted.** I merged `4ee0cd106` into this head in a
throwaway worktree: clean auto-merge, no conflict markers, **merged suite `38 passed, 0 failed`**,
and #2760's source guard finds zero hits in the merged file.

---

**One thing to know rather than change: do not close the `GLOB_SEED` leak with a `trap`.**

The explicit `rm -rf "$GLOB_SEED"` leaves a narrow leak if `set -e` aborts between the `mktemp`
and the `rm`. The obvious fix is a `trap ... EXIT` — but #2760 already installs one for its own
`GLOB_CWD`, and a second `trap ... EXIT` **silently replaces the first**:

```
bash -c 'trap "echo FIRST" EXIT; trap "echo SECOND" EXIT; true'   ->   SECOND
```

In the merged file those are lines 213 and 355. Adding the obvious fix here would stop #2760's
temp directory from ever being cleaned, and nothing would report it. The explicit `rm -rf` is
the right shape; if the leak is ever worth closing it needs one shared cleanup function that both
call sites append to.

---

**Cosmetic, in the design record rather than the code.** The BLOCK row of the header table picked
up a stray comment marker and a column of padding when its label grew from two lines to three:

```
#   `expected` NAMES it         #                               did not. That is a positive report of failure
```

The sentence "control before it answered on its port, and then did not" is now split by a `#`
mid-line. Harmless to the shell, but that table is the artefact this PR exists to make readable.

```
controls 37/37   migrate 68/68   git-sync 16/16   exit 0
shellcheck   lib/controls.sh clean; controls.test.sh identical set to origin/dev at the real path
HEAD 738e8bf2c asserted in the same shell as every run, tree clean by git status --porcelain
```

Three checks were still in flight at approval — `Analyze (javascript-typescript)`, `grype` and
`Vulnerability scan (grype)`. Nothing red. Confirm `mergeStateStatus` reads CLEAN before merging.

Issue #2764, opened 2026-09-06
Before the next API deploy: the startup-controls gate has never run


aupyay:
The cutover gate on startup controls landed across #2755, #2759, #2757 and
#2763. `scripts/deploy/api-deploy.sh` now boots the bundle on a smoke port,
reads `/health/controls`, and stops before cutover if a named control reports
`failed` — or, since #2763, if the bundle declares a control in `expected` and
never reports it.

**The gate has been merged. It has never run.** Both API hosts currently serve
a body with no `authentication` control and no `expected` key, and the `at`
stamp on the one control they do report predates the bundle that records
`authentication`. So the processes running today started before any of this
existed, and every branch of the new rule is inert against them.

Two consequences worth deciding before the next deploy rather than during it:

1. **The next deploy is the first time the gate can block.** The smoke boot runs
   against the real environment, so if the auth environment on that host is
   incomplete, the deploy stops before cutover with the control body printed.
   That is the gate working exactly as designed — and it will look like a broken
   deploy to whoever is watching. Is that the behaviour we want on the first
   run, or should the first deploy be treated as a dry run of the gate?

2. **Nobody has yet observed the gate's own output on a real host.** Every
   result so far is from fixtures or from the two live bodies, both of which
   exercise only the ship branches. The blocking branches have never been
   exercised outside a test.

The question this issue is really asking: **who dispatches the first deploy
after this, and do they know what a gate-stop looks like?** Deploys are
`workflow_dispatch` against a live database with no blue/green, so this is a
human's call rather than something to schedule.

Not blocking anything. Filed so it is not only in a chat thread.

---

**Update — the dispatch history changes the operative instruction.**

Measured from the Actions API rather than reasoned: the Deploy API workflow has
**24 `workflow_dispatch` runs ever, of which 6 concluded `failure`**, and **zero
dispatches since the gate landed**. Every other run of that workflow is a
`pull_request` event where the deploy job itself is skipped.

So a red deploy is already a familiar sight on this workflow. **The first gate
stop will not stand out as unusual — it will look like the deploy failing
again**, and the natural response to a familiar failure is to re-dispatch.

That makes the instruction **"read the failing step before re-running"**, not
"expect something that looks broken".

The diagnosis is not only in the control body — each cutover stop prints the
failing control and its detail at the point of failure:

```
smoke boot came up with a security control that did not apply:
  authentication: failed (auth env incomplete)
NOT cutting over. Rollback sha: …
```

[...]

PR #2765, opened 2026-09-06
fix(ci): annotate every cutover stop so a gate stop is not read as a flaky deploy

aupyay:
Refs #2764.

## The problem, measured rather than argued

```
all 6 workflow_dispatch failures this workflow has ever had   step "Deploy over SSH"
  33627412850  33627289875  33627112116  33620397620  33567704148  32783760381
a cutover stop                                                step "Deploy over SSH"
```

`api-deploy.sh` runs inside that step, so **the step name, the job name and the red X of a gate
stop are identical to every failure this workflow has ever had.** The natural response to a
familiar failure is to re-run it rather than read it, and without an annotation the reason is
only inside an expanded log.

**All three stops, not just the startup-controls one.** `deploy-api.yml` annotates its three
preflight key checks (`:108`, `:129`, `:144`) and none of the three cutover stops
(`api-deploy.sh:280/287/297`) did. **The asymmetry was the finding** — fixing only the row this
line of work added would have left the other two silent for the same reason.

## Two decisions that are not obvious

**Stdout, with the human-readable lines left on stderr.** Workflow commands are documented as
reaching the runner over stdout; stderr is not mentioned. Raised in review: the stderr
precedents in `deploy-api.yml` **have never fired**, so they are an unexercised convention rather
than evidence — the nearest thing this file has to a precedent is the one thing that cannot
corroborate itself. There is a guard pinning the stream, not only the presence.

**Unconditional, not gated on `$GITHUB_ACTIONS`.** That variable is *not* forwarded across the
`ssh` in `deploy-api.yml` — the only one that crosses is `REQUIRE_PROMOTED_FROM`:

```
ssh … "chmod +x /tmp/yc-deploy/api-deploy.sh && REQUIRE_PROMOTED_FROM='$PROMOTION_BRANCH' …"
```

So gating on it would switch this off permanently and silently, which is the failure mode the
file exists to avoid. Run by hand, the line is one extra line of log beside the message it
duplicates.

## Stated limit: the over-SSH case is unverified and cannot be verified from here

The annotation is written by a script running on the remote host; its stdout returns through the
local `ssh` process, which is the step's stdout. **I believe the runner parses it. Nobody here
has observed that specific path. The first dispatch is what confirms it.**

The failure mode is benign: an unparsed `::error::` is a literal line in the log, which is
exactly where the message already is. So this is safe to ship unproven — but it is unproven, and
the stdout decision above matters precisely so that a first dispatch failing to annotate points
at the SSH boundary rather than at a stream choice made against the docs.

> **Corrected 2026-09-06, after merge.** This section originally read *"no `::error::` in this
> repo has ever produced an annotation — every failed run examined shows only `Process completed
> with exit code N`."* **That is false**, and it was an unscoped negative drawn from the runs I
> happened to examine.
>
> `promotion-guard.yml:105,115,120` are plain `echo "::error::…"` inside a `run:` step with no
> `>&2` — the same construct as the three cutover stops — and three failed runs each carry the
> workflow's own message as an annotation:
>
> ```
> 33304690565  job 99239026328   main only accepts merges from dev, but this PR's head is
>                                'claude/install-skills-a09751'. …
> 31963982037  job 95206177187   A fork branch (OctoBored/Yosemite-Crew:…) cannot merge into
>                                main. Target dev instead.
> 31719017835  job 94510951098   main only accepts merges from dev, but this PR's head is
>                                'test/guard-probe'. …
> ```
>
> **So a `run:` step's shell writing `::error::` to stdout annotates in this repository, and the
> shell path is the observed-working part.** The last paragraph above stands and is now
> established rather than merely argued: **the ssh boundary is the only unverified element**, and
> if the first dispatch does not annotate, that is where to look. Independently re-derived by a
> second reader. Detail in #2764.

## The half that keeps this from being critical

The diagnosis is not only in the printed body. Each stop already carries its reason at the point
of failure, and **a re-dispatch is idempotent** — the same bundle stops the same way and reprints
it. A re-run costs a person's second attempt, not the information.

## Guards and mutations

```
controls.test.sh   44 passed, 0 failed   (was 38)
migrate 68/68      git-sync 16/16        exit 0
```

| Mutation | Reddens |
|---|---|
| annotation removed from the smoke-boot status stop | 2 — its own row and parity |
| annotation removed from the request-body stop | 2 — its own row and parity |
| annotation removed from the startup-controls stop | 2 — its own row and parity |
| a **fourth** stop added with no annotation | **1 — parity only** |
| an annotation redirected to `>&2` | **1 — the stream row only** |
| the region extractor anchored on nothing | 4 — the canary and the three named rows |

**Row 4 is why parity is not subsumed** by the three named rows: they describe today, and a stop
added later is the case the guard exists for.

**Row 6 is why the vacuity canary earns its line, and the number that shows it is the one that
stayed green:**

```
region extractor matches nothing:
  the cutover region was actually extracted            FAIL
  the three named stop rows                            FAIL
  every cutover stop that exits emits an annotation    ok    <- 0 stops, 0 annotated IS parity
```

**A parity check over an empty set passes.** Without the canary, breaking the extractor would
report a completely unguarded file as guarded.

## Verification

```
shellcheck -x   api-deploy.sh    identical set to origin/dev at the real path (SC1091/2015/2034)
                controls.test.sh identical set to origin/dev at the real path (2x SC1091)
annotation streams, driven       stdout: the ::error:: line   stderr: the human line
the joined control list          single line, 0 newlines, no trailing separator, 1 and 3 entries
hooks ran                        gitleaks, secretlint, lint-staged, commitlint
HEAD asserted in the same shell as every run; tree clean
```

**Two of my own mistakes this round, both caught by a guard rather than by review.** A mutation
of the region extractor **failed to apply** and said so instead of returning a green — the assert
on the substitution count is the only reason that was visible. And I introduced **2x SC2016** by
spelling a literal `$CODE` inside a single-quoted awk pattern; rewritten with `-v` and `index()`,
which is also the form the per-stop guards already used.

### DIFF
--- a/scripts/deploy/api-deploy.sh
+++ b/scripts/deploy/api-deploy.sh
@@ -269,6 +269,31 @@ echo "  body parse -> $POST_CODE"
 CONTROLS_BODY="$(curl -s --max-time 10 "http://127.0.0.1:$SMOKE_PORT/health/controls" || echo '')"
 echo "  controls -> ${CONTROLS_BODY:-<unreachable>}"
 CONTROL_FAILURES="$(deploy_blocking_control_failures "$CONTROLS_BODY" "$DEPLOY_BLOCKING_CONTROLS")"
+# Every cutover stop below emits a GitHub Actions error annotation as well as
+# the human lines. Not decoration: all six workflow_dispatch failures this
+# workflow has ever had failed on the step "Deploy over SSH", and so does every
+# stop below, because this script runs inside that step. Step name, job name and
+# red X are therefore identical to a routine deploy failure, and the natural
+# response to a familiar failure is to re-run it rather than read it. An
+# annotation is what puts the reason on the run summary instead of only inside
+# an expanded log.
+#
+# STDOUT, and the human lines stay on stderr. GitHub documents workflow commands
+# as reaching the runner over stdout and does not mention stderr. The three
+# preflight key checks in deploy-api.yml are the nearest precedent in this file
+# and they write to stderr - but raised in review: not one of them has ever
+# fired, so that is an unexercised convention rather than evidence. The rest of
+# the repo (promotion-guard.yml, _migration.yaml, the release workflows) uses
+# stdout, which is also what the docs say.
+#
+# Emitted unconditionally rather than gated on $GITHUB_ACTIONS. That variable is
+# NOT forwarded across the ssh in deploy-api.yml - only REQUIRE_PROMOTED_FROM is
+# - so gating on it would switch the annotation off permanently and silently,
+# which is the failure mode this whole file exists to avoid. Run by hand, the
+# line is one extra line of log next to the message it duplicates.
+#
+# On stdout, not stderr, and single-line: annotation data may not contain a raw
+# newline, which is why the control list is joined below rather than looped.
 # Reports, never gates - the `|| true` is what makes that true, and it is
 # deliberate: this is a second signal for whoever reads the log, and POST_CODE
 # above is the gate.
@@ -278,6 +303,7 @@ SMOKE_PID=""
 sleep 2
 if [ "$CODE" != "200" ]; then
   echo "smoke boot failed - NOT cutting over. Rollback sha: $ROLLBACK_SHA" >&2
+  echo "::error::Smoke boot did not answer 200 (got $CODE). NOT cutting over. Rollback sha: $ROLLBACK_SHA"
   exit 1
 fi
 
@@ -285,6 +311,7 @@ if [ "$POST_CODE" != "404" ]; then
   echo "smoke boot could not parse a request body ($POST_CODE, expected 404)" >&2
   echo "- every GET probe passes a bundle that 500s on every write." >&2
   echo "NOT cutting over. Rollback sha: $ROLLBACK_SHA" >&2
+  echo "::error::Smoke boot could not parse a request body ($POST_CODE, expected 404) - every GET probe passes a bundle that 500s on every write. NOT cutting over. Rollback sha: $ROLLBACK_SHA"
   exit 1
 fi
 
@@ -295,6 +322,7 @@ if [ -n "$CONTROL_FAILURES" ]; then
   done <<< "$CONTROL_FAILURES"
   echo "- the process starts and /health answers 200 with the control absent." >&2
   echo "NOT cutting over. Rollback sha: $ROLLBACK_SHA" >&2
+  echo "::error::Security control did not apply: $(printf '%s' "$CONTROL_FAILURES" | tr '\n' ';' | sed 's/;/; /g; s/; $//'). NOT cutting over. Rollback sha: $ROLLBACK_SHA"
   exit 1
 fi
 
--- a/scripts/deploy/api-deploy.sh
+++ b/scripts/deploy/api-deploy.sh
@@ -322,7 +322,7 @@ if [ -n "$CONTROL_FAILURES" ]; then
   done <<< "$CONTROL_FAILURES"
   echo "- the process starts and /health answers 200 with the control absent." >&2
   echo "NOT cutting over. Rollback sha: $ROLLBACK_SHA" >&2
-  echo "::error::Security control did not apply: $(printf '%s' "$CONTROL_FAILURES" | tr '\n' ';' | sed 's/;/; /g; s/; $//'). NOT cutting over. Rollback sha: $ROLLBACK_SHA"
+  echo "::error::Security control did not apply: $(printf '%s' "$CONTROL_FAILURES" | awk 'NR > 1 { printf "; " } { printf "%s", $0 }'). NOT cutting over. Rollback sha: $ROLLBACK_SHA"
   exit 1
 fi
 
--- a/scripts/deploy/tests/controls.test.sh
+++ b/scripts/deploy/tests/controls.test.sh
@@ -593,14 +593,46 @@ done
 
 # Stream, not just presence. GitHub documents workflow commands as reaching the
 # runner over stdout and never mentions stderr, so an annotation redirected to
-# stderr may be a plain log line - which would be the quietest possible failure
-# for a change whose entire purpose is to be more visible. Raised in review,
-# along with the finding that the stderr precedents in deploy-api.yml have never
-# fired, so they are an unexercised convention rather than evidence.
-CUTOVER_ANNOTATIONS_ON_STDERR="$(grep -c '::error::.*>&2' <<< "$CUTOVER_REGION" || true)"
-check "the annotations go to stdout, where workflow commands are read" \
-  "0 redirected to stderr" \
-  "$CUTOVER_ANNOTATIONS_ON_STDERR redirected to stderr"
+# stderr may be a plain log line - the quietest possible failure for a change
+# whose entire purpose is to be more visible.
+#
+# Redirection is NOT line-local, and the first version of this guard was. Three
+# spellings put the same annotation on stderr and only one is on the same line
+# as it:
+#
+#   echo "::error::…" >&2          line-local, trailing
+#   >&2 echo "::error::…"          line-local, leading
+#   { echo "::error::…" ; } >&2    ENCLOSING - and this is the house style:
+#                                  deploy-api.yml:106-121 wraps all three of its
+#                                  preflight annotations in exactly this shape.
+#
+# So the form most likely to be written here is the one a line-local grep cannot
+# see, and someone writing it would be following the convention this file's own
+# comment names. Raised in review.
+#
+# The enclosure check tracks ONE level of brace group, which is all this region
+# has and all the house style uses. A group nested inside another group would
+# not be seen; that is a limit of a source guard, not a case anyone writes here.
+# It deliberately fires only on a group that CONTAINS an annotation, so a future
+# brace group wrapping ordinary human lines is not a false positive.
+CUTOVER_STDERR_INLINE="$(grep -cE '(::error::.*>&[12]|>&[12].*::error::)' \
+  <<< "$CUTOVER_REGION" || true)"
+
+CUTOVER_STDERR_ENCLOSED="$(awk '
+  /^[[:space:]]*\{[[:space:]]*$/ { inside = 1; buf = ""; next }
+  inside && /^[[:space:]]*\}[[:space:]]*>&[12]/ {
+    if (buf ~ /::error::/) { bad++ }
+    inside = 0; buf = ""; next
+  }
+  inside { buf = buf $0 "\n"; next }
+  END { print bad + 0 }
+' <<< "$CUTOVER_REGION")"
+
+check "no annotation is redirected to stderr on its own line" \
+  "0 inline" "$CUTOVER_STDERR_INLINE inline"
+
+check "no annotation is redirected to stderr by an enclosing group" \
+  "0 enclosed" "$CUTOVER_STDERR_ENCLOSED enclosed"
 
 # The parity guard, and it is not subsumed by the three above: a FOURTH stop
 # added later without an annotation reddens only this one. That is the case
--- a/scripts/deploy/tests/controls.test.sh
+++ b/scripts/deploy/tests/controls.test.sh
@@ -605,22 +605,36 @@ done
 #   { echo "::error::…" ; } >&2    ENCLOSING - and this is the house style:
 #                                  deploy-api.yml:106-121 wraps all three of its
 #                                  preflight annotations in exactly this shape.
+#   ( echo "::error::…" ) >&2      ENCLOSING, subshell. Same construct, different
+#                                  bracket; raised in review after the first
+#                                  enclosure check tracked braces only.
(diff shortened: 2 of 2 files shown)


ankit-yc (dismissed):
Approved at `8ebb3a330`.

The parity-over-an-empty-set canary is the right instrument and the whole matrix reproduces. One
finding on the stream guard — it pins one spelling of three, and the spelling it misses is the one
this repository already uses.

**Your six, re-derived, each mutation printed after landing, both files restored between runs:**

```
annotation removed from stop 1        2 red   its own named row + parity
annotation removed from stop 3        2 red   its own named row + parity
annotation redirected with >&2        1 red   the stream row only
a fourth stop added, unannotated      1 red   PARITY ONLY - not subsumed
region extractor anchored on nothing  4 red   canary + all three named rows
pristine                              44 passed, 0 failed
```

The canary earns its place: with the extractor broken, the parity row alone reports
`0 stops, 0 annotated` and passes. Arithmetic parity over an empty set is the same failure as a
grep-based wiring guard matching its own comment.

---

**Finding: `grep -c '::error::.*>&2'` is line-local, and the hazard is not.**

Two forms put an annotation on stderr with the guard green:

```
  {
    echo "::error::Smoke boot did not answer 200 ..."
  } >&2                                                    44 passed, 0 failed
  >&2 echo "::error::Smoke boot did not answer 200 ..."     44 passed, 0 failed
```

The brace-group form is not hypothetical. **It is exactly the shape `deploy-api.yml:106-121` uses
for its three `::error::` preflight blocks** — the precedent this PR's own comment cites three
lines above the guard. Someone grouping the annotation with the human lines it duplicates would be
following the house style, and the annotation would silently stop reaching the runner.

Two greps over the region you already extract close all three, verified against every form:

```bash
ANN_ON_STDERR="$(grep -cE '(::error::.*>&2|>&2.*::error::)' <<< "$CUTOVER_REGION" || true)"
GROUP_REDIRECT="$(grep -cE '^[[:space:]]*[})][[:space:]]*>&[12]' <<< "$CUTOVER_REGION" || true)"
```

```
                 shipped guard   proposed pair
pristine             0 (pass)      PASS
trailing >&2         1 (red)       FAIL   caught by both
{ ... } >&2          0 (GREEN)     FAIL   caught only by the pair
>&2 echo ...         0 (GREEN)     FAIL   caught only by the pair
annotation deleted   0 (pass)      PASS   correctly silent - that is the parity row's job
```

The last row matters: the pair does not steal the failure from the guard that names the cause.

**Not blocking, but I rate this higher than the boundary I declined to chase on #2760**, and for a
concrete reason: there the evading form was new control flow nobody writes, here it is the existing
convention in the same repository.

---

**Two things I checked rather than took.**

`GITHUB_ACTIONS` is genuinely not forwarded. The ssh command in `deploy-api.yml` sets exactly one
variable — `REQUIRE_PROMOTED_FROM='$PROMOTION_BRANCH'` — so gating the annotation on it would
switch it off permanently. Unconditional is right.

And the reason stdout works here at all: **the `ssh` invocation is bare — not piped, not captured
into a variable** — so the remote script's stdout is the step's stdout. Worth stating because my
own stdout recommendation would have been wrong if that call were `OUT="$(ssh ...)"`.

**Cosmetic, in the join:** `tr '\n' ';' | sed 's/;/; /g'` also spaces semicolons already inside a
detail, so `failed (a; b)` renders as `failed (a;  b)`. Only reachable through a control detail
containing a semicolon.

```
controls 44/44   migrate 68/68   git-sync 16/16   exit 0
shellcheck   both files identical finding sets to origin/dev
HEAD 8ebb3a330 asserted in the same shell as every run, tree clean by git status --porcelain
```

aupyay:
## Round 2 — `8ebb3a330` → `803affd8a`. This dismisses the approval; the guard I was asked for pinned one spelling out of three.

**The finding, reproduced before taking it.** `grep -c '::error::.*>&2'` is line-local and
redirection is not:

```
echo "::error::…" >&2          44 passed, 0 failed   caught
>&2 echo "::error::…"          44 passed, 0 failed   GREEN, on stderr
{ echo "::error::…" ; } >&2    44 passed, 0 failed   GREEN, on stderr
```

**The third is the house style.** `deploy-api.yml:106-121` wraps all three preflight annotations
in exactly `{ … } >&2`. So the form most likely to be written into this file is the one the guard
could not see — and whoever wrote it would be following the convention **this file's own comment
names**. That is the unexercised-precedent problem from round 1, pointed at the guard instead of
the source.

## Enclosure detection rather than a blanket group-redirect ban

The proposed pair was `(::error::.*>&[12]|>&[12].*::error::)` plus
`^[[:space:]]*[})][[:space:]]*>&[12]`. I took the first verbatim. **The second reddens any group
redirect in the region, including one that contains no annotation at all** — a false positive on
a change that keeps the annotations working, in a file where a `{ … } >&2` around ordinary human
lines is the established style.

So the second check tracks one level of brace group and fires **only when the group contains an
annotation**:

```
                          shipped   line-local pair   this PR
trailing  >&2               red           red           red   (inline row)
leading   >&2              GREEN          red           red   (inline row)
{ … } >&2                  GREEN         GREEN          red   (enclosure row)
{ … } >&2 with NO           pass          pass          pass  <- the false-positive control
  annotation inside                       FAIL                   under the blanket grep
```

**One level, and that is stated in the source as a limit** — a group nested inside another group
is not seen. That is a limit of a source guard, not a case this region writes.

## Full matrix at `803affd8a`

```
controls.test.sh   45 passed, 0 failed   (was 44, was 38 on dev)
migrate 68/68      git-sync 16/16        exit 0
```

| Mutation | Reddens |
|---|---|
| annotation removed, each of the three stops | 2 each — own row and parity |
| a **fourth** stop added, unannotated | **1 — parity only** |
| annotation redirected `>&2`, trailing | **1 — the inline row only** |
| annotation redirected `>&2`, leading | **1 — the inline row only** |
| annotation wrapped in `{ … } >&2` | **1 — the enclosure row only** |
| a `{ … } >&2` group containing **no** annotation | **0 — control, correctly silent** |
| region extractor anchored on nothing | 4 — the canary and the three named rows |

The last row is still the one worth reading, and the two new guards join the same shape: with the
region empty, **`0 inline`, `0 enclosed` and `0 stops, 0 annotated` all pass.** Three arithmetic
checks over an empty set, all green, and only the canary says the file is unguarded.

## Also in this commit

The control list is joined with `awk 'NR>1{printf "; "}{printf "%s",$0}'` rather than
`tr '\n' ';' | sed 's/;/; /g'`. The sed spaced semicolons **already inside a detail**, so
`failed (a; b)` rendered as `failed (a;  b)`. Now the separator is the only thing inserted:

```
one entry     authentication: failed (auth env incomplete)
two entries   authentication: failed (a; b); stream-upload-policy: declared but never reported
newlines      0 in both
```

## Verification

```
shellcheck -x   api-deploy.sh    identical set to origin/dev at the real path
                controls.test.sh identical set to origin/dev at the real path (2x SC1091)
hooks ran       gitleaks, secretlint, lint-staged, commitlint
HEAD asserted in the same shell as every run; tree clean
```

**And a correction I would not have made without review.** The reason stdout works here at all is
that the `ssh` invocation is bare — not piped, not captured — so the remote script's stdout is the
step's stdout. **That was true by luck rather than by checking** when the stream was chosen; had
that call been `OUT="$(ssh …)"`, moving the annotation to stdout would have moved it somewhere
that reaches nothing at all. It is right, and it is now right for a reason someone looked at.

ankit-yc (dismissed):
Approved at `803affd8a`.

Your objection to my second grep is right and the control proves it: a brace group wrapping only
human lines is `45 passed` here and would have been a false positive under
`^[[:space:]]*[})][[:space:]]*>&[12]`. Containment is the correct question and I was matching on
the closing line alone.

**The matrix reproduces, and the control is the row that matters:**

```
trailing >&2                     1 red   the inline row only
leading  >&2                     1 red   the inline row only
{ ... ; } >&2 on one line        1 red   the inline row (the >&2 is on the same line)
{ ... } >&2 multi-line           1 red   the ENCLOSURE row only
{ ... } >&2 with NO annotation    0      correctly silent - your control, reproduced
pristine                         45 passed, 0 failed
```

---

**One thing to take if you want it, and it is one character class.** The awk tracks `{` and `}`
only, so the subshell spelling of the same construct evades:

```
  (
    echo "::error::Smoke boot did not answer 200 ..."
  ) >&2                                   45 passed, 0 failed
```

Measured rather than read - the annotation really is on stderr, with a plain `echo` as the
positive control on the same probe:

```
plain echo            stdout=[::error::X]  stderr=[]
( ... ) >&2           stdout=[]            stderr=[::error::X]
```

Widening the two anchors to `[({]` and `[)}]` closes it and changes nothing else, including your
control:

```
                          shipped   widened
{ ... } >&2                  1         1
( ... ) >&2                  0         1     <- closed
{ ... } >&2, NO annotation   0         0     <- control still silent
trailing / leading / inline  0         0     <- still the inline guard's job
pristine                     0         0
```

The containment test does the work either way; only the bracket characters change.

**Two spellings I am NOT asking you to chase, stated so the boundary is written down.** Both put
the annotation on stderr and both are green under either version:

```
{ echo "::error::..."          the opening brace is not alone on its line, so the
  echo "second" ; } >&2        one-level tracker never opens

exec >&2                       redirects the whole shell; takes out all three
echo "::error::..."            annotations at once and is invisible to a per-site guard
```

Neither is a formatting anyone writes here, and `exec >&2` in this region would be a deliberate
act rather than a tidy-up. A source guard that reads lines cannot follow a shell's file
descriptors, and that is the boundary rather than a hole.

**Marginal:** `>&[12]` in the inline grep also reddens `echo "::error::..." >&1`, which is a no-op
to stdout and correct. Nobody writes it; noting it only because the row's name says stderr.

**The join is fixed at the cause.** `awk 'NR > 1 { printf "; " }'` inserts only the separator -
driven on one, two and a semicolon-bearing detail, `failed (a; b)` now renders as itself.

**The canary getting weaker as guards accumulate is the observation I would keep from this PR.**
With the region empty, `0 inline`, `0 enclosed` and `0 stops, 0 annotated` all pass - three
independent guards, every one of them satisfied by the thing they guard not existing, and one
canary standing between that and a green suite over an unguarded file.

```
controls 45/45   migrate 68/68   git-sync 16/16   exit 0
shellcheck   both files identical finding sets to origin/dev
HEAD 803affd8a asserted in the same shell as every run, tree clean by git status --porcelain
```

ankit-yc (approved):
Approved at `74e70377d`. The character class is taken verbatim and both false-positive controls
hold, including the new bracket. Your `exec` row was worth taking and your reason for it is the
sharper half: it is not a site, so every per-site guard is silent by construction.

**Reproduced, mutation printed after landing, both files restored from pristine between runs:**

```
trailing >&2 / leading >&2      1 red each   inline row only
{ ... } >&2                      1 red        enclosure row
( ... ) >&2                      1 red        enclosure row - closed
either bracket, NO annotation    0            controls, still silent
exec >&2 in api-deploy.sh        1 red        the new row
pristine                         46 passed, 0 failed
```

---

**Three spellings still evade the `exec` row, and one of them is the scope you just moved out of,
one level further.** All four suppress stdout, measured with a plain `echo` as the control:

```
plain echo (CONTROL)   stdout=[::error::X]
exec >&2               stdout=[]      caught
exec 1>&2              stdout=[]      46 passed   numbered fd, as idiomatic as the bare form
exec >/dev/stderr      stdout=[]      46 passed   path form
exec >&2 in lib/       stdout=[]      46 passed   SOURCED, so it applies to the whole shell
```

`api-deploy.sh:71,73,75` sources `lib/git-sync.sh`, `lib/migrate.sh` and `lib/controls.sh`. A
redirect in any of them applies from that point onward in the same shell - so the guard reads the
file that would be *affected* rather than every file that can *cause* it. That is the same
observation you made about the region, moved out by one: `DEPLOY_CODE` is not the scope of a shell
redirect either.

One pattern over `api-deploy.sh` plus `lib/*.sh` catches all four and is `0` on pristine:

```
grep -cE '^[[:space:]]*exec[[:space:]]+[0-9]*[><]'

                   shipped   widened + lib
exec >&2              1            1
exec 1>&2             0            1
exec >/dev/stderr     0            1
exec >&2 in lib/      0            1
pristine              0            0
```

Stated cost: it also flags file-descriptor juggling that never touches stdout, `exec 3>&1` being
the obvious one. There is no `exec` anywhere in those four files today, so it is `0` either way.

**Not blocking, and I would take it only if you are in the file for another reason.** The returns
here are now a guard protecting a guard protecting an annotation whose primary mechanism - whether
a workflow command written over an ssh connection is parsed at all - is still unverified and stays
that way until a dispatch. Three spellings of a line nobody has written are a smaller risk than
the one already accepted and documented.

```
controls 46/46   migrate 68/68   git-sync 16/16   exit 0
shellcheck   both files identical finding sets to origin/dev at the real path
HEAD 74e70377d asserted in the same shell as every run, tree clean by git status --porcelain
```

PR #2768, opened 2026-09-06
chore(repo): ignore derivatives of .env, not just its canonical names

aupyay:
## What

`.gitignore` covered `.env`, `.env.local` and `.env.*.local`. It did not cover `.env.bak`,
`.env.backup`, `.env.old` or `.env.save`.

A derivative is what `cp .env .env.bak` produces, which is the ordinary way somebody preserves a
working config before editing it. **Two such files were found untracked and unignored in a working
tree of this repo, and `git add -A` in that tree would have staged both.**

## Nothing has been exposed — checked, not assumed

```
tracked on origin/dev                                  0
tracked on origin/main                                 0
ever added, all refs, --diff-filter=A -- '*.env.bak'   0
```

Latent, not an incident. No rotation is needed and neither file was read. This closes the class
rather than the instance.

## Why `.env.*` plus `!.env.example`

Exactly three tracked files have a basename starting with `.env`, and all three are
`.env.example`. `apps/mobileAppYC/ios/.xcode.env` is tracked and does **not** match `.env.*`.

A `.gitignore` rule cannot untrack what is already tracked, so the existing samples were never at
risk either way — but without the re-allow a *future* `.env.example` could not be added, and the
intent would be unstated.

## Driven, before and after, in this worktree

```
BEFORE (dev's rules)   git add -An stages   .env.bak  .env.backup  .env.old  .env.save
                                            apps/backend/.env.bak
                                            packages/database/.env.bak
AFTER                  git add -An stages   only .gitignore

samples                apps/backend/.env.example                            visible
                       apps/frontend/.env.example                           visible
                       apps/mobileAppYC/config-templates/env/.env.example   visible
                       apps/mobileAppYC/ios/.xcode.env                      visible
new sample             apps/probe/.env.example (new directory)              visible
```

All six probe files were removed afterwards and verified absent by name.

## Scope

`.gitignore` only. `.env.*.local` is now subsumed by `.env.*` and left in place rather than
removed, to keep the diff to the thing being fixed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,13 @@ dist
 dist-ssr
 *.local
 .env
+# Any derivative of a real env file, because `cp .env .env.bak` before an edit is the
+# ordinary way somebody preserves a working config - and `.bak`, `.backup`, `.old` and
+# `.save` matched none of the rules that used to be here. Two such files were found
+# untracked and unignored in a working tree; `git add -A` would have staged both.
+.env.*
+# Sample files are the exception and stay visible, including ones added later.
+!.env.example
 .env.local
 .env.*.local
 .cache


ankit-yc (approved):
Approving. Driven in a detached worktree at `8c07bdf69` rather than read, with both
directions and a negative control.

**After this PR — every derivative ignored, every sample and `.xcode.env` still visible:**

```
.env.bak .env.backup .env.old .env.save        IGNORED
.env.test  .env.example.bak                    IGNORED   (not in the PR body, same rule)
apps/backend/.env.example                      visible
apps/frontend/.env.example                     visible
apps/mobileAppYC/config-templates/env/.env.example   visible
apps/mobileAppYC/ios/.xcode.env                visible
apps/probe/.env.example  (NEW dir, new file)   visible

git add -An  ->  add 'apps/probe/.env.example'      <- the only thing stageable
```

**Before, with `origin/dev`'s rules, same probes in the same tree:**

```
git add -An  ->  .env.bak  .env.backup  .env.old  .env.save  .env.test
                 .env.example.bak  apps/probe/.env.bak  apps/probe/.env.example
```

Nine entries down to one. Tracked-file check at this head: exactly three files whose
basename starts with `.env`, all `.env.example`, so nothing currently tracked changes
status. `apps/mobileAppYC/ios/.xcode.env` is tracked and does not match `.env.*` —
catching that is the difference between this and a fix that breaks the iOS build.

**The one-character near-miss worth naming, because it is what most people would type.**
`.env*` — without the dot — would also have swallowed `.envrc`, which direnv users keep
and which nothing here ignores today. Verified: `.envrc` stays visible under `.env.*`.
Negative control `notenv.bak` also stays visible, so the pattern is anchored at the
basename rather than matching anywhere.

**Two non-blocking notes.**

1. The re-allow covers the exact name `.env.example` only, so `.env.test.example` and
   `.env.sample` are ignored. Both are plausible future *sample* names and neither
   exists today. If one is ever wanted, extend the re-allow rather than reaching for
   `git add -f` — the failure mode is a sample file that silently never gets committed.

2. ~~`.env.local` (line 18) and `.env.*.local` (line 19) are now fully subsumed by
   `.env.*` at line 15. Harmless, and leaving them documents the original intent, but
   they read as live rules and are not.~~

   **Corrected after review.** The last clause is wrong: they *are* live rules that
   currently have nothing to do. Keep them. If `.env.*` is ever narrowed or removed,
   `.env.local` is still covered by its own line; deleting them converts three
   independent protections into one, over the file most likely to hold a real
   credential. Redundancy in an ignore file is defence in depth — the clutter reading
   is right for code and wrong here.

Checks: 21 passing, 8 skipping, zero non-passing — `BLOCKED` was the review requirement
alone. Probe files removed; `git status --porcelain` empty afterwards.

aupyay:
Merged as `1203d952b`. Recording the limits raised in review so they are on the artefact rather
than only in the originating thread.

**Verified on `dev` after the merge, from my own extraction:**

```
.env.bak .env.backup .env.old .env.save      IGNORED
apps/backend/.env.bak                        IGNORED
.envrc                                       visible
notenv.bak                                   visible
all three .env.example                       visible
apps/mobileAppYC/ios/.xcode.env              visible
```

**`.env.*` versus `.env*` is the correctness of this change and it was not stated anywhere.** A
bare `.env*` also matches `.envrc`, direnv's config, which nothing here ignores. The dot anchors
the pattern to a real suffix. Neither the author nor the reviewer who proposed the rule reasoned
about it at the time — it was right by construction rather than by argument, which is worth
having in the record.

**Stated limit 1 — the re-allow is the exact name only.** `.env.test.example` and `.env.sample`
would both be ignored. Neither exists today; the failure mode is a future sample file that is
silently never committed. A measured half-fix exists if it ever matters:

```
                    !.env.example only    plus !.env.*.example
.env.test.example   IGNORED               visible
.env.sample         IGNORED               IGNORED     (does not end in .example)
.env.example.bak    IGNORED               IGNORED     (the widening reopens nothing)
```

**Stated limit 2 — `.env.local` and `.env.*.local` are now subsumed by `.env.*`, and they were
deliberately left in place.** They read as dead rules but they are live rules that currently have
nothing to do: if `.env.*` is ever narrowed or removed, `.env.local` remains covered by its own
line. Deleting them would convert three independent protections into one, over the file most
likely to hold a real credential. Redundancy in an ignore file is defence in depth.

aupyay:
**One operational note that belongs here rather than in a thread: this rule does not take effect
in a worktree until that worktree is updated past `1203d952b`.**

Measured after the merge, on a machine with 19 worktrees of this repo:

```
worktrees containing 1203d952b                 0 of 19
YC-qa-auth (HEAD f455595d4, predates the fix)
  .gitignore in that checkout                  .env  .env.local  .env.*.local   (the old three)
  git check-ignore .env.bak                    not ignored
  git add -An                                  add 'apps/backend/.env.bak'
                                               add 'packages/database/.env.bak'
```

So the class is closed on `dev` and closed in nothing that already exists on disk. The two files
this PR was filed about are still stageable in the tree that holds them. Nothing is tracked and
nothing has ever been committed — that still holds — but "the exposure is closed" is a statement
about the repository and not about any checkout.

**A second consequence, raised in review, which matters for anyone auditing later.** The sweep
that found these files used the predicate *untracked AND not-ignored*. After a worktree updates,
that predicate returns zero for exactly the files it exists to find — the files stay on disk and
become invisible to it. The two states will coexist on every machine for as long as older
worktrees live:

```
tree at/after 1203d952b   exposure closed,  untracked-and-not-ignored sweep BLIND
tree before it            exposure open,    that sweep still works
```

The dangerous transition is a tree that updates while still holding such a file: the exposure
closes and the audit goes blind in the same instant, and nothing announces either. **Sweep the
filesystem — `find` by name, ignore-status independent — and prove the detector fires by planting
a file before trusting a zero.**

PR #2774, opened 2026-09-06
fix(repo): raise the nodemailer and js-yaml override floors

aupyay:
## What

Three `pnpm.overrides` entries had stopped admitting patches, holding two packages below their fixes.

```
- "js-yaml@4": "4.3.1",                       exact pin, one patch below the fix
+ "js-yaml@3": ">=3.15.2 <4",                 NEW: the 3.x line had no override at all
+ "js-yaml@4": ">=4.3.2 <5",
- "nodemailer": "9.0.5",                      NO version selector: captured every request
+ "nodemailer@<9.1.1": ">=9.1.1 <10",
```

## Why it had not been fixed by an ordinary bump

Aikido reports three advisories against `nodemailer` 9.0.5 (patched 9.1.0 / 9.1.1, due 2026-09-09) and one against `js-yaml` naming 3.15.2 and 4.3.2 as patched. **None was reachable while the overrides pinned exact versions** - a bump PR cannot pass a pin, and neither can Dependabot.

And the obvious remedy is the harmful one:

```
supertokens-node@24.0.3   declares   nodemailer: ^8.0.2
```

With **no** override the tree resolves 8.x, so deleting the rule is a downgrade dressed as a fix. Raising the floor keeps the control and takes the patch. `nodemailer` is transitive through `supertokens-node`, which is the auth mailer path - verification and password-reset mail.

## Two causes, one commit

- **frozen pin** - `nodemailer`, `js-yaml@4`: the rule exists and no longer admits patches.
- **absent override** - `js-yaml@3`: no rule at all, simply stale at 3.15.1.

They are fixed together because one lockfile regeneration serves both; the commit message separates them.

## Acceptance check

The condition for a dependency change on a 116-override graph is that the diff is reviewable as one package's fix. Comparing **all 1816 resolved entries** in `pnpm-lock.yaml` before and after:

```
nodemailer  9.0.5  -> 9.1.1
js-yaml     3.15.1 -> 3.15.2
js-yaml     4.3.1  -> 4.3.2
anything else:  NOTHING
```

Six resolution lines, three packages, no other package's resolution changed.

## Verification

```
pnpm install                      full, so husky's prepare ran and the repo gates were armed
pre-commit                        lint-staged, commitlint, secretlint  PASSED
pre-push                          turbo run lint + builds               PASSED
core.hooksPath                    captured before, restored after, asserted equal
```

The first push attempt failed on `@prisma/client has no exported member PrismaClient` - the Prisma client had never been generated in this fresh worktree. Fixed by running `prisma:generate`, **not** by bypassing the hook; the gate then ran and passed.

- **The other 113 exact-version overrides.** They are the same shape and are a hazard for the *next* advisory against each package, not for this one - an exact pin is precisely what a regeneration cannot move. Folding them in would be a dependency sweep and would destroy the reviewability of a security fix.
- **Three dead selectors** (`ajv@7`, `glob@11.0.3`, `jws@3.2.2`) whose target versions no longer appear in the graph. Each has a live sibling rule holding the same package, so they are clutter rather than lapsed controls. Tidying, and not for this diff.
- **The undeclared-major question**: the override forces `nodemailer` above the `^8.0.2` that `supertokens-node` declares. That was true before this change and after it, since the move is inside 9.x. Worth owning separately - it surfaces at runtime rather than at build time.


aupyay:
Closing as a duplicate of #2775, which does the same three override changes with a stronger verification: it classifies every changed line of the lockfile (`js-yaml` 29, `nodemailer` 4, integrity 6, anything else 0) rather than only the resolved-version entries, and it records the intermediate re-resolve proving the `js-yaml@3` rule was necessary rather than inferred.

We built this simultaneously without either of us checking for open PRs first - and neither of us found #2762, the Aikido autofix that has covered these packages since 02:17Z. That one is the better base still (8 packages against 3) and its problem is the *shape* of its values, which I have raised there rather than duplicating a third time.

Nothing here is lost: the branch `fix/override-pin-inversion` stays if anyone wants the diff, and the two findings it carried are recorded where they act -
- the floor-raise-versus-pin form, and the missing `js-yaml@3` rule, on #2762;
- the classifier for the remaining 115 pins, on YosemiteCrew/SuperAdmin#322.

Issue #321, opened 2025-06-11
Create Page Admin Empty Dashboard


dunexploration-frontend:
**Is your feature request related to a problem? Please describe.**
When an admin logs in and no data has been added yet (projects, users, tasks, etc.), the dashboard currently displays a blank screen or a broken layout. This creates confusion and lacks guidance for first-time users or new administrators.

**Describe the solution you'd like**
Implement an "Empty State" design for the Admin Dashboard. This page should include:

- A welcoming message (e.g., "Welcome, Admin 👋")
- A short description of what this page is for
- A visual illustration or icon (optional, but helpful)
- Clear CTA buttons like:

- “Add New Project”
- “Invite Team Members”
- “Configure Settings”

- Proper layout structure so the page doesn’t look broken or unfinished

This helps guide new users/admins through the initial steps of setting up their dashboard.

**Describe alternatives you've considered**

- Leaving the dashboard blank (not user-friendly)
- Using generic placeholders without clear direction (still confusing)
- Redirecting to other setup pages (adds more complexity)

PR #2775, opened 2026-09-06
fix(repo): raise the nodemailer and js-yaml overrides off the versions they pinned

aupyay:
## What

Three packages move and nothing else does.

```
/js-yaml@3.15.1   -> /js-yaml@3.15.2
/js-yaml@4.3.1    -> /js-yaml@4.3.2
/nodemailer@9.0.5 -> /nodemailer@9.1.1
```

Five advisories, four on nodemailer and one on js-yaml, due 2026-09-09 and 2026-09-11. None of them was reachable by a bump, because an override held each package at the version being reported.

## nodemailer - the override was the whole reason it sat

```
- "nodemailer": "9.0.5",
+ "nodemailer@<9.1.1": ">=9.1.1 <10",
```

The old rule had no version selector at all, so it captured every request from every dependent and resolved it to exactly 9.0.5. No bump PR could pass it and Dependabot could not either.

Deleting it would have been worse, not better:

```
npm view supertokens-node@24.0.3 dependencies.nodemailer  ->  ^8.0.2
```

With no override the tree resolves 8.x. The override is what lifts nodemailer into 9 in the first place, so raising its floor keeps the control and takes the patch. This is the same fix already merged on the SuperAdmin panel as #321 there, against the same parent package.

## js-yaml - one advisory row, two different causes

The scanner reports `installed=4.3.1`, `patched=['3.15.2', '4.3.2']`. Both lines are in this tree and they got there differently:

```
- "js-yaml@4": "4.3.1",
+ "js-yaml@4": ">=4.3.2 <5",
+ "js-yaml@3": ">=3.15.2 <4",
```

- **4.x** was pinned at 4.3.1 by an exact override, one patch below the fix.
- **3.x** resolved at 3.15.1 with **no override covering it at all** - simply stale.

This was checked rather than assumed. I raised the `@4` rule first and re-resolved on its own: `js-yaml@4.3.1 -> 4.3.2` and `3.15.1` did not move. So a fix that trusted the single reported row would have gone green, closed it, and left a vulnerable 3.x in the tree with nothing reporting it. The `js-yaml@3` rule is new and exists only because the measurement said it was needed.

The two are kept in one commit because the commit message separates the causes and the whole change is three lines of `package.json`; happy to split them if a reviewer would rather see them independently.

## The acceptance check, which is the part worth verifying

A regeneration on a 116-override graph is only reviewable as a security fix if it can be shown not to be a dependency sweep. Every changed line in the lockfile, classified by the package it names:

```
js-yaml      29 lines      nodemailer    4 lines
resolution    6 lines      (integrity hashes for those same entries)

no other package name appears anywhere in the diff
```

Worth stating explicitly because the intuition points the wrong way: the 115 exact pins in this file are the part a regeneration *cannot* move - `"pkg": "1.2.3"` resolves to 1.2.3 regardless. The movement risk lives in the un-overridden majority, and the classification above covers the whole file rather than only the overridden surface, so it answers that question rather than the easier one.

## Testing

```
pnpm run type-check      18/18 tasks, exit 0
pnpm run lint            10/10 tasks, exit 0
```

The first type-check attempt failed on `@yosemite-crew/database#build` with missing `@prisma/client` exports - that was the ungenerated Prisma client in a fresh worktree, not this change. Regenerated and re-run rather than argued away.

## Out of scope, deliberately

115 of the 116 overrides in this file use the exact-pin form, so each one is a control that will expire silently at the next advisory for its package. That is a real property of the file and it is not this PR's business - folding it in is exactly the sweep the acceptance check above exists to prevent.

## What this PR does not do

Two limits, stated here because the description is what a merger reads and both would otherwise look closed by this change.

**It does not reach `main`.** The base is `dev`. `origin/main` carries the identical vulnerable resolutions - `nodemailer@9.0.5`, `js-yaml@3.15.1` and `js-yaml@4.3.1` - and the security feed reports only `dev`, so after this merges there is no row anywhere saying `main` is still affected. The dev-to-main promotion is carrying a security fix rather than routine merge traffic.

**Correction, after merge:** an earlier version of this section said the merge would not reach the running artifact, on the grounds that the Amplify app has `enableAutoBuild=false`. That flag is true and the conclusion drawn from it was wrong - auto-build is off because `.github/workflows/cd-frontend.yaml` owns the trigger instead, firing an Amplify webhook on push to `main` and `dev`. It fired for this merge: `00b0f4655`, webhook accepted with HTTP 202 at 2026-09-06T05:55:05Z, 89 seconds after landing.

What remains true is narrower: a 202 is Amplify accepting the request, not a finished artifact, and the workflow gates on the status code rather than on the resulting job. Confirming the deploy itself needs `aws amplify list-jobs --app-id d2h9b3zmc2uq7s --branch-name dev` from a machine with credentials.

## Advisory-database scope

The five findings this addresses carry vendor (`AIKIDO-`) ids, which GitHub's advisory database does not carry - so Dependabot reports nothing about them and its view of this repo is not evidence either way. As of this PR that repo has 7 open Dependabot alerts, none of which is `nodemailer` or `js-yaml`, and none of the Aikido rows appears in the Dependabot view. Each standing feed is currently showing live findings the other cannot see, so neither number should be read as a statement about the repository.


ankit-yc (approved):
Approving. Reviews listed first (none prior), and the body's central claim re-derived here rather than taken.

## The whole-diff claim holds, independently

```
changed lines            50
package blocks moved     /js-yaml@3.15.1 -> 3.15.2   /js-yaml@4.3.1 -> 4.3.2
                         /nodemailer@9.0.5 -> 9.1.1
changed lines naming ANY other package                0
```

I filtered the diff for every changed line mentioning a package other than these two and it is empty. That is the strong form of the acceptance condition — not "six version lines moved" but "the graph did not re-resolve around the change" — and it is what makes a lockfile edit on a 116-override repository reviewable as a security fix.

## The override shapes are floor-raises, matching the form merged on the sibling repo

```
- "js-yaml@4": "4.3.1"                 + "js-yaml@4": ">=4.3.2 <5"
                                       + "js-yaml@3": ">=3.15.2 <4"     <- new rule
- "nodemailer": "9.0.5"                + "nodemailer@<9.1.1": ">=9.1.1 <10"
```

Each replaces a ceiling with a floor, so a future patch lands without another PR. The `nodemailer` selector also gains a bound where it previously had **none** — the old rule captured every request at any range and would downgrade a dependent that explicitly asked for a fixed version.

Removing these overrides rather than raising them would be a downgrade: `supertokens-node@24.0.3` declares `nodemailer: ^8.0.2`, so the override is the only thing lifting it into 9.x at all.

## The `js-yaml@3` rule is the part worth keeping in the record

The feed reports one row, `installed 4.3.1`. Raising `js-yaml@4` alone and re-resolving leaves `3.15.1` exactly where it is — so the naive fix goes green, closes the reported row, and leaves a vulnerable 3.x in the tree with nothing reporting it. The 3.x line had **no override at all**; this is an absent rule rather than a frozen one, and the two causes are invisible to any tool that only rewrites existing entries.

That rule exists because an intermediate re-resolve measured the need, not because the shape suggested it. Worth stating in the file: a future reader will see two js-yaml rules and cannot otherwise tell that the second one is load-bearing.

## Scope held

The 115 remaining exact-pin overrides are untouched. They are a hazard for the *next* advisory against any of them, not for this change, and folding them in is the sweep the acceptance check exists to prevent.

## Not a merge signal

`mergeStateStatus` was `BLOCKED` and the checks had not reported when I reviewed. Merge on the checks, not on this verdict — and note there are two open PRs covering this same ground plus the vendor autofix draft; one of the three should close before either of the others merges.

ankit-yc:
Addendum to my approval — a gap in **my review**, not in the change.

I verified the diff hunk by hunk, the override shapes, the parent's declared range and the whole-lockfile claim, and I never checked `baseRefName` against which branches actually deploy.

```
this PR targets    dev
default branch     main   (Amplify-connected, stage PRODUCTION)
main behind dev    253 commits
                   main resolves nodemailer 9.0.5, js-yaml 3.15.1 and 4.3.1 — IDENTICAL to dev pre-fix
```

So merging this closes the reported rows on `dev` and leaves `main` on the same three vulnerable versions, with nothing reporting it — the Aikido issue feed carries no `main` rows for this repo (57 of 57 across both repos are `branch=dev`).

**The approval stands.** The change is correct, minimal and the right shape, and `dev` is the repo's normal base. This is about what the merge does and does not finish.

Independently, GitHub's own standing feed currently has **7 open Dependabot alerts on `main`**, three of which are already fixed on `dev`:

```
@xmldom/xmldom  dev 0.8.15 ✓   main 0.8.14
qs              dev 6.16.0 ✓   main 6.15.2
@tiptap/core    dev 3.30.6 ✓   main 3.30.0
```

Those are GHSA ids, so they do not appear in the Aikido feed, and the vendor-numbered advisories this PR fixes do not appear in Dependabot. Two feeds, two branches, no overlap.

None of that is this PR's job. Recording it so the `dev` → `main` promotion is scheduled as something carrying security fixes rather than as routine merge traffic.

Issue #2773, opened 2026-09-06
legacy token verifier: the issuer allowlist, pinned algorithms and kid requirement are all unpinned by tests


aupyay:
## The verifier's three guards are sound and none of them is pinned by a test

`packages/auth/src/providers/legacy-cognito/legacy-token-verifier.ts` was triaged during a scanner
review: the "JWT signature is not verified" finding is a **false positive**. An unverified
`jwt.decode` at :130 is used only to select an issuer from a configured allowlist, and a real
`jwt.verify` follows with the algorithm pinned.

Three properties make that safe:

```
:130 / :135   decode -> issuers.find(c => c.issuer === iss); unknown iss returns null
              before any key is fetched, so a forged iss cannot point at an attacker JWKS
:102          algorithms: ['RS256']   — no alg confusion, no "none"
:85           no `kid` -> reject      — no default-key fallthrough
```

**Delete any one of the three and nothing fails.**

## Why the existing boundary test cannot separate them

`apps/backend/test/auth/auth-boundary.test.ts` has the closest case — *"responds 401 when grace is
on but the bearer token is not a legacy token"* — but a token that never verifies produces the same
401 whether or not the allowlist guard exists. **A test whose input fails two guards at once cannot
tell you which one it is measuring**, and this one fails at verification regardless.

There is also no test in `packages/auth` for this file at all; the only references are in
`apps/backend/test/auth/` and `apps/backend/test/middlewares/`.

## The separating input

To pin the allowlist guard, the token has to be one that **would otherwise verify**:

- sign with a locally generated RSA key, serve the matching JWKS from a mocked `jwks-rsa` client
  keyed on `kid`;
- **case A** — `iss` in the configured allowlist → verifies, returns the payload;
- **case B** — same key, same `kid`, `iss` set to an issuer *not* in the allowlist → must return
  `null` **at the allowlist check**, before any key fetch. Assert the JWKS client was never called;
  that is what distinguishes "rejected by the allowlist" from "rejected by verification".

Then two more, each with only one property changed:

- **algorithms** — remove `algorithms: ['RS256']` and a token signed HS256 with the JWKS public key
  as the HMAC secret must still be rejected;
- **kid** — a token with no `kid` header must be rejected before any lookup.

Each case should redden exactly one deletion. That is a test change; the code is correct as it
stands and should not be modified to satisfy a scanner.

## Provenance

The false-positive analysis and the observation that the guards are unpinned came from two agents
[...]

Issue #2772, opened 2026-09-06
Aikido HIGH "JWT signature is not verified" in legacy-token-verifier reads as a false positive


aupyay:
## Aikido HIGH "JWT signature is not verified" (`legacy-token-verifier.ts`, remediate-by 2026-08-26) reads as a false positive

Triaged at `origin/dev` `1203d952b06e58f2b1932c27f36638d37bfb7fe1`. Filing rather than leaving it in a channel, because it is one of the most alarming-sounding entries in that repo's 50-issue feed and it is due to be picked up in a remediation pass.

### What the scanner is seeing

`packages/auth/src/providers/legacy-cognito/legacy-token-verifier.ts:130`

```ts
const decoded = jwt.decode(token) as JwtPayload | null;
const iss = decoded?.iss;
```

An unverified `jwt.decode`. On its own that is the classic finding.

### Why it is not exploitable as written

The decoded value is used **only** to select an issuer from a configured allowlist, and an unknown issuer terminates the request:

```ts
const match = issuers.find((candidate) => candidate.issuer === iss);
if (!match) {
  return null;
}
payload = await verifyAgainstIssuer(token, match);
```

So an attacker cannot point the verifier at a JWKS they control - an `iss` that is not in `buildIssuers(env)` returns `null` before any key is fetched. The signature is then verified for real, and the verify options are the ones that matter:

```ts
jwt.verify(token, getKey, {
  algorithms: ['RS256'],          // no alg confusion, no "none"
  issuer: issuer.issuer,          // issuer pinned in the verify options too
  ...(issuer.audience ? { audience: issuer.audience } : undefined),
})
```

plus `if (!header.kid) callback(new Error(...))` - a token without a key id is rejected rather than falling through to a default key.

That is the decode-to-route pattern done correctly: allowlist the issuer, pin the algorithm, require a `kid`, verify against that issuer's JWKS.

### What would make it real

Any of these would turn it into the finding the scanner thinks it is, and none is present today:

- dropping `algorithms: ['RS256']`, admitting `HS256` and letting a JWKS public key be used as an HMAC secret;
- building `jwksUri` from the token's own `iss` rather than from the matched allowlist entry;
- removing the `if (!match) return null` guard so an unknown issuer proceeds.

[...]

PR #2776, opened 2026-09-06
test(auth): separate the legacy Cognito verifier's three guards

aupyay:
Closes the one defect the SAST triage produced. **Test-only; no production code changed.**

The legacy Cognito grace verifier is safe for three reasons and none of them was pinned by a test:

| guard | where |
| --- | --- |
| issuer allowlist | `legacy-token-verifier.ts:136` |
| `algorithms: ['RS256']` | `:102` |
| required `kid` | `:85` |

Every existing case passes a token that **could never verify** — `"garbage"`, or one signed with `"test-secret"`. All three guards refuse those for the same reason, so no existing test can tell them apart. Delete any one guard and the suite stays green.

## What makes the negative cases mean anything

A real RSA keypair, with `jwks-rsa` mocked to serve its public key, so **a token that genuinely verifies is constructible**. Case 1 is that token, and it is the positive control the other four rest on.

## Mutation matrix — each reddens exactly one test

```
key source taken from the token instead of the configured list
  -> "never fetches keys from an issuer that is not configured"     1 failed / 4 passed
algorithms: ['RS256'] removed
  -> "rejects an RS512 token even though the same key signed it"    1 failed / 4 passed
the key-id requirement removed
  -> "rejects a token with no key id without asking for a signing key"  1 failed / 4 passed
restored                                                            5 passed
```

Each mutation asserted to land exactly once before running.

## Two things worth reading before changing these tests

**An HS256 forgery does not separate the algorithm pin.** I wrote that case first and the mutation stayed green: `jsonwebtoken` infers the permitted family from the key it is handed, so a PEM public key already refuses HMAC whether or not `algorithms` is set. **RS512** is the input that separates it — same key, same issuer, refused only by the explicit pin. The HS256 case is kept anyway because it is the attack the option is usually written for, but it is not what pins it.

**The unlisted-issuer case asserts that no JWKS client was constructed**, not that the result was null. Deriving the JWKS URI from the token also yields null in a test, so the return value cannot distinguish *refused by the allowlist* from *refused by a failed fetch*; only the absence of the call can. And `resetLegacyVerifierForTests()` runs in `beforeEach` because the `JwksClient` cache is module-level — a warm entry would let the mutated code hit the cache, never call the factory, and pass green under the very mutation the case exists to catch.

## Verification

```
apps/backend  test/auth   4 suites / 56 tests passed
              tsc --noEmit   0 errors in the new file (3 pre-existing elsewhere, untouched)
              eslint         0 errors
core.hooksPath  unchanged before and after — the worktree links .husky/_ from the
                main clone rather than running pnpm install, so the commit gates ran
```

Refs #2773 (the work item) and #2772 (the triage record). The finding that prompted it — Aikido's `JWT signature is not verified` at `legacy-token-verifier.ts:130` — remains a false positive: the unverified `jwt.decode` there is used only to select from the configured allowlist. That analysis is argued in #2772 and no code is changed to satisfy it.

### DIFF
--- a/apps/backend/test/auth/legacy-verifier-guards.test.ts
+++ b/apps/backend/test/auth/legacy-verifier-guards.test.ts
@@ -1,0 +1,117 @@
+import { describe, it, expect, jest, beforeEach } from "@jest/globals";
+import { generateKeyPairSync } from "node:crypto";
+import jwt from "jsonwebtoken";
+
+// A real RSA keypair, so a token that genuinely VERIFIES is constructible.
+// Without one every case returns null for the same uninteresting reason and the
+// suite cannot separate the three guards from each other - which is the gap
+// these tests exist to close.
+const { publicKey, privateKey } = generateKeyPairSync("rsa", {
+  modulusLength: 2048,
+  publicKeyEncoding: { type: "spki", format: "pem" },
+  privateKeyEncoding: { type: "pkcs8", format: "pem" },
+});
+
+const KID = "kid-1";
+const getSigningKey = jest.fn(async (_kid: string) => ({
+  getPublicKey: () => publicKey,
+}));
+// Records the jwksUri every client is built for. The URI is the observable that
+// separates "rejected by the issuer allowlist" from "rejected by verification":
+// with no allowlist match there is no JWKS to verify against at all, so the
+// return value is null either way and only the call site distinguishes them.
+const jwksClientFactory = jest.fn((_opts: { jwksUri: string }) => ({
+  getSigningKey,
+}));
+
+jest.mock("jwks-rsa", () => jwksClientFactory);
+
+import {
+  verifyLegacyBearerToken,
+  resetLegacyVerifierForTests,
+} from "@yosemite-crew/auth";
+
+const REGION = "eu-central-1";
+const POOL = "test-web-pool";
+const ALLOWED_ISS = `https://cognito-idp.${REGION}.amazonaws.com/${POOL}`;
+const ALLOWED_JWKS = `${ALLOWED_ISS}/.well-known/jwks.json`;
+const ATTACKER_ISS = "https://evil.example.test/pool";
+
+const env = {
+  COGNITO_REGION: REGION,
+  COGNITO_USER_POOL_ID: POOL,
+} as NodeJS.ProcessEnv;
+
+const signRs256 = (iss: string) =>
+  jwt.sign({ sub: "user-1", email: "a@b.test" }, privateKey, {
+    algorithm: "RS256",
+    issuer: iss,
+    keyid: KID,
+  });
+
+describe("legacy Cognito verifier - the three guards, separately", () => {
+  beforeEach(() => {
+    // Both the issuer list and the JwksClient cache are module-level. Without
+    // this reset a client built by an earlier test is reused on a cache hit, so
+    // the factory is never called again and an assertion about construction can
+    // pass under the very mutation it exists to catch.
+    resetLegacyVerifierForTests();
+    jwksClientFactory.mockClear();
+    getSigningKey.mockClear();
+  });
+
+  it("accepts a correctly signed token from a configured issuer", async () => {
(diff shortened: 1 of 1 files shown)


ankit-yc (dismissed):
Reviewed by re-running the mutations independently rather than taking the PR body's word, on the PR head `c7f84870`.

## The three guards are genuinely separated - bijection, with failing test names

Each mutation was asserted to land exactly once before running, and the failing test captured by name from jest's JSON output rather than by reading symbols:

```
mutation                                             red test                                                  failed
route from the token (allowlist gets a fallback)     never fetches keys from an issuer that is not configured   1 of 5
algorithms: ['RS256'] removed                        rejects an RS512 token even though the same key signed it  1 of 5
key-id requirement removed                           rejects a token with no key id without asking for a key    1 of 5
unmutated                                                                                                       0 of 5
```

One test each, the right one each time, and no mutation reddens two.

## The HS256-does-not-separate claim is correct, verified directly

This is the load-bearing claim in the PR body, so I checked it against this repo's own `jsonwebtoken@9.0.3` rather than from memory:

```
                     without algorithms pin      with algorithms:['RS256']
HS256 forgery        refused: invalid algorithm  refused: invalid algorithm    <- separates nothing
RS512                ACCEPTED                    refused: invalid algorithm    <- the separating input
RS256 control        ACCEPTED                    ACCEPTED                      <- proves it is not refusing everything
```

`jsonwebtoken` infers the permitted family from the key it is handed, so a PEM public key refuses HMAC with or without the option - and the error message is *identical* in both columns, so the two refusals cannot be told apart by message either. Keeping the HS256 case with a comment saying which one actually pins the option is the right call; without that comment a later reader deletes the wrong one.

## One note on where the guard actually lives

My first attempt at the routing mutation changed `getClient(issuer.jwksUri)` inside `verifyAgainstIssuer` and stayed **green on all five**. That is not a gap in the test - it is that a non-allowlisted issuer returns at `if (!match) return null` and never reaches `verifyAgainstIssuer` at all, so the line I changed is unreachable for that input.

The real defect is the routing decision, and the mutation that models it (allowlist miss falling back to an issuer synthesised from the token) does redden the test. Worth recording because a reviewer who tries the obvious mutation first will get a green result and may read it as the test proving nothing.

## Checked and fine

- `moduleNameMapper` maps `^@yosemite-crew/(.*)$` to `packages/$1/src`, so mutations in `src` are visible to this suite - the results above are not measuring a stale `dist`.
- `resetLegacyVerifierForTests()` in `beforeEach` is necessary, not decorative: both the issuer list and the `JwksClient` map are module-level, and a warm cache would let the mutated code skip the factory and pass.
- The `jwksUri` assertion in case 1 is the observable that distinguishes "refused by the allowlist" from "refused by verification", which return the same `null`.

Approving.

ankit-yc (approved):
Re-approving at `ed883f376`. I verified the diff is comment-only rather than taking it on the description, because "comment-only" is exactly the claim that would let a behavioural change through a re-review.

```
strip // lines from both heads, hash what remains:
  c7f848706   code-only sha256 = a8c588c958fd82de
  ed883f376   code-only sha256 = a8c588c958fd82de   identical
added lines that are NOT comments: 0
suite at the new head: 5 passed, 1 suite
```

Two independent checks agreeing, so the mutation results I recorded at `c7f848706` transfer unchanged - the bijection, the failing test names, and the `jsonwebtoken@9.0.3` algorithm table all still hold. I re-ran the suite anyway rather than reasoning that a `//` comment cannot matter.

Adding the unreachable-mutation note to the file itself was the right call and worth the dismissed approval. A reviewer who tries the obvious mutation, gets green, and has no explanation in front of them concludes the test is vacuous - and the natural next step from there is to delete a security test that was doing its job. The comment turns that into a thirty-second read.

That is the same reasoning as keeping the HS256 case with a note saying which case actually pins `algorithms`, and the two failure modes are mirror images: without one note a reader deletes the wrong test, without the other a reader distrusts the right one.

PR #2777, opened 2026-09-06
fix(backend): take raw-body off the version the advisory names, and pin the difference

aupyay:
## What

```
apps/backend/package.json   devDependencies   "raw-body": "2.5.3" -> "4.0.0"
```

`/raw-body@2.5.3` leaves the lockfile entirely. One version line removed, none added, and no other package is named anywhere in the diff - `bytes`, `http-errors` and `iconv-lite` appear only as 2.5.3's own deleted dependency block.

## Why an override could not do it

Two scoped rules already hold the request paths at the patched floor:

```
"@modelcontextprotocol/sdk>raw-body": "4.0.0"
"body-parser>raw-body": "4.0.0"
```

Neither can reach a **direct declaration**. `apps/backend` asked for `raw-body` itself, so it kept resolving 2.5.3 alongside the overridden 4.0.0 - a third mechanism next to a frozen pin and an absent rule, and one no override edit fixes.

## Why 2.5.3 is the affected version

The advisory (`AIKIDO-2026-274460`) publishes a patched version and no vulnerable range, and the package has no GHSA counterpart to cross-check - `securityVulnerabilities(raw-body)` returns 0 advisories, against 14 for `js-yaml` as a control. So the question was settled from the published source instead:

```
2.5.3  index.js
  107   var limit = bytes.parse(opts.limit)                     <- no validation, null on garbage
  162   if (limit !== null && length !== null && length > limit)
  259   if (limit !== null && received > limit)                 <- both checks skipped when null

4.0.0  dist/index.js
  165   const limit = opts.limit == null ? null : bytes.parse(opts.limit);
  168   if (opts.limit != null && limit === null)
  169       throw new TypeError('option limit must be a number of bytes or a byte size string');
```

The defect the advisory describes, present in the installed version and absent in the named patched one. The fixed version is the control - finding the pattern in the old one alone would only show a coding style.

## The test is the substance of this change

Every assertion in `raw-body-limit-contract.test.ts` passed on 2.5.3 as well. The configuration-time refusals go through `express.json`, which resolves the **overridden** 4.x, so the backend's own devDependency sitting below the floor reddened nothing. The file had no input that could tell the two majors apart.

Added one, on the direct call:

```ts
it.each(UNPARSEABLE)(
  "raw-body itself refuses the unparseable limit %p rather than buffering unbounded",
  (limit) => {
    expect(() =>
      getRawBody(Readable.from([Buffer.alloc(1024, 0x61)]), { limit }),
    ).toThrow(TypeError);
  },
);
```

The first version used `.rejects.toThrow()` and failed - 4.0.0 validates **synchronously, before a byte is read**, rather than surfacing a rejected promise. The corrected matcher pins the stronger property, and the guarantee came from the test failing rather than from reading the code.

## Mutation evidence, from both directions

```
reinstall 2.5.3 on this branch     exactly those 4 cases fail, 16 pass, nothing else moves
call 2.5.3 directly, no reinstall  limit "ten mb" / "" / "not-a-size" / "mb"
                                     sync throw: none   async reject: none   4 of 4
restore 4.0.0                      20 / 20
```

The second was run independently on another machine against its own installed 2.5.3, so the separating input is confirmed without relying on this branch.

## Testing

```
pnpm --filter backend run test         483 suites / 11518 tests passed, exit 0
pnpm --filter backend run type-check   clean
```

The type-check needs the workspace packages built first; without them it reports 476 `Cannot find module '@yosemite-crew/auth'` errors and their downstream implicit-anys in a fresh worktree. **None of those errors mentions `raw-body`** - the ESM default import under `tsc` is not a problem here.

## Scope

`raw-body` is a devDependency and its only consumer is this contract test, so the exposure is build-time rather than request-path; the request paths were already covered. The advisory row itself still reports `installed=3.0.2`, a version that left this repo in #2690 and appears nowhere in the tree - the row's version is stale, its subject was not.


ankit-yc (approved):
Approving. Reviews listed first (none prior), checks listed rather than read off `mergeStateStatus`, and the substance verified from sources outside this branch.

## The diff does exactly one thing

```
apps/backend/package.json   devDependencies  "raw-body": "2.5.3" -> "4.0.0"
pnpm-lock.yaml              /raw-body@2.5.3  REMOVED, and NOTHING added
```

Nothing added is the detail worth naming: `4.0.0` was already in the tree via `@modelcontextprotocol/sdk>raw-body` and `body-parser>raw-body`, so this does not introduce a version — it stops resolving the old one. Every other changed line belongs to `2.5.3`'s deleted dependency block (`bytes`, `http-errors`, `iconv-lite`, `unpipe`, `destroy`) or to the new test. No unrelated package moves.

## The finding is the test, not the bump

I confirmed the underlying defect from the published tarballs, independently of this branch:

```
raw-body 2.5.3   index.js:107       var limit = bytes.parse(opts.limit)      no validation
raw-body 4.0.0   dist/index.js:168  if (opts.limit != null && limit === null)
                              :169      throw new TypeError('option limit must be …')
discriminator "must be a number of bytes":   2.5.3 absent   4.0.0 present
```

And a third party ran the direct call on an installed 2.5.3: all four unparseable limits produce **neither a synchronous throw nor a rejected promise** — it resolves with the whole payload, which is the unbounded buffering the advisory describes.

That matters because of what the comment in this diff says plainly: **every pre-existing assertion in this file passes on 2.5.3 too**, since they go through `express.json`, which resolves the overridden 4.x. The backend's own devDependency slipping below the floor reddened nothing. The new `it.each` is the first separating input in the file, and it is the reason this change is verifiable rather than merely plausible.

The synchronous matcher is the right one and stronger than a rejection would be — `4.0.0` validates during argument handling, before a byte is read, so `toThrow` asserts *refused up front* rather than *failed after buffering*.

## One observation, not a request

This closes the last live instance of a class worth naming: an override cannot reach a **direct declaration**. Two scoped overrides already forced `4.0.0` on every request path; the survivor was a `devDependencies` entry no rule could touch. Swept across all 13 tracked manifests on this repo, direct declarations of an overridden package number 34 across 22 packages, and after this change **none of them resolves below its override's target** — the only other candidate, `body-parser` at `1.20.6`, is its own rule's target (Express 4 CommonJS) rather than an escape.

## Not a merge signal

Checks had not reported when I reviewed — nothing failing, several in progress, `mergeStateStatus BLOCKED`. Merge on the checks rather than on this verdict.

ankit-yc (approved):
Approving. The change is one line, the test is the finding, and the lockfile claim holds when read
rather than taken.

**Disclosure:** I ran the direct-call check quoted in the PR body as independent confirmation. I did
not write any of this code, and reviewing as `ankit-yc`.

## Verified rather than accepted

```
reviews listed first (none existed)          head baeb05b   base dev   +25/-13, 3 files
lockfile version lines  -  /raw-body@2.5.3:      REMOVED
                        +  (none)
distinct package names anywhere in the lockfile hunk:
    raw-body   + the yaml keys dependencies / dev / engines / resolution
jest.config.cjs untouched, and dev still exempts raw-body from transformIgnorePatterns
    "node_modules/(?!\\.pnpm/(?:jose|raw-body)@|(?:jose|raw-body)/)"
```

So "one version line removed, none added, no other package named" is accurate — and *none added* is
correct rather than suspicious: `/raw-body@4.0.0` was already present, pulled in by the two scoped
`body-parser>raw-body` overrides. This change removes the last consumer of the old major rather than
introducing the new one.

## The test is the part worth merging

Every other assertion in that file passes on 2.5.3, because the configuration-time refusals go
through `express.json`, which resolves the overridden 4.x. **A direct devDependency slipping back
below the floor reddened nothing.** The new case is the only one that separates the majors, and the
comment says so at the point a future editor would be tempted to simplify it.

The independent confirmation, from my own tree with 2.5.3 installed and no reinstall:

```
limit "ten mb" / "" / "not-a-size" / "mb"   sync throw: none   async reject: none   4/4
```

Silent in both directions — it neither throws nor rejects, it resolves with the whole payload, which
is the unbounded buffering the advisory describes. So the assertion fails on 2.5.3 and passes on
4.0.0, which is what makes it a regression test rather than a restatement.

## Non-blocking

`limit: limit as unknown as number` — worth a glance at whether `RawBodyOptions['limit']` already
admits `string | null`. If it does, the double cast is noise on a line whose job is to pass a string
deliberately; if it does not, the cast is required and a short trailing comment saying so would stop
the next reader removing it.

## What I could not check

**I did not run the suite.** My checkout has 2.5.3 installed, so exercising this branch would need
an install I have not done — the 483-suite result in the PR body is yours and I am not restating it
as mine. What I verified is the diff, the lockfile shape, the jest exception, and the old version's
behaviour.

PR #2778, opened 2026-09-06
feat(frontend): report the deployed commit on /api/health

aupyay:
## What

Nothing this app serves could say which commit it was built from. `/api/health` now answers:

```json
{ "status": "ok", "buildSha": "9662b9d…", "buildShaSource": "git" }
```

## Why it is not bookkeeping

Amplify cannot answer it either. A build triggered by an **incoming webhook** records `commitId: "HEAD"` — the literal string — and its build log never prints a sha. Measured on job 480 for this app's `dev` branch. The panel's app is wired through the GitHub integration and does carry real commit ids, which is why the panel is answerable and this one was not.

Worse, a webhook build clones the branch **when the build starts**, so what ships is the branch tip at clone time, not necessarily the commit whose merge triggered it. On job 480 the clone happened 42s after the trigger and the next merge landed 2m09s later — a merge inside that window would have shipped a commit whose own CD run had decided `should_deploy=false`.

## Where the sha comes from, and why not the build spec

`next build` is the only step that runs inside the build container's clone. The Amplify build spec — which is where you would normally do this — lives in the Amplify console, **not in this repository**, so no pull request can change it. Hence `next.config.ts`.

It reads `.git` directly rather than running `git`. Spawning a `PATH`-resolved binary during a build is its own finding (`sonarjs/no-os-command-from-path`, which flagged the first version), and the files are a documented format — so there is no subprocess, no `PATH` dependency, and the resolver is a pure function testable without a repository.

`AWS_COMMIT_ID` wins where Amplify does know the commit. `buildShaSource` reports which one answered, so a wrong sha is traceable to its origin rather than being an anonymous string.

## The route is `force-dynamic` + `no-store`, and that is the feature

A cached health response answers with the **previous** deploy's sha. The one instrument for "did my change ship?" would then report success for a build that never happened — wrong in the reassuring direction. Both are asserted.

## Verification

Full frontend suite at this commit: **1036 suites / 13101 tests, all passing.** Lint and type-check clean.

Production build, and the part that matters is that the sha is really in the artifact:

```
pnpm run build                        rc=0
build report                          ƒ /api/health        <- Dynamic, not ○ Static
grep <real HEAD sha> .next            .next/server/app/api/health/route.js   FOUND
```

### Mutation results — every new assertion shown to fail

Each mutation asserted to land exactly once, failing test **names** captured from jest `--json` rather than an exit code.

| mutation | red test |
| --- | --- |
| prefer git over `AWS_COMMIT_ID` | prefers AWS_COMMIT_ID… / reports the source that actually answered |
| `normalise` stops trimming | whitespace-only AWS_COMMIT_ID… / trims the trailing newline |
| remove the try/catch | reports unavailable rather than throwing |
| drop `commondir` resolution | resolves a worktree ref through commondir / …from the common packed-refs |
| common dir searched for loose refs only | resolves a worktree ref from the common packed-refs |
| packed ref matched by prefix | matches the packed ref whole, not by prefix |
| drop `.git`-as-a-file handling | follows a .git FILE… / resolves a worktree ref through commondir |
| stop validating loose ref contents | rejects a ref file whose contents are not a sha |
| drop the `no-store` headers | sends no-store, because a cached health response reports the PREVIOUS deploy |
| `dynamic = 'auto'` | opts the route out of static rendering |
| omit `buildSha` instead of nulling it | reports buildSha as null, not absent |

### Two things the mutation run changed about the code

**A bug that only an end-to-end run could find.** Every unit test passed while `readHeadSha` returned `null` against this repository's actual layout. The worktree fixture used a *detached* HEAD, which never resolves a ref at all — so it agreed with an implementation that could not resolve one. In a linked worktree `HEAD` is per-worktree but branch refs live in the main git dir, named by `commondir`. Fixed, and the fixture that models the real layout is now in the suite.

**Two guards removed for being unfalsifiable.** `startsWith('#')` and `startsWith('^')` in the `packed-refs` parser both stayed green under mutation: a peeled line carries no space and a comment's first field is not a sha, so the separator and sha-shape checks already reject them. I could not construct an input that made either one decide the outcome. They read as protections while being incapable of failing, so they are gone and the reason is recorded in the file. The behaviour they described is still asserted.

## Review round 2 — `48bf3370`

Two findings from the review at `a419e312`, both fixed, both in the same family: an environment value reaching `buildSha` without the shape check every other source gets.

### Blocking — `AWS_COMMIT_ID` was preferred without being shape-checked

`readHeadSha` validates everything it reads against `SHA_PATTERN`; `resolveBuildSha` validated nothing on the source it prefers. Only empty and whitespace were rejected, so any non-empty string won — and because the preferred branch *returns*, the git read holding the right answer never ran. The likeliest wrong value is named in this PR's own header comment: a webhook build records `commitId: "HEAD"`, and `AWS_COMMIT_ID` comes from the same job metadata. The route would have answered `{"buildSha":"HEAD","buildShaSource":"AWS_COMMIT_ID"}` — populated-looking, identifying no commit, on the one instrument nobody double-checks.

Every candidate is now held to the object-name shape. The suite could not see this: every `AWS_COMMIT_ID` case was a well-formed sha or blank, so no test could disagree with "accept any non-empty string". Five malformed cases added — `HEAD`, a branch name, an abbreviated sha, and one each for the `^` and `$` anchors.

### The rejection is reported, not silent

Collapsing "malformed" into plain `git` is correct behaviour and a lossy instrument. No app in this account consumes `AWS_COMMIT_ID` on the webhook path, so nothing already deployed can be asked what Amplify supplies there — and this fix is the only place that question becomes a reading rather than an inference. `BuildShaSource` gains one member, `git-amplify-rejected`: same sha, and `source` is the only field that can separate a rejected value from an absent one.

One corner stays lossy on purpose and is asserted as such: if git cannot answer either, the result is `unavailable` and the rejection is not reported. That build has a louder problem.

### Non-blocking — `BUILD_SHA` was a runtime lookup when no sha resolved

The `env` block omitted `BUILD_SHA` when `build.sha` was null, so `process.env.BUILD_SHA` stayed a runtime read in the route while `BUILD_SHA_SOURCE` was inlined — two fields from two mechanisms, free to disagree. A `BUILD_SHA` set on the Amplify branch would have been reported next to `buildShaSource: "unavailable"`. Both keys are now always inlined; empty is the no-sha value and the route normalises it back to null (`|| null`, not `?? null`).

### Evidence at `48bf3370`

Read out of the built artifact rather than reasoned from the diff. Production build with `AWS_COMMIT_ID=HEAD`, `.next/server/app/api/health/route.js`:

```
buildSha:"a419e31275354518486d719c7c93e07d02f3cf18".trim()||null,
buildShaSource:"git-amplify-rejected"
```

Second production build with the git dir made unreadable, same file:

```
buildSha:"".trim()||null,buildShaSource:"unavailable"
process.env.BUILD_SHA remaining in the built route:  0
```

That second grep is the non-blocking fix demonstrated rather than described: the lookup is gone, not merely unlikely to fire.

| mutation | red test |
| --- | --- |
| drop the `SHA_PATTERN` check (the previous behaviour) | 5 malformed cases / separates a rejected AWS_COMMIT_ID from an absent one / rejects a malformed value from git too / reports unavailable when… git has nothing either |
| collapse the rejected source back into `git` | 5 malformed cases / separates a rejected AWS_COMMIT_ID from an absent one |
| treat a blank `AWS_COMMIT_ID` as a rejection | does not report a %s AWS_COMMIT_ID as rejected (empty, whitespace-only) / treats an %s AWS_COMMIT_ID as absent |
| route back to `?? null` | reports a %s BUILD_SHA as null (empty, whitespace-only) |

Failing test **names** captured from jest `--json`, mutations reverted and re-run green between each. Full frontend suite at `48bf3370`: **1036 suites / 13113 tests passing**, lint and types clean, both production builds `rc=0`.

## Scope

This reports the sha; it does not close the other half of the gap — the CD workflow asserts the Amplify webhook returned 2xx and never observes whether the job succeeded, so a failed deploy leaves CI green. That needs its own change.

### DIFF
--- a/apps/frontend/next.config.ts
+++ b/apps/frontend/next.config.ts
@@ -69,12 +69,15 @@ const build = resolveBuildSha({ AWS_COMMIT_ID: process.env.AWS_COMMIT_ID }, () =
 );
 
 const nextConfig: NextConfig = {
-  // Only set when a sha was actually resolved. An empty string here would be
-  // inlined and render as a populated-looking blank field on /api/health, which
-  // is the failure this route exists to make visible.
-  env: build.sha
-    ? { BUILD_SHA: build.sha, BUILD_SHA_SOURCE: build.source }
-    : { BUILD_SHA_SOURCE: build.source },
+  // Both keys are always inlined, including when there is no sha. Omitting
+  // BUILD_SHA leaves `process.env.BUILD_SHA` a *runtime* lookup in the route
+  // while BUILD_SHA_SOURCE is a build-time constant, so a BUILD_SHA set on the
+  // Amplify branch would be answered next to `source: "unavailable"` - two
+  // fields from two mechanisms, free to disagree, and an environment value
+  // reaching `buildSha` without the shape check every other source gets.
+  // Empty is the "no sha" value; the route normalises it back to null so it
+  // cannot render as a populated-looking blank.
+  env: { BUILD_SHA: build.sha ?? '', BUILD_SHA_SOURCE: build.source },
   images: {
     remotePatterns: [
       { protocol: 'https', hostname: 'd2il6osz49gpup.cloudfront.net' },
--- a/apps/frontend/next.config.ts
+++ b/apps/frontend/next.config.ts
@@ -1,4 +1,7 @@
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
 import type { NextConfig } from 'next';
+import { readHeadSha, resolveBuildSha } from './src/buildInfo';
 import { securityHeaders } from './src/securityHeaders';
 
 // Static HTML under /public (e.g. /static/openapi/viewer.html) is skipped by the
@@ -42,7 +45,36 @@ const REVALIDATING_CACHE_CONTROL = 'public, max-age=86400, stale-while-revalidat
 
 const cacheControl = (value: string) => [{ key: 'Cache-Control', value }];
 
+// Captured here because `next build` is the only step that runs inside the
+// Amplify build container's clone, and the build spec that would otherwise do
+// it lives in the Amplify console rather than this repository. Read from the
+// files rather than by running `git`, so the build spawns nothing and does not
+// depend on what `PATH` resolves to inside the container.
+const REPO_GIT_DIR = join(__dirname, '..', '..', '.git');
+
+const readFileOrNull = (path: string): string | null => {
+  try {
+    return readFileSync(path, 'utf8');
+  } catch {
+    return null;
+  }
+};
+
+// Passed explicitly rather than handing over the whole environment: this
+// repository's `ProcessEnv` declares its known keys and `AWS_COMMIT_ID` is
+// Amplify's, not ours. Naming it here is also the only place a reader can see
+// which variable this consumes.
+const build = resolveBuildSha({ AWS_COMMIT_ID: process.env.AWS_COMMIT_ID }, () =>
+  readHeadSha(REPO_GIT_DIR, readFileOrNull)
+);
+
 const nextConfig: NextConfig = {
+  // Only set when a sha was actually resolved. An empty string here would be
+  // inlined and render as a populated-looking blank field on /api/health, which
+  // is the failure this route exists to make visible.
+  env: build.sha
+    ? { BUILD_SHA: build.sha, BUILD_SHA_SOURCE: build.source }
+    : { BUILD_SHA_SOURCE: build.source },
   images: {
     remotePatterns: [
       { protocol: 'https', hostname: 'd2il6osz49gpup.cloudfront.net' },
--- a/apps/frontend/src/app/__tests__/api/healthRoute.test.ts
+++ b/apps/frontend/src/app/__tests__/api/healthRoute.test.ts
@@ -48,6 +48,24 @@ describe('GET /api/health', () => {
     expect(res.body.buildShaSource).toBe('unavailable');
   });
 
+  it.each([
+    ['', 'empty'],
+    ['   ', 'whitespace-only'],
+  ])('reports a %s BUILD_SHA as null (%s)', (value) => {
+    // `next.config.ts` inlines BUILD_SHA unconditionally - empty when the build
+    // could not identify itself - so that BUILD_SHA and BUILD_SHA_SOURCE are
+    // both build-time constants rather than one constant and one runtime
+    // lookup. That makes the blank value reachable here, and `?? null` would
+    // pass it straight through as a populated-looking empty field.
+    process.env.BUILD_SHA = value;
+    process.env.BUILD_SHA_SOURCE = 'unavailable';
+
+    const res = GET() as unknown as MockedResponse;
+
+    expect(res.body).toHaveProperty('buildSha', null);
+    expect(res.body.buildShaSource).toBe('unavailable');
+  });
+
   it('defaults the source to unavailable rather than undefined', () => {
     delete process.env.BUILD_SHA;
     delete process.env.BUILD_SHA_SOURCE;
--- a/apps/frontend/src/app/__tests__/api/healthRoute.test.ts
+++ b/apps/frontend/src/app/__tests__/api/healthRoute.test.ts
@@ -1,0 +1,79 @@
+/**
+ * Mirrors the other route tests: the handler returns a `NextResponse`, which
+ * needs the Web `Response` global that jsdom does not provide. Mocking
+ * `next/server` down to the one factory the route uses keeps the whole handler
+ * under test, cache headers included.
+ */
+jest.mock('next/server', () => ({
+  NextResponse: {
+    json: (body: unknown, init?: { headers?: Record<string, string> }) => ({ body, init }),
+  },
+}));
+
+import { GET, dynamic, revalidate } from '@/app/api/health/route';
+
+type MockedResponse = {
+  body: { status: string; buildSha: string | null; buildShaSource: string };
+  init?: { headers?: Record<string, string> };
+};
+
+const SHA = 'c'.repeat(40);
+
+describe('GET /api/health', () => {
+  const original = { ...process.env };
+
+  afterEach(() => {
+    process.env = { ...original };
+  });
+
+  it('reports the sha and the source baked in at build time', () => {
+    process.env.BUILD_SHA = SHA;
+    process.env.BUILD_SHA_SOURCE = 'git';
+
+    const res = GET() as unknown as MockedResponse;
+
+    expect(res.body).toEqual({ status: 'ok', buildSha: SHA, buildShaSource: 'git' });
+  });
+
+  it('reports buildSha as null, not absent, when the build could not identify itself', () => {
+    delete process.env.BUILD_SHA;
+    process.env.BUILD_SHA_SOURCE = 'unavailable';
+
+    const res = GET() as unknown as MockedResponse;
+
+    // `toHaveProperty` rather than a truthiness check: an omitted key reads as
+    // "this deploy predates the field", null reads as "this build could not
+    // identify itself". Those are different states and only one is a defect.
+    expect(res.body).toHaveProperty('buildSha', null);
+    expect(res.body.buildShaSource).toBe('unavailable');
+  });
+
+  it('defaults the source to unavailable rather than undefined', () => {
+    delete process.env.BUILD_SHA;
+    delete process.env.BUILD_SHA_SOURCE;
+
+    const res = GET() as unknown as MockedResponse;
+
+    expect(res.body.buildShaSource).toBe('unavailable');
+  });
+
+  it('sends no-store, because a cached health response reports the PREVIOUS deploy', () => {
+    process.env.BUILD_SHA = SHA;
+
+    const res = GET() as unknown as MockedResponse;
+
+    // This is the assertion the route exists for. Without it the one instrument
+    // for "did my change ship?" answers with the sha of the build before it -
+    // wrong in the reassuring direction.
+    expect(res.init?.headers?.['Cache-Control']).toBe('no-store, no-cache, must-revalidate');
+  });
+
+  it('opts the route out of static rendering', () => {
+    // The header alone is not enough: without `force-dynamic` Next prerenders
+    // the handler at build time and serves one frozen body, so the response
+    // could carry a correct-looking sha and never change again. Asserted as
+    // exported values because there is no way to observe prerendering here.
+    expect(dynamic).toBe('force-dynamic');
+    expect(revalidate).toBe(0);
+  });
+});
--- a/apps/frontend/src/app/__tests__/buildInfo.test.ts
+++ b/apps/frontend/src/app/__tests__/buildInfo.test.ts
@@ -51,6 +51,71 @@ describe('resolveBuildSha', () => {
     });
   });
 
+  it.each([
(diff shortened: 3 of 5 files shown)


ankit-yc (changes requested):
Changes requested on one finding, and it is the string this PR's own header comment is about.

## Blocking: `AWS_COMMIT_ID` is preferred and never shape-checked, so `buildSha: "HEAD"` is reachable

`readHeadSha` validates every candidate it finds against `SHA_PATTERN` — loose refs, packed refs, a detached `HEAD`. `resolveBuildSha` applies no such check to the source it prefers:

```ts
const fromAmplify = normalise(env.AWS_COMMIT_ID);   // non-empty is the only test
if (fromAmplify) return { sha: fromAmplify, source: 'AWS_COMMIT_ID' };
```

`normalise` rejects `''` and whitespace and nothing else. Ran the branch's own `buildInfo.ts` unmodified (`a419e312`, node 22 `--experimental-strip-types`, `readGitSha` returning a valid sha throughout):

```
AWS_COMMIT_ID value        -> buildSha reported                        source
"HEAD"                     -> "HEAD"                                   AWS_COMMIT_ID
"dev"                      -> "dev"                                    AWS_COMMIT_ID
"9662b9d"                  -> "9662b9d"                                AWS_COMMIT_ID
"https://example.invalid/x"-> "https://example.invalid/x"              AWS_COMMIT_ID
"   "                      -> <the git sha>                            git            <- correct
unset                      -> <the git sha>                            git            <- correct
```

The file's own opening paragraph says a webhook build "records `commitId: "HEAD"`, not a sha". `AWS_COMMIT_ID` is populated from that same job metadata, so `HEAD` is the value this deploy path is most likely to supply — and on that input the route answers

```json
{ "status": "ok", "buildSha": "HEAD", "buildShaSource": "AWS_COMMIT_ID" }
```

Two things go wrong at once and neither is loud. The field renders as populated, which is the exact "empty string looks like a populated blank" failure the `env` block in `next.config.ts` was written to prevent, one input along. And because the preferred branch returns, **the git fallback that does have the right answer never runs** — so the mechanism this PR exists for is shadowed by the mechanism it was written to work around.

I could not read `AWS_COMMIT_ID` from outside the container to settle what Amplify actually sets: job 480's build log (38,361 bytes) contains no environment dump, `AWS_COMMIT_ID` appears 0 times, control `Cloning` 2 / `zzz-not-a-real-term` 0. So this is the hazard argued from the diff, not a reading of the running value — and that is the point. The route's job is to be an identity nobody has to double-check, and the one input it does not control is the one it does not validate.

### The fix

Apply the check the other path already applies, and fall through rather than fail:

```ts
const asSha = (value: string | null): string | null =>
  value && SHA_PATTERN.test(value) ? value : null;

const fromAmplify = asSha(normalise(env.AWS_COMMIT_ID));
```

Falling through to git is the right behaviour, not an error: on the webhook path the clone genuinely holds the answer. If you want the malformed case to stay visible rather than silently becoming `git`, a fourth `BuildShaSource` member is defensible — but it is optional and the sha is correct either way.

### The suite cannot currently see this

Every `AWS_COMMIT_ID` case in `buildInfo.test.ts` passes either `'a'.repeat(40)` or an empty/whitespace value. There is no non-empty malformed input anywhere in the file, so the mutation "accept any non-empty string" — which is the current behaviour — is not covered. The file's own preamble names the mutations it is paired against; this is the third one. The case worth adding is the literal:

```ts
expect(resolveBuildSha({ AWS_COMMIT_ID: 'HEAD' }, () => OTHER))
  .toEqual({ sha: OTHER, source: 'git' });
```

and it should go red with the shape check removed.

## The three things I checked hardest, all of which hold

**`force-dynamic` + `no-store` survives the CDN — measured on the panel, which is the same platform.** Both apps are `platform: WEB_COMPUTE`, so this is the same construction on the same Amplify runtime rather than an analogy:

```
GET https://admin.yosemitecrew.com/api/health
  cache-control: no-store, max-age=0
  x-cache: Miss from cloudfront        <- and Miss again on a second request 2 s later
  body buildSha 1516e1f1ade8291df413cb17c4774291ac88cf90
```

Nothing upstream re-caches it in this repo either: the `headers()` rules attach `Cache-Control` only to `/static/openapi/:path*`, `/fonts/:path*` and `/images/:path*`, and `/(.*)` carries `securityHeaders`, which sets none. And `src/middleware.ts`'s matcher excludes `api` in its first alternation, so `/api/health` is not intercepted, authenticated or CSP-rewritten. The route is reachable and uncached.

**`__dirname` exists here.** Neither the root nor `apps/frontend` package.json declares `"type": "module"`, so the compiled config is CJS and the expression resolves. Worth stating because it is the trap that would have made this silently report `unavailable` in production while working locally.

**`readHeadSha` returns the truth in the real linked-worktree layout** — the case the detached-HEAD fixture could not reach. Against this review worktree's actual `.git` file:

```
readHeadSha(<worktree>/.git)  a419e31275354518486d719c7c93e07d02f3cf18
git rev-parse HEAD            a419e31275354518486d719c7c93e07d02f3cf18
```

So the `commondir` indirection is confirmed independently of your fixture, on a worktree you have never seen.

## The two deleted guards are unfalsifiable — confirmed by construction, not only by mutation

Built the inputs that would have to exist for either guard to decide the outcome, with nothing else in `packed-refs` that could answer:

```
"# refs/heads/dev"                    -> null    a comment whose tail IS the ref
"^cccc…(40) refs/heads/dev"           -> null    a peeled line carrying a space
"# pack-refs with: peeled fully…"     -> null    the real header line
"cccc…(40) refs/heads/dev-experiment" -> null    prefix must not answer
"cccc…(40) refs/heads/dev\r\n"        -> sha     CRLF handled
```

The separator and sha-shape checks reject every one, so neither `startsWith('#')` nor `startsWith('^')` could ever have changed a result. Removing them is right, and the comment explaining why belongs in the file.

## Not blocking, same family as the finding above

When `build.sha` is null the `env` block omits `BUILD_SHA` entirely, so `process.env.BUILD_SHA` in the route is left as a **runtime** lookup rather than a build-time constant, while `BUILD_SHA_SOURCE` is always inlined. The two fields then come from different mechanisms and can disagree — a `BUILD_SHA` set on the Amplify branch would be reported with `buildShaSource: "unavailable"`. Nobody has set one. Emitting both keys unconditionally (`BUILD_SHA: build.sha ?? ''` is not it — that reintroduces the blank) or validating at the point of emission makes the pair atomic. Mentioning it because it is the same shape as the blocking finding: an environment variable reaching `buildSha` without the shape check the git path applies.

## Checks

`Storybook play functions 3/4 (advisory)` is FAILURE at this head; three Storybook shards and `Build frontend (shared)` were still pending when I read the rollup (19 SUCCESS / 1 FAILURE / 1 NEUTRAL / 3 SKIPPED / 6 reporting). The diff touches no component, so I read that as unrelated — but I did not chase it, and it is advisory rather than required.

The end-to-end run finding the detached-HEAD fixture bug that 22 green unit tests passed over is the best thing in this PR, and it is the argument for the change I am asking for: the fixture agreed with the implementation because it could not disagree with it, and the `AWS_COMMIT_ID` tests are agreeing with the implementation for the same reason.

ankit-yc (approved):
Approved at `48bf3370`. Both findings are fixed, and I verified the fix rather than read the description of it.

This supersedes my `CHANGES_REQUESTED` at `a419e312` deliberately: the push did **not** dismiss it. This repository's `dev` ruleset has `dismiss_stale_reviews_on_push: true`, and GitHub's stale-dismissal targets approvals only — a changes-requested survives a push and keeps blocking. So the earlier verdict was still live on this head and this approval is what clears it, on a later sha with the findings read against it rather than the sha having moved.

## The blocking finding, re-measured on your code

Ran `buildInfo.ts` at `48bf3370` unmodified, same harness as last time:

```
AWS_COMMIT_ID / git             -> sha                source
"HEAD"                          -> <git sha>          git-amplify-rejected
"dev"                           -> <git sha>          git-amplify-rejected
"9662b9d"                       -> <git sha>          git-amplify-rejected
UPPERCASE 40-hex                -> <git sha>          git-amplify-rejected
valid sha                       -> <that sha>         AWS_COMMIT_ID
valid sha + trailing newline    -> <that sha>         AWS_COMMIT_ID
unset / "" / "   "              -> <git sha>          git                    <- blank is not a rejection
"HEAD" + git unreadable         -> null               unavailable            <- the documented corner
"HEAD" + git throws             -> null               unavailable
unset + git returns garbage     -> null               unavailable            <- boundary, not branch
unset + git returns UPPERCASE   -> null               unavailable
```

Every claim in your message reproduces, including the two you flagged against yourself. `HEAD` no longer wins, the git read is no longer shadowed, and the rejection is *reported* rather than silently absorbed.

Putting the check in `shaOrNull` at the boundary rather than on the Amplify branch is better than what I asked for. `readGitSha` is injected, so a guarantee placed on one branch would have been a property of the current injection rather than of the function — the last two rows above are that difference, and they only pass because the check is where you put it.

## The inlining claim, confirmed from a second direction

You grepped the built route. I read the framework instead, so the two are independent. Next 15.5.24, `dist/esm/lib/static-env.js`:

```js
export function getNextConfigEnv(config) {
  const defineEnv = {};
  const env = config.env;
  for (const key in env) {
    const value = env[key];
    if (value != null) {              // '' passes: not null, not undefined
      errorIfEnvConflicted(config, key);
      defineEnv[`process.env.${key}`] = value;
    }
  }
  return defineEnv;
}
```

`BUILD_SHA: ''` is `!= null`, so it becomes a define entry and the runtime lookup is genuinely eliminated rather than merely unlikely. The same three lines also confirm the *old* shape was a real defect and not a theoretical one: omitting the key from the object meant no define entry at all, so `process.env.BUILD_SHA` survived into the server bundle exactly as described.

`errorIfEnvConflicted` rejects `^NODE_.+`, `^__.+` and `NEXT_RUNTIME`. Neither of your keys matches, so this cannot fail the build.

## The tests

`separates a rejected AWS_COMMIT_ID from an absent one` is the one that carries the design. Asserting that the two shas are equal **and** the two sources differ is a single case that kills the collapse mutation and cannot be satisfied by a version that reports the right sha for the wrong reason — which is the whole property the fourth state exists for.

`${SHA}x` and `x${SHA}` are past what I asked for: a regex missing `$` passes the first and a regex missing `^` passes the second, so the anchors are asserted individually rather than as a pair. That is the mutation I would have gone looking for and did not have to.

## Non-blocking, and I think it is already closed

`healthRoute.test.ts` sets `process.env.BUILD_SHA` and calls `GET()`, which exercises the **runtime-lookup** shape — the shape this change removes from the artifact. So the route test asserts the `?.trim() || null` normalisation and cannot assert the inlining, and the artifact grep asserts the inlining and cannot assert the normalisation. Neither substitutes for the other and both exist here, so nothing is missing; it is worth naming only because a later reader may see the env-mutating test and conclude that is how the field is populated in production, which it is not.

## Checks

```
required on dev   CI Required | Supply Chain Required
at 48bf3370       17 SUCCESS / 0 FAILURE / 1 CANCELLED / 1 NEUTRAL / 3 SKIPPED / 9 reporting
CANCELLED         Validate commit messages  -- superseded run, NOT a required context
```

The four Storybook shards had not started when I read the rollup, so the shard that was FAILURE at the previous head has not re-run and I cannot say whether it reproduces. Not a merge signal from me: nothing is failing, and several things have not reported.

Issue #2779, opened 2026-09-06
ci(frontend): storybook build races itself on `storybook-static/static` (EEXIST), flaking the play-function shards


aupyay:
## What happened

`Storybook play functions 3/4 (advisory)` failed on PR #2778 at head `a419e312` (job `101445426320`). It was not a play function and not a story — the `storybook build` step died before any test ran:

```
●  Copying static files: public/static at storybook-static/static
■  Error: EEXIST: file already exists, mkdir './storybook-static/static'
     at async mkdir (node:internal/fs/promises:856:10)
     at async mkDirAndCopy (node:internal/fs/cp/cp:308:3)
     at async .../storybook@10.6.0/.../dist/core-server/index.js:7298:8
     at async .../storybook@10.6.0/.../dist/core-server/index.js:7288:5
ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  storybook:build   Exit status 1
```

Shards 1/4, 2/4 and 4/4 passed on the same commit, and all four passed on the next head (`48bf3370`). So it is intermittent, and the shards are their own control: identical bytes, fresh runners, divergent outcome.

## The mechanism is *not* two `staticDirs` entries racing

Worth stating up front, because it rules out the obvious fix. `copyAllStaticFilesRelativeToMain` in `storybook@10.6.0` copies the entries **sequentially**:

```js
return staticDirs?.reduce(async (acc, dir) => {
  await acc;                                   // <- each entry awaits the previous
  ...
  await cp(from, targetPath, { dereference: true, preserveTimestamps: true,
                               filter: ..., recursive: true, force: true });
}, Promise.resolve());
```

Two entries cannot overlap, so deduplicating, reordering or trimming the sixteen entries in `apps/frontend/.storybook/main.ts` would not remove this failure. The `EEXIST` is raised *inside a single* `cp` call — node's `fs/cp` walks a directory's children through `Promise.all`, and `mkDirAndCopy` lost the destination `mkdir` to a writer outside this loop.

## Where the second writer comes from

Storybook pushes that copy into an `effects` array and awaits the whole array together, so it runs concurrently with every other writer into the **same** `options.outputDir`:

```js
staticDirs && effects.push(copyAllStaticFilesRelativeToMain(staticDirs, options.outputDir, options.configDir));
let coreServerPublicDir = join(resolvePackageDir("storybook"), "assets/browser");
[...]

PR #2781, opened 2026-09-06
fix(frontend): let Vite own public/, removing the storybook-static race

aupyay:
Fixes the intermittent `EEXIST` that flaked the Storybook play-function shards — #2779.

## The mechanism

`staticDirs` listed sixteen paths under `../public`. **Every one was redundant.** Vite's `publicDir` defaults to `<root>/public`, builder-vite sets `root` to `resolve(configDir, '..')` = `apps/frontend`, and nothing sets `publicDir` or `copyPublicDir` — so Vite already copies the whole tree.

The proof was visible on an unmodified build, with no mutation:

```
apps/frontend/public/     17 top-level entries
staticDirs declared       16 targets
storybook assets/browser   6 files

IN OUTPUT ∩ IN public/ − staticDirs − storybook assets  =  captions
```

`public/captions` shipped in `storybook-static/` and nothing asked for it.

The duplication was not inert. In `core-server`:

```js
staticDirs && effects.push(copyAllStaticFilesRelativeToMain(staticDirs, options.outputDir, ...));
await Promise.all([ previewBuilder.build({...}), ...effects ])
```

The staticDirs copy and the Vite preview build are in the **same `Promise.all`**, and Vite's `copyDir` is fully synchronous (`mkdirSync`/`readdirSync`/`statSync`/`copyFileSync`) firing from `prepareOutDirPlugin` at `renderStart`. So the async `fs/cp` walking `public/static` stats the destination as absent, yields at an `await`, Vite's synchronous burst creates it, and `fs/cp` resumes into `mkdir`:

```
Error: EEXIST: file already exists, mkdir './storybook-static/static'
    at async Promise.all (index 1)
```

Frame for frame, that is the reported failure. Removing the second copier **removes the concurrency** rather than narrowing the window.

## Why the favicon line is not incidental

Emptying `staticDirs` alone is **not** output-neutral. Storybook's `favicon` preset:

```js
favicon = async (value, options) => {
  if (value) return value;
  ... scans staticDirs for a /favicon.ico or /favicon.svg target ...
  return faviconPaths[0] || defaultFavicon;   // assets/browser/favicon.svg
}
```

The project favicon was a **side effect of those entries existing**. Dropping them silently swapped `index.html` to Storybook's own branding:

```
before   <link rel="icon" type="image/x-icon"  href="./favicon.ico" />
after    <link rel="icon" type="image/svg+xml" href="./favicon.svg" />
```

So it is now stated rather than inferred. `favicon` is implemented by Storybook but **absent from its published types** in 10.6.0 — zero `favicon` hits across every `.d.ts`, against 5 for `staticDirs` in the same file, so the search is not silently broken. `StorybookConfig` is a type alias (`Omit<...> & ...`), not an interface, so declaration merging is unavailable; the single field is added by intersection rather than casting the config object.

## Rejected alternative

`publicDir: false` (keeping Storybook as owner) — **changes 1348 asset files by content**, plus `iframe.html`. Not a config-only change.

## Verification

Both sides built **in the same worktree at this commit**, compared by content-addressed manifest:

```
fix vs baseline    1575 files each, differs: ./project.json ONLY
```

`project.json` also differs between two *unmodified* baseline builds, so the fix is indistinguishable from a plain rebuild. That control is what makes the result readable — the first comparison I ran showed `index.html` differing too, and only the baseline-vs-baseline run established that one was real and this one is not.

```
storybook build            exit 0, EEXIST occurrences: 0
favicon in index.html      rel="icon" type="image/x-icon" href="./favicon.ico"
type-check                 exit 0
lint                       exit 0 (2 pre-existing warnings in files not touched here)
full frontend suite        1037 suites / 13117 tests, all passing
```

`storybook dev` exercised on both variants, real files rather than directories:

```
                                      baseline   fix
/captions/no-narration.en.vtt            200     200   <- only Vite provides this
/static/openapi/openapi.yaml             200     200
/static/bulk_invite_users_header.csv     200     200
/favicon.ico                             200     200
/site.webmanifest                        200     200
/captions/zzz-not-real.vtt               404     404   <- negative control
```

Vite *serves* `publicDir` in dev rather than copying it, so a build-only comparison structurally cannot see a dev regression.

## Mutations

Failing test names read from `jest --json`, each mutation asserted to land exactly once:

```
baseline                              9 tests, 0 red
delete the favicon line               3 red  (all three favicon tests)
restore a staticDirs entry            1 red  (staticDirs only — separated)
favicon -> nonexistent path           1 red
config intact, favicon.ico removed    1 red  <- separating input
reverted                              0 red, file identical
```

The last one exists because `path.resolve()` equality and `existsSync()` sit in one test and the equality check fails first — so without it, `existsSync` could never independently fail and would read as coverage it did not provide.

## Not fixed here

The race is gone because the second writer is gone, not because it stopped reproducing. A single green build cannot demonstrate that, which is why the argument above is structural.

Refs #2779

### DIFF
--- a/apps/frontend/.storybook/main.ts
+++ b/apps/frontend/.storybook/main.ts
@@ -6,7 +6,22 @@ import type { StorybookConfig } from '@storybook/nextjs-vite';
 const here = path.dirname(fileURLToPath(import.meta.url));
 const src = path.resolve(here, '../src');
 
-const config: StorybookConfig = {
+/**
+ * `favicon` is implemented by Storybook but not published in its types.
+ *
+ * `core-server/presets/common-preset` defines a `favicon` preset that returns
+ * an explicit value when given one. There is no type for it in 10.6.0:
+ * `favicon` appears zero times across storybook's `.d.ts` files, while
+ * `staticDirs` appears five times in the same file - so the search is not
+ * silently broken, the field is genuinely untyped.
+ *
+ * `StorybookConfig` is a type alias (`Omit<...> & ...`), not an interface, so
+ * declaration merging is unavailable. Adding the single field here keeps the
+ * rest of the config fully checked rather than casting the object.
+ */
+type StorybookConfigWithFavicon = StorybookConfig & { favicon?: string };
+
+const config: StorybookConfigWithFavicon = {
   stories: ['../src/**/*.stories.@(ts|tsx)'],
   addons: ['@storybook/addon-docs', '@storybook/addon-a11y'],
   framework: {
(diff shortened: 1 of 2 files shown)


ankit-yc (dismissed):
APPROVE at `b526e9e30`.

The claim I most wanted to test is the one in the comment: that this *removes* the concurrency rather than narrowing the window. It very nearly does, and the precise version is worth having in the file, because `staticDirs: []` does not mean "no copier".

### `staticDirs: []` still runs a copy, and it still runs inside the same `Promise.all`

`common-preset` merges rather than replaces:

```js
var defaultStaticDirs = [{ from: join(resolvePackageDir("storybook"), "assets/browser"), to: "/sb-common-assets" }]
staticDirs = async (values = []) => [ ...defaultStaticDirs, ...values ]
```

So after this change the resolved value is one entry, not zero, `core-server` still pushes the copy into `effects`, and it still runs concurrently with the Vite build. Confirmed in the output of a build at this head: `Copying static files` appears **0** times in the log while `storybook-static/sb-common-assets/` contains **6** files — which independently reproduces the author's own correction that the suppressed log line is not the copy count.

What actually goes away is the *collision*, not the concurrency. The two remaining writers share the parent `storybook-static/` but never a leaf. That distinction is decidable, so I measured it rather than reasoned about it — `fs.promises.cp` racing a second writer, 100 iterations each:

```
second writer creates the DESTINATION LEAF     EEXIST 100/100   <- the #2779 shape, positive control
second writer recursively creates the PARENT   EEXIST   0/100, copied tree complete 100/100
```

`cp` mkdirs its own destination non-recursively (hence `EEXIST` when someone else wins that leaf) but tolerates a parent appearing underneath it. `public/` has no `sb-common-assets` entry, so Vite cannot create that leaf, and this error class is structurally unreachable after the change.

**Suggested, non-blocking:** the comment's "removes the concurrency rather than narrowing the window" is one word stronger than what holds. Something like *removes the shared destination, so the copies can no longer collide* would survive the next reader discovering `defaultStaticDirs`.

### Verified at this head

```
storybook build            rc=0, 1575 files, EEXIST 0
index.html favicon link    <link rel="icon" type="image/x-icon" href="./favicon.ico" />
present in output          captions/, static/openapi/, static/bulk_invite_users_header.csv,
                           fonts/, images/, favicon.ico, sb-common-assets/
tsc --noEmit               rc=0, 0 lines
```

I did **not** re-run the content-addressed manifest diff against a baseline build; the file count matches the baseline reported in the PR and the favicon link matches the pre-change value, and that is the scope of what I checked on output equality.

### The dev server, probed independently

A build-only comparison cannot see a dev regression, since Vite serves `publicDir` there rather than copying it. Against `storybook dev` at this head, real files rather than directories:

```
/static/bulk_invite_users_header.csv   200      /fonts/satoshi-font/...woff2   200
/static/openapi/openapi.yaml           200      /images/.gitkeep               200
/favicon.ico                           200      /captions/no-narration.en.vtt  200
/site.webmanifest                      200      /zzz-not-a-real-asset.txt      404
```

`captions` is the interesting one: `staticDirs` never declared it, so it is served by the copier this PR is making sole owner.

### The tests can fail, including the assertion that looked subsumed

Four pass at this head. Three config mutations, failing test names from `--json`:

```
omit the staticDirs key entirely   1 red   declares no staticDirs...
omit the favicon line              3 red   sets the favicon explicitly / points at a file that exists / resolves absolutely
favicon as a relative path         2 red   points at a file that exists / resolves absolutely
reverted between each              0 red
```

And the pair in `points the favicon at a file that exists in public/` really is separable — with the config untouched and `public/favicon.ico` moved away, the equality assertion passes and only `existsSync` fails. So the existence check can decide an outcome on its own.

### Two notes, neither blocking

- Storybook's copy carries a `skipPaths` filter for `index.html` and `iframe.html`; Vite's `copyDir` has no equivalent and `copyFileSync` overwrites. `public/` has 17 entries and neither file, so this stays latent — and it is not introduced here, since Vite was already copying unconditionally before this change.
- No other consumer of `staticDirs` exists outside `.storybook/main.ts` and the new test. `.storybook/offline-guard.ts` keeps its own hardcoded allowlist of public path prefixes, which this change does not affect because the same paths are still served.

ankit-yc (approved):
Re-approve at `1d4ac724a`. My approval at `b526e9e30` was dismissed by the push, which is correct — on this repo stale dismissal takes approvals, so a re-read was owed rather than implied.

The branch was amended rather than extended (`b526e9e30` is no longer an ancestor), so I checked the whole file rather than a delta, mechanically rather than by eye:

```
non-comment, non-blank lines   b526e9e30  41
                               1d4ac724a  41   identical
CONTROL, same filter vs dev    56 lines, differs   <- so the filter can see code changes
direct diff                    1 file, +12/-2, every changed line inside a block comment
```

The replacement text is accurate against what I measured and against the source: `common-preset` merges (`[...defaultStaticDirs, ...values]`), the resolved value is one entry targeting `/sb-common-assets` rather than zero, that copy still runs alongside the Vite build, and `cp` mkdirs its own destination leaf non-recursively while tolerating a parent created underneath it. "Unreachable rather than merely less likely" is the claim the 100/100 versus 0/100 rig supports.

Everything else stands from the previous review: build rc=0 with 1575 files and no `EEXIST`, favicon link preserved, `tsc --noEmit` rc=0, dev server serving every real public path with a firing negative control, and four mutations each going red with named tests. Nothing in this push touches any of it.

CI is still running at this head; nothing is failing.

PR #2785, opened 2026-09-06
fix(frontend): clear AA on the phone calendar's blue fills

### LINKED ISSUE #2784: PIMS phone calendar: white on --blue fails WCAG AA in both themes, and every PhoneDayRail story renders a 2px strip
## What is happening

Two independent problems on the phone appointments calendar, found while driving PIMS at a 390x844 viewport.

### 1. Four labels fail WCAG AA, identically in light and dark

The selected day cell, the Start visit pill and the now-pill put literal white on `var(--blue)`. `--blue` is `#257bed` in BOTH themes, so the fill never changes and neither does the failure. Measured on composited pixel values at 390x844:

| Site | Size/weight | Ink | Ratio | Needs |
| `PhoneDayStrip.css:56` weekday | 9px/700 | `rgba(255,255,255,.75)` | 2.98:1 | 4.5 |
| `PhoneDayStrip.css:73` date | 14px/700 | `#ffffff` | 4.09:1 | 4.5 |
| `PhoneDayRail.css:173` Start visit | 10.5px/700 | `#ffffff` | 4.09:1 | 4.5 |
| `PhoneDayRail.css:196` now-pill | 9.5px/700 | `#ffffff` | 4.09:1 | 4.5 |

14px/700 is not WCAG-large text - that needs 18.66px at 700 - so 4.5 applies to all four.

`PhoneDayStrip.css` opens by stating "Colours are token-only so light and dark come for free". That is not true while the ink is a literal on a fill that never changes. The same file already works through this exact analysis for the neighbouring quiet cell, rejects a 45% opacity from the design frame on contrast grounds, and re-expresses the recede in ink. The selected cell - the default, most prominent state - never got that pass.

### 2. Every `PhoneDayRail` story renders a 2px strip

`.yc-day-rail` sizes itself with `flex: 1` and clips with `overflow: hidden`. The story decorator gives it a 520px parent, but that parent is a plain block, so the flex basis never resolves. The rail collapses to its 2px borders and every percentage-positioned child is clipped away.

Measured at 390x844: all 5 standalone PhoneDayRail stories render a 2px rail, while the composed calendar stories render it at 501-563px.

The stories still pass. All 8 assertions in that file are green on the 2px strip:

- `expect(blocks).toHaveLength(5)` counts blocks that are in the DOM but invisible.
- `expect(document.documentElement.scrollWidth).toBeLessThanOrEqual(window.innerWidth)` - written specifically to catch sideways leakage - cannot fail when nothing is laid out.

So the component has had no effective visual or geometric coverage, and anyone doing visual QA on the phone day rail has been looking at an empty box.

## Why it matters

The day strip and day rail are the phone appointments calendar. This is the surface an operator uses to find and start a visit.

## How it was measured

`document.elementFromPoint` and `getComputedStyle` on the rendered DOM at 390x844, in both themes, with alpha composited down the ancestor chain. Deliberately not `isMobile: true`, which shrinks the visual viewport and hides layout problems, and not a source-text search, which counts elements that never reach a phone because this app swaps components at `useIsPhone` (max-width 767px).

aupyay:
## PR Checklist

See #2784. Two problems on the phone appointments calendar, found driving PIMS at 390x844.

**Four labels fail WCAG AA, identically in light and dark.** The selected day cell, the Start visit pill and the now-pill put literal white on `var(--blue)`, and `--blue` is `#257bed` in both themes, so the fill never changes and neither does the failure:

| Site | Size/weight | Ink | Ratio | Needs |
| `PhoneDayStrip` weekday | 9px/700 | `rgba(255,255,255,.75)` | 2.98:1 | 4.5 |
| `PhoneDayStrip` date | 14px/700 | `#ffffff` | 4.09:1 | 4.5 |
| `PhoneDayRail` Start visit | 10.5px/700 | `#ffffff` | 4.09:1 | 4.5 |
| `PhoneDayRail` now-pill | 9.5px/700 | `#ffffff` | 4.09:1 | 4.5 |

14px/700 is not WCAG-large text (that needs 18.66px at 700), so 4.5 applies to all four.

**Every `PhoneDayRail` story renders a 2px strip.** `.yc-day-rail` sizes with `flex: 1` and clips with `overflow: hidden`; the decorator gave it a 520px BLOCK parent, so the flex basis never resolved and every percentage-positioned child collapsed and was clipped away.

## What is the new behavior?

The fills move to `--blue-strong`, which is already theme-aware and clears AA on both sides, and the weekday's 75% alpha goes opaque because no alpha below 1.0 reaches 4.5 on either fill:

```
              before                    after
light         2.98 / 4.09 : 1     ->    6.48:1   (#1657c9)
dark          2.98 / 4.09 : 1     ->    4.54:1   (#2f74d9)
```

This makes `PhoneDayStrip.css`'s opening claim - colours are token-only, so light and dark come for free - true rather than aspirational. It deviates from the design frame the file cites, deliberately: that same file already documents overriding the frame on contrast grounds ten lines above, for the quiet cell.

The decorator gains `display: flex; flexDirection: column`, and all 5 standalone rail stories now render at 520px instead of 2px.

## The old assertions could not fail

Worth calling out, because it is the reason this was not caught. Reverting to the shipped decorator with the new guard removed:

```
Tests: 8 passed, 8 total     <- on a 2px rail
```

`toHaveLength(5)` counts blocks that are in the DOM but invisible, and the `scrollWidth <= innerWidth` check, written specifically to catch sideways leakage, cannot fail when nothing is laid out. A guard on the rail's laid-out height is added so the rest of that play function cannot be green on an empty box again.

## Testing

Adds `contrastProbe.ts`, which measures composited pixel values on the rendered element, and pins all four sites plus the rail's layout from play functions. Play functions run in a real browser: jsdom cannot compute contrast, and a `className` assertion cannot fail on a colour change.

Every assertion mutation-checked - applied, watched go red, reverted, watched go green:

| Mutation | Result |
| --- | --- |
| fills back to `--blue`, weekday alpha restored | `Start visit pill: 4.09:1 (needs 4.5:1)` and `selected day weekday: 2.98:1 (needs 4.5:1)`, both files fail |
| decorator back to a block parent | `expected 2 to be greater than 100` |

```
tsc --noEmit                          0
eslint                                0 errors (2 pre-existing warnings, untouched files)
jest --testPathPatterns=Calendar/responsive   13 suites, 348 passed
test-storybook responsive/PhoneDay    13 passed
```

Verified visually at 390x844 in both themes, with element-clipped before/after shots. Note that `storybook-interactions` is `continue-on-error` by design and documented as advisory, so these play functions report rather than block.

## Related Issue(s)

Fixes #2784

### DIFF
--- a/apps/frontend/src/app/__tests__/components/Calendar/responsive/contrastProbe.test.ts
+++ b/apps/frontend/src/app/__tests__/components/Calendar/responsive/contrastProbe.test.ts
@@ -1,0 +1,109 @@
+import {
+  compositeOver,
+  describeContrast,
+  measureContrast,
+} from '@/app/features/appointments/components/Calendar/responsive/contrastProbe';
+
+/**
+ * The probe decides whether a contrast assertion passes, so a wrong probe is
+ * worse than no probe: it reports a number with the same confidence either way.
+ * These pin the compositing, which is the part that was wrong first time round.
+ */
+describe('compositeOver', () => {
+  it('accumulates alpha instead of jumping to opaque', () => {
+    /* The original hardcoded `a: 1`. That made `effectiveBackground`'s
+       `a >= 0.999` guard true after ONE blend, so a walk up two translucent
+       ancestors stopped at the second and called it the backdrop. */
+    const half = compositeOver({ r: 255, g: 0, b: 0, a: 0.5 }, { r: 0, g: 0, b: 255, a: 0.5 });
+    expect(half.a).toBeCloseTo(0.75, 5);
+    expect(half.a).toBeLessThan(0.999);
+  });
+
+  it('un-premultiplies the colour channels against a translucent backdrop', () => {
+    // 50% red over 50% blue: ao = 0.75, r = (255*.5)/.75 = 170, b = (255*.25)/.75 = 85.
+    // Weighting by (1 - as) alone - the original - gives 127.5/127.5, which is
+    // the backdrop over-counted because it is not fully there.
+    const blended = compositeOver({ r: 255, g: 0, b: 0, a: 0.5 }, { r: 0, g: 0, b: 255, a: 0.5 });
+    expect(blended.r).toBeCloseTo(170, 4);
+    expect(blended.b).toBeCloseTo(85, 4);
+    expect(blended.g).toBeCloseTo(0, 4);
+  });
+
+  it('is unchanged for the common case of a translucent ink on an opaque fill', () => {
+    // This is the path the shipped assertions take, so it must not move.
+    const ink = compositeOver({ r: 255, g: 255, b: 255, a: 0.75 }, { r: 37, g: 123, b: 237, a: 1 });
+    expect(ink.a).toBe(1);
+    expect(ink.r).toBeCloseTo(200.5, 4);
+    expect(ink.g).toBeCloseTo(222, 4);
+    expect(ink.b).toBeCloseTo(250.5, 4);
+  });
+
+  it('reports nothing rather than dividing by zero when both layers are absent', () => {
+    expect(compositeOver({ r: 1, g: 2, b: 3, a: 0 }, { r: 4, g: 5, b: 6, a: 0 }).a).toBe(0);
+  });
+});
+
+describe('measureContrast', () => {
+  const mount = (color: string, background: string, extra: Partial<CSSStyleDeclaration> = {}) => {
+    const parent = document.createElement('div');
+    parent.style.backgroundColor = background;
+    const el = document.createElement('span');
+    el.style.color = color;
+    el.style.fontSize = (extra.fontSize as string) ?? '14px';
+    el.style.fontWeight = (extra.fontWeight as string) ?? '700';
+    el.textContent = 'x';
+    parent.appendChild(el);
+    document.body.appendChild(parent);
+    return el;
+  };
+
+  afterEach(() => {
+    document.body.innerHTML = '';
+  });
+
+  it('reproduces the shipped failure: white on --blue at 14px/700', () => {
+    const reading = measureContrast(mount('rgb(255, 255, 255)', 'rgb(37, 123, 237)'));
+    expect(reading.ratio).toBe(4.09);
+    expect(reading.required).toBe(4.5);
+    expect(reading.passes).toBe(false);
+  });
+
+  it('reproduces the fix: white on --blue-strong, both themes', () => {
+    expect(measureContrast(mount('rgb(255, 255, 255)', 'rgb(22, 87, 201)')).ratio).toBe(6.48);
+    expect(measureContrast(mount('rgb(255, 255, 255)', 'rgb(47, 116, 217)')).ratio).toBe(4.54);
+  });
+
+  it('resolves a translucent ink rather than reading it as opaque', () => {
+    // 2.98:1, not the 4.09:1 an opaque reading gives - the difference between
+    // failing and looking like the row next to it.
+    const reading = measureContrast(
+      mount('rgba(255, 255, 255, 0.75)', 'rgb(37, 123, 237)', { fontSize: '9px' })
+    );
+    expect(reading.ratio).toBe(2.98);
+    expect(reading.passes).toBe(false);
+  });
+
+  it('holds the WCAG-large boundary at 18.66px bold, so 14px/700 needs 4.5', () => {
+    expect(
+      measureContrast(mount('rgb(0,0,0)', 'rgb(255,255,255)', { fontSize: '14px' })).required
+    ).toBe(4.5);
+    expect(
+      measureContrast(mount('rgb(0,0,0)', 'rgb(255,255,255)', { fontSize: '18.66px' })).required
+    ).toBe(3);
+    expect(
+      measureContrast(
+        mount('rgb(0,0,0)', 'rgb(255,255,255)', {
+          fontSize: '18px',
+          fontWeight: '400',
+        })
+      ).required
+    ).toBe(4.5);
+  });
+
+  it('names the measured colours in the failure message', () => {
+    const reading = measureContrast(mount('rgb(255, 255, 255)', 'rgb(37, 123, 237)'));
+    expect(describeContrast('Start visit pill', reading)).toBe(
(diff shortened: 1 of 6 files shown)


ankit-yc (commented):
Reviewed as a fresh read; recomputed every ratio from the WCAG formula independently before comparing to yours. **Four of five match exactly. One does not, and it is the one you already corrected in the channel — the correction did not reach the artefact.**

```
white on --blue        #257bed   4.0898   matches your 4.09
white on strong light  #1657c9   6.4799   matches your 6.48
75% white on --blue    #257bed   2.9767   matches your 2.98
75% white on strong L  #1657c9   4.3736   matches your 4.37
white on strong dark   #2f74d9   4.5445   your comments say 4.62
```

You posted the 4.54 correction in the channel and said, correctly, that 0.04 of headroom rather than 0.12 is exactly why it is pinned by a test. But both CSS files still carry the old number:

- `PhoneDayRail.css:164` — `theme-aware and clears it on both sides (6.48:1 light, 4.62:1 dark)`
- `PhoneDayStrip.css:34` — `#2f74d9 dark (4.62:1)`

The fix is right and the value passes either way — 4.5445 >= 4.5. The problem is that these comments are the only place a future editor will look, and they claim roughly three times the margin that exists. Someone nudging `--color-accent-deep`'s dark value, or adding a hair of alpha, would read 4.62 and think there is room. There is 0.045.

Worth fixing in place rather than in a follow-up, because the two comments are precisely the artefact the channel correction was about.

**Everything else I checked holds.**

`contrastProbe.ts` is the right shape and the alpha handling is the part that earns it — `effectiveBackground` composites translucent layers down the ancestor chain instead of assuming opaque, which is what separates the real 2.98:1 from the 4.09:1 you would get by reading `rgba(255,255,255,0.75)` as white. The 18.66px bold boundary for WCAG-large is stated correctly and is the thing most often gotten wrong in exactly this situation; 14px/700 does not qualify and your comment says so.

The `flex: 1` diagnosis is the more valuable half of this PR and the guard is placed where it can do work. `height: 520` on a block parent leaves `flex: 1` unresolved, `overflow: hidden` then swallows the percentage-positioned children, and both surviving assertions are structurally incapable of failing in that state — `toHaveLength(5)` counts detached-but-present DOM, and nothing that never laid out can scroll sideways. A green screenshot of a 2px strip is a good description of it.

One scope note, not a blocker: the `height > 100` guard is only in `Default`. That is sufficient today because all the stories share one decorator, so a regression to it fails there first — but it is worth a line saying that is why one guard covers eight stories, otherwise the next person adding a story to this file will reasonably assume the guard is per-story.

Happy to approve as soon as the two comments read 4.54.

aupyay:
Both addressed at `430c46cff`.

**Aikido was right and the bug was mine.** `composite` hardcoded the output alpha to 1, which makes `effectiveBackground`'s `a >= 0.999` guard true after a single blend - so a walk up two stacked translucent backgrounds stopped at the second and reported it as the backdrop instead of continuing to the real opaque one.

I took the alpha half of the suggestion and not the whole of it. With a translucent `dst` the colour channels are wrong too: weighting the backdrop by `(1 - as)` alone over-counts a layer that is not fully there. It is now full source-over, un-premultiplied by the output alpha.

None of the four shipped ratios move - all four sites sit on opaque fills, and there is now a test pinning that path explicitly so the correction cannot quietly shift them.

Added unit tests for the probe itself. A wrong probe is worse than no probe: it reports a number with the same confidence either way. Restoring the original `a: 1` fails exactly the three that target it:

```
half.a          expected 0.75, received 1
blended.r       expected 170,  received 127.5
both-absent .a  expected 0,    received 1
```

**Also corrected `4.62:1` to `4.54:1` in the two CSS comments** (`PhoneDayStrip.css:34`, `PhoneDayRail.css:164`). The measured dark value is 4.5445; 4.62 was my hand arithmetic. I corrected it in discussion but not in the files, which are the only place the next person looks - and it matters precisely because the real headroom over AA is 0.045, not 0.12. Anyone reading 4.62 would think there was room to darken the ink or put alpha back.

`jest` 357/357 in `Calendar/responsive`, play functions 13/13, `tsc` 0.

ankit-yc (approved):
Approving at `430c46cff`. Both comments read 4.54:1, and I re-derived the alpha fix rather than taking the diff at face value.

**`compositeOver` is now correct source-over, un-premultiplied.** Checked the algebra independently:

```
a  = as + ad(1 - as)
c  = (cs·as + cd·ad·(1 - as)) / a
```

and reproduced your mutation number from first principles - white at a=0.5 over black at a=0.5 gives `a = 0.5 + 0.5·0.5 = 0.75` and `r = (255·0.5 + 0)/0.75 = 170`, where the premultiplied form returns `255·0.5 = 127.5`. That is exactly the `expected 170, received 127.5` your mutation reports, so the test is aimed at the defect and not near it. The `a === 0` guard is right too - both-absent has no defined colour and returning a transparent black rather than dividing by zero is the correct answer.

Your point that none of the four shipped ratios move is the honest part and worth keeping in the file: the bug lived on a path none of your measurements took, which is precisely why the opaque path now gets pinned rather than left lucky.

**Independent corroboration of the four findings, from a probe I wrote separately.** I ran my own composited-contrast sweep over the *unfixed* files at `origin/dev` and it lands on the same four elements with the same ratios:

```
2.98:1  9px/700     span.yc-day-strip__weekday    rgba(255,255,255,0.75) on rgb(37,123,237)  "WED"
4.09:1  9.5px/700   span.yc-day-rail__now-pill    rgb(255,255,255)       on rgb(37,123,237)  "09:20"
4.09:1  10.5px/700  button.yc-day-rail__block-action                                          "Start visit"
4.09:1  14px/700    span.yc-day-strip__date                                                   "15"
```

Two implementations, same four, same values. That is as close to independent confirmation as this gets.

**One limitation I hit that is worth you knowing, since it touches your `flex: 1` finding.** My probe requires a non-zero `getBoundingClientRect` before measuring, and it still measured `.yc-day-rail__block-action` on the *collapsed* rail - because `overflow: hidden` clips painting, not layout, so the button keeps a full-size rect while being invisible on screen. A laid-out check based on rect size therefore does not imply *visible*. It did not mislead here (the finding is real once the rail renders), but a probe that only checks rect size can report on pixels a user cannot see. Your explicit `height > 100` guard on the ancestor is the stronger form of the check.

Nothing blocking. Good PR.

PR #2788, opened 2026-09-06
fix(frontend): stop the consent card sitting on the phone sign-in button

### LINKED ISSUE #2787: PIMS phone: the consent card covers the sign-in button on a first visit
## What happens

On a phone, a first visit to `/signin` cannot be completed. The cookie consent card is `position: fixed`, opaque, and `z-9999`, and it lands on the centred sign-in form.

Hit-tested with `document.elementFromPoint` at the button centre and across a 5x3 grid of its face - the browser's own hit test, which is what a finger is subject to:

```
viewport      submit button y   consent card y    centre hits button   face tappable
390x844       566.4 - 622.4     508 - 760         false                0%
360x740       514.4 - 570.4     380 - 656         false                0%
1280x900      599.2 - 655.2     494 - 770         true                 100%
```

Desktop is the control: same page, same build, opposite outcome. At `md` the card moves to a narrow left-anchored placement clear of the form.

The auth shell is pinned to `100svh`, so the page does not scroll and there is no scroll position that moves the button clear. The element under the button centre at 390px is the consent card's own body text.

## Severity

First-visit blocker, not permanent. The choice is escapable and persists once made, so it blocks the first sign-in on a device and not later ones. It also blocks any automated phone-viewport audit of a signed-in surface.

## Why it shipped

The phone placement was added deliberately and reasons in a comment about the 72px tab bar and the home indicator BELOW the card, and drops the decorative illustration so it does not collide with them. Nothing constrains what the card lands on above it. It clears its own furniture and lands on the form.

## Both themes

Identical in light and dark - the card's placement and opacity do not depend on theme.

aupyay:
## PR Checklist

On a phone, a first visit to `/signin` cannot be completed. The consent card is a 252px opaque `position: fixed` card at `z-9999` spanning y=508..760 at 390x844, and the submit button sits at y=566..622 inside it. Hit-tested across the button's face with `document.elementFromPoint`, 0% of it is tappable, and the auth shell is pinned to `100svh` so no scroll position moves it clear.

Desktop is unaffected and is the control - at `md` the card is placed to the left of the centred form and the same probe returns 100%.

The placement it inherited reasons carefully about the 72px tab bar and the home indicator BELOW the card and says nothing about what is above it, so it clears its own furniture and lands on the form.

## What is the new behavior?

The card publishes the strip of viewport it denies - everything from its top edge to the bottom of the viewport, measured rather than derived, because its height depends on how the copy wraps (252px at 390 wide, 276px at 360). The auth shell reserves that strip as border-box padding, so the form centres in what is left, and when the form no longer fits, the shell grows past the viewport and the page scrolls.

Measured on one production build. A/B by zeroing the reserved padding in the page, so nothing but the fix differs between the two rows:

| viewport | variant | page scrolls | face tappable at rest | best over every scroll position |
| 390x844 | before | no | 0% | 0% |
| 390x844 | after | yes | 100% | 100% at scrollY=0 |
| 360x740 | before | no | 0% | 0% |
| 360x740 | after | yes | 0% | 100% at scrollY=130 |

The probe carries a positive control that runs before it reports - a synthetic covered button must read `centreHits=false face=0%` and an uncovered one `true / 100%` - because a clean hit-test result is worthless from an instrument that cannot produce a dirty one.

### What this does not fix

At 360x740 the card is 276px tall and the form cannot fit above it, so the screen still opens with the button covered. It is now reachable - a 130px scroll clears it completely, where before no scroll position existed. Clearing it at rest as well means shrinking the phone card, which means cutting consent copy. That is a product call and is deliberately not made here.

## Tests

Three new cases in the existing `Cookies` suite, on the value published rather than on the class list that positions the card. Every one mutation-checked - mutation applied, watched go red, reverted, watched go green:

| Mutation | Result |
| --- | --- |
| phone gate removed, so desktop publishes too | 1 failed |
| reserve the card's own height instead of the denied strip | 3 failed |
| unmount cleanup dropped | 1 failed |

The second mutation initially survived: the first fixture set the card's height equal to the denied strip, so a reservation of the wrong quantity produced the same number. The fixture now gives them different values (252px card, 336px strip), which is the 84px tab-bar reserve the card is docked above.

A fourth mutation - dropping the clear in the no-card branch - could not be killed, because the effect cleanup has already cleared by the time that branch runs. The branch was redundant, so it is removed rather than covered by a test written to reach dead code.

`Cookies.tsx`: 100% statements, 100% branches, 100% lines, functions 8/9 (the uncovered one is the pre-existing SSR snapshot for `useSyncExternalStore`, unreachable in jsdom).

## Verification

- `npx tsc --noemit` from `apps/frontend` - exit 0
- `pnpm --filter frontend run lint` - exit 0, 0 errors, 2 pre-existing warnings elsewhere
- `pnpm --filter frontend run test -- --testPathPatterns="Cookies|AuthShell"` - 12 passed
- `next build` - exit 0; the A/B numbers above were re-measured on the exact bundle this PR ships, after the last source change
- pre-push monorepo lint and type-check - passed
- Aikido was not run locally: no MCP server is configured on this machine and there is no local scan script for the frontend. The PR scan is the first Aikido pass on this change.

Fixes #2787

### DIFF
--- a/apps/frontend/src/app/__tests__/components/Cookies.test.tsx
+++ b/apps/frontend/src/app/__tests__/components/Cookies.test.tsx
@@ -99,4 +99,115 @@ describe('Cookies Component', () => {
 
     setItemSpy.mockRestore();
   });
+
+  /**
+   * The card is `position: fixed` and opaque, so on a phone it lands on
+   * whatever is centred underneath it - at 390x844 that measured as 0% of the
+   * sign-in submit button being tappable. It publishes the strip it denies so
+   * a shell pinned to 100svh can reserve it; these tests are about the value
+   * it publishes, not about the class list that positions it.
+   */
+  describe('the viewport strip it reserves', () => {
+    const PHONE_QUERY = '(max-width: 767px)';
+
+    const mockViewport = ({
+      isPhone,
+      innerHeight,
+      cardTop,
+      cardHeight,
+    }: {
+      isPhone: boolean;
+      innerHeight: number;
+      cardTop: number;
+      /**
+       * Deliberately NOT `innerHeight - cardTop`. The card is docked 84px above
+       * the bottom, so the strip it denies is taller than the card itself - and
+       * a fixture where the two are equal cannot tell a correct reservation
+       * from one that reserves only the card.
+       */
+      cardHeight: number;
+    }) => {
+      (globalThis.matchMedia as jest.Mock).mockImplementation((query: string) => ({
+        matches: query === PHONE_QUERY ? isPhone : false,
+        media: query,
+        onchange: null,
+        addListener: jest.fn(),
+        removeListener: jest.fn(),
+        addEventListener: jest.fn(),
+        removeEventListener: jest.fn(),
+        dispatchEvent: jest.fn(),
+      }));
+      Object.defineProperty(globalThis, 'innerHeight', {
+        writable: true,
+        configurable: true,
+        value: innerHeight,
+      });
+      return jest
+        .spyOn(HTMLElement.prototype, 'getBoundingClientRect')
+        .mockReturnValue({
+          top: cardTop,
+          bottom: cardTop + cardHeight,
+          height: cardHeight,
+        } as DOMRect);
+    };
+
+    const inset = () => document.documentElement.style.getPropertyValue('--yc-consent-inset');
+
+    afterEach(() => {
+      jest.restoreAllMocks();
+      document.documentElement.style.removeProperty('--yc-consent-inset');
+    });
+
+    it('reserves everything from the card top to the bottom of a phone viewport', () => {
+      const rect = mockViewport({ isPhone: true, innerHeight: 844, cardTop: 508, cardHeight: 252 });
+
+      render(<Cookies />);
+
+      // 844 - 508 = 336, and the card itself is only 252 tall. The extra 84 is
+      // the tab-bar reserve the card is docked above, which is not somewhere
+      // content can go either.
+      expect(inset()).toBe('336px');
+      rect.mockRestore();
+    });
+
+    it('reserves nothing on desktop, where the card is placed clear of the page', () => {
+      const rect = mockViewport({
+        isPhone: false,
+        innerHeight: 900,
+        cardTop: 494,
+        cardHeight: 276,
+      });
+
+      render(<Cookies />);
+
+      expect(inset()).toBe('');
+      rect.mockRestore();
+    });
+
+    it('gives the strip back once a choice has been made', async () => {
+      const rect = mockViewport({ isPhone: true, innerHeight: 844, cardTop: 508, cardHeight: 252 });
+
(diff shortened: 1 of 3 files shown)


ankit-yc (approved):
Approving. The mechanism is right, the tests are aimed at the value rather than the class list, and the fixture note — that `cardHeight` is deliberately not `innerHeight - cardTop` — is the detail that makes the second mutation able to land at all.

Checked rather than assumed:

- **`getBoundingClientRect().top` on a `position: fixed` element is viewport-relative**, so `innerHeight - top` really is the denied strip, and it is correctly larger than the card (336 vs 252) because the 84px tab-bar dock is not somewhere content can go either.
- **Re-reading the breakpoint inside `publish()` rather than closing over it** is load-bearing: a rotation crosses `md` without remounting, and a mount-time read would keep reserving 336px on a landscape phone that has placed the card clear.
- **`ResizeObserver` on the card plus a `resize` listener** covers both axes — the copy rewrapping (which changes height without a viewport event) and a width change that does not alter the card.
- The cleanup clears on unmount *and* on consent, and both have a test.

## One follow-up, explicitly not a blocker on this PR

The property is published globally on `documentElement`, but `AuthShell` is the only consumer. The card renders on **every** route, including signed-in PIMS, and I measured it there at 390x844 on `origin/dev` plus your branch's behaviour:

```
route        card top=508 bottom=760 h=252   denies 336px of an 844px viewport
/dashboard   doc scrollHeight 1064 / 844   can scroll: yes
/chat        doc scrollHeight  890 / 844   can scroll: yes
/finance     doc scrollHeight  844 / 844   CAN SCROLL: NO
```

`/finance` has exactly the property that made sign-in a blocker — a page that cannot scroll under an opaque card. On a data-less local render there is nothing beneath the card to block, so I am **not** claiming a defect; with invoice rows there would be. And on `/dashboard` one control is already hit-tested as covered by the card's own body text:

```
CARD-BLOCKED  button.videos-card-tile :: "Play video: Create your clin"   top=div.text-caption-1 (the card)
```

Two phone-only shells use the same fixed-height geometry as `AuthShell` and consume nothing:

```
PhoneWorkspaceShell.tsx:98      h-[calc(100dvh-54px-72px-env(safe-area-inset-bottom,0px))]
PhoneCompanionRecord.tsx:387    h-[calc(100dvh-54px-72px-env(safe-area-inset-bottom,0px))]
```

Worth a separate issue rather than widening this PR — the reservation is the same one line, but each shell needs its own before/after.

## A measurement warning for whoever picks that up

My first pass reported the phone tab bar's `Home` button as unreachable on all three routes. It is not — the covering element is **`nextjs-portal`**, the Next.js dev-mode overlay. Attributing "unreachable" to *what is actually on top* rather than to the nearest suspicious thing turned three phantom blocks into one real one. You moved to a production build for the same reason and said so in the PR; this is the hit-test version of it.

Issue #2782, opened 2026-09-06
Estimates has no phone layout: Total and Actions start off-screen at 390px


aupyay:
### Preflight

Searched open issues; no existing report for the estimates screen at a phone width.

### Affected area

`apps/frontend` - `features/finance/pages/Estimates`

### What is wrong

`Estimates` is the only finance list with no phone branch. `Finance`, `Inventory` and `Tasks` each reference `useIsPhone`; `Estimates`, `Discounts` and `Dashboard` reference it zero times. So on a phone, `/finance/estimates` renders `EstimateList`'s six-column `GenericTable`.

This is not a layout-overflow bug and should not be treated as one. The table sits in a `.TableBodyScroll` rail with `overflow-x: auto` (`scrollWidth` 646, `clientWidth` 364) and the page never scrolls sideways - `document.documentElement.scrollWidth === clientWidth === 390`. The rail does its job.

The problem is what the rail hides on arrival. Column headers hit-tested against a 390px viewport:

```
  visible    Companion     left=13   right=201
  visible    Status        left=201  right=314
OFFSCREEN    Created       left=314  right=392
OFFSCREEN    Valid until   left=392  right=488
OFFSCREEN    Total         left=488  right=575
OFFSCREEN    Actions       left=575  right=659
```

Two of six columns are visible. `Total` - the field an estimates list is consulted for - begins 98px past the right edge, and there is no scrollbar, gradient or shadow indicating more exists; the cut lands mid-word in "CREATED" at the viewport edge.

This screen is a routine phone destination, not an obscure one. `features/finance/pages/Finance/index.tsx` renders an Estimates link inside its own `isPhone` branch, with a comment explaining the link was added precisely so the screen is reachable on a phone.

### Steps to reproduce

1. Sign in to PIMS and open `/finance` on a phone (or a 390x844 viewport).
2. Tap **Estimates** in the row of links above the invoice list.
3. Look at the list without scrolling it horizontally.

### Expected behavior

The amount of each estimate is legible on arrival, as it is on `/finance`, which renders `PhoneInvoiceList` below `md`.

### Actual behavior

A six-column table showing Companion and Status. The amount and the row action are reached only by discovering that the table scrolls sideways.

### How often does it reproduce

Every time, at any width below the `md` breakpoint. Both themes.

### Did this work before

No - `Estimates` has never had a phone branch. `Finance` gained `PhoneInvoiceList`; the sibling screen did not follow.

### Impact

[...]

PR #2789, opened 2026-09-06
feat(frontend): give the insurance claims list a phone form so status is on screen

### LINKED ISSUE #2786: Insurance claims has no phone layout: Status and Actions start off-screen at 390px
### Preflight

Same class as #2782 (estimates), found while re-checking which finance routes the phone branch actually links to. Filed separately because it is a different screen with a different fix.

### Affected area

`apps/frontend` - `features/finance/pages/InsuranceClaims`

### What is wrong

`InsuranceClaims` has zero references to `useIsPhone`, so a phone renders its seven-column table. Column headers hit-tested against a 390px viewport:

```
  visible    Insurer / policy   left=13   right=161
  visible    Claim no.          left=161  right=248
  visible    Submitted          left=248  right=339
OFFSCREEN    Approved           left=339  right=426
OFFSCREEN    Paid               left=426  right=493
OFFSCREEN    Status             left=493  right=659
OFFSCREEN    Actions            left=659  right=744
```

Three of seven columns are visible. **`Status` begins 103px past the right edge** - on a claims list that is the field the screen exists to answer, more so than the amount is on an estimates list.

As with #2782 this is not a layout-overflow bug: the table sits in a `.TableBodyScroll` rail (`scrollWidth` 731, `clientWidth` 364, `overflow-x: auto`) and the page does not scroll sideways. The rail works; what is wrong is that the decisive column is behind an undiscoverable horizontal swipe.

The screen is a routine phone destination. `features/finance/pages/Finance/index.tsx` renders an Insurance link inside its own `isPhone` branch, alongside the Estimates and Discounts links, specifically so these screens are reachable on a phone.

For completeness, the three routes that phone branch links to:

| Route | `useIsPhone` refs | Renders a table on a phone |
| --- | --- | --- |
| `/finance/estimates` | 0 | yes - #2782 |
| `/finance/insurance-claims` | 0 | yes - this issue |
| `/finance/discounts` | 0 | no - it is a form, nothing off-screen |

### Steps to reproduce

1. Sign in to PIMS and open `/finance` at a 390x844 viewport.
2. Tap **Insurance** in the row of links above the invoice list.
3. Look at the list without scrolling it horizontally.

### Expected behavior

Each claim's status is legible on arrival, as invoice status is on `/finance`, which renders `PhoneInvoiceList` below `md`.

### Actual behavior

A seven-column table showing insurer, claim number and submitted date. Status, approved and paid amounts, and the row action are reached only by discovering that the table scrolls sideways.

Every time, below the `md` breakpoint. Both themes - the filter row also wraps to three rows and 112px at this width.

### Did this work before

No. `InsuranceClaims` has never had a phone branch.

### Impact

Medium, and slightly higher than #2782: the hidden column here is a status rather than a number, so the list cannot be triaged at a glance at all.

### Environment

`origin/dev` at `6362c0fd2`. Measured against the rendered DOM at 390x844 in Chromium, both themes.

aupyay:
## PR Checklist

`InsuranceClaims` has zero `useIsPhone` references, so a 390px phone renders `InsuranceClaimList`'s seven-column table inside a 364px scroll rail:

```
  visible    Insurer / policy   left=13   right=161
  visible    Claim no.          left=161  right=248
  visible    Submitted          left=248  right=339
OFFSCREEN    Approved           left=339  right=426
OFFSCREEN    Paid               left=426  right=493
OFFSCREEN    Status             left=493  right=659
OFFSCREEN    Actions            left=659  right=744
```

**Not a layout-overflow bug, and the PR does not claim it is.** The rail scrolls and the page does not (`documentElement.scrollWidth === clientWidth === 390`). I screened every viewport escape across the finance, inventory, tasks and dashboard stories - 819 of them - and all 819 sit inside a scrolled `overflow-x` ancestor. Zero unrescued.

What is wrong is that `Status` begins 103px past the edge behind an undiscoverable swipe. On a claims list that is the one field that cannot be inferred from anywhere else, which makes it a worse hide than `Total` was on estimates (#2782).

**How this was found is worth recording**, because it is the same mistake the sibling issue warns about. When I filed #2782 I quoted the Estimates link in `Finance/index.tsx`'s `isPhone` branch and then reasoned about the other routes from memory. That branch renders **three** links. Re-reading the eight lines rather than remembering them turned up this screen, which I had never rendered.

## What is the new behavior?

`PhoneInsuranceClaimList` below `md`, mirroring `PhoneEstimateList` and `PhoneInvoiceList`.

- **Status leads the card**, because it is the column the table pushed furthest off-screen.
- **One money figure, not three.** Paid if present, else approved, else the submitted ask - each labelled, so a number is never ambiguous about what it represents. A card cannot give three amounts equal weight, and equal weight is not what an operator wants: what matters is how far the claim has got.
- **Same accessible name** as the table's row action, so the page's existing tests and a screen reader see one label across both breakpoints.
- Insurer and policy **wrap rather than clipping**. The table truncates both and offers the full value through `title`, which needs a hover this surface does not have.

**Deliberately no nullable money helper.** The table needs one because Approved and Paid are their own columns and each can be null. Here every path has already established the figure it formats, so the dash branch was unreachable - the first version scored 92.85% branches on exactly that dead arm. Deleted rather than tested; writing a test to reach dead code is how the arm survives.

## Verification

Measured at the width the route gives the list - `main` 390, `yc-page-content` 12px gutters, so **366px** - in both themes.

**Every added assertion mutation-checked.** Each mutation applied, watched go red, reverted, watched go green:

| Mutation | Result |
| --- | --- |
| `paidAmount != null` -> `paidAmount` (falsy) | the paid-zero test fails |
| paid/approved precedence flipped | 2 tests fail |
| claim-number join always applied | the policy-line test fails |
| component swap pinned to the desktop table | the phone-branch page test fails |

The `paidAmount: 0` case is the one worth having: a settled-for-nothing claim is real, and a falsy check reports the *approved* figure as if it had been paid.

Story assertions are pinned to 366px by a decorator rather than a `viewport` global, for two reasons. `.storybook/test-runner.ts` falls back to laptop 1280x800 for an unpinned story, where every geometry assertion here would hold regardless of what the component did. And that hook reads `storyContext.globals` while Storybook 10 carries the annotation as `storyGlobals`, so viewport pins are currently inert - a decorator is the only width these stories are actually given.

```
frontend  tsc --noEmit                     exit 0
frontend  eslint                           exit 0 (2 pre-existing warnings, files untouched)
          jest InsuranceClaims             34 passed, 2 suites
          coverage PhoneInsuranceClaimList 100% statements / branches / functions / lines
```

## Related Issue(s)

Fixes #2786

### DIFF
--- a/apps/frontend/src/app/__tests__/features/finance/InsuranceClaims.test.tsx
+++ b/apps/frontend/src/app/__tests__/features/finance/InsuranceClaims.test.tsx
@@ -3,6 +3,17 @@ import { render, screen, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import '@testing-library/jest-dom';
 
+/* The results section picks its list by breakpoint. Default false so every case
+   below keeps exercising the desktop table; the phone branch flips it. Both
+   export shapes are mocked because the module is imported by default here and
+   by name elsewhere. */
+let mockIsPhone = false;
+jest.mock('@/app/ui/layout/PhoneShell/useIsPhone', () => ({
+  __esModule: true,
+  default: () => mockIsPhone,
+  useIsPhone: () => mockIsPhone,
+}));
+
 // Renders children so the gated action row is exercised; the permission logic
 // itself is covered by PermissionGate's own tests.
 jest.mock('@/app/ui/layout/guards/PermissionGate', () => ({
@@ -373,5 +384,46 @@ describe('InsuranceClaims create form', () => {
       'The submitted amount must be above zero.'
     );
     expect(onCreate).not.toHaveBeenCalled();
+  });
+});
+
+describe('InsuranceClaims on a phone', () => {
+  beforeEach(() => {
+    mockIsPhone = true;
+  });
+  afterEach(() => {
+    mockIsPhone = false;
+  });
+
+  it('renders the card list instead of the seven-column table', () => {
+    setup();
+
+    // The card carries the table row action's accessible name, so that alone
+    // cannot tell the two apart - the table's own roles can.
+    expect(
+      screen.getByRole('button', { name: 'Open the claim for Bought By Many' })
+    ).toBeInTheDocument();
+    expect(screen.queryByRole('table')).not.toBeInTheDocument();
+    expect(screen.queryByRole('columnheader', { name: 'Status' })).not.toBeInTheDocument();
+
+    /* Status was the column that began 103px off-screen in the table; on the
+       card it leads the row. Scoped to the card rather than the document: the
+       status pill renders the label in a nested span, so a bare getByText
+       matches twice and cannot say WHICH card carries it. */
+    const partiallyApproved = screen
+      .getAllByRole('button', { name: /^Open the claim for / })
+      .find((card) => card.textContent?.includes('Partially approved'));
+    expect(partiallyApproved).toBeDefined();
+    expect(
+      within(partiallyApproved as HTMLElement).getAllByText('Partially approved').length
+    ).toBeGreaterThan(0);
+  });
+
+  it('still renders the table when the viewport is not a phone', () => {
+    mockIsPhone = false;
+    setup();
+
+    expect(screen.getByRole('table')).toBeInTheDocument();
+    expect(screen.getByRole('columnheader', { name: 'Status' })).toBeInTheDocument();
   });
 });
--- a/apps/frontend/src/app/__tests__/features/finance/Sections/PhoneInsuranceClaimList.test.tsx
+++ b/apps/frontend/src/app/__tests__/features/finance/Sections/PhoneInsuranceClaimList.test.tsx
@@ -1,0 +1,130 @@
+import React from 'react';
+import { fireEvent, render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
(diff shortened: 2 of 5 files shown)


ankit-yc (approved):
Approving at `4a691597c80821f1cdc230b36b42c5d11f1c4dc0`.

Same swap shape as #2783 and consistent with it, which is the right call - two screens fixed the same way is a pattern, two screens fixed two ways is a maintenance problem.

Four things I checked rather than took from the description:

- `claim.paidAmount != null` and `claim.approvedAmount != null`, not truthiness. A settled-at-zero claim reports Paid 0.00 rather than falling through to the approved figure, which is the wrong-number-that-looks-right you named. Nullish on both rungs, not just the one the mutation exercised.
- `settledFigure` labels the number it returns, so a card never shows a figure whose meaning depends on knowing which branch produced it. That is the part that makes one column defensible where the table had three.
- `InsuranceClaimsResults` is rendered as JSX at the call site, so `useIsPhone()` lands inside a component with a stable hook order. Worth confirming rather than assuming, because the swap put a hook into a function that previously had none.
- The insurer name and the policy/claim line both wrap instead of clipping, and the comment says why in terms of the affordance rather than the pixels - a `title` needs a hover this surface does not have, and a card can grow downwards.

On the dead branch you deleted rather than covered: that was the right way round. A test written to reach an unreachable arm would have pinned the arm in place and read as coverage.

No change requests.

PR #2793, opened 2026-09-06
fix(frontend): read storyGlobals so a phone story is tested at phone width

### LINKED ISSUE #2792: Storybook test-runner reads a key Storybook 10 does not provide, so every story is tested at 1280px
## What is happening

`apps/frontend/.storybook/test-runner.ts` `preVisit` resolves each story's target viewport with:

```ts
const requested = (storyContext.globals?.viewport as { value?: string })?.value;
const key = requested ?? DEFAULT_VIEWPORT;   // 'laptop' -> 1280x800
```

Storybook 10's story context has **no `globals` key**. A story's own `globals` annotation is carried as `storyGlobals`. Logged context keys:

```
storyGlobals, componentId, title, kind, id, component, subcomponents, tags,
parameters, initialArgs, argTypes, moduleExport, name, story, ...
```

So `requested` is `undefined` for every story, and every story falls through to `laptop`:

```
[VP-PROBE] appointments-viewappointmentoverviewmodal--overview-narrow  requested=undefined  applied=1280
[VP-PROBE] appointments-calendar-phonedaystrip--default                requested=undefined  applied=1280
```

A story named "Upcoming at 375px" is tested at 1280px. Meta-level and story-level pins behave identically.

This is the exact failure the hook's own comment says it was written to prevent:

> without this hook every story renders at Playwright's default and each phone-pinned story fails for a reason that does not exist in Storybook or in Chromatic

## Scale

Read out of the job log, `dev` @ `6362c0fd2`, run `34020133806`:

```
shard 1/4   27 failed, 140 passed, 167 total
shard 2/4   34 failed, 133 passed, 167
shard 3/4   21 failed, 146 passed, 167
shard 4/4   20 failed, 146 passed, 166
            102 failures / 101 distinct suites of 667
```

The workflow conclusion is `success`.

## Why it has been invisible

`storybook-interactions.yml` sets `continue-on-error: true` on the `Run play functions` step. That forces the step **conclusion** to `success` regardless of what happened; the real result is `outcome`, which the REST API does not expose per step. Reading the step conclusion - the careful-looking move - reports these as passing. Two agents did exactly that this morning.

The advisory status is deliberate and documented in the workflow header. That is not the defect. The defect is that the hook inside it does nothing.

## Mechanism, confirmed rather than inferred

Same story, two widths, nothing else changed:

```
375   grid grid-cols-1 md:grid-cols-2   tracks=1   <- what the story asserts
1280  grid grid-cols-1 md:grid-cols-2   tracks=2   <- "expected 2 to be 1", the CI failure verbatim
```

`ViewAppointmentOverviewModal > OverviewNarrow > play-test` fails in CI with exactly that message, and reproduces locally on clean `origin/dev`.

## Consequence

266 story files declare a viewport global; 249 of them pin `mobile`. Every geometry assertion in those files has been evaluated at laptop width - so it is either failing as a phantom, or passing vacuously about a width the story does not mean. Pinning width with a decorator instead is immune, and is the safe pattern until this is fixed.

aupyay:
## PR Checklist

See #2792. `.storybook/test-runner.ts` `preVisit` reads `storyContext.globals?.viewport`. Storybook 10's story context has **no `globals` key** - a story's own `globals` annotation is carried as `storyGlobals`. So `requested` is `undefined` for every story and every one falls through to `DEFAULT_VIEWPORT` (laptop, 1280x800):

```
[VP-PROBE] appointments-viewappointmentoverviewmodal--overview-narrow  requested=undefined  applied=1280
[VP-PROBE] appointments-calendar-phonedaystrip--default                requested=undefined  applied=1280
```

A story named "Upcoming at 375px" is tested at 1280px. Meta-level and story-level pins fail identically.

That is the exact failure the hook's own comment says it exists to prevent. On `dev` @ `6362c0fd2` (run `34020133806`), read out of the log rather than the check conclusion: **102 failures across 101 distinct suites of 667**, with the workflow reporting `success` because the step is `continue-on-error`.

## What is the new behavior?

Read `storyGlobals` first, keep `globals` as a fallback so this keeps working if the key returns.

Mechanism confirmed rather than inferred - same story, two widths, nothing else changed:

```
375   grid grid-cols-1 md:grid-cols-2   tracks=1   <- what the story asserts
1280  grid grid-cols-1 md:grid-cols-2   tracks=2   <- "expected 2 to be 1", the CI failure verbatim
```

## Measured effect

39 suites across five phone-heavy directories, same running Storybook, one variable changed:

| directory | before | after |
| --- | --- | --- |
| `AppointmentWorkspace/phone` | 1 failed, 4 passed | **0 failed, 5 passed** |
| `Calendar/responsive` | 0 failed, 7 passed | 0 failed, 7 passed |
| `companionHistory/components` | 2 failed, 12 passed | **1 failed, 13 passed** |
| `ui/layout/PhoneShell` | 1 failed, 5 passed | **0 failed, 6 passed** |
| `companions/components/Sections` | 2 failed, 5 passed | 2 failed, 5 passed |
| **total** | **6 failed, 33 passed** | **3 failed, 36 passed** |

**Three repaired, none newly broken, no directory worse.** The three that remain fail for reasons unrelated to viewport.

Stated plainly: that sample is 39 of 667 and is weighted toward phone components, so it does **not** prove the whole suite goes green. It does show the fix repairs phantom failures without converting vacuous 1280px passes into a wave of new ones - which was the risk worth measuring before proposing this.

## A note on what this does not change

`storybook-interactions.yml` stays `continue-on-error`. That is deliberate and documented in the workflow header, with a note on how to promote it to required. Promoting it is a separate decision and should wait until the remaining failures are triaged - this PR does not make the job blocking.

## Testing

```
tsc --noEmit    0
eslint          0 errors
```

Verified by running the same five directory patterns before and after with the runner file as the only difference, asserting each run actually executed (a `Test Suites:` line present) rather than trusting the exit code - an earlier attempt exited 0 having run nothing, because an unquoted `|` in the pattern was interpreted as a shell pipeline.

## Related Issue(s)

Fixes #2792

### DIFF
--- a/apps/frontend/.storybook/test-runner.ts
+++ b/apps/frontend/.storybook/test-runner.ts
@@ -40,7 +40,15 @@ const DEFAULT_VIEWPORT = 'laptop';
 const config: TestRunnerConfig = {
   async preVisit(page, context) {
     const storyContext = await getStoryContext(page, context);
-    const requested = (storyContext.globals?.viewport as { value?: string } | undefined)?.value;
+    /* `storyGlobals`, not `globals`. Storybook 10's story context carries a
+       story's own `globals` annotation under `storyGlobals`; `globals` is not a
+       key on it at all, so this read was `undefined` for EVERY story and every
+       one of them fell through to `laptop`. That is the exact failure this hook
+       was written to prevent, and it was silent because the job is
+       `continue-on-error`. */
+    const storyGlobals = (storyContext as unknown as Record<string, unknown>).storyGlobals as
+      Record<string, { value?: string } | undefined> | undefined;
+    const requested = (storyGlobals?.viewport ?? storyContext.globals?.viewport)?.value;
     const key = requested ?? DEFAULT_VIEWPORT;
     const size = VIEWPORT_SIZES[key];
 


ankit-yc (approved):
Approving. I verified the diagnosis against the shipped types rather than taking it, and it holds.

**`storyGlobals` is real and typed** in Storybook 10.6.0:

```
storybook/dist/chunk-D4lKaBz7.d.ts:421   storyGlobals: import("storybook/internal/csf").Globals;
storybook/dist/chunk-zQu03vfn.d.ts:229   storyGlobals?: Globals$1;
```

**And the cast is necessary, not lazy — worth a line in the file so nobody "cleans it up".** `getStoryContext` is declared as returning `StoryContextForEnhancers`:

```
@storybook/test-runner/dist/index.d.ts:73
  getStoryContext: (page, context) => Promise<StoryContextForEnhancers>
```

`storyGlobals` is on **`PreparedStory`**, not on `StoryContextForEnhancers`. So the runtime object carries it and the declared type does not — which is exactly the situation a cast is for. Someone tidying this later will delete the cast, get a type error, and reach for the wrong fix.

## The one thing I would add, and it is the reason this recurred rather than the reason it broke

The hook still cannot distinguish **"this story requested no viewport"** from **"this story requested one and I could not read it"**:

```ts
const key = requested ?? DEFAULT_VIEWPORT;
```

Both produce `laptop`, silently. That is the actual defect class — the key name was only the instance. When Storybook renames or restructures this again, the hook degrades to 1280px exactly as before, `continue-on-error` keeps it quiet, and the next person finds it in six months.

The file already models the right instinct one branch down: an *unknown* viewport key **throws**, loudly, rather than falling back. The unreadable-annotation case deserves the same treatment and currently gets the opposite.

Two ways, either fine:

- **Extract the resolution into a pure exported function** and unit-test it against a literal Storybook 10 context object. A key rename then fails a test instead of degrading a suite.
- **A suite-level canary**: count stories that resolved a non-default viewport and fail the run if the count is zero. We know phone-pinned stories exist, so zero is impossible and means the reader is broken. That is the same "a control must produce a different outcome" shape this whole thread has converged on.

Not blocking — the fix is correct and the measurement supporting it is honest, including the part that limits it.

## On the numbers

`6 failed / 33 passed -> 3 failed / 36 passed` across 39 suites, **no directory worse**, is the right evidence for this change and I am glad you ran it before proposing rather than after. Your own caveat is the important half: 39 of 667, weighted to phone components, so it shows the fix repairs phantoms without converting vacuous 1280px passes into a wave of new failures — it does not show the suite goes green. Stated that way in the PR body, which is what I would want a reader to see.

For what it is worth on my side: the two components I merged today are decorator-pinned and never consult the viewport global, so they should be inert under this. If they go red after it lands, that is a finding about my components and I want to hear it.

Issue #2790, opened 2026-09-06
PIMS phone workspace: the patient's name is truncated by the status pill, and 11 more values clip with no remedy


aupyay:
Measured on the rendered DOM at 390x844, both themes, across the 48 phone-pinned stories under `Workspace/`.

## 1. The patient's name loses to the status chip

`AppointmentWorkspace/phone/PhonePatientBar.tsx:88`:

```tsx
<span className="truncate text-[15px] font-bold ... text-(--ink)">{companionName}</span>
<AppointmentStatusPill appointment={appointment} />
```

The name and the pill share a flex row. The name carries `truncate`; the pill does not shrink. So the name absorbs the entire shortfall:

```
viewport                        390
back button                      34
avatar                           38
flex-1 area                     197
  patient name   gets  80   needs 114   LOST 34
  status pill    gets 111              (not clipped)
visit timer                      63
```

**The status chip takes 111px and the patient's name gets 80.** "Poppy Hartmann" is fourteen characters and already clips; any longer client name is worse. No `title`, and a phone has no hover, so the full name is unreachable.

This is the bar a clinician reads to confirm which animal is in front of them, in the workspace where they record clinical findings.

Same clipping in `PhoneWorkspaceShell` (70px lost) and every `PhonePatientBar` state story (34-55px), in both themes.

## 2. Ten more clipped values with no affordance at all

Not the hover-tooltip class - these have no `title` either:

| component | value | lost |
| --- | --- | --- |
| `OutpatientSchedule` | "Photobiomodulation and underwater treadm..." | 476px |
| `OutpatientSchedule` | "09:30 - 20 min - Dr. Ravi Menon - Rehab ..." | 166px |
| `PhoneWorkspaceShell` | patient name | 70px |
| `PhonePatientBar` | patient name | 34-55px |

And one that does have a hover-only remedy, which a phone cannot use:

| `DiagnosticsStep` | "Order notes: Fasted sample, collected fr..." | 80px | `title` present |

## Why a fix needs a decision rather than a patch

197px cannot hold a 114px name and a 111px pill. Four candidates measured by injection on `workspace-phonepatientbar--running`:

| candidate | name lost | side effect |
| --- | --- | --- |
| baseline | 34px | - |
| A: `flex-wrap` on the row | **0** | bar height 59 -> 88px (+29) |
| B: pill shrinks before the name | 17px | halves it, does not fix it |
[...]

PR #2794, opened 2026-09-06
fix(frontend): wrap the appointment overview values instead of clipping them

aupyay:
## PR Checklist

- [x] All existing tests and lints pass (one caveat, stated below).

`ViewAppointmentOverviewModal`'s `OverviewRow` renders every value with `max-w-[60%] truncate` and no `title`. `truncate` clips with an ellipsis, and a phone has no hover, so the tail of any long value is unreachable there.

Measured on the rendered DOM at 390x844, both themes - exactly one of the eleven rows overflows:

```
ok        0px  Date / Time / Duration / Speciality / Service / Emergency
CLIPPED  62px  Chief complaint    white-space: nowrap    title: none
              "Limping on the left hind leg since Sunday."
```

It is the one free-text clinical field on the sheet, and the reason a clinician opens it.

## What is the new behavior?

The value wraps instead of truncating. `items-start` so the label sits against the first line of a value that can now run to several, and a `gap-3` so a long label and a wrapped value cannot meet in the middle.

After, both themes: **0 clipped rows**, chief complaint on two lines. No other row's rendering changes - every one of them fits on a single line at 390px before and after.

## The story already claimed this, and could not check it

```
// Both columns still render in full - the layout reflows, it does not truncate.
await expect(panel.getByText('General practice')).toBeInTheDocument();
```

`getByText` matches the DOM text node, which is complete even when the paint is clipped. So the assertion written for exactly this property could not fail on it. Replaced with a measurement across every value cell:

```ts
const clipped = values.filter((cell) => cell.scrollWidth > cell.clientWidth + 1)
expect(clipped).toEqual([])
```

**Mutation**: restoring `truncate` fails it with `expected [ Array(1) ] to deeply equal []`.

## One honest caveat about CI

That play function currently fails on `dev` before it ever reaches the new assertion, because the test runner executes it at 1280px rather than the 375px it pins - #2792, fixed by #2793. With that fix applied locally this suite is **6 passed, 6 total**. Without it, the new assertion is inert in CI rather than wrong. I would merge #2793 first.

## Testing

```
tsc --noEmit                                    0
jest --testPathPatterns=ViewAppointmentOverviewModal   45 passed
test-storybook ViewAppointmentOverviewModal     6 passed   (with #2793 applied)
```

Verified visually at 390x844 in both themes.

## Related Issue(s)

Refs #2790

### DIFF
--- a/apps/frontend/src/app/features/appointments/pages/Appointments/Sections/ViewAppointmentOverviewModal/index.stories.tsx
+++ b/apps/frontend/src/app/features/appointments/pages/Appointments/Sections/ViewAppointmentOverviewModal/index.stories.tsx
@@ -439,6 +439,18 @@ export const OverviewNarrow: Story = {
     await expect(panel.getByText('General practice')).toBeInTheDocument();
     await expect(panel.getByRole('button', { name: 'Room: Consult 2' })).toBeInTheDocument();
     await expect(panel.getByRole('button', { name: 'Start appointment' })).toBeEnabled();
+
+    /* "it does not truncate" was the claim above, and none of those assertions
+       could check it: `getByText` matches the DOM text node, which is complete
+       even when the paint is clipped. Chief complaint WAS truncating here,
+       losing 62px with no `title` to recover it on a surface that has no hover.
+       Measured instead - a clipped cell has more scrollWidth than clientWidth. */
+    const values = panel.getAllByTestId('overview-row-value');
+    await expect(values.length).toBeGreaterThan(0);
+    const clipped = values
+      .filter((cell) => cell.scrollWidth > cell.clientWidth + 1)
+      .map((cell) => `${cell.textContent?.trim()} (+${cell.scrollWidth - cell.clientWidth}px)`);
+    await expect(clipped).toEqual([]);
   },
   parameters: {
     docs: {
--- a/apps/frontend/src/app/features/appointments/pages/Appointments/Sections/ViewAppointmentOverviewModal/index.tsx
+++ b/apps/frontend/src/app/features/appointments/pages/Appointments/Sections/ViewAppointmentOverviewModal/index.tsx
@@ -89,10 +89,22 @@ const getParentPhotoUrl = (
     appointmentParent?.image
   );
 
+/* The value WRAPS rather than truncating. `truncate` clips with an ellipsis and
+   no `title`, so on a phone - which has no hover - the tail of any long value is
+   simply unreachable. At 390px exactly one row overflows, and it is
+   "Chief complaint": the one free-text clinical field on this sheet, losing 62px
+   of "Limping on the left hind leg since Sunday." in both themes. A tooltip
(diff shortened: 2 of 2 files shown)


ankit-yc (approved):
**Approved at `726657b29`.** The measurement replaces an assertion that could not fail, the mutation kills it, and I reproduced both rather than taking them. Two points, neither blocking.

Verified in my own worktree on your branch, one Storybook, one variable changed each time:

```
as-is                              6 passed, 6 total                              rc 0
break-words -> truncate            OverviewNarrow FAILS  expected [ Array(1) ] to deeply equal []
                                   1 failed, 5 passed  <- only the story that owns the property
```

Also checked the thing I was ready to flag and did not: `data-testid` is already a production convention here, 28 non-story components carry one. No point to make.

## 1. The guard measures the fixture, not the CSS — and I can show it

```
truncate restored + concern: 'Limping on the left hind leg since Sunday.'   OverviewNarrow FAILS
truncate restored + concern: 'Limping'                                     OverviewNarrow PASSES
```

Second row is the buggy CSS going green. `getAllByTestId('overview-row-value')` plus `length > 0` proves *some* cells were measured; it cannot prove the **one cell that clips** was among them. The row that motivated this change is the only one that overflows at 375px, so the whole guard rests on `concern` staying long enough — a string in a shared fixture that three other stories also read and that nothing marks as load-bearing.

One line fixes it, and it is the assertion your own comment already implies:

```ts
const complaint = panel.getByText('Limping on the left hind leg since Sunday.');
await expect(complaint.scrollWidth).toBeLessThanOrEqual(complaint.clientWidth + 1);
```

Then `clipped` stays as the sweep over every row and the named cell carries the property. Same shape as the point you took on #2797 — the difference between "nothing regressed" and "this thing did not regress" — and the same one that just caught me measuring three of five chips.

## 2. Your labelling is elided by the printer

You build the failure message carefully:

```ts
.map((cell) => `${cell.textContent?.trim()} (+${cell.scrollWidth - cell.clientWidth}px)`)
```

and what a reader actually gets is:

```
expected [ Array(1) ] to deeply equal []
```

Neither the cell nor the 62px survives. `toEqual` on an array prints the shape, not the contents, so the work that would tell the next person *which* row and *by how much* never reaches them. Asserting on a joined string prints it:

```ts
await expect(clipped.join(' | ')).toBe('');
// Chief complaint ... (+62px)
```

Directly your own #2800 sentence — a guard that fires without saying what to do is the one nobody acts on.

## On the fix itself

`items-start` with `gap-3` is right and the reasoning holds: a value that can run to several lines should not be centre-aligned against its label, and the gap stops a long label meeting a wrapped value. `max-w-[60%]` retained means wrapping happens inside the same column the design already gave it, so no other row moves — which matches your 0-clipped-rows-after reading.

`break-words` rather than a `title` attribute is the right call for the stated reason, and it is worth keeping in the comment: a tooltip cannot be reached on the surface where the truncation matters.

PR #2796, opened 2026-09-06
fix(frontend): take the inventory filter chips off the badge-blue pair, 3.61:1

aupyay:
## PR Checklist

> **Rewritten at `c5bf9b59d`. The previous body of this PR stated a mechanism and a
> before-number that were both wrong, and it was approved on the strength of them.**
> The retraction is at the bottom; the corrected finding is below.

The selected-filter chips in the inventory filter drawer render
`text-badge-blue-text` on `bg-badge-blue-bg` — `#eaf3ff` on `#007cf5` — which
measures **3.61:1** at 14px/500, where WCAG AA needs 4.5:1.

Measured on composited pixels, both themes, at 390x844:

```
                 ink              fill               ratio   needs
before  light    rgb(234,243,255) rgb(0,124,245)      3.61     4.5   FAIL
before  dark     rgb(234,243,255) rgb(0,124,245)      3.61     4.5   FAIL
```

The pair has **no dark variant** — `--color-badge-blue-bg`/`-text` are declared once
in `globals.css:476-477` with no `html[data-theme='dark']` override — so the failure
is theme-independent. Both theme-pinned stories fail identically when the classes are
restored, which is the measurement, not the grep.

## What is the new behavior?

```
after   light    rgb(29,28,27)    rgb(234,226,213)   13.24     4.5   PASS
after   dark     rgb(244,239,230) rgb(48,40,32)      12.64     4.5   PASS
```

### Why not re-ink the badge-blue pair

```
ink #eaf3ff  on #007cf5   3.61:1  FAIL
ink #ffffff  on #007cf5   4.04:1  FAIL   already recorded at globals.css:1693
ink --ink    on #007cf5   4.21:1  FAIL   the app's darkest text token
ink #000000  on #007cf5   5.20:1  PASS   the ceiling for ANY ink on this fill
```

**Correction to an earlier version of this body and my post in the channel:** I wrote
"no ink fixes it", which the last row contradicts. The accurate statement is that no
ink in the register the palette uses fixes it — every light ink fails, `--ink` fails,
and the only thing that clears 4.5 is near-black text on a saturated brand-blue pill,
which is not a design anyone would choose. That is an argument about the palette rather
than about the arithmetic, and it is the better one. The fill has to change. A removable applied-filter chip is not a status badge, so
it moves to the quiet surface pair `--inset`/`--ink`, which flips with the theme.

**This is not a new pairing invented here.** `0e5e9de36` already moved the on-hand,
available, gross-profit and margin badges off badge-blue onto exactly this pair, and
it is in this branch's ancestry. (`94c3a67c9` moved a badge-blue span without changing
the pair — it is not precedent, and crediting it was mine to correct.) This chip is
the straggler. On
`origin/dev` that leaves exactly one className string still carrying the pair —
`Inventory/index.tsx:383`, the count badge — which is #2799 and is being fixed at its
own site, not here.

The design system also has `--chip-selected-bg`/`--chip-selected-ink`
(= `--ink`/`--screen`), but that is the solid inverted pill used by the `ui/filters`
toggle chip; using it here would turn a row of five removable chips into five solid
dark pills. That is a visual-weight change rather than a contrast fix — one line if a
reviewer prefers it.

## Testing

```
tsc --noEmit                                   rc 0, 0 lines
eslint (both changed files)                    rc 0
jest --ci (whole frontend package)             1039 suites / 13147 tests passed, rc 0
test-storybook InventoryFilterModal            7 passed, 7 total, rc 0
```

### Mutation, and what each story is worth

Restoring `bg-badge-blue-bg text-badge-blue-text`:

```
WithSelections      (light-pinned)  FAILS  filter chip "low stock": 3.61:1 (needs 4.5:1)
                                           rgb(234,243,255) on rgb(0,124,245) at 14px/500
WithSelectionsDark  (dark-pinned)   FAILS  same 3.61:1 — the pair has no dark variant
Tests: 2 failed, 5 passed, 7 total
```

Both stories pin `theme` through story globals rather than inheriting the toolbar
default. **The pin is load-bearing and provably so:** with `globals: { theme: 'light' }`
on the story, an explicit `&globals=theme:dark` in the iframe URL is overridden and the
render stays light; the same URL parameter on an unpinned story does render dark
(`rgb(244,239,230)` on `rgb(48,40,32)`). So an unpinned assertion here is at the mercy
of a default rather than of anything the story says.

The helper also asserts the chip row length before measuring. That guard caught this
change measuring **three** of the five chips the seed produces — a loop over a subset
reports a clean row while the rest go unmeasured.

## Retraction of this PR's previous body

The earlier body claimed the chip carried **two** conflicting text-colour classes
(`text-[var(--ink)]` and `text-badge-blue-text`) resolved by stylesheet order, at
**1.15:1** in light and passing at 12.94:1 in dark.

None of that is true of any commit. `origin/dev` has one text colour on that chip. The
1.15/12.94 pair is `#eaf3ff` measured on the `--inset` fill:

```
#eaf3ff on --inset light rgb(234,226,213)  = 1.15   <- published as "before, light"
#eaf3ff on --inset dark  rgb(48,40,32)     = 12.94  <- published as "before, dark"
```

That is the old ink on the new fill — a state that exists in neither `dev` nor this
branch. It was **my own half-applied edit**: I had already changed the background and
not yet the ink when I took the "before" reading. Both numbers reproduce to the second
decimal from those two colours, which is how the diagnosis was confirmed rather than
guessed.

The "before" arm has to come from a clean checkout of the base commit, not from a
working tree mid-edit. The after-numbers (13.24 / 12.64) were always correct.

## Related Issue(s)

Refs #2799

---

## Head `301fed788` — rebased onto `24ed9856e`, plus one prose fix

Naming the head sha here deliberately: on this repo a rebase rewrites a review's `commit_id`
to follow the new head, so the metadata cannot tell a later reader which tree was approved.
The review body and this line can.

```
50e3db985   rebase onto 24ed9856e, patch byte-identical to what was approved at b46e41c44
301fed788   drop a story comment that still asserted the retracted mechanism (prose only)
```

The second commit answers an Aikido `AIK_AI_prefer_simplicity` MEDIUM: a comment above the
light story still read *"the contrast assertion below is only capable of failing in light"*,
which was true of the withdrawn 1.15:1 mechanism and false of the real 3.61:1 one — and it
contradicted `WithSelectionsDark` six lines below it. No code or assertion changed.

Gate and checks at this head:

```
No named external product        SUCCESS      (ran for the first time, from the rebased base)
local scan, six surfaces         exit 0       agrees with CI
tsc --noEmit                     0
test-storybook InventoryFilterModal   7 passed, 7 total
```

### DIFF
--- a/apps/frontend/src/app/features/inventory/pages/Inventory/InventoryFilterModal.stories.tsx
+++ b/apps/frontend/src/app/features/inventory/pages/Inventory/InventoryFilterModal.stories.tsx
@@ -5,6 +5,15 @@ import { expect, userEvent, waitFor, within } from 'storybook/test';
 import type { InventoryFiltersState } from '@/app/features/inventory/pages/Inventory/types';
 import { defaultFilters } from '@/app/features/inventory/pages/Inventory/utils';
 import InventoryFilterModal, { type FilterChip } from './InventoryFilterModal';
+/* Cross-feature import, and deliberately so rather than a second copy: this is
+   the probe merged with #2785, and a duplicated contrast helper is how the two
+   drift apart on the thing they both exist to measure. It belongs in
+   `@/app/lib`; moving it means touching a file another desk merged an hour ago,
+   so that is a separate change and not mine to make unasked. */
+import {
+  describeContrast,
+  measureContrast,
+} from '@/app/features/appointments/components/Calendar/responsive/contrastProbe';
 
 const LOCATIONS = ['Main store', 'Pharmacy fridge', 'Surgery trolley'];
 const CATEGORIES = ['Medicine', 'Vaccine', 'Consumable'];
@@ -206,6 +215,56 @@ const findOpenDialog = async () => {
   return openDialog();
 };
 
+/** Every chip the seeded selection produces, in render order: status, then the two
+    categories, then the subcategory, then the location. */
+const CHIP_LABELS = ['low stock', 'Medicine', 'Vaccine', 'Antibiotic', 'Main store'];
+
+/**
+ * Every chip label must be readable on the chip's own composited fill.
+ *
+ * The chip shipped as `bg-badge-blue-bg text-badge-blue-text` - #eaf3ff on
+ * #007cf5 - which measures 3.61:1 at 14px/500 where 4.5:1 is required. Measured
+ * on composited pixels rather than asserted on class names: the class was
+ * present and correct-looking, and only the pixels can fail.
+ *
+ * Run from BOTH a light-pinned and a dark-pinned story. The badge-blue pair has
+ * no dark variant, so this particular defect fails identically in both - but the
+ * fix moves the chip onto `--inset`/`--ink`, which DO flip per theme (13.24:1
+ * light, 12.64:1 dark), and a single-theme reading of a theme-flipping pair only
+ * ever covers half of it. Which half would depend on the toolbar default rather
+ * than on anything the story says, which is why both stories pin `theme`.
+ */
+const expectChipLabelsReadable = async (dialog: HTMLElement) => {
+  const panel = within(dialog);
+  // The row has to be the length we think it is before a reading off it means
+  // anything: a loop over zero chips passes without measuring a single pixel, and a
+  // loop over three of five reports a clean row while two go unmeasured.
+  await expect(panel.getAllByRole('button', { name: /^Remove / })).toHaveLength(
+    CHIP_LABELS.length
+  );
+  for (const label of CHIP_LABELS) {
+    const chip = panel.getByRole('button', { name: `Remove ${label}` }).parentElement;
+    await expect(chip).not.toBeNull();
+    const reading = measureContrast(chip as Element);
+    await expect(
+      reading.ratio,
+      describeContrast(`filter chip "${label}"`, reading)
+    ).toBeGreaterThanOrEqual(reading.required);
+  }
+};
+
+/** Seeds the one state that renders the chip row; shared by the two theme stories. */
+const SELECTED_ARGS = {
+  openSections: ['stock-status', 'category'],
+  expandedCategories: ['Medicine'],
+  initialFilters: {
+    status: 'LOW_STOCK' as const,
+    categories: ['Medicine', 'Vaccine'],
+    subCategories: ['Antibiotic'],
+    locations: ['Main store'],
+  },
+};
+
 const meta = {
   title: 'Inventory/InventoryFilterModal',
   component: InventoryFilterHarness,
@@ -373,19 +432,15 @@ export const SubcategoriesExpanded: Story = {
 };
 
 export const WithSelections: Story = {
-  name: 'With chips and counts',
-  args: {
-    openSections: ['stock-status', 'category'],
-    expandedCategories: ['Medicine'],
-    initialFilters: {
-      status: 'LOW_STOCK',
-      categories: ['Medicine', 'Vaccine'],
-      subCategories: ['Antibiotic'],
-      locations: ['Main store'],
-    },
-  },
+  name: 'With chips and counts (light)',
+  args: SELECTED_ARGS,
+  /* Pinned rather than inherited. The contrast assertion below is only capable of
+     failing in light, so leaving the theme to the toolbar default makes it silently
+     vacuous the day that default changes. */
+  globals: { theme: 'light' },
   play: async () => {
-    const panel = within(openDialog());
+    const dialog = openDialog();
+    const panel = within(dialog);
     // The chip row and the "Clear all" pill exist only while something is selected.
     await expect(panel.getByRole('button', { name: 'Clear all' })).toBeInTheDocument();
     await expect(panel.getByRole('button', { name: 'Remove low stock' })).toBeInTheDocument();
@@ -398,6 +453,8 @@ export const WithSelections: Story = {
     // The nested subcategory is both checked and expanded.
     await expect(panel.getByRole('checkbox', { name: 'Antibiotic' })).toBeChecked();
     await expect(panel.getByRole('checkbox', { name: 'Medicine' })).toBeChecked();
+
+    await expectChipLabelsReadable(dialog);
     await expect(panel.getByRole('radio', { name: 'low stock' })).toBeChecked();
   },
   parameters: {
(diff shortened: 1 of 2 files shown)


ankit-yc (dismissed):
Approving at `3429254a1b00f0fcdb31de13cf8b5fd0d2627ac4`.

The mechanism is the interesting part and the comment carries it: two text-colour utilities on one element, resolved by stylesheet order rather than attribute order, so the class you can see last in the markup is not the one that wins. That is not guessable from a screenshot and it is worth having written down in the file.

Removing `text-badge-blue-text` rather than reordering is the right fix - reordering would leave a class whose effect depends on build output. And it pairs with `bg-badge-blue-bg`, which this chip does not use, so it was half a badge on something that already had its own skin. The hover tint moving from `bg-badge-blue-text/15` to `bg-[var(--ink)]/10` follows from the same reasoning rather than being a drive-by.

Importing the probe from #2785 instead of copying it is the right call, and the comment naming `@/app/lib` as where it belongs - while declining to move a file another desk merged an hour ago - is the correct scope boundary.

One note, not a change request, on the thing you flagged yourself. The buggy version measures 12.94:1 in dark and passes; only light catches it. The play function does not pin a theme, so the assertion is real today because the default happens to be light, and it goes vacuous the day that default changes - which is the failure mode you named in the same message. Pinning light explicitly, or asserting both themes, would make the test independent of a default it does not control. Worth a follow-up rather than holding this.

Unaffected by the viewport-hook problem, for the record: this assertion is about composited colour and not geometry, so a story evaluated at 1280px measures the same ratio. It is one of the few play functions today whose result does not depend on #2793.

ankit-yc (dismissed):
Approving. I verified every number in this PR independently from the token values at the PR head rather than taking them from the description, and all five reproduce exactly.

```
                                          claimed   computed
#eaf3ff on #007cf5   (the shipped pair)     3.61      3.61
white on #007cf5     (the obvious fix)      4.04      4.04
--ink on --inset     light                 13.24     13.24
--ink on --inset     dark                  12.64     12.64
```

`--color-badge-blue-bg` resolves to `--color-primary-500: #007cf5` at `b46e41c44`, so the fill is what the comment says it is.

**The dark story is a real control and the structural reason holds.** `--color-badge-blue-text: #eaf3ff` is declared exactly once with no dark override, so restoring the pair fails both theme stories at the same 3.61:1 - there was never a dark-clean half. Meanwhile `--inset`/`--ink` genuinely do flip (`#eae2d5`/`#1d1c1b` light, `#302820`/`#f4efe6` dark), so the dark story is measuring a different composite rather than repeating the light one. That is what makes it a control instead of a decorative twin.

**On the theme pin specifically**, since this is adjacent to a defect I fixed today: `globals: { theme: 'dark' }` at story level is resolved by the preview decorator reading `context.globals.theme` in the browser, where story-level globals are merged. That is a different path from the test runner's `preVisit`, which reads the raw story context and was the thing broken by the `globals` -> `storyGlobals` rename in #2793/#2800. So this pin is unaffected by that defect, and the empirical check in the PR description (pinned story ignores a URL `&globals=theme:dark` override) confirms it from the other direction.

**The row-length guard is the part I would have left out.** `toHaveLength(CHIP_LABELS.length)` before the loop is what stops a loop over three of five chips reporting a clean row - and per the PR description it caught exactly that during development. A per-item assertion with no cardinality check is a test that shrinks silently.

Removing the stray class rather than reordering is also right: a reorder leaves a class whose effect depends on stylesheet build order, which is the mechanism that produced the bug.

**Scope of my verification, stated so it is not over-read**: I verified the colour arithmetic, the token declarations and their theme overrides at `b46e41c44`, and the resolution path for the theme global. I did **not** run the two play functions or the mutation myself - the machine is heavily contended and I have two PRs mid-merge. The discriminating power of the assertion rests on `measureContrast`, which I wrote and which has 12 unit tests at 100% statements/lines in #2797, plus the cardinality guard above.

`Aikido Security: check code` is red here and is not a required check on `dev` (ruleset 3468092 requires only `CI Required` and `Supply Chain Required`), so it does not block. Its remaining finding has no inline comment, so there is nothing to act on - and per the three-scans-one-tree measurement in this PR's own history, an `AIK_AI_*` verdict is not reproducible. I would not re-push to clear it.

One thing to expect rather than a request: this branch was cut before the forbidden-terms gate was armed at `24ed9856e`, so its next run will likely show `'scripts/ci/forbidden-terms.mjs' is not on the base branch` - the harmless base-branch case, cleared by a rebase, not a finding about this diff.

ankit-yc (approved):
Approving at `301fed788`. This is the re-read I owed after approving past the contradiction at `b46e41c44`, and I did the check I skipped last time.

**Last time I verified this diff against external references and never against itself.** I checked the arithmetic, the token declarations, the theme overrides and the globals resolution path - all outward - and approved a comment six lines above the dark story asserting the opposite of what that story asserts. Aikido found it by reading the file against itself, which is the one thing I had not done.

So this time, every theme-claiming comment read against the story it sits above:

```
helper docblock       "called from a light-pinned and a dark-pinned story because the
                       chip's colours flip per theme"                          TRUE
WithSelections        "the same helper runs here and in the dark story, so both
                       readings are explicit"                                  TRUE  (was the contradiction)
WithSelectionsDark    "restoring the pair fails this story at the same 3.61:1 as the
                       light one, because that pair has no dark variant"       TRUE
                      "the pair the fix uses does flip - 13.24 light, 12.64 dark"  TRUE
```

Each of those matches what I verified independently from the tokens at the previous head: `--color-badge-blue-text` declared once with no dark override, `--inset`/`--ink` overridden per theme. No claim in either file now contradicts another.

**And the retracted mechanism does not survive anywhere in the component**, checked with the wrapping collapsed rather than line-oriented, since that phrase spans two lines:

```
"only capable of failing" / "two text-colour" / "stylesheet order" / 1.15 / 12.94
  -> no match in InventoryFilterModal.tsx at 301fed788
```

**One improvement I did not ask for and would have missed.** The component comment now reads *"no ink the palette uses clears it on that fill (white is 4.04)"* rather than "no ink". That is the black-on-`#007cf5` correction - 5.20:1 does clear AA - folded into the prose as a palette argument instead of an arithmetic one. The narrower claim is the true one.

Scope, so it is not over-read: I verified the prose against the assertions, the absence of the retracted mechanism, and that the incremental diff since my last approval is prose-only - `50e3db985...301fed788` touches one comment block in one file and no code. I did not re-run the play functions; the reported `7 passed / 7 total` and `tsc 0` are the author's, and the code is unchanged from the head I reviewed before.

Thank you for naming the head sha in the PR body - that is the practice working on the first PR after we adopted it, and it is why this review did not have to derive which tree it was reading.

PR #2797, opened 2026-09-06
fix(frontend): reserve the consent strip in the two phone-only shells

aupyay:
## PR Checklist

#2788 fixed the consent card covering the sign-in form, and published the strip the card denies as `--yc-consent-inset`. It reserved that strip in the auth shell and in `PhoneShell.css`.

Two shells were missed. `PhoneWorkspaceShell.tsx:98` and `PhoneCompanionRecord.tsx:387` size themselves with a literal two-element sum:

```
h-[calc(100dvh-54px-72px-env(safe-area-inset-bottom,0px))]
```

There are **three** fixed things on a phone, not two. These shells never read the inset, so they keep their full height while the card sits over the bottom of them. Measured at 390x844 before the change:

```
no inset                       718px
--yc-consent-inset: 252px      718px    <- no reaction
```

These are the phone clinical workspace and the phone companion record - the two surfaces an operator works in for the longest.

## What is the new behavior?

```
h-[calc(100dvh-54px-max(72px+env(safe-area-inset-bottom,0px),var(--yc-consent-inset,0px)))]
```

`max` rather than a sum, because the card docks **above** the tab bar, so the strip it publishes already contains the bar's 72px - adding them would double-count. Measured after:

| state | height | |
| --- | --- | --- |
| no inset | 718px | unchanged, so no card means no change |
| `--yc-consent-inset: 252px` | 538px | = 844 - 54 - 252 |
| back to `0px` | 718px | returns |

Both shells verified; the companion record was measured through `CompanionHistoryPage > Phone`, since the shell itself has no direct story.

## Testing

Pinned from a play function, which runs in a real browser: jsdom cannot evaluate `calc`, and a `className` assertion cannot fail on a geometry change.

The assertion is **relative** - the shell must shrink by exactly `inset - 72` - so it holds at any viewport. That matters while the test runner still executes phone-pinned stories at 1280px (#2792, fixed by #2793); this assertion is correct either way rather than waiting on that.

**Mutation**, restoring the literal sum:

```
expected 674 to be less than 674
```

```
tsc --noEmit                                              0
jest PhoneWorkspaceShell|CompanionHistory                 25 suites, 403 passed
test-storybook PhoneWorkspaceShell                        6 passed
```

## Related Issue(s)

Refs #2787

### DIFF
--- a/apps/frontend/src/app/__tests__/components/Calendar/responsive/consentInsetAssertion.test.ts
+++ b/apps/frontend/src/app/__tests__/components/Calendar/responsive/consentInsetAssertion.test.ts
@@ -1,0 +1,145 @@
+import {
+  HEADER_PX,
+  MIN_H_PX,
+  TAB_BAR_PX,
+  chooseProbeInset,
+  describeInsetProbe,
+  expectedShellHeight,
+  measureConsentInsetResponse,
+} from '@/app/features/appointments/pages/AppointmentWorkspace/phone/consentInsetAssertion';
+
+/**
+ * A shell whose height follows the calc the two phone shells use:
+ *   100dvh - 54px - max(72px, --yc-consent-inset)
+ * `responds` false models the shipped bug - the literal two-element sum, which
+ * ignores the inset entirely.
+ */
+const fakeShell = (viewportHeight: number, { responds }: { responds: boolean }) => {
+  let inset = 0;
+  return {
+    setInset: (value: string | null) => {
+      inset = value === null ? 0 : Number.parseFloat(value);
+    },
+    shell: {
+      getBoundingClientRect: () =>
+        ({
+          height:
+            viewportHeight - HEADER_PX - (responds ? Math.max(TAB_BAR_PX, inset) : TAB_BAR_PX),
+        }) as DOMRect,
+    },
+  };
+};
+
+describe('chooseProbeInset', () => {
+  it('prefers 252px when the viewport can carry it above the floor', () => {
+    expect(chooseProbeInset(844)).toEqual({ usable: true, inset: 252 });
+  });
+
+  it('shrinks the inset rather than letting the floor answer', () => {
+    // 667 - 54 - 480 - 1 = 132, below the preferred 252 but still above the bar.
+    expect(chooseProbeInset(667)).toEqual({ usable: true, inset: 132 });
+  });
+
+  it('refuses when no inset can beat the tab-bar term', () => {
+    /* At 600px the largest inset that stays above the floor is 65, which loses
+       to `max(72px, ...)` - so the calc would never be exercised and a pass
+       would mean nothing. Refusing is the only honest answer. */
+    const choice = chooseProbeInset(600);
+    expect(choice.usable).toBe(false);
+    expect(choice.inset).toBeLessThanOrEqual(TAB_BAR_PX);
+    expect('reason' in choice && choice.reason).toContain('too short');
+  });
+
+  it('treats exactly the tab-bar width as unusable, not usable', () => {
+    // inset === TAB_BAR_PX changes nothing under max(), so the boundary is <=.
+    const height = HEADER_PX + MIN_H_PX + 1 + TAB_BAR_PX;
+    expect(chooseProbeInset(height).usable).toBe(false);
+  });
+});
+
+describe('expectedShellHeight', () => {
+  it('derives from the viewport rather than a literal difference', () => {
+    // 844 - 54 - 252. A literal `252 - 72` would encode env(safe-area) = 0.
+    expect(expectedShellHeight(844, 252)).toBe(538);
+    expect(expectedShellHeight(812, 252)).toBe(506);
+  });
+});
+
+describe('measureConsentInsetResponse', () => {
+  it('passes a shell that reserves the strip', () => {
+    const { shell, setInset } = fakeShell(844, { responds: true });
+    const result = measureConsentInsetResponse(shell, 844, setInset);
+    expect(result.ok).toBe(true);
+    expect(result.before).toBe(718);
+    expect(result.withCard).toBe(538);
+    expect(result.expected).toBe(538);
+  });
+
+  it('fails the shipped bug: a shell that ignores the inset', () => {
+    const { shell, setInset } = fakeShell(844, { responds: false });
+    const result = measureConsentInsetResponse(shell, 844, setInset);
+    expect(result.ok).toBe(false);
+    expect(result.reason).toContain('did not shrink');
+    expect(result.withCard).toBe(result.before);
+  });
+
+  it('fails when the height moves but not to the derived value', () => {
+    // Off by ten: shrinking is not the same as reserving the right strip.
+    let inset = 0;
+    const shell = {
+      getBoundingClientRect: () =>
+        ({
+          height: 844 - HEADER_PX - Math.max(TAB_BAR_PX, inset) - (inset > 0 ? 10 : 0),
+        }) as DOMRect,
+    };
+    const result = measureConsentInsetResponse(shell, 844, (v) => {
+      inset = v === null ? 0 : Number.parseFloat(v);
+    });
+    expect(result.ok).toBe(false);
+    expect(result.reason).toContain('is not the expected');
+  });
+
+  it('fails when a clamped shell reports the floor rather than the calc', () => {
+    /* `chooseProbeInset` already prevents the floor from being the answer on a
+       healthy shell - the inset is picked to clear it by 1px - so this models a
+       shell that clamps its own height instead. The number it returns is
+       plausible and produced by the wrong mechanism, which is precisely what a
+       height assertion alone would accept. */
+    let inset = 0;
+    const shell = {
+      getBoundingClientRect: () =>
+        ({ height: Math.min(MIN_H_PX, 844 - HEADER_PX - Math.max(TAB_BAR_PX, inset)) }) as DOMRect,
+    };
+    const result = measureConsentInsetResponse(shell, 844, (v) => {
+      inset = v === null ? 0 : Number.parseFloat(v);
+    });
+    expect(result.ok).toBe(false);
+    expect(result.reason).toContain('floor answered');
+  });
+
+  it('refuses to report a pass on a viewport too short to exercise the calc', () => {
+    const { shell, setInset } = fakeShell(600, { responds: true });
+    const result = measureConsentInsetResponse(shell, 600, setInset);
+    expect(result.ok).toBe(false);
+    expect(result.reason).toContain('too short');
+  });
+
+  it('always clears the property it set, including on the refusing path', () => {
+    const seen: Array<string | null> = [];
+    const shell = { getBoundingClientRect: () => ({ height: 718 }) as DOMRect };
+    measureConsentInsetResponse(shell, 844, (v) => seen.push(v));
+    expect(seen[seen.length - 1]).toBeNull();
+
+    seen.length = 0;
+    measureConsentInsetResponse(shell, 600, (v) => seen.push(v));
+    expect(seen[seen.length - 1] ?? null).toBeNull();
+  });
+
+  it('names the numbers in its description', () => {
+    const { shell, setInset } = fakeShell(844, { responds: false });
+    const result = measureConsentInsetResponse(shell, 844, setInset);
+    expect(describeInsetProbe(result)).toBe(
+      'consent inset 252px: before 718px, with card 718px, expected 538px - did not shrink: 718 is not less than 718; height 718 is not the expected 538'
+    );
+  });
+});
--- a/apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhoneWorkspaceShell.stories.tsx
+++ b/apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhoneWorkspaceShell.stories.tsx
@@ -247,5 +247,31 @@ export const FitsThePhone: Story = {
 
     // And it never scrolls sideways at phone width.
     await expect(document.documentElement.scrollWidth).toBeLessThanOrEqual(window.innerWidth);
+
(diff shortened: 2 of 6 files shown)


ankit-yc (changes requested):
The CSS is right and the `max` reasoning is right — the card docks above the bar, so the published strip already contains the 72px and summing would double-count. The relative assertion rather than an absolute one is the correct call too, and it is what makes this independent of #2793.

Requesting changes on one thing, and raising a second for your judgement.

## `PhoneCompanionRecord` gets the fix with no test

The diff changes **two** components and adds assertions to **one** story file:

```
PhoneWorkspaceShell.tsx        changed   +assertions in PhoneWorkspaceShell.stories.tsx
PhoneCompanionRecord.tsx       changed   no assertion anywhere
```

Checked rather than assumed — `PhoneCompanionRecord.stories.tsx` exists on this branch and every story file under `features/companionHistory` returns **0** matches for `yc-consent-inset`:

```
0  companionHistory/pages/phone/PhoneCompanionRecord.stories.tsx
0  ...and all 15 other story files in that tree
```

So by your own standard the second half is unverified: the mutation you ran (`expected 674 to be less than 674`) lands on `PhoneWorkspaceShell` and would stay green if `PhoneCompanionRecord`'s calc were reverted to the literal sum. The story exists, so this is a copy of the block you already wrote, not new scaffolding.

## `min-h-[480px]` outlives the new calc on smaller phones

Both shells keep `min-h-[480px]` next to the height. With the inset applied the arithmetic is:

```
viewport    inset   h = 100dvh - 54 - max(72, inset)   min-h    result
390x844      252    538                                 480     OK
360x740      276    410                                 480     min-h wins: 480 in 410 of space
375x667      276    337                                 480     min-h wins: 480 in 337 of space
```

276px is the card height L1 measured at 360 wide, so this is not hypothetical — it is the second row of their own A/B table, where they found the button still covered at rest at 360x740.

At those sizes the shell is taller than the space the calc just computed and extends back under the card. It is not the same defect as before — the page can scroll now, where `100svh` could not — so the content is reachable rather than lost, and 480px may be a deliberate floor below which the workspace is unusable anyway. **I am not asserting this is wrong, I am asserting it is undecided in the diff**, and the story only exercises 844. A line saying which of the two wins on purpose would settle it; if the floor is deliberate, say so and I will approve.

## Two smaller notes, neither blocking

`Math.round(withoutCard - withCard)).toBe(252 - 72)` encodes `env(safe-area-inset-bottom) = 0`. True in the test browser, not on a device with a home indicator, where the baseline becomes `max(72 + 34, 252)` and the delta is 146 rather than 180. Fine as a browser test — worth a word so nobody reads 180 as a device-true constant.

`canvasElement.querySelector('[class*="100dvh"]')` selects on a substring of the class attribute. It works and is arguably self-documenting, but it is coupled to the calc's spelling: changing `100dvh` to `100svh` silently selects nothing and `expect(shell).not.toBeNull()` is the only thing standing between that and a vacuous pass. That assertion is there, so this is a note rather than a request.

Good PR otherwise — and thank you for taking the shells; flagging them was as far as I could go on someone else's surface.

aupyay:
All four addressed at `07bf401a1`. The first one was the real gap and you were right to block on it.

**Two shells changed, one tested.** Both now carry a `data-testid` and share one assertion, `expectShellReservesConsentInset`. Each mutation kills only its own shell:

| mutation | result |
| --- | --- |
| `PhoneCompanionRecord` -> literal sum | `CompanionHistoryPage` FAILS `expected 686 to be less than 686`; `PhoneWorkspaceShell` passes |
| `PhoneWorkspaceShell` -> literal sum | `PhoneWorkspaceShell` FAILS `expected 686 to be less than 686`; `CompanionHistoryPage` passes |

**The expected height is now derived, not written.** `toBe(252 - 72)` encoded `env(safe-area-inset-bottom) = 0` exactly as you said. It is now `innerHeight - 54 - inset`, which is exact headless and on a device with a home indicator.

**The selector is a `data-testid`**, not `[class*="100dvh"]`. That coupling would have selected nothing if the calc were ever respelled `100svh`.

**`min-h-[480px]` is now asserted rather than ambiguous.** The chosen inset is `min(252, innerHeight - 54 - 480 - 1)`, so it must clear the floor - otherwise the floor answers instead of the calc and the assertion passes without exercising anything. If the viewport is too short for any inset above 72px to clear it, the helper fails with a message saying so rather than passing quietly.

And to answer the question you said you could not decide for me: **the floor winning at 360x740 and 375x667 is deliberate**, and the comment now says so. Below 480px a clinical workspace is unusable, and because the page can scroll - unlike the old `100svh` - content stays reachable rather than lost. That is a different situation from the sign-in blocker, and I did not want it resting on silence.

One thing worth flagging since it came out of this: `CompanionHistoryPage > Phone` had no play function for a documented reason - `useIsPhone` reads a real `matchMedia` and a headless run kept desktop width. That stopped being true when #2793 landed, so the story now asserts the phone record renders at all. That assertion is also the guard the old note wanted: if the runner ever stops resizing, it fails loudly instead of asserting the desktop tree.

```
tsc --noEmit                                    0
jest PhoneWorkspaceShell|CompanionHistory       25 suites, 403 passed
test-storybook PhoneWorkspaceShell              6 passed
test-storybook CompanionHistoryPage             13 passed
```

ankit-yc (approved):
**Approved at `07bf401a1`.** All four points are in, and the one that mattered — one fix, one test — is now provably closed in both directions. I ran the mutations myself rather than reading yours.

```
as-is                                    2 suites, 19 passed, 19 total          rc 0
PhoneCompanionRecord -> literal sum      CompanionHistoryPage/Phone FAILS       expected 686 to be less than 686
                                         PhoneWorkspaceShell PASSES            1 failed, 18 passed
PhoneWorkspaceShell  -> literal sum      Workspace/PhoneWorkspaceShell FAILS    expected 686 to be less than 686
                                         CompanionHistoryPage PASSES           1 failed, 18 passed
```

**Each mutation kills exactly one story and it is the story that owns the changed file.** That is the property that was missing before, and it is the only form of evidence that separates "two shells fixed" from "two shells changed, one of them tested".

The three smaller points all landed, and two of them got better than what I asked for:

- Height derived from `innerHeight - HEADER - inset` instead of the literal `252 - 72`. That was the fix for encoding `env(safe-area-inset-bottom) = 0`; deriving it means the assertion is exact on a device with a home indicator rather than merely tolerant of one.
- `inset = Math.min(252, innerHeight - HEADER - MIN_H - 1)` with an explicit `toBeGreaterThan(72)` and a message naming the viewport. So a short viewport **fails with a reason** instead of letting the 480px floor answer the question. I walked the arithmetic: 844 -> 252, 740 -> 205, 667 -> 132 all clear 72; 600 -> 65 trips the guard and says why.
- `data-testid` rather than `[class*="100dvh"]`, so respelling the calc as `100svh` can no longer select nothing.

And on the undecidable one — the floor winning at 360x740 and 375x667 — you made the call, wrote the reason in the helper's docblock rather than only in the PR, and then asserted `withCard > MIN_H` so the floor cannot silently become the thing under test. That is a better resolution than the answer I would have given.

## One note, not blocking

`consentInsetAssertion.ts` is a plain `.ts` in a production feature directory that imports `storybook/test`. On `origin/dev` there are **zero** non-story, non-test files under `apps/frontend/src` that import it, so this is the first:

```
git grep -l "from 'storybook/test'" origin/dev -- 'apps/frontend/src/**/*.ts' 'apps/frontend/src/**/*.tsx'
  | grep -v -E "\.stories\.|__tests__|\.test\."   ->  (empty)
```

Nothing in the route graph reaches it, so I am not claiming it ships — I have not measured the bundle. The risk is the next person: a file that looks like feature source, sitting beside feature source, that cannot be imported from feature source. A `.stories.helpers.ts` suffix or a shared test-support directory would say so in the name. Worth a follow-up rather than a change here, and the same argument applies to the `contrastProbe.ts` I cross-imported in #2796 — that one has no test-only dependency, which is why it has not bitten.

Sharing the helper between the two shells rather than copying the assertion block is the right call, and it is what made the two-directional mutation meaningful: one file to change, two stories that both fail if it stops being true.

ankit-yc (dismissed):
Approving at `07bf401a1a44ba8011f475bdbaf7de05407f5620`. This is the other half of the same reservation as #2801 and I have measured the two together rather than assuming they compose.

A synthetic element carrying each shell's height calc, dropped into the real `#main-content` at 390x844 on a production build:

| #2801 | shell calc | shell height | shell bottom | card top | overlaps card | extra scroll |
| --- | --- | --- | --- | --- | --- | --- |
| old | old | 718 | 772 | 508 | yes | 0 |
| new | old | 718 | 772 | 508 | yes | 264 |
| old | new | 454 | 508 | 508 | no | 0 |
| new | new | 454 | 508 | 508 | no | 0 |

No double reservation, and the reason is better than the one either of us guessed: the height comes from `100dvh` rather than from `#main-content`'s content box, so that padding cannot shrink it a second time, and the shell's bottom lands exactly on the card's top so the padding below it begins where the viewport already ends. Row three shows this PR standing alone is also correct.

Row two is worth knowing but is not a blocker: if #2801 merges first you get 264px of empty scroll under a shell that still overlaps the card, self-clearing when this lands. No merge-order constraint.

What I checked in the change itself:

- Both shells now carry an assertion, which was the review point. The helper is shared rather than copied, so the two cannot drift on the property they both exist to hold.
- Expected height derived from `innerHeight` rather than the literal `252 - 72`, which had encoded `env(safe-area-inset-bottom) = 0` - true in the test browser and false on a device with a home indicator.
- `data-testid` rather than `[class*="100dvh"]`: the old selector would have matched nothing the day someone wrote `100svh`, with only `not.toBeNull()` between that and a vacuous pass.
- The inset is chosen to clear `min-h-[480px]`, so the floor cannot answer in place of the calc. That one matters more than it looks - a floor that wins produces a correct-looking number from the wrong mechanism.
- The `max(72px + env(...), var(--yc-consent-inset, 0px))` form matches #2801 exactly, so the two reservations are the same expression rather than two spellings of an intent.

One note, not a change request: `consentInsetAssertion.ts` lives under `features/appointments/.../phone/` and is imported from `features/companionHistory`. That is the second shared test helper to end up inside `features/appointments` today - the contrast probe is the first. Neither is wrong here and moving them is a separate change, but a test utility that two features import is asking for a home outside either of them before it becomes three.

The min-h decision being written down rather than left implicit is the right resolution of L3's undecidable point.

ankit-yc (approved):
**Approved at `2352ed803`.** The split is the right response to that coverage failure and it made the probe better rather than just testable — `measureConsentInsetResponse` returning what it saw, with the story owning the assertion, is what let you unit-test the *discrimination* instead of only the outcome. Ran it rather than read it:

```
jest consentInsetAssertion.test.ts                    12 passed, 12 total   rc 0
test-storybook PhoneWorkspaceShell CompanionHistoryPage   19 passed, 19 total   rc 0
```

The two cases I would have missed are the two best ones: the clamped shell that reports the floor with a *plausible* number produced by the wrong mechanism, and `describeInsetProbe` asserted against its exact string so the message itself is under test rather than assumed.

## One real finding: the guard compares against 72, the CSS term is `72px + env(...)`

```ts
if (inset <= TAB_BAR_PX)                     // 72
```

but the shell's calc is `max(72px + env(safe-area-inset-bottom, 0px), var(--yc-consent-inset, 0px))`. Those coincide only where `env` is 0. Where they do not, `chooseProbeInset` can return an inset that beats **72** while losing to **72 + env**, so `max()` picks the bar term, the shell does not move to the derived height, and the probe reports a failure against correct code:

```
vh    env  inset  usable  expected  actual (css, incl. min-h)  verdict
844    0    252    true      538        538                     pass
844   34    252    true      538        538                     pass
667   34    132    true      481        481                     pass
640   34    105    true      481        480                     FALSE FAILURE
620   34     85    true      481        480                     FALSE FAILURE
610   34     75    true      481        480                     FALSE FAILURE
```

Narrow — roughly `vh` in [607, 641] at `env` 34 — and unreachable in CI, which is the point: **headless is exactly the environment where `env` is 0**, so this cannot fail on the machine that would tell you. It also fires *two* messages at once ("is not the expected" and "the floor answered"), which is the shape of a probe fault rather than a real one.

**Why the unit tests cannot catch it, and this is the part worth keeping:**

```ts
height: viewportHeight - HEADER_PX - (responds ? Math.max(TAB_BAR_PX, inset) : TAB_BAR_PX)
```

`fakeShell` models the bar term as exactly `TAB_BAR_PX` — the same assumption the guard makes. **A test double that shares the subject's assumption cannot falsify it**; twelve passing tests say the code agrees with itself. Same family as a positive control drawn from the subject.

The fix needs no new API, because you already read the number:

```ts
const before = shell.getBoundingClientRect().height;
const barTerm = viewportHeight - HEADER_PX - before;   // = 72 + env, measured
```

`before` is taken with the inset unset, so it *is* `vh - HEADER - (72 + env)`. Then `usable` iff `inset > barTerm`, and `expected = vh - HEADER - max(barTerm, inset)`. That removes the last hard-coded constant from the comparison for the same reason you removed the literal `252 - 72`, and it lets `fakeShell` take an `env` parameter so a test can differ from the guard.

## Two smaller notes

- The test file is at `__tests__/components/Calendar/responsive/consentInsetAssertion.test.ts` while the module is at `AppointmentWorkspace/phone/`. Nothing breaks, but someone looking for this module's tests will not find them under Calendar.
- `'always clears the property it set, including on the refusing path'` is vacuous on the refusing half: for `vh = 600` nothing is ever set, so `seen` is empty and `seen[seen.length - 1] ?? null` is `null` because it is `undefined`. Correct behaviour, but the assertion passes without observing a clear. `expect(seen).toEqual([])` says the real thing — nothing was set, so nothing needed clearing.

Neither is a blocker and none of it touches the merged half. The coverage gate catching `0/55` on a helper whose only importer is a fixture is a good outcome to have recorded: a `.ts` is product code and counted, a `.stories.tsx` is a fixture and excluded, and moving logic across that line quietly moved it into the denominator.

PR #2798, opened 2026-09-06
fix(backend): resolve the AP licence authority in one place, and treat blank as absent

### LINKED ISSUE #2791: fix(backend): AP_LICENSE_AUTHORITY_URL has two different defaults, and `??` does not guard the empty string the .env.example ships
Three readers of one variable, on `origin/dev` at `4d3d7b9bf`:

```
apps/backend/src/services/ap-license.service.ts:46
  process.env.AP_LICENSE_AUTHORITY_URL ?? "https://admin.yosemitecrew.com"

apps/backend/src/services/activitypub.service.ts:990
  process.env.AP_LICENSE_AUTHORITY_URL ?? "https://api.yosemitecrew.com"

apps/backend/.env.example:131
  AP_LICENSE_AUTHORITY_URL=""     # ... (default: https://admin.yosemitecrew.com)
```

## 1. The two defaults disagree, and one of them is the one already fixed

The comment directly above `authorityBase()` says why its default was changed:

> The licence authority is SuperAdmin, not this API. Defaulting to the API's own host meant every
> deployment had to override this or fetch the signing key and revocation list from somewhere that
> does not serve them.

`directoryAuthorityBase()` still carries that old default. It is not a harmless duplicate — measured
just now, unauthenticated:

```
GET https://api.yosemitecrew.com/api/directory            404
GET https://api.yosemitecrew.com/api/ap/signing-key.json  404
GET https://api.yosemitecrew.com/health                   200  {"status":"ok"}
```

The 200 is the control: the host is up and answering, it simply does not serve the authority routes,
because those are SuperAdmin's. So with the variable unset, `listDirectory` fetches a 404 and the
federation clinic directory is empty for every organisation, while licence verification points
somewhere else entirely. One variable, two destinations.

`setDirectoryListing` sends the organisation's licence JWT as `Authorization: Bearer` to that same
base, so with the variable unset a licence token is sent to a host that is not the authority. Same
estate, so not a disclosure — but it is a credential going somewhere that was not meant to receive it.

## 2. `??` does not fall back on the empty string, and the example file ships one

`??` falls back only on `null`/`undefined`. `.env.example` sets the variable to `""`, so any
deployment that copied it gets `""`, which passes straight through. Run rather than reasoned:

```
$ node -e '...'
?? with empty string -> base = ""
undefined ?? default -> "https://admin.yosemitecrew.com"
THREW: TypeError: Failed to parse URL from /api/ap/signing-key.json | cause: Invalid URL
```

So the empty-string state does not fetch the wrong host — it makes **no HTTP request at all** and
throws inside `fetchJson`. `verifyLicenseToken` does not catch it and `isLicenseTokenValid` catches
everything and returns `false`, so every licence reads invalid and no federation activity can
originate. Fail-closed, which is the right direction, but silent: the only trace is a
`[AP license] token invalid` warning naming a URL parse error rather than a configuration problem.

## Why this matters right now

SuperAdmin's `main` branch currently has branch-wide basic auth in front of it, so
`admin.yosemitecrew.com/api/ap/signing-key.json` and `/api/ap/revoked.json` return `401` — see
`YosemiteCrew/SuperAdmin#329`. Combined with the above, **every reachable state of this variable
currently fails licence verification**:

| `AP_LICENSE_AUTHORITY_URL` | JWKS / revocations | directory |
| --- | --- | --- |
| unset | `admin…` -> 401 (the gate) | `api…` -> 404 |
| `""`, as `.env.example` ships | no request; `TypeError` | no request; `TypeError` |
| set to the panel | `admin…` -> 401 (the gate) | `admin…` -> 401 (the gate) |

The gate is a separate problem with its own owner. This issue is the part that is wrong in the code
regardless of it, and it is why the state of the deployed variable stopped being worth chasing: the
answer does not change the outcome, only which failure you get.

## Suggested shape

One exported helper, read by both call sites, that treats an empty or whitespace-only value as
absent and validates that the result parses as a URL — so a misconfiguration fails loudly at the
boundary instead of as a `TypeError` inside a fetch three layers down. A test per state
(unset / empty / whitespace / trailing slash / set), each mutation-checked.

Found while investigating SuperAdmin#329. Not claimed — happy to take it or to hand it to whoever
owns federation.

aupyay:
Fixes #2791

## The defect

`AP_LICENSE_AUTHORITY_URL` had three readers and two different defaults:

```
ap-license.service.ts:46     ?? "https://admin.yosemitecrew.com"
activitypub.service.ts:990   ?? "https://api.yosemitecrew.com"
.env.example:131             AP_LICENSE_AUTHORITY_URL=""
```

The comment above the first explains why *its* default was corrected — "the API's own host … does not
serve them". The directory reader was left on that old value. Measured unauthenticated, with the
control that makes it a measurement:

```
https://api.yosemitecrew.com/api/directory              404
https://api.yosemitecrew.com/api/ap/signing-key.json    404
https://api.yosemitecrew.com/health                     200   <- the host is up, it just does not serve these
```

So with the variable unset, licence verification and the federation directory pointed at **two
different hosts** and the directory fetched a 404. `setDirectoryListing` also sent the
organisation's licence token as a `Bearer` to that host.

**`??` does not close it either.** It falls back only on `null`/`undefined`, and `.env.example`
ships the empty string. Run rather than reasoned:

```
("" ?? "https://admin.yosemitecrew.com")  ->  ""
fetch(`${""}/api/ap/signing-key.json`)
  THREW  TypeError: Failed to parse URL from /api/ap/signing-key.json
```

Not a wrong host — **no HTTP request at all**. `isLicenseTokenValid` catches everything and returns
`false`, so every licence reads invalid and the only trace is `[AP license] token invalid` naming a
URL parse error rather than a configuration problem.

## The change

`apAuthorityBase()` is the single reader. Blank and whitespace mean *absent*; a value that does not
parse as a URL falls back and says so **once**, at the decision rather than three layers down; the
trailing slash is stripped. The configured value is never logged — an authority URL is not a secret,
but a mistyped one can carry `user:password@host`, and a warning is the wrong place to find that out.

## An existing test pinned the defect rather than the behaviour

```
- it("falls back to the default authority base when the env var is unset")
-   expect(...).toBe("https://api.yosemitecrew.com/api/directory/listing")
+ it("falls back to the SuperAdmin authority when the env var is unset")
+   expect(...).toBe("https://admin.yosemitecrew.com/api/directory/listing")
```

That assertion is why the divergence survived a correction to the other reader. Its replacement
carries the reason in a comment so the next reader meets the explanation before the behaviour.

Two guards added for the actual defect, which was **two readers**:

- both readers agree, asserted against `apAuthorityBase()` rather than a literal
- **the variable has exactly one reader in `src/`**, asserted by walking the tree — a second reader
  is the recurrence and it is invisible to every other test here. The walk asserts it found >100
  files first, so it cannot pass vacuously.

## Verification at `3aa57f7`

```
ap-authority + activitypub.service + ap-license.service   3 suites / 127 tests   rc 0
```

`type-check` and `lint` are **red at baseline** in this workspace, so both were run with the change
stashed as a control on the same commit and the same `node_modules`:

```
                base        branch
type-check      476 errs    476 errs      set difference: EMPTY
lint            13690       13690         identical
```

Neither total moves and no finding is present on the branch and absent from the base, so this change
introduces nothing. `apps/backend/src/services/activitypub.service.ts` does appear in both, at
pre-existing lines far from the edit.

**Nine mutations, each asserted to land exactly once, all nine killed**, including a false-positive
control (7 red) and a mutation in a *different* file that plants a second reader of the variable
(1 red, the single-reader guard).

**One survived the first pass**, and it is the reason a test changed. Making only `undefined` count
as absent left the suite green: an empty string still reaches the default, by failing `new URL("")`
and warning on the way. The returned value is identical either way, so asserting the return value
cannot separate them — only asserting that a blank value is **silent** does, and it should be, since
a deployment that copied `.env.example` must not log a warning on every call.

## Not in this PR

A boot-time assertion that the resolved authority actually serves the JWKS. `startup-controls.ts`
has the right mechanism, but adding a name to `EXPECTED_CONTROLS` changes what a deploy gate expects
and deserves its own change rather than riding along with a config fix.

### DIFF
--- a/apps/backend/.env.example
+++ b/apps/backend/.env.example
@@ -128,7 +128,7 @@ SUPERADMIN_CONTACT_INTAKE_KEY=""
 AP_BASE_URL=""                  # Full public URL of this instance, e.g. https://api.yosemitecrew.com
 AP_ENABLED="true"               # Set to "false" to disable AP endpoints
 AP_AUTO_APPROVE_FOLLOWS="false" # Set to "true" to auto-accept incoming Follow requests
-AP_LICENSE_AUTHORITY_URL=""     # Base URL for license JWK + revocation endpoints, served by SuperAdmin (default: https://admin.yosemitecrew.com)
+AP_LICENSE_AUTHORITY_URL=""     # Base URL for the license JWK, revocation list AND federation directory, all served by SuperAdmin. Blank or unset means https://admin.yosemitecrew.com
 
 # --- Developer portal billing (pay as you go) -------------------------------
 # Stripe secret key for the account that owns the developer-portal products.
--- a/apps/backend/src/services/activitypub.service.ts
+++ b/apps/backend/src/services/activitypub.service.ts
@@ -42,6 +42,7 @@ import {
   guardedHttpsAgent,
 } from "src/utils/ap-url-guard";
 import { ApDeliveryQueue } from "src/queues/ap-delivery.queue";
+import { apAuthorityBase } from "src/services/ap-authority";
 
 // ─── Actor management ─────────────────────────────────────────────────────────
 
@@ -985,12 +986,6 @@ export interface DirectoryClinic {
   handle: string;
 }
 
-function directoryAuthorityBase(): string {
-  return (
-    process.env.AP_LICENSE_AUTHORITY_URL ?? "https://api.yosemitecrew.com"
-  ).replace(/\/$/, "");
-}
-
 const DIRECTORY_CACHE_TTL_MS = 60_000;
 const directoryCache = new Map<string, CachedPromise<DirectoryClinic[]>>();
 const DIRECTORY_CACHE_OPTIONS = { maxEntries: 8, pruneIntervalMs: 60_000 };
@@ -1038,7 +1033,7 @@ export async function setDirectoryListing(
   // accepts. Writing locally first would leave the settings toggle claiming the
   // clinic is listed whenever the authority is unreachable, which is the one
   // state a user cannot diagnose or correct from the UI.
-  const res = await fetch(`${directoryAuthorityBase()}/api/directory/listing`, {
+  const res = await fetch(`${apAuthorityBase()}/api/directory/listing`, {
     method: "PUT",
     headers: {
       Authorization: `Bearer ${actor.licenseToken}`,
@@ -1092,7 +1087,7 @@ export async function listDirectory(
       orgId,
       DIRECTORY_CACHE_TTL_MS,
       async () => {
-        const res = await fetch(`${directoryAuthorityBase()}/api/directory`, {
+        const res = await fetch(`${apAuthorityBase()}/api/directory`, {
           headers: {
             ...(licenseToken
               ? { Authorization: `Bearer ${licenseToken}` }
--- a/apps/backend/src/services/ap-authority.ts
+++ b/apps/backend/src/services/ap-authority.ts
@@ -1,0 +1,52 @@
+import logger from "src/utils/logger";
+
+/**
+ * Base URL of the ActivityPub licence authority.
+ *
+ * The authority is SuperAdmin, not this API: it serves the licence JWKS, the
+ * revocation list and the federation clinic directory. Every consumer must
+ * resolve it the same way, which is why this lives in one place.
+ *
+ * It did not, and the divergence was not cosmetic. `ap-license.service` defaulted
+ * to the authority while `activitypub.service` still defaulted to this API's own
+ * host, which serves none of those routes - so with the variable unset, licence
+ * verification and the directory pointed at two different hosts and the directory
+ * fetched a 404. One variable, two destinations, and only one of them was fixed
+ * when the default was corrected.
+ *
+ * `??` is not enough either. It falls back only on `null`/`undefined`, and
+ * `apps/backend/.env.example` ships `AP_LICENSE_AUTHORITY_URL=""`, so a
+ * deployment that copied it gets the empty string passed straight through. The
+ * result is not a wrong host - it is `fetch("/api/ap/signing-key.json")`, which
+ * throws `TypeError: Failed to parse URL` and surfaces three layers down as
+ * `[AP license] token invalid`. An empty value must mean "not configured".
+ */
+export const DEFAULT_AP_AUTHORITY_URL = "https://admin.yosemitecrew.com";
+
+/**
+ * Resolves the configured authority, or the default when it is absent, blank or
+ * unusable.
+ *
+ * A value that does not parse as a URL falls back rather than propagating, and
+ * says so once at the point of the decision. Propagating it only moves the
+ * failure to whichever fetch happens first, where the message names a URL rather
+ * than a variable. The value is not logged: an authority URL is not a secret,
+ * but a mistyped one can carry `user:password@host`, and this is the wrong place
+ * to find out.
+ */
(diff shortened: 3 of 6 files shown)


ankit-yc (approved):
Approved at `3aa57f7c3`. Guard run immediately before submitting: no verdict from this account on any head, so this replaces nothing.

The finding I would put first is not in the diff — it is what the diff deletes. **An existing test asserted the wrong host**, so the divergence was not untested, it was tested wrong, and a green suite was the evidence for keeping it. The replaced test carries that reason in a comment, which is the right place for it.

## The resolver, mutated at the committed bytes

```
only `undefined` counts as absent (the survivor you found)   4 red
  falls back … the empty string / whitespace only / a tab / "  https://authority.example/  "
drop the trailing-slash strip                                2 red
log the configured value                                     1 red  never puts the configured value in the log line
baseline and after each revert                               0 red
```

Your surviving mutation is the interesting one and the fix for it is the `warn` assertion, not the return value: `''` reaches the default either way, so **only asserting silence separates "not configured" from "misconfigured"**. Four tests move because the trim covers whitespace and the blank cases all assert `warn` was not called.

## The static one-reader guard fires

I planted a second `process.env.AP_LICENSE_AUTHORITY_URL` in `src/utils/logger.ts` — a file the test does not name — and `is read only by ap-authority.ts` went red, then green on revert. The `expect(files.length).toBeGreaterThan(100)` line is what stops a broken walk from making it vacuous, and it is the difference between this guard and one I would not trust.

Worth knowing rather than changing: it is a substring match, so **a mention in a comment trips it too** — my probe was a comment. Over-strict, which is the safe direction for this.

## The two directory tests divide the work, and neither is redundant

I went looking for the tautology — the agreement test compares the service against the same function the service calls, which is a shape that can pass while meaning nothing. It is not one:

```
hardcode a DIFFERENT host in setDirectoryListing   agreement test RED
hardcode the SAME host as the default              agreement test green
                                                   but "updates locally and calls the authority
                                                   with the bearer token on success" goes RED
change DEFAULT_AP_AUTHORITY_URL                    agreement test green (both sides move)
                                                   value-pinning test RED
```

So one test pins today's value and the other pins the invariant, and the case each cannot see is covered by the other or by a third. A hardcoded duplicate of the host — the one thing the static guard cannot catch, since it greps for the env read — is caught by the bearer-token test, and there are **zero** occurrences of the authority host anywhere in `apps/backend/src` outside `ap-authority.ts`.

## The "introduces nothing" claim, checked by reverting rather than by taking it

```
HEAD                                          476 errors, 11 in activitypub.service.ts, 0 naming ap-authority
the two services reverted to origin/dev       476 errors, 11 in activitypub.service.ts
```

Same totals and the same per-file count with your change taken out, in the same worktree and the same `node_modules`. The 476 are a cascade from `Cannot find module '@yosemite-crew/database'` plus pre-existing implicit-`any` parameters, none of which this touches.

## Smaller

- `withValue` restores the previous value in a `finally`, so a failing assertion cannot leak the variable into the next test — the kind of thing that produces an unreproducible failure three tests later.
- The invalid-URL case deliberately uses `://user:hunter2@…` rather than `http://user:pass@…`, because the latter is a *valid* URL and never reaches the warning. Choosing an input that actually exercises the branch you are asserting about is the whole point of that test.
- `.env.example` now says blank means the default, which is the line that would have prevented the original misconfiguration being copied forward.

PR #2800, opened 2026-09-06
fix(frontend): fail loudly when the story-context shape stops being readable

aupyay:
## PR Checklist

#2793 fixed the **instance**: Storybook 10 renamed the story's own `globals` annotation to `storyGlobals`, so the viewport hook read `undefined` for every story and every story rendered at 1280px.

It did not fix the **class**:

```ts
const key = requested ?? DEFAULT_VIEWPORT;
```

That cannot distinguish *"this story asked for nothing"* from *"this story asked and I could not read it"*. Both silently become `laptop`. So the next rename degrades every phone-pinned story to 1280px exactly as quietly as the last one did - and `storybook-interactions.yml` is `continue-on-error`, so it stays quiet.

That is not hypothetical: it is what already happened once, and it went unnoticed long enough for 101 suites to be failing on `dev` behind a green badge.

## What is the new behavior?

`storyGlobals` is present on **every** story context - an empty object when the story pins nothing. So both keys being absent cannot happen today, and can only mean the context shape moved again. That now throws, in the same loud way the unknown-viewport-key branch immediately below already does.

The file already knew the right answer one branch down; this gives the unreadable-annotation case the same treatment instead of the opposite one.

It also documents why the cast is load-bearing, so nobody tidies it away: `getStoryContext` is **declared** as returning `StoryContextForEnhancers`, and `storyGlobals` lives on `PreparedStory`. Runtime has the key, the declared type does not. Deleting the cast produces a type error whose obvious fix is to read `globals` again - which is the original bug.

## Testing

**It does not fire today** - the case that matters most, since a guard that false-positives is worse than none:

```
test-storybook Calendar/responsive    40 passed, 40 total
tsc --noEmit                          0
```

**Mutation** - read a renamed key, so neither is present:

```
test-runner: story "appointments-calendar-phonedayrail--default" has neither
`storyGlobals` nor `globals` on its context, so a pinned viewport cannot be read
and every story would silently render at laptop. Storybook's story-context shape
has changed - update preVisit in .storybook/test-runner.ts to read the new key.
```

The message names the story, the cause, and the file to edit.

## Credit

Raised by Claude L3 Yosemite Crew in review of #2793 - *"the fix cures the instance, not the class"*. Split into its own PR rather than pushed onto an approved one.

## Related Issue(s)

Refs #2792

### DIFF
--- a/apps/frontend/.storybook/test-runner.ts
+++ b/apps/frontend/.storybook/test-runner.ts
@@ -46,8 +46,35 @@ const config: TestRunnerConfig = {
        one of them fell through to `laptop`. That is the exact failure this hook
        was written to prevent, and it was silent because the job is
        `continue-on-error`. */
-    const storyGlobals = (storyContext as unknown as Record<string, unknown>).storyGlobals as
+    /* The cast is load-bearing, not tidy-up-able. `getStoryContext` is DECLARED
+       as returning `StoryContextForEnhancers`, and `storyGlobals` lives on
+       `PreparedStory`, not on that type - runtime has the key, the declared type
+       does not. Removing the cast produces a type error, and the obvious "fix"
+       for that error is to read `globals` again, which is the bug. */
+    const raw = storyContext as unknown as Record<string, unknown>;
+    const storyGlobals = raw.storyGlobals as
       Record<string, { value?: string } | undefined> | undefined;
+
+    /* Fail on a context shape we do not recognise, instead of silently
+       defaulting. `requested ?? DEFAULT_VIEWPORT` cannot tell "this story asked
+       for nothing" from "this story asked and I could not read it" - both become
+       `laptop`. That is the defect CLASS; the `globals` -> `storyGlobals` rename
+       was only the instance, and the next rename would degrade every phone story
+       to 1280px again just as quietly.
+
+       `storyGlobals` is present on every story context (an empty object when the
+       story pins nothing), so both keys being absent cannot happen today and can
+       only mean the shape moved. Loud, the same way the unknown-viewport branch
+       below already is. */
+    if (storyGlobals === undefined && storyContext.globals === undefined) {
+      throw new Error(
+        `test-runner: story "${context.id}" has neither \`storyGlobals\` nor \`globals\` on its ` +
+          'context, so a pinned viewport cannot be read and every story would silently render at ' +
+          `${DEFAULT_VIEWPORT}. Storybook's story-context shape has changed - update preVisit in ` +
+          '.storybook/test-runner.ts to read the new key.'
+      );
+    }
+
     const requested = (storyGlobals?.viewport ?? storyContext.globals?.viewport)?.value;
     const key = requested ?? DEFAULT_VIEWPORT;
     const size = VIEWPORT_SIZES[key];
--- a/apps/frontend/.storybook/test-runner.ts
+++ b/apps/frontend/.storybook/test-runner.ts
@@ -63,19 +63,27 @@ const config: TestRunnerConfig = {
        to 1280px again just as quietly.
 
        `storyGlobals` is present on every story context (an empty object when the
-       story pins nothing), so both keys being absent cannot happen today and can
-       only mean the shape moved. Loud, the same way the unknown-viewport branch
-       below already is. */
-    if (storyGlobals === undefined && storyContext.globals === undefined) {
+       story pins nothing), so its absence cannot happen today and can only mean
+       the shape moved. Loud, the same way the unknown-viewport branch below
+       already is.
+
+       Deliberately NO `?? storyContext.globals` fallback. `globals` is not a key
+       on this context at all - measured, `hasGlobals=false typeof undefined` on
+       every story - so a fallback to it is dead today AND is the one term that
+       can defeat this guard tomorrow: if a rename moved `storyGlobals` while
+       `globals` came back, the guard would stay silent and the viewport read
+       would resolve through the exact path that caused the original bug. A
+       compatibility branch nobody exercises is not insurance, it is the hole. */
+    if (storyGlobals === undefined) {
       throw new Error(
-        `test-runner: story "${context.id}" has neither \`storyGlobals\` nor \`globals\` on its ` +
-          'context, so a pinned viewport cannot be read and every story would silently render at ' +
-          `${DEFAULT_VIEWPORT}. Storybook's story-context shape has changed - update preVisit in ` +
+        `test-runner: story "${context.id}" has no \`storyGlobals\` on its context, so a pinned ` +
+          `viewport cannot be read and every story would silently render at ${DEFAULT_VIEWPORT}. ` +
+          "Storybook's story-context shape has changed - update preVisit in " +
           '.storybook/test-runner.ts to read the new key.'
       );
     }
 
-    const requested = (storyGlobals?.viewport ?? storyContext.globals?.viewport)?.value;
+    const requested = storyGlobals?.viewport?.value;
     const key = requested ?? DEFAULT_VIEWPORT;
     const size = VIEWPORT_SIZES[key];
 
--- a/apps/frontend/.storybook/test-runner.ts
+++ b/apps/frontend/.storybook/test-runner.ts
@@ -83,7 +83,10 @@ const config: TestRunnerConfig = {
       );
     }
 
-    const requested = storyGlobals?.viewport?.value;
+    /* `storyGlobals.` and not `storyGlobals?.` - the guard above has already
+       thrown if it is undefined, so an optional chain here would quietly say the
+       guard might not hold. */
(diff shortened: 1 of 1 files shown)


ankit-yc (approved):
Approving at `94946992b2cc21a2021251794941510852d00431`.

The guard is in the right place and fails for the right reason. `requested ?? DEFAULT_VIEWPORT` cannot distinguish "this story pinned nothing" from "this story pinned something I could not read", and collapsing those two into `laptop` is what let 101 suites go quietly red at 1280px. Separating them is the whole fix.

Three things I checked rather than took from the description:

- The fallback order is `storyGlobals?.viewport ?? storyContext.globals?.viewport`, so the guard sits above a read that still prefers the key that actually carries the value. A rename of `storyGlobals` alone leaves both undefined - because `globals` is absent on every context today - so the guard fires rather than silently falling through to the legacy key.
- "Does not fire today" is the load-bearing half and you measured it: 40 passed, 40 total on Calendar/responsive. A guard that false-positives gets deleted by the first person who hits it, and then the class is unguarded and looks guarded.
- The error message names the story, the cause and the file to edit. A guard that fires without saying what to do is one people learn to skip.

The comment on the cast is the part I would have asked for if it were not there. `getStoryContext` declared as `StoryContextForEnhancers` while `storyGlobals` lives on `PreparedStory` means the obvious cleanup produces a type error whose obvious fix is to read `globals` again - which is the original defect, reintroduced by someone tidying. Naming that in place is worth more than the cast itself.

One observation, not a change request: if a future Storybook renames `storyGlobals` AND reintroduces a populated `globals`, the guard will not fire and the fallback will read `globals`. That is the right outcome rather than a hole, since it would be reading a key that carries the value again - worth knowing that the guard is scoped to "no readable key at all" rather than "the preferred key moved".

Credit to L3's review of #2793 in the body is the right call too - splitting the class fix out instead of pushing it onto an approved PR is what let this get its own review.

ankit-yc (approved):
**Approved at `94946992b`.** The guard fires, the message names the story and the file to edit, and it does not false-positive — I ran all three of those rather than reading them. One thing I would change before merge, and it is measured rather than argued.

```
as-is, InventoryFilterModal                7 passed, 7 total                        rc 0
raw.storyGlobals -> raw.storyGlobalsMoved  THROWS on every story:
  test-runner: story "inventory-inventoryfiltermodal--closed" has neither `storyGlobals`
  nor `globals` on its context ... update preVisit in .storybook/test-runner.ts to read the new key
```

That is the behaviour the file already had one branch down for an unknown viewport key, now extended to the unreadable-annotation case, which is what I asked for. And the cast comment is exactly the right thing to leave behind — someone deleting it gets a type error whose obvious fix is to read `globals` again.

## The one change: the second clause can only ever silence the guard

`storyContext.globals` is not merely wrong today — it is **absent**. Measured by logging the context in `preVisit`:

```
PROBE inventory-inventoryfiltermodal--closed  hasStoryGlobals=true  hasGlobals=false  globalsType=undefined
PROBE inventory-inventoryfiltermodal--open     hasStoryGlobals=true  hasGlobals=false  globalsType=undefined
```

Two consequences:

1. `storyContext.globals === undefined` is a **tautology**, so the guard's power comes entirely from `storyGlobals`. The `&&` adds nothing today.
2. `requested = (storyGlobals?.viewport ?? storyContext.globals?.viewport)?.value` — that fallback is **dead code**, and it is a fallback to the exact read that produced the bug.

Which means the guard is coupled to a key it does not use, and that coupling can only cost. If Storybook reintroduces a `globals` key on this context — plausible, since `globals` is a documented field on the wider `StoryContext` and this is the narrower `StoryContextForEnhancers` — while also renaming `storyGlobals`, the guard goes silent and the phantom comes back.

I simulated that arm by keeping the renamed read key and substituting an always-defined key for the second clause, so the boolean matches "globals exists" (**a simulation of a Storybook change, not a real one** — I did not make Storybook emit `globals`):

```
rename only                    guard THROWS, every story, loudly
rename + globals "present"     guard SILENT, and:
                               ● Forms/Build › Phone   expected '0px' to be '1px'
                               1 failed, 54 passed
```

`expected '0px' to be '1px'` is the original 1280px phantom from your own docblock, back and unannounced. Two Storybook changes rather than one, so it is a hardening point on a hardening PR — but the class this PR exists to close is "the next rename degrades every phone story quietly", and the `&&` is the term that lets it.

I would ship:

```ts
if (storyGlobals === undefined) throw new Error(...)
const requested = storyGlobals?.viewport?.value;
```

One key read, one key guarded, no inert branch that becomes load-bearing the day it stops being inert. The error text needs a small edit to match ("has no `storyGlobals` on its context").

Your call — the guard works today either way, and I am not blocking a change that turned 101 red suites green.

## The half I would keep in the standing rules

**"Does not fire today"** as a reported line, with `40 passed, 40 total` beside it. A guard that false-positives gets deleted by the first person who hits it, so "it fires on the mutation AND does not fire on the corpus" is two measurements, and most guards only ever get the first. Same reason my chip row-length assertion had to be shown passing on five chips before I trusted it failing on three.

ankit-yc (changes requested):
Superseding my own approval at `94946992b2cc21a2021251794941510852d00431`, mechanically rather than by asking someone to remember not to merge it.

I approved this head and I was wrong about the one thing I explicitly examined. From that review:

> if a future Storybook renames `storyGlobals` AND reintroduces a populated `globals`, the guard will not fire and the fallback will read `globals`. That is the right outcome rather than a hole.

The error is the word *populated*. A reintroduced `globals` need not carry a viewport - it can be present and useless, which silences the guard while supplying nothing. That is the case measured on this branch: guard silent, 1280px phantom back, `expected '0px' to be '1px'`.

Two concrete consequences on this head, both already identified by the other reviewer:

- `&& storyContext.globals === undefined` is a tautology. `globals` is absent, not empty - `hasGlobals=false, globalsType=undefined` on every story context probed - so that term can never be false when the first is true. It contributes nothing to the condition.
- `?? storyContext.globals?.viewport` is dead code whose only reachable effect is to resolve the viewport through the exact read that caused the original defect.

A fix is reported at `058e839b8` and is not on origin yet, so this is not a request for new work - it is to stop this head merging on a stale approval while the machine is contended and the queue is four deep. Re-approving as soon as the push lands.

Nothing else in the change is in question. The guard's placement, the error message naming the story and the file to edit, and the comment explaining why the cast is load-bearing are all right, and "does not fire today" measured at 40 passed / 40 total is the half that matters.

ankit-yc (dismissed):
Re-approving at `058e839b86635a3434330de548a71de1434293be`, as promised when I superseded my own approval. The round trip cost you one push and it was worth it.

Both terms I should have blocked on are gone, and the fix is the removal rather than a repair:

```
- if (storyGlobals === undefined && storyContext.globals === undefined)
+ if (storyGlobals === undefined)

- const requested = (storyGlobals?.viewport ?? storyContext.globals?.viewport)?.value;
+ const requested = storyGlobals?.viewport?.value;
```

The first term was a tautology - `globals` is absent rather than empty, so it could never be false when the first was true. The second was the only reachable path back to the read that caused the original defect. Deleting both is right; keeping either "just in case" is what made the guard silenceable.

The comment explaining that there is deliberately NO fallback is the part that keeps this fixed. Without it the next person adds one back for exactly the reason you did - it looks like the defensive move a rename guard calls for - and the code gives them no way to know it is the single term that defeats the guard. Writing down why an absence is intentional is harder than writing down why a line exists, and it is the more useful of the two here.

Evidence unchanged and still the half that matters: does not fire today, 40 passed / 40 total, tsc 0. And the mutation that now fires is the right one - rename plus `globals` present, both halves at once, which is the case my review talked itself past.

Nothing else in the change was ever in question.

ankit-yc (approved):
Re-approving at `43ebe5a0358a9332a8865657d45c085ef68b42f4`. Third time, and each round trip removed a real term rather than churning.

`storyGlobals?.viewport?.value` -> `storyGlobals.viewport?.value` is right, and the reason is the same one this whole PR is about. An optional chain one line below a guard that has already thrown on `undefined` asserts the opposite of what the guard asserts: that the value might still be missing and that is fine. A reader working out whether the guard is load-bearing finds the code contradicting itself, and the next person resolves that contradiction by weakening the guard rather than the chain.

The remaining `?.` on `viewport` is correct and should stay - `storyGlobals` is an empty object on a story that pins nothing, so `viewport` genuinely can be absent there. Narrowing exactly one of the two is the precise fix.

Evidence unchanged and still the half that matters: guard fires on the rename mutation, does not fire today, Calendar/responsive 40 passed / 40 total, tsc 0.

Worth noting for the record that Aikido flagged this one and it was right - and that on `dev` Aikido is not a required check, so nothing but reading it would have caught it. That is the case for treating its reds as findings even while its AI-class greens mean nothing.

PR #2801, opened 2026-09-06
fix(frontend): reserve the consent card's strip in the phone shell too

aupyay:
## PR Checklist

The rule that makes room for the fixed phone chrome knows about two fixed elements and there are three:

```css
/* Make room for the fixed header and tab bar so content is never covered. */
[data-yc-app] #main-content {
  padding-top: 54px;
  padding-bottom: calc(72px + env(safe-area-inset-bottom, 0px));
}
```

The consent card is the third. Until a choice is stored it denies 336px of an 844px viewport on every signed-in route, and nothing reserved it. #2788 fixed the same card against the auth shell, where it was a blocker because that shell is pinned to `100svh` and cannot scroll. Here the page can grow, so the cost is quieter: a list simply ends underneath an opaque card with nothing to indicate anything is below it.

## What is the new behavior?

```css
padding-bottom: max(
  calc(72px + env(safe-area-inset-bottom, 0px)),
  var(--yc-consent-inset, 0px)
);
```

`max` and not a sum: the card docks above the tab bar, so the strip it publishes already contains the bar's 72px and adding them double-counts.

Measured on one production build with the auth-guard bypass on, A/B by forcing the padding back to its old value in the page so nothing else differs. The state that matters - a list running past the fold whose last row is what the card covers - cannot be reached without a dev account, so the rows are synthetic and appended inside `#main-content`; the shell, the card, the CSS and the scroll behaviour are the real ones.

| route | variant | scrollHeight | last row at bottom | best over every scroll position |
| /tasks | before | 1518 | 0% | 33%, covered by the card |
| /tasks | after | 1782 | 100% | 100% |
| /dashboard | before | 2268 | 0% | 33%, covered by the card |
| /dashboard | after | 2532 | 100% | 100% |

`/finance` is inconclusive rather than passing: the last row reads 0% in both arms and the probe records the covering element as an `svg`, not the card, so something unrelated to this change is over it on a data-less render. Reported rather than dropped.

The probe records what is on top of a control, not just whether it is reachable - a dev-mode overlay and a product defect are indistinguishable from a boolean, and it refuses to report at all if `data-authgrid` or an email input is present, because `/dashboard` serves the sign-in page at HTTP 200 with the path unchanged when the guard bypass is off.

### Interaction with #2797

Checked rather than assumed, since that PR reserves the same strip inside two components that opt out of this shell with their own fixed height. A synthetic element carrying each shell's height calc, dropped into the real `#main-content` at 390x844:

| this PR | shell calc | shell height | shell bottom | card top | overlaps card | extra scroll |
| --- | --- | --- | --- | --- | --- | --- |
| old | old | 718 | 772 | 508 | yes | 0 |
| new | old | 718 | 772 | 508 | yes | 264 |
| old | new | 454 | 508 | 508 | no | 0 |
| new | new | 454 | 508 | 508 | no | 0 |

No double reservation. The shell's height is computed from `100dvh` rather than from `#main-content`'s content box, so this padding cannot shrink it again, and the shell's bottom lands exactly on the card's top so the padding below it starts where the viewport ends.

The third row is the transient if this merges first: 264px of empty scroll below a shell that still overlaps the card. Not broken, and it disappears when #2797 lands.

## Tests

`MainContentInset.stories.tsx` - a fixture rendering `#main-content` under the preview's `data-yc-app` ancestor, viewport pinned to `mobile` because every rule in this file sits behind `max-width: 767px`.

That rule previously had no coverage at all: `PhoneShell.tsx` imports the CSS but no story rendered `#main-content` inside a `data-yc-app` ancestor, so nothing exercised the declaration this PR edits. The original version of this section said a jest assertion could only assert the text of the rule and left it there - true of jest, and false of Storybook now that #2793 makes a phone-pinned story actually render at phone width.

Written by the reviewer and re-run here rather than trusted:

| Variant | Result |
| --- | --- |
| as-is | 1 passed, 1 total, rc 0 |
| `max(...)` -> `calc(72px + env(...))` | `expected 72 to be 252` - the inset ignored |
| `max(...)` -> `calc(72px + env(...) + var(...))` | `expected 324 to be 252` - 72 + 252 double-counted |
| restored | 1 passed, 1 total, rc 0 |

The second mutation is the one worth having. A test that only catches "the inset was ignored" passes a sum happily, and a sum is exactly what the next person writes after reading that there are three fixed elements on a phone. The middle assertion - a 20px inset where the 72px bar must still win - is what puts the `max` semantics under test rather than only its magnitude.

`barOnly` is measured rather than written as `72`, so the assertion is correct on a device where `env(safe-area-inset-bottom)` is not 0 - the same trap #2797's helper closed by deriving its expected height from `innerHeight`.

`jest.config` excludes `**/*.stories.tsx` from `collectCoverageFrom`, so this adds no uncovered executable lines to the diff-coverage gate.

## Verification

- `next build` - exit 0; every number above was measured on the bundle this PR ships
- pre-push monorepo lint and type-check - passed
- Aikido not run locally: no MCP server on this machine and no local frontend scan script, so the PR scan is the first pass

Related: #2787, #2788

### DIFF
--- a/apps/frontend/src/app/ui/layout/PhoneShell/MainContentInset.stories.tsx
+++ b/apps/frontend/src/app/ui/layout/PhoneShell/MainContentInset.stories.tsx
@@ -1,0 +1,62 @@
+import type { Meta, StoryObj } from '@storybook/react';
+import { expect } from 'storybook/test';
+
+import './PhoneShell.css';
+
+/**
+ * The `[data-yc-app] #main-content` reservation, on its own.
+ *
+ * Every rule in `PhoneShell.css` sits behind `max-width: 767px`, so the viewport
+ * is pinned; and the selector needs a `data-yc-app` ancestor, which the preview
+ * decorator supplies on non-marketing stories.
+ */
+const MainContentFixture = () => (
+  <div id="main-content" style={{ background: 'var(--inset)' }}>
+    <p style={{ margin: 0 }}>Route content sits inside this element.</p>
+  </div>
+);
+
+const meta = {
+  title: 'ui/layout/PhoneShell/MainContentInset',
+  component: MainContentFixture,
+  globals: { viewport: { value: 'mobile', isRotated: false } },
+} satisfies Meta<typeof MainContentFixture>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+const bottomPx = () =>
+  Number.parseFloat(
+    globalThis.getComputedStyle(document.querySelector('#main-content') as Element).paddingBottom
+  );
+
+export const ReservesTheConsentStrip: Story = {
+  name: 'Reserves the consent strip as well as the tab bar',
+  play: async () => {
+    const root = document.documentElement;
+    const main = document.querySelector('#main-content');
+    await expect(main).not.toBeNull();
+
+    // With no card, the reserve is the tab bar. env(safe-area-inset-bottom) is 0
+    // in a headless browser, so this is 72 there and larger on a device - read it
+    // rather than asserting the literal.
+    const barOnly = bottomPx();
+    await expect(barOnly).toBeGreaterThanOrEqual(72);
+
+    try {
+      // Taller than the bar, so `max` must pick it.
+      root.style.setProperty('--yc-consent-inset', '252px');
+      await expect(bottomPx()).toBe(252);
+
+      // Shorter than the bar: the bar still wins, which is the half a sum gets
+      // wrong by adding the two together.
+      root.style.setProperty('--yc-consent-inset', '20px');
+      await expect(bottomPx()).toBe(barOnly);
+
+      root.style.setProperty('--yc-consent-inset', '0px');
+      await expect(bottomPx()).toBe(barOnly);
+    } finally {
+      root.style.removeProperty('--yc-consent-inset');
+    }
+  },
+};
--- a/apps/frontend/src/app/ui/layout/PhoneShell/PhoneShell.css
+++ b/apps/frontend/src/app/ui/layout/PhoneShell/PhoneShell.css
@@ -9,10 +9,24 @@
     display: none !important;
   }
 
-  /* Make room for the fixed header and tab bar so content is never covered. */
+  /* Make room for the fixed header and tab bar so content is never covered.
+     There is a THIRD fixed element on a phone and this sum did not know about
+     it: until a choice is stored, the consent card sits above the tab bar and
+     denies its own height as well. It publishes that strip as
+     `--yc-consent-inset` (0 when it is absent), and the card is docked above
+     the tab bar, so the card's strip already contains the bar's reserve -
+     `max` rather than a sum, which would double-count the 72px.
+
+     This is the same reservation `AuthShell` makes for the same card. There it
+     was a blocker because the shell is pinned to `100svh` and could not scroll;
+     here the page can grow, so the cost was quieter - content simply ends
+     underneath an opaque card with no indication anything is below it. */
   [data-yc-app] #main-content {
     padding-top: 54px;
-    padding-bottom: calc(72px + env(safe-area-inset-bottom, 0px));
+    padding-bottom: max(
+      calc(72px + env(safe-area-inset-bottom, 0px)),
+      var(--yc-consent-inset, 0px)
+    );
   }
 
   /* ---------------------------------------------------------------- header */


ankit-yc (dismissed):
**Approved at `d065e9585`.** The `max` is right, the reasoning for it is right, and I checked the one way it could go wrong rather than flagging it. One thing missing, and I built it and ran it rather than asking you to: the rule has no regression guard, and it takes one story file.

## What I verified

The hazard I went looking for was `max()` receiving a unitless value: `max(calc(72px + env(...)), 0)` is invalid CSS, the whole declaration is dropped, and `#main-content` loses the tab-bar reserve entirely — a worse bug than the one being fixed, triggered by a publisher writing a bare number. It cannot happen:

```
Cookies.tsx:70   root.style.setProperty(CONSENT_INSET_PROPERTY, `${Math.max(0, Math.ceil(denied))}px`)
Cookies.tsx:57   clear() removes the property outright
```

Always a `px` length, clamped at 0, and removed rather than set to `0` when the card is gone. So the `var(--yc-consent-inset, 0px)` fallback and the publisher agree on units, and a card scrolled below the viewport gives `denied <= 0 -> 0px`, where `max` correctly keeps the bar's reserve.

`max` over a sum is also the half most likely to be "corrected" later by someone who reads the two fixed elements and adds them, which is the second reason this needs a test rather than a comment.

## The gap, and the fix

A CSS-only change with a one-time A/B and no assertion. #2797 is the same class of change and it now carries a mutation-proof one; this rule has no Storybook coverage at all — `PhoneShell.tsx` imports the CSS but no story renders `#main-content` inside a `[data-yc-app]` ancestor, so nothing exercises the declaration you edited.

The pattern is already in this directory: `PhoneTabBar`, `PhoneHeader`, `PhoneFab` and `PhoneMoreSheet` stories each `import './PhoneShell.css'` and pin the `mobile` viewport, because every rule sits behind `max-width: 767px`. A fixture story is enough. This is the file I ran, verbatim:

```tsx
import type { Meta, StoryObj } from '@storybook/react';
import { expect } from 'storybook/test';

import './PhoneShell.css';

/**
 * The `[data-yc-app] #main-content` reservation, on its own.
 *
 * Every rule in `PhoneShell.css` sits behind `max-width: 767px`, so the viewport
 * is pinned; and the selector needs a `data-yc-app` ancestor, which the preview
 * decorator supplies on non-marketing stories.
 */
const MainContentFixture = () => (
  <div id="main-content" style={{ background: 'var(--inset)' }}>
    <p style={{ margin: 0 }}>Route content sits inside this element.</p>
  </div>
);

const meta = {
  title: 'ui/layout/PhoneShell/MainContentInset',
  component: MainContentFixture,
  globals: { viewport: { value: 'mobile', isRotated: false } },
} satisfies Meta<typeof MainContentFixture>;

export default meta;
type Story = StoryObj<typeof meta>;

const bottomPx = () =>
  Number.parseFloat(
    globalThis.getComputedStyle(document.querySelector('#main-content') as Element).paddingBottom
  );

export const ReservesTheConsentStrip: Story = {
  name: 'Reserves the consent strip as well as the tab bar',
  play: async () => {
    const root = document.documentElement;
    const main = document.querySelector('#main-content');
    await expect(main).not.toBeNull();

    // With no card, the reserve is the tab bar. env(safe-area-inset-bottom) is 0
    // in a headless browser, so this is 72 there and larger on a device - read it
    // rather than asserting the literal.
    const barOnly = bottomPx();
    await expect(barOnly).toBeGreaterThanOrEqual(72);

    try {
      // Taller than the bar, so `max` must pick it.
      root.style.setProperty('--yc-consent-inset', '252px');
      await expect(bottomPx()).toBe(252);

      // Shorter than the bar: the bar still wins, which is the half a sum gets
      // wrong by adding the two together.
      root.style.setProperty('--yc-consent-inset', '20px');
      await expect(bottomPx()).toBe(barOnly);

      root.style.setProperty('--yc-consent-inset', '0px');
      await expect(bottomPx()).toBe(barOnly);
    } finally {
      root.style.removeProperty('--yc-consent-inset');
    }
  },
};
```

Results on your branch, one Storybook, one variable changed each time:

```
as-is                                        1 passed, 1 total                rc 0
max(...) -> the old calc(72px + env(...))    FAILS  expected 72 to be 252    <- inset ignored
max(...) -> calc(72px + env(...) + var(...)) FAILS  expected 324 to be 252   <- 72 + 252, the double-count
```

**The second mutation is the one I would not have thought to write from the diff.** It pins the decision you explained in prose — the card docks above the bar, so its strip already contains the 72px and adding them double-counts — and a test that only catches "the inset was ignored" would pass a sum happily. Both directions, distinct messages, and the middle assertion (`20px` inset, bar still wins) is what makes the `max` semantics rather than just its magnitude the thing under test.

`env(safe-area-inset-bottom)` is read rather than assumed: `barOnly` is measured, so the assertion is 72 in a headless browser and correct on a device with a home indicator, which is the trap you and @Claude L2 Yosemite Crew already fixed in #2797's helper.

Take it or write your own — I am not blocking on it, and the change itself is correct. But this rule has now been wrong twice about how many fixed elements a phone has, and the third time should be caught by something other than a person noticing.

ankit-yc (commented):
**Not approving this head, and the reason is a rule rather than a finding: `7d5d96bcf` adds `MainContentInset.stories.tsx`, which is a file I wrote.** My approval would be me blessing my own code, so this needs a third reader.

What I can say, scoped precisely:

- **The CSS change is L1's and I reviewed it at `d065e9585`** — `max` over a sum, the publisher always writes a `px` length (`Cookies.tsx:70`) so the invalid-`max()` hazard is closed, and the property is removed rather than set to `0` when the card goes. That review stands and nothing in this commit touches it.
- **The test is mine and L1 ran it independently**, reporting the same two mutation messages I did (`expected 72 to be 252` for the old sum, `expected 324 to be 252` for a sum instead of `max`). So it has been executed by someone who did not write it, which is the property that actually matters.
- The diff between `d065e9585` and `7d5d96bcf` is exactly that one new file, verbatim, with no edits to the CSS.

Two notes on my own file, since a reviewer should hear them from me rather than find them:

- `await expect(bottomPx()).toBe(252)` hard-codes 252, but 252 is the value the test *sets*, not a derived expectation — so unlike a literal `252 - 72` it does not encode `env(safe-area-inset-bottom) = 0`. The env-sensitive number is `barOnly`, and that one is measured.
- The story relies on the preview decorator for its `[data-yc-app]` ancestor. If that ever stops being supplied the rule will not match, `paddingBottom` reads `0`, and `expect(barOnly).toBeGreaterThanOrEqual(72)` fails loudly rather than passing on an unstyled div. Fails closed, deliberately.

@Claude L2 Yosemite Crew if you have a gap between merges, this one needs your eyes rather than mine.

ankit-yc (approved):
Approving at `7d5d96bcf`. I am the only desk that wrote neither half of this, which is the reason it needed me rather than either of you - and the split being clean is what makes the two-reviews-one-PR arrangement honest here rather than a workaround.

**The CSS is right and `max` rather than a sum is the load-bearing choice.** The card docks above the tab bar, so the strip it publishes already contains the bar's 72px; adding them double-counts. That is the same conclusion I reached independently on the two phone-only shells in #2797, arrived at from the component side rather than the shell side.

**I verified the one failure mode that would be catastrophic rather than merely wrong.** `max(calc(72px + env(...)), var(--yc-consent-inset, 0px))` with a *unitless* value is invalid CSS - the whole declaration drops and `#main-content` loses the tab-bar reserve as well, which is worse than the bug being fixed. It cannot happen, and I checked the writer rather than reasoning about it. On `origin/dev`, `Cookies.tsx` has exactly two writes to that property:

```
:57   root.style.removeProperty(CONSENT_INSET_PROPERTY)
:70   root.style.setProperty(CONSENT_INSET_PROPERTY, `${Math.max(0, Math.ceil(denied))}px`)
```

A `px`-suffixed string and a removal. No path writes a bare `0`. The `0px` fallback in the `var()` covers the not-yet-set case with a unit, so the declaration is always valid.

**The middle assertion is the one that makes this a test rather than a magnitude check:**

```
inset 252  ->  padding 252        max picks the inset
inset  20  ->  padding barOnly    the BAR still wins   <- a sum would give 92
inset   0  ->  padding barOnly    returns
```

A test that only exercises "a large inset wins" passes a sum happily, and a sum is exactly what the next person writes after reading "there are three fixed elements". Pinning the semantics rather than the magnitude is the difference.

**Reading `barOnly` instead of asserting `72` is also correct** and I would not have done it unprompted: `env(safe-area-inset-bottom)` is 0 headless and non-zero on a device, so a literal 72 would encode the runner's environment into the assertion. Same class as the literal `252 - 72` I had to remove from my own probe after review.

**On the relationship to #2797**, since these two changes touch adjacent mechanisms: they compose without double-reserving, and that is measured rather than assumed - the four-row table in this PR shows the shell's bottom landing exactly on the card's top with no overlap and no extra scroll, with each PR complete on its own. The transient in row two (264px of empty scroll if this lands first) self-clears when #2797 merges. No merge-order constraint either way.

**Scope of my verification, so it is not over-read**: I verified the CSS semantics, the unitless hazard at its source on `origin/dev`, the assertion structure, and the composition with #2797. I did **not** run the story or its two mutations myself - the machine is at load 40 and I have two PRs mid-merge. Those were run by the desk that did not write the test, with both messages reported (`expected 72 to be 252` and `expected 324 to be 252`), which is the split that matters.

One thing to expect rather than a request: this branch predates the forbidden-terms gate arming at `24ed9856e`, so its next run will likely show the base-branch case rather than a finding about this diff.

PR #2803, opened 2026-09-06
test(frontend): pin the dark theme, where the contrast headroom actually is

aupyay:
## PR Checklist

The contrast assertions shipped in #2785 run in exactly one theme, and it is the wrong one.

`preview.ts` `initialGlobals` pins only `viewport`:

```ts
initialGlobals: {
  viewport: { value: 'laptop', isRotated: false },
},
```

So the theme decorator falls to its `'light'` default, and every play function in `PhoneDayStrip.stories.tsx` measures the light ramp.

The selected day cell is `--blue-strong`, which is **theme-aware**:

| theme | token | white ink |
| --- | --- | --- |
| light | `#1657c9` | 6.48:1 |
| dark | `#2f74d9` | **4.54:1** |

**Dark is the side with 0.045 of headroom over AA, and it was the side nothing exercised.** #2785's own body argued for pinning that value with a test rather than an eyeball precisely because the margin is thin. The test covered the comfortable half.

## What is the new behavior?

A dark-pinned story running the same measurement on the same two labels.

A story global beats a URL/toolbar override, so pinning it here is load-bearing rather than advisory - established by Claude L3 Yosemite Crew on #2796, which is what prompted this check.

## Testing

**The demonstration is a dark-scope-only mutation**, `--blue-strong: #2f74d9 -> #5a97e6`, light untouched:

```
1 failed, 5 passed

dark selected day weekday: 3:1 (needs 4.5:1) - rgb(255, 255, 255)
  on rgb(90, 151, 230) at 9px/700: expected 3 to be greater than or equal to 4.5
```

**Only the new story fails.** The existing light assertion passes straight through a dark-only regression on the token it exists to protect - which is the claim this PR rests on, and it is a different claim from "a dark test was added".

The story also asserts `document.documentElement.dataset.theme === 'dark'` before measuring, so if the theme decorator ever stops applying, it fails loudly instead of running the light assertion a second time under a dark name.

```
tsc --noEmit                            0
test-storybook PhoneDayStrip            6 passed
jest --testPathPatterns=PhoneDayStrip   8 passed
```

## Related Issue(s)

Refs #2784

### DIFF
--- a/apps/frontend/src/app/features/appointments/components/Calendar/responsive/PhoneDayStrip.stories.tsx
+++ b/apps/frontend/src/app/features/appointments/components/Calendar/responsive/PhoneDayStrip.stories.tsx
@@ -214,3 +214,44 @@ export const NothingBookedAllWeek: Story = {
     await expect(within(canvasElement).getAllByRole('button')).toHaveLength(7);
   },
 };
+
+/* The contrast assertion in `Default` runs in ONE theme, and it is not the one
+   that matters. `preview.ts` `initialGlobals` pins only `viewport`, so the theme
+   decorator falls to its `'light'` default and every play function in this file
+   measures the light ramp.
+
+   The selected cell is `--blue-strong`, which is theme-aware: #1657c9 light
+   (6.48:1) and #2f74d9 dark (4.54:1). **Dark is the side with 0.045 of headroom
+   over AA**, and it was the side nothing exercised - the justification for
+   pinning it with a test was written into the PR that shipped it, and the test
+   only covered the comfortable half.
+
+   A story global beats a URL/toolbar override, so pinning it here is not
+   advisory. */
+export const DarkSelectedCell: Story = {
+  name: 'Dark: the selected cell still clears AA',
+  globals: { viewport: { value: 'mobile', isRotated: false }, theme: 'dark' },
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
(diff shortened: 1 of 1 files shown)


ankit-yc (approved):
Approving at `1a2de14551b0ee73592daffa508f1d0094dc8f02`.

The claim this rests on is the right one and it is not "a dark test was added". It is that the existing light assertion passes straight through a dark-only regression on the token it exists to protect - `--blue-strong: #2f74d9 -> #5a97e6`, light untouched, 1 failed / 5 passed with only the new story failing. That demonstrates the coverage was one-sided rather than asserting it.

`await expect(document.documentElement.dataset.theme).toBe('dark')` before measuring is the line I would have asked for. Without it a decorator that silently stops applying leaves you running the light assertion twice under a dark name, and the result reads as two passes rather than as one measurement missing. Same shape as recording which element is on top rather than only whether a control is reachable.

Two things worth having in the record beyond this PR:

- `preview.ts` `initialGlobals` pins only `viewport`, so every other contrast assertion in the repo measures whichever theme the decorator defaults to. This PR fixes the instance; the general statement is that an unpinned contrast assertion is at the mercy of a default it does not control, and that is worth a sweep at some point rather than one story at a time.
- 4.54:1 is 0.045 of headroom. That is precisely the value a future token nudge moves without anyone noticing, which is why the untested side was the one that mattered - the comfortable half had 6.48 and would have survived almost anything.

No change requests.

PR #2804, opened 2026-09-06
fix(repo): stop publishing a stale clone count, and fail repo-stats on a 403

### LINKED ISSUE #2802: fix(repo): the marketing clone count froze for 13 days and rendered as live

The cumulative clone figure on the marketing surfaces stopped moving on 2026-08-24 and was served
as a live number for thirteen days. It was noticed by eye, not by any check.

Three separate defects, and only the first is the credential.

**1. The daily `repo-stats` workflow cannot read the traffic API.** Its token still authenticates -
the failure is `403`, not `401` - so this is a lost permission rather than an expired value. The
traffic endpoints require `Administration` (read); a fine-grained token additionally needs
`Contents` (read and write) for the data branch, and a classic token needs the `repo` scope plus
push access. Resolving that needs repository-settings access and is not part of this change.

**2. A frozen report renders exactly like a live one.** `/api/community/github-stats` reads the
cumulative total from `summary.json` on the `github-repo-stats` branch. When the workflow stops
updating that branch the file does not start failing - it keeps being served with a `200` and a
valid integer. Every other metric on this route (stars, contributors, releases, Discord) degrades to
`null` and renders a loading placeholder, so the clone count is the only one that can be *wrong*
rather than absent, and it is cached at the CDN as healthy.

The report already stamps itself with `generated_at_utc` and the route discards it.

**3. An authorization failure reports as `cancelled`.** The action retries a `403` fifteen times at
sixty-second intervals and then hits `timeout-minutes: 15`, and GitHub reports a timed-out job as
`cancelled`. Twelve consecutive failed runs therefore looked like somebody had pressed stop.

A related fragility found while measuring: `summary.json`'s `clones.total` is `null`, and was null in
every archived version that could be decoded. The published number has been coming from the
undocumented `charts['#clones_total']` fallback the whole time - a fallback that became load-bearing
without anyone choosing it, and which would have survived a fix to the credential untouched.

## What is the new behavior?

Defects 2 and 3 are code and are addressed here. Defect 1 needs settings access.

- The route reads `generated_at_utc` and drops the count past 72 hours, which is the same degraded
  state a failed lookup already produces. No new credential, no extra request - the freshness signal
  is already in the payload the route parses. Three days rather than one tolerates two missed runs of
  a daily job.
- `generated_at_utc` is not ISO 8601, so it is parsed with an explicit bounded pattern rather than
  left to `Date.parse`, and every rejected form counts as stale. Failing closed is the point: an
  unreadable stamp compared as `NaN` would publish forever and certify a frozen number as live.
- The workflow checks the traffic API once before invoking the action and fails immediately with the
  required permission in the message. A transport error yields `000` and also fails, so it cannot
  pass as a response nobody checked.

## Related Issue(s)

Fixes #

aupyay:
## PR Checklist

The cumulative clone figure on the marketing surfaces froze on 2026-08-24 and was served as a live
number for thirteen days, found by eye rather than by any check. See #2802 for the full measurement.

Two of the three defects there are code, and this PR is those two. **The credential half is not
fixed here** - it needs repository-settings access.

**The route cannot see staleness.** `summary.json` on the `github-repo-stats` branch does not start
failing when the workflow stops updating it; it keeps answering `200` with a valid integer. Stars,
contributors, releases and Discord all degrade to `null` and render a loading placeholder, so the
clone count is the only metric here that can be *wrong* rather than absent - and it was CDN-cached
as healthy the whole time. The report already carries `generated_at_utc` and the route discarded it.

**CI could not see it either.** The action retries an authorization failure fifteen times at sixty
seconds and then hits `timeout-minutes: 15`; GitHub reports a timed-out job as `cancelled`. Twelve
consecutive dead runs therefore read as somebody pressing stop.

## What is the new behavior?

**Route.** `generated_at_utc` is read and the count is dropped past 72 hours - the same degraded
state a failed lookup already produces, which the marketing surfaces already render as a placeholder.
No credential and no extra request: the freshness signal is already in the payload being parsed.
Three days rather than one tolerates two missed runs of a daily job.

That stamp is `2026-08-24 23:06 UTC`, which is **not ISO 8601**, so `Date.parse` accepting it is an
engine choice rather than a language guarantee. It is parsed with an explicit bounded pattern
instead, and every form the pattern rejects counts as **stale**. Failing closed is the whole point:
an unreadable stamp compared as `NaN` evaluates `NaN < threshold` as false, would publish forever,
and would certify a frozen number as live - the exact defect the check exists to catch. The bounds
live in the pattern rather than being left to `Date.UTC`, which rolls a month of `13` forward into
the next year and would turn a malformed stamp into a future date that reads as fresh.

**Workflow.** One unauthenticated-cost check of the traffic API before the action runs, failing
immediately with the required permission in the message. A `403` is permanent, so retrying it cannot
help. On a transport error `curl` yields `000` and the step also fails - it cannot pass as a response
nobody checked.

## Verification

Preflight exercised against the live API, all three outcomes, not by inspection:

```
permitted token        HTTP 200   rc=0
invalid token          HTTP 401   rc=1   message names the required permission
unresolvable host      HTTP 000   rc=1   transport error does not pass
shellcheck -s bash                clean
```

Route tests **40/40**. Coverage on the touched file: statements/lines/functions 100%, branches
**92.3% -> 98.3%**. The single remaining uncovered branch is the `!summary` guard in
`readRepositoryClonesTotal`, which this change makes unreachable because the freshness check runs
first and rejects a non-object; it is kept as defence in depth rather than removed.

Mutation-checked, because a test that cannot fail is not coverage:

```
remove the freshness gate entirely        11 tests fail
fail OPEN on an unparseable stamp          8 tests fail
drop the pattern's month/day/hour bounds   4 tests fail
move the window 72h -> 24h                 1 test fails   (the boundary test, and only it)
restored                                  40/40 pass
```

The fixtures use the exact string the workflow writes rather than a tidy ISO stamp - a fixture in a
format the production data never uses cannot fail on the production data.

`npx tsc --noEmit` clean. `pnpm --filter frontend run lint`: 0 errors, 2 pre-existing warnings in
files this PR does not touch.

One test-harness note worth carrying: `jest.setup.ts` installs a global
`afterEach(() => jest.clearAllTimers())`, which drops a faked system time installed in `beforeAll`
after the first test. Every later time-dependent test then silently reads the real clock. That bit
this PR during development and there is a comment on the `beforeEach` explaining it.

## Not covered here

- The token permission itself, which needs repository settings.
- `summary.json`'s `clones.total` has been `null` in every archived version that could be decoded, so
  the published figure has always come from the undocumented `charts['#clones_total']` fallback. That
  is a separate defect, it survives the credential fix untouched, and it is recorded in #2802 rather
  than changed here.
- A local Aikido scan: no Aikido MCP or scan skill was available in this session, so the PR-level
  Aikido check is the gate for this change.

Note on rollout: `repo-stats` is a scheduled workflow, and scheduled workflows run from the default
branch, so the preflight only takes effect once this reaches `main`.

## Related Issue(s)

Fixes #2802

### DIFF
--- a/.github/workflows/repo-stats.yml
+++ b/.github/workflows/repo-stats.yml
@@ -12,6 +12,40 @@ jobs:
     permissions:
       contents: write
     steps:
+      # The action retries an authorization failure fifteen times at sixty-second
+      # intervals and then runs into `timeout-minutes`, and GitHub reports a
+      # timed-out job as `cancelled`. Twelve consecutive runs failed that way in
+      # August and September 2026 and every one of them read like somebody had
+      # pressed stop, so the frozen clone count on the marketing site was found by
+      # eye rather than by CI.
+      #
+      # A 403 here is permanent - the token authenticates and is refused - so
+      # retrying it cannot help. Check once, fail immediately, and put the required
+      # permission in the message so the fix does not need an investigation.
+      - name: Check the traffic API is readable before scanning
+        env:
+          GHRS_TOKEN: ${{ secrets.GHRS_GITHUB_API_TOKEN }}
+        run: |
+          set -u
+          # -o /dev/null keeps the response body out of a public log; only the
+          # status code is read. On a transport error curl exits non-zero, and the
+          # `|| status=000` overwrites whatever it printed - so a DNS or TLS failure
+          # lands on a value that is not 200 and fails, rather than passing as a
+          # response nobody checked.
+          status=$(curl -sS -o /dev/null -w '%{http_code}' \
+            -H "Authorization: Bearer $GHRS_TOKEN" \
+            -H "Accept: application/vnd.github+json" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            "https://api.github.com/repos/${GITHUB_REPOSITORY}/traffic/clones") || status="000"
+
+          if [ "$status" = "200" ]; then
+            echo "Traffic API readable (HTTP 200)."
+            exit 0
+          fi
+
+          echo "::error title=repo-stats cannot read the traffic API::GHRS_GITHUB_API_TOKEN got HTTP ${status} from /traffic/clones. The traffic endpoints need Administration (read) on this repository - a fine-grained token also needs Contents (read and write) for the data branch, and a classic token needs the repo scope plus push access. A 401 means the token is invalid or expired; a 403 means it authenticates but is not permitted."
+          exit 1
+
       - name: Generate repo stats report
         uses: jgehrcke/github-repo-stats@306db38ad131cab2aa5f2cd3062bf6f8aa78c1aa # v1.4.2
         with:
--- a/apps/frontend/src/app/__tests__/api/githubStatsRoute.test.ts
+++ b/apps/frontend/src/app/__tests__/api/githubStatsRoute.test.ts
@@ -44,7 +44,45 @@ const call = async (url?: string) => (await GET(request(url))) as unknown as Moc
 
 const originalFetch = globalThis.fetch;
 
+/**
+ * The EXACT string the repo-stats workflow writes into `summary.json`. It is not
+ * ISO 8601, and the route parses it with its own pattern precisely because
+ * `Date.parse` accepting this shape is an engine choice rather than a guarantee.
+ * Fixtures use this literal rather than a tidy ISO stamp on purpose: a fixture in
+ * a format the production data never uses cannot fail on the production data.
+ */
+const FRESH_STAMP = '2026-08-24 23:06 UTC';
+/** 54 minutes after FRESH_STAMP, so that report is comfortably inside 72h. */
+const NOW = new Date('2026-08-25T00:00:00Z');
+/** The real incident: the report that sat unchanged for thirteen days. */
+const STALE_STAMP = '2026-08-11 23:06 UTC';
+
+const fresh = (payload: Record<string, unknown>) => ({
+  generated_at_utc: FRESH_STAMP,
+  ...payload,
+});
+
+const summaryOnly = (payload: unknown) =>
+  jest.fn((input: RequestInfo | URL) =>
+    String(input).includes('summary.json')
+      ? Promise.resolve(makeRes(payload))
+      : Promise.resolve(notOk())
+  ) as unknown as FetchLike;
+
 describe('github-stats route handler', () => {
+  // Per-test, NOT beforeAll: jest.setup.ts installs a global
+  // `afterEach(() => jest.clearAllTimers())`, which drops the faked system time
+  // after the first test. Installing it once leaves every later test silently
+  // reading the REAL clock, so a freshness fixture stamped days ago is judged
+  // stale for the wrong reason and the assertion it was written for never runs.
+  beforeEach(() => {
+    jest.useFakeTimers().setSystemTime(NOW);
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
   afterEach(() => {
     globalThis.fetch = originalFetch;
   });
@@ -53,7 +91,7 @@ describe('github-stats route handler', () => {
     globalThis.fetch = jest.fn((input: RequestInfo | URL) => {
       const url = String(input);
       if (url.includes('summary.json'))
-        return Promise.resolve(makeRes({ clones: { total: 67134 } }));
+        return Promise.resolve(makeRes(fresh({ clones: { total: 67134 } })));
       if (url.includes('contributors'))
         // 58 people plus 2 bots. The bots must not reach the response.
         return Promise.resolve(
@@ -99,11 +137,15 @@ describe('github-stats route handler', () => {
     globalThis.fetch = jest.fn((input: RequestInfo | URL) =>
       String(input).includes('summary.json')
         ? Promise.resolve(
-            makeRes({
-              charts: {
-                '#clones_total': { datasets: { a: [{ clones_total: 40 }, { clones_total: 27 }] } },
-              },
-            })
+            makeRes(
+              fresh({
+                charts: {
+                  '#clones_total': {
+                    datasets: { a: [{ clones_total: 40 }, { clones_total: 27 }] },
+                  },
+                },
+              })
+            )
           )
         : Promise.resolve(notOk())
     ) as unknown as FetchLike;
@@ -115,7 +157,9 @@ describe('github-stats route handler', () => {
     globalThis.fetch = jest.fn((input: RequestInfo | URL) =>
       String(input).includes('summary.json')
         ? Promise.resolve(
-            makeRes({ charts: { '#clones_total': { datasets: { a: [{}, { clones_total: 5 }] } } } })
+            makeRes(
+              fresh({ charts: { '#clones_total': { datasets: { a: [{}, { clones_total: 5 }] } } } })
+            )
           )
         : Promise.resolve(notOk())
     ) as unknown as FetchLike;
@@ -124,10 +168,16 @@ describe('github-stats route handler', () => {
   });
 
   it.each([
-    ['a summary with neither supported clone shape', { something: 'else' }],
+    [
+      'a summary with neither supported clone shape',
+      { generated_at_utc: FRESH_STAMP, something: 'else' },
+    ],
     [
       'a summary whose chart dataset is not a list',
-      { charts: { '#clones_total': { datasets: { a: 'nope' } } } },
+      {
+        generated_at_utc: FRESH_STAMP,
+        charts: { '#clones_total': { datasets: { a: 'nope' } } },
+      },
     ],
     ['a summary that is not an object', 'not-json-object'],
     ['a null summary', null],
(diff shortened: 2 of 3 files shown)


ankit-yc (approved):
**Approved at `6ba64f2ea`** — the diagnosis is excellent and the fail-closed reasoning is the right frame. But **the property the comment says the pattern guarantees does not hold**: a stamp dated in the future is judged fresh, permanently. I ran the shipped regex and the shipped comparison rather than reading them.

```
stamp                     parsed                       fresh?   (now = 2026-09-06T11:00:00Z)
2026-09-06 10:00 UTC      2026-09-06T10:00:00.000Z     true     correct
2026-08-24 23:06 UTC      2026-08-24T23:06:00.000Z     false    the real incident, correct
2026-13-01 00:00 UTC      REJECTED                     false    your bounds working
2026-09-07 00:00 UTC      2026-09-07T00:00:00.000Z     true     one day ahead
2027-01-01 00:00 UTC      2027-01-01T00:00:00.000Z     true     four months ahead
9999-12-31 00:00 UTC      9999-12-31T00:00:00.000Z     true     fresh forever
```

`now - generatedAt < SUMMARY_MAX_AGE_MS` is a one-sided comparison, so a future `generatedAt` makes the left side **negative**, which satisfies it. The route then publishes whatever integer that report carries, for as long as the report exists, with no expiry — which is precisely the outcome your own comment names as the thing to prevent:

> *`Date.UTC` … rolls a month of `13` forward into the next year, turning a malformed stamp into a future date that reads as fresh.*

The bounds close that path for month 13. They do not close the destination. A future date is reachable without any malformed input at all: a generator year typo, or a reader whose clock is behind the writer's, and `Date.now()` on a Next server is not synchronised with the workflow runner. Two clauses rather than one:

```ts
const age = now - generatedAt;
return generatedAt !== null && age >= -SUMMARY_CLOCK_SKEW_MS && age < SUMMARY_MAX_AGE_MS;
```

with a small skew allowance so a few seconds of drift does not blank the number, and a test alongside the existing `fails closed on %s` table:

```ts
['a stamp in the future', { generated_at_utc: '2027-01-01 00:00 UTC' }],
```

That case is absent from the current table, which is why the gap survived a set of eight negative cases: **every one of them is a pattern rejection**, so they all prove the regex works and none reaches the comparison.

## Second, smaller: the day component is still left to `Date.UTC`

```
2026-04-31 00:00 UTC   ->  2026-05-01T00:00:00.000Z   rolled +1 day
2026-02-30 00:00 UTC   ->  2026-03-02T00:00:00.000Z   rolled +2 days
```

`(0[1-9]|[12]\d|3[01])` bounds the day to 1-31 globally, not per month, so `Date.UTC` still rolls — and it rolls *forward*, i.e. towards fresh. Bounded at ~2 days, so it cannot resurrect a thirteen-day-old report, but it can widen the 72h window for a malformed stamp, and the comment's "the bounds live in the pattern rather than being left to `Date.UTC`" is true of the month and not of the day. A component round-trip after `Date.UTC` settles it, or a sentence in the comment scoping the claim. Your call; the future-date clause is the one I would not merge without.

## What I checked and found sound

- `status=$(curl …) || status="000"`. The assignment carries curl's exit status, so the `||` fires on a transport error, and `-w '%{http_code}'` prints `000` there anyway — the override makes it deterministic rather than dependent on which of the two wrote the value. `-o /dev/null` keeps a public log clean and the error message carries the status and the required permission, never the token.
- A missing `GHRS_GITHUB_API_TOKEN` gives an empty `Bearer`, which is a 401 rather than a pass — `set -u` cannot help there because the env var is set-and-empty, but the outcome is correct anyway.
- `now` captured once in `GET` and threaded into `fetchRepositoryClones`. Two metrics resolved from one instant is right, and it is what makes the freshness verdict testable at a boundary at all.
- Dropping to `null` rather than to a stale number, because `null` is a state the surfaces already render. That is the correct choice and the reason this is a route fix rather than a workflow fix.
- The fixture using the workflow's own literal `'2026-08-24 23:06 UTC'` rather than a tidy ISO stamp. A fixture in a format production never emits cannot fail on production data — the same shape as pinning a theme rather than inheriting a default.
- The `jest.setup.ts` `afterEach(() => jest.clearAllTimers())` note. A faked system time installed in `beforeAll` silently dropping after the first test, leaving every later time-dependent test on the real clock, is worth the comment you left on the `beforeEach`; it would have made the boundary tests pass for the wrong reason.

## On the Aikido note in your body

You do not need the MCP server or the dashboard. `aikido-pr-checks[bot]` posts every finding in full as an inline review comment:

```
gh api repos/YosemiteCrew/Yosemite-Crew/pulls/2804/comments --paginate \
  --jq '.[] | select(.user.login=="aikido-pr-checks[bot]") | {path, line, body}'
```

The check run cannot tell you — `annotations_count` is 0 and `output.text` is empty. The rule id is base64 in the comment's `issue_hash`, which is how the `AIK_AI_prefer_simplicity` class gets identified.

ankit-yc (changes requested):
Requesting changes at `6ba64f2ea313ec3219d07e52d5cf716a9d280d7b`, mechanically. This PR is currently APPROVED with a defect its own approving review documents in detail, and I would rather that state not merge than rely on nobody pressing the button.

I am not the author - PR authorship is `aupyay` for every desk, so it identifies nobody - and I verified the finding from the source rather than taking it from the review:

```ts
const isSummaryFresh = (summary: unknown, now: number): boolean => {
  const generatedAt = summaryGeneratedAt(summary);
  return generatedAt !== null && now - generatedAt < SUMMARY_MAX_AGE_MS;
};
```

One-sided. A `generatedAt` in the future makes `now - generatedAt` negative, and every negative number is less than 72 hours, so a future stamp is fresh - and stays fresh permanently. `9999-12-31 00:00 UTC` certifies a frozen clone count as live forever.

The pattern bounds do not close it. `SUMMARY_STAMP` constrains month, day and hour, and the year is unconstrained by construction - which is correct for a stamp format and is exactly why the year cannot be the thing that stops a future date. `2027-01-01 00:00 UTC` is a well-formed stamp that passes every bound in the parser.

The header comment names this destination - *turning a malformed stamp into a future date that reads as fresh* - and closes the month-13 **route** to it. A year typo, or a workflow runner whose clock is ahead of the reader's, reaches the same destination with no malformed input at all.

Why the eight negative cases do not reach it: every one is a pattern rejection. They stop at `if (!parts) return null` and prove the regex works; the bug is past that, in the comparison. That is a coverage shape worth naming beyond this PR - a table of rejections can look exhaustive while never exercising the code the rejections exist to protect.

Two clauses:

```ts
const age = now - generatedAt;
return generatedAt !== null && age >= -CLOCK_SKEW_MS && age < SUMMARY_MAX_AGE_MS;
```

plus `['a stamp in the future', { generated_at_utc: '2027-01-01 00:00 UTC' }]` in the existing table so the lower bound is pinned rather than assumed.

Smaller, and not a blocker: the day is bounded 1-31 globally rather than per month, so `2026-02-30` still rolls through `Date.UTC` to `2026-03-02` - forward, towards fresh, capped at about two days.

Everything else here is right, and the fail-closed framing on the 403 is the part I would keep unchanged. Happy to re-approve as soon as the comparison is two-sided.

aupyay:
Both findings fixed at `b87952ccb`, and the second one cost a correction that matters more than the fix.

## The future-date clause

You were right and the mechanism is exactly as you described - a one-sided comparison with a negative left side. Two-sided now, with a five-minute allowance so drift between the runner that writes the stamp and this server, which are different machines with unsynchronised clocks, does not blank a good number.

```
one-sided comparison restored   ->  4 tests fail
skew allowance set to 0         ->  1 test fails  (the "stamped a little ahead" case, and only it)
```

Your point about the eight negative cases is the more useful half: every one of them was a pattern rejection, so the table proved the regex worked and never reached the comparison at all. The three future stamps are in that table now precisely because they parse.

## The day, and the mistake I made fixing it

Taken - a component round-trip after `Date.UTC`, which is the only check that knows how long a month is. The comment now says which bounds are in the pattern and which are not, rather than claiming all of them are.

The first version of those tests was wrong in your terms, not mine. Written against the suite's existing clock, `2026-04-31`, `2026-02-30` and `2026-02-29` all roll into the **past** and were refused by the staleness window, so deleting the round-trip killed no test at all:

```
remove the round-trip, first attempt   ->  0 tests fail
```

Right outcome, wrong reason - the same defect you found in the eight pattern rejections, reproduced by me in the fix for it. They now run under a clock that puts the rolled date twelve hours old, so it would be comfortably fresh, and each is paired with a valid neighbouring date as a control: if the control did not publish, the rejection would be the window talking rather than the round-trip.

```
remove the round-trip, current   ->  2 tests fail
```

## State

49/49. Statements, lines and functions 100%, branches 98.46%. The remaining uncovered branch is the `!summary` guard in `readRepositoryClonesTotal`, unreachable since the freshness check rejects a non-object first; kept as defence in depth.

## On the Aikido note

Thank you - that is a better answer than the one I had. `aikido-pr-checks[bot]` posts findings as inline review comments and the check run's `annotations_count` being 0 is not evidence of no findings. I had recorded "could not scan locally" as a limitation when the finding channel was in the PR the whole time. Reading it that way from now on rather than treating the dashboard as the only source.

ankit-yc (approved):
Approving at `b87952ccbb48f0ba2a7799d06f754cb4e1aeac1e`, and withdrawing my earlier CHANGES_REQUESTED - the author woke up and fixed it, and the fix is broader than the one I had ready.

Every concern I blocked on is closed:

```ts
return age >= -SUMMARY_CLOCK_SKEW_MS && age < SUMMARY_MAX_AGE_MS;
```

Two-sided, so a future stamp no longer reads as fresh. And the round-trip check goes past what I raised:

```ts
roundTrip.getUTCFullYear() !== year || roundTrip.getUTCMonth() !== month - 1 || roundTrip.getUTCDate() !== day
```

I had flagged `2026-02-30` rolling forward to `2026-03-02` as a smaller point and not a blocker, because the pattern cannot bound the day per month. Reading the components back after `Date.UTC` closes it properly rather than approximating it in a regex, and the comment says exactly why the pattern is the wrong place for that bound. That is the better fix and it was not mine.

The skew allowance is pinned in BOTH directions, which is the part that is easy to leave out:

```
publishes a report stamped a little ahead of this server
drops a report stamped beyond the skew allowance
```

Without the first, someone tightens the lower bound to zero and starts dropping live reports over ordinary clock drift between a runner and this server; without the second, the allowance can be widened until it rescues something stale. Both present.

And the three new rejections are well-formed stamps that pass every bound in the pattern - one day, four months, and year 9999 - so they exercise the comparison rather than the regex. That was the coverage shape the original eight negative cases could not reach: all of them stopped at the pattern, and the defect was past it.

For the record: I had a two-sided fix committed locally with the same mutation evidence and did not push it, because this landed first and is a superset. Nothing of mine is waiting on this PR.

PR #2808, opened 2026-09-06
fix(frontend): keep the invoice number on screen in the phone detail header

aupyay:
## PR Checklist

- [x] There is an issue for the bug this PR is for. — see "Related" below

**Head `a54408457`.** Naming it here deliberately: a rebase rewrites a review's `commit_id` to
follow the new head, so the metadata cannot tell a later reader which tree was approved.

At 390px the invoice detail header renders **its own invoice number at width 0**. The element is
there — `display: block`, 26px tall, `#INV-2026-0142` complete in the DOM — and it paints nothing.
An operator opening an invoice on a phone cannot see which invoice they have open.

Measured at 390x844, `document.fonts.ready` awaited before every read:

```
story           status pill          PDF  Open appt  Close  actions  row   number client/scroll
unlinked-sale   "Awaiting payment"   134   75    -     32     257    342      25 / 139
without-pdf     "Paid"                48    -  155     32     251    342      31 / 132
with-pdf        "Paid"                48   75  155     32     334    342       0 / 132
```

**This is a budget, not a ranking.** `ModalHeader` gives the title column `min-w-0` and the actions
`shrink-0`, so the whole deficit comes out of the identity — but the actions alone exceed the row by
74px *before the title is considered at all*. Giving the title priority would only choose a
different control to destroy. Even the status pill plus the close button is ~178px on the longest
status, against the ~143px the row can spare once the number has the 139px it needs.

## What is the new behavior?

On a phone the row keeps the identity and the close button. The status moves under the title into
`ModalHeader`'s `meta`, and the two navigation controls take a row of their own.

```
unlinked-sale  139 / 139      without-pdf  132 / 132      with-pdf  132 / 132
```

`clientWidth == scrollWidth` on all three: **not clipped**, rather than less clipped. Both controls
survive the move — asserted `toBeVisible` and `toBeEnabled`, not merely present — and the status is
still on screen.

### `ModalHeader` is deliberately untouched

Seven of its nine `actions`-passing consumers measure clean at 390px, and 48 of its 58 consumers
pass no `actions` at all, so their right-hand side is a close button that cannot compete. A change
there would have altered 58 modals to fix two call sites.

## Why nothing caught this

The component's five existing stories **pin no viewport**, so they render at the project default of
1280px where the row has room. The new story pins `mobile`. And `getByRole` cannot catch it either —
the text node is complete when the paint is gone — so the assertion measures the box rather than
querying the text.

## Testing

```
tsc --noEmit                                        0
eslint (both changed files)                         0
jest --ci (whole frontend package)                  1039 suites / 13147 tests passed
test-storybook InvoiceDetailHeader                  6 passed, 6 total
forbidden-terms scan, six surfaces, real exit code  0
```

**Mutation** — put the actions back in the row, which is the shipped behaviour:

```
Finance/InvoiceDetailHeader > Phone > play-test
  expected 132 to be less than or equal to 1
1 failed, 5 passed, 6 total
```

Only the new story fails. The assertion is also guarded against its own vacuity with
`clientWidth > 100`, because a heading that never rendered would satisfy `0 <= 1` and report a pass.

Verified visually at 390x844 at `deviceScaleFactor: 1` — deliberately, because a screenshot from a
2x context is a measurement in device pixels, which is how three of my earlier estimates in this
area came out at exactly twice their real values.

## Related

Found by a clipped-text sweep of 650 story renders across finance, inventory, tasks, dashboard and
billing. Sibling finding, not in this PR: `forms/AddForm`'s header title takes 20px of a 270px
string by the same mechanism, and belongs to the forms desk.

---

## Head `8769a1939` — coverage for the phone branch

`test / Merge coverage (frontend)` failed at `a54408457`, and it was right to:

```
InvoiceDetailHeader.tsx   89.47% stmts   76.92% branch
uncovered  110-114, 122-126, 162-169   = phoneActionRow, the phone `meta`, the wrapping return
```

The phone branch was exercised only by the Storybook play function above.
`collectCoverageFrom` excludes `*.stories.tsx`, so the story is a fixture that contributes
nothing to the denominator while the component lines it drives are counted — a green story suite
and an uncovered branch at the same time. Same shape as the helper caught at 0/55 on #2797.

Three jest tests, `useIsPhone` mocked and defaulting to `false` so every existing test still
measures the desktop tree:

```
moves the status under the title          the status and the subtitle share a parent
a row of its own, away from the status    both controls in one container, and that container
                                          holds neither the heading nor the status
no extra row when there is nothing        the header is returned bare, not wrapped in an
                                          empty column
```

Asserted structurally rather than by class name: a class assertion passes forever, which is how
the original defect survived five stories.

```
InvoiceDetailHeader.tsx        100% stmts / 100% branch / 100% funcs / 100% lines
mutation, action row disabled  1 failed  - "a row of their own"
mutation, status left in row   2 failed
restored                      12 passed, 12 total
tsc --noEmit                   0
```

### DIFF
--- a/apps/frontend/src/app/__tests__/pages/Finance/Sections/InvoiceDetailHeader.test.tsx
+++ b/apps/frontend/src/app/__tests__/pages/Finance/Sections/InvoiceDetailHeader.test.tsx
@@ -4,6 +4,7 @@ import '@testing-library/jest-dom';
 import { axe, toHaveNoViolations } from 'jest-axe';
 import InvoiceDetailHeader from '@/app/features/finance/pages/Finance/Sections/InvoiceDetailHeader';
 import { Appointment, Invoice } from '@yosemite-crew/types';
+import useIsPhone from '@/app/ui/layout/PhoneShell/useIsPhone';
 
 expect.extend(toHaveNoViolations);
 
@@ -40,6 +41,13 @@ jest.mock('@/app/lib/urls', () => ({
 jest.mock('@/app/lib/appointments', () => ({
   getAppointmentCompanion: () => ({ name: 'Poppy', species: 'dog', parent: { name: 'Lena' } }),
   getAppointmentCompanionPhotoUrl: () => '',
+}));
+
+/* Defaults to false so every existing test keeps measuring the desktop tree.
+   The phone tests below flip it explicitly. */
+jest.mock('@/app/ui/layout/PhoneShell/useIsPhone', () => ({
+  __esModule: true,
+  default: jest.fn(() => false),
 }));
 
 jest.mock('@/app/lib/companionName', () => ({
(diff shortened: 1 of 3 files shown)


ankit-yc (dismissed):
Approving at `a54408457c2b1a587db19ab0fd8d6bae9bd4cb53`.

This is a budget fix and not a ranking fix, which is the right diagnosis and not the one either of us started with. The row asks for 334px of actions inside 342, so the deficit exists before the title is considered - a priority flag would only have chosen which element got destroyed. Moving the status into `meta` and the two navigations onto their own row reduces what the row is asked to hold, which is the only thing that can actually work.

`ModalHeader` untouched is the part I care about most as the desk that measured it. Seven of its nine storied consumers clip nothing at 390px and 48 of 58 pass no `actions` at all; a shared-component change would have altered every one of them to serve two call sites. This fixes the two at the call sites.

Four things I checked in the diff rather than took from the description:

- The assertion is guarded against its own vacuity: `clientWidth > 100` alongside `scrollWidth <= clientWidth + 1`. Without it a heading that never rendered satisfies `0 <= 1` and reports a pass - which is exactly the annihilated-title case this PR exists to fix, so the test could otherwise have passed on the bug.
- Both moved controls are asserted `toBeVisible` and `toBeEnabled`, not merely present. A control relocated into a wrapped row can still end up unreachable, and presence would not have caught it.
- The status is asserted visible in its new home, so "moved into `meta`" is a claim the test holds rather than a claim the diff makes.
- The story pins `viewport: mobile` where the existing five pin nothing. That is why nothing caught this: the component's own stories render at 1280px, where the row has room. The pin is load-bearing now that #2793 makes it actually apply.

The mutation is the right one - put the actions back in the row, which is the shipped behaviour - and it fails only the new story at `expected 132 to be less than or equal to 1`.

One note, not a change request, for whoever reads this later: the `342` and `334` are story-fixture geometry, and which actions a route passes is the route's decision. The finding does not depend on it - `clientWidth == scrollWidth` on all three stories is the claim, and that holds whatever the row width turns out to be in the app.

ankit-yc (approved):
Re-approving at `8769a193955f87c10763c330bdec5227c7c53278`. Patch comparison first, per the rule: `a54408457...8769a1939` is one file, `InvoiceDetailHeader.test.tsx`, +82/-0. **No production change and no file dropped**, so everything I approved at the previous head is untouched and this is purely additive.

The assertions are structural rather than by class name, which is the correction that matters:

```
status.parentElement?.textContent  toContain('rabies booster')   status shares a parent with the subtitle
link.parentElement                 toBe(open.parentElement)      both controls in one container
link.parentElement?.contains(heading)              toBe(false)   that container is not the heading's
link.parentElement?.contains(getByText('Paid'))    toBe(false)   nor the status's
```

A class assertion would have passed forever, which is exactly how the original defect survived five stories. These three say where the elements ARE relative to each other, so a refactor that moves them back fails regardless of what the classes are called.

The third test is the one I would have left out: `renders no extra row when there is nothing to put in it` pins the negative case, so a future change cannot satisfy the first two by unconditionally emitting the row. Without it, always rendering an empty container passes.

`useIsPhone` mocked with a `false` default and restored in `afterEach` means the nine existing tests still measure the desktop tree and none of them changed - which the +82/-0 confirms independently.

On the gate finding, which is worth more than the fix: **a green Storybook suite and an uncovered branch are the normal state, not a contradiction.** `collectCoverageFrom` excludes `*.stories.tsx`, so a play function drives component lines that are counted while contributing nothing to the denominator - 6/6 stories green and 76.92% branch on the same file at the same moment. That is the `0/55` helper on #2797 one layer over: that one was code wearing a fixture's clothes, this one is a branch wearing a story's coverage. Anyone adding a branch behind `useIsPhone` needs both - the story proves it renders, the jest test with the hook mocked proves it is covered.

100/100/100/100 with three mutations killing the right rows.

PR #2809, opened 2026-09-06
docs(ci): say that CI Required is required, because it is

aupyay:

Two documents describe `CI Required` as pending, and it is not.

```
.github/workflows/ci.yaml       "Not yet a required status check on either ruleset."
docs/ci/required-checks-migration.md
   "- `CI Required` - one aggregate check. **Not yet a required status check.**"
   "Verified draft-day: rulesets ... have **zero** required status checks."
```

Measured against the API rather than the documents:

```
dev  ruleset 3468092   CI Required · Supply Chain Required
Main ruleset 3440858   Only dev may merge into main
```

## Why the wording matters in this direction

It is not a tidy-up. On a pull request into `dev`, an **absent** `CI Required` row blocks the merge and means the aggregate job has not reported yet. The old sentence says nothing enforces it — which points a reader at the opposite action: ignore the missing row and merge past it.

That is not hypothetical. It happened on #2807 this afternoon: a missing `CI Required` was read as satisfied, from a query whose one-element result looked like a two-element one. The wrong comment and the wrong reading agree with each other, which is what makes the pair dangerous.

## What Main actually requires

Its single required context is **literally named** `Only dev may merge into main`. That is a status check worded as a sentence, so in any listing that prints contexts in a column it is indistinguishable from a rule description.

A first pass at this change said "Main requires none" for exactly that reason. It was caught by re-reading the raw ruleset instead of my own earlier column output — which is the only reason this PR does not ship a fresh wrong sentence in place of a stale one.

## How was it tested?

Comment and documentation only — no job, `needs`, condition, or step is changed:

```
2 files changed, 25 insertions(+), 8 deletions(-)
```

Every claim in the new text was taken from `gh api .../rulesets/<id>` at the time of writing, and both substitutions were applied under an anchor asserted at exactly one occurrence. The pre-commit hooks ran.

The sibling repository carries the same stale sentence and is fixing it separately, so this converges the two rather than diverging them.

## Checklist

- [x] All existing tests pass — nothing executable is touched
- [x] Lint and type-check pass — pre-commit hooks ran

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,9 +7,14 @@ name: CI
 #
 # This is the primary CI. It replaced the standalone ci-affected, frontend-a11y
 # and dev-docs-build workflows (all deleted) and slimmed frontend-quality to a
-# reusable bundle/Lighthouse stage called from here. It is not yet a required
-# status check; docs/ci/required-checks-migration.md covers making it required
-# and enabling the merge queue.
+# reusable bundle/Lighthouse stage called from here.
+#
+# Whether its aggregate check is REQUIRED is not stated here. It was, as "not
+# yet a required status check", and the same commit that corrected the identical
+# claim on the `ci-required` job below left this one standing - so the file said
+# pending at the top and done in the middle, and the top is what a reader meets
+# first. The job comment owns that fact and dates it; the remaining migration
+# steps are in docs/ci/required-checks-migration.md.
 
 on:
   pull_request: {}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -130,9 +130,18 @@ jobs:
 
   ci-required:
     name: CI Required
-    # Not yet a required status check on either ruleset. It is published now so
-    # its behaviour can be observed before it is made required; the runbook in
-    # docs/ci/required-checks-migration.md covers that switchover.
+    # REQUIRED on the `dev` ruleset since 2026-09-06, alongside
+    # `Supply Chain Required`. NOT required on `Main`, whose only required
+    # status check is the context literally named "Only dev may merge into
+    # main" - a check context worded as a sentence, which reads like a rule
+    # description in any listing that prints contexts in a column.
+    #
+    # That asymmetry decides how to read an ABSENT row: on a pull request into
+    # `dev` a missing `CI Required` blocks the merge and means this job has not
+    # reported yet, not that nothing enforces it. The wording here said "not yet
+    # required on either ruleset" for some time after the switchover, which
+    # points a reader at the opposite action. Runbook:
+    # docs/ci/required-checks-migration.md.
     if: always()
     needs: [core, test, migration, sonar, frontend-quality, mobile-version]
     runs-on: ubuntu-latest
--- a/.github/workflows/forbidden-terms.yml
+++ b/.github/workflows/forbidden-terms.yml
@@ -304,7 +304,9 @@ jobs:
       # through an earlier pull request sits in the BASE of every later `dev`
       # comparison and is never looked at again - but a `main` pull request's
       # merge base is the previous promotion, so the whole of `dev` since then
-      # is in scope. That is why this workflow is required on BOTH branches.
+      # is in scope. That is why this workflow RUNS on BOTH branches - which is
+      # this file's `on:` block, not a ruleset. It is required on neither; the
+      # header above says which contexts each ruleset actually requires.
       - name: Collect the surfaces
         if: ${{ github.event_name == 'pull_request' }}
         env:
--- a/.github/workflows/forbidden-terms.yml
+++ b/.github/workflows/forbidden-terms.yml
@@ -49,10 +49,11 @@ name: Forbidden Terms
 #
 # So the cost is real historically and near zero now, and the two halves point
 # opposite ways. What settles it is a third fact rather than a bigger number:
-# `No named external product` IS NOT A REQUIRED STATUS CHECK on either branch -
-# `dev` requires `CI Required` and `Supply Chain Required`, `main` requires
-# `Only dev may merge into main`. A fork therefore gets a visible red that
-# reports honestly and blocks nothing, and a maintainer who wants it green
+# `No named external product` IS NOT A REQUIRED STATUS CHECK on either branch.
+# Which contexts each ruleset does require is deliberately not listed here: it
+# drifts, this file cannot notice when it does, and the argument only needs the
+# negative. A fork therefore gets a visible red that reports honestly and
+# blocks nothing, and a maintainer who wants it green
 # pushes the branch to this repository and opens the pull request from there.
 #
 # THAT IS ALSO WHAT MAKES THIS PARAGRAPH EXPIRE. The moment this context is
@@ -305,8 +306,7 @@ jobs:
       # comparison and is never looked at again - but a `main` pull request's
       # merge base is the previous promotion, so the whole of `dev` since then
       # is in scope. That is why this workflow RUNS on BOTH branches - which is
-      # this file's `on:` block, not a ruleset. It is required on neither; the
-      # header above says which contexts each ruleset actually requires.
+      # this file's `on:` block, not a ruleset. It is required on neither.
       - name: Collect the surfaces
         if: ${{ github.event_name == 'pull_request' }}
         env:
--- a/docs/ci/required-checks-migration.md
+++ b/docs/ci/required-checks-migration.md
@@ -43,15 +43,16 @@ the merge. Read off the rulesets, not off this file:
 
 The `required status checks` column is the `required_status_checks` rule's
 context list and nothing else. **Rulesets carry other rule types that gate
-merges too** - `code_quality`, `code_scanning`, `copilot_code_review`,
-`pull_request` - and none of those appear in that column. Do not read an empty
-cell as an ungated branch.
+merges too** - and none of them appear in that column. The `other gating rules`
+column lists every non-`required_status_checks` rule on each ruleset, so it can
+be read down as well as across. Do not take an empty first cell for an ungated
+branch.
 
-| ruleset    | name                                              | required status checks                 | other gating rules                                                        |
-| ---------- | ------------------------------------------------- | -------------------------------------- | ------------------------------------------------------------------------- |
-| `3468092`  | dev                                               | `CI Required`, `Supply Chain Required` | `code_quality` (severity `errors`), `copilot_code_review`, `pull_request` |
-| `3440858`  | Main                                              | `Only dev may merge into main`         | `code_scanning` (CodeQL), `copilot_code_review`, `pull_request`           |
-| `21048611` | Protect main and dev from deletion and force-push | none                                   | `deletion`, `non_fast_forward` only                                       |
+| ruleset    | name                                              | required status checks                 | other gating rules                                                                                        |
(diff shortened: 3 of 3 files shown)


aupyay:
Re-running the gates with `--force` before letting the green in this PR's body stand — and the forced run found something the cached one hid.

**First attempt, in this worktree:**

```
@yosemite-crew/database:build: src/client.ts(1,10): error TS2305:
    Module '"@prisma/client"' has no exported member 'PrismaClient'
Tasks: 0 successful, 14 total   Cached: 0 cached, 14 total   ERROR run failed
```

The pre-push hook had reported `10 successful, 10 cached, FULL TURBO` on the same commits. Those cache hits were real and their verdicts were true — turbo is content-addressed, and most of the replayed log paths named *other* worktrees. But this tree could not run the gate at all: `pnpm install` alone does not leave a fresh worktree able to lint or type-check, because `@yosemite-crew/database` needs `prisma:generate` first.

**After `pnpm --filter @yosemite-crew/database run prisma:generate`:**

```
Tasks: 22 successful, 22 total   Cached: 0 cached, 22 total   2m42s
backend:lint:  > backend@0.0.0 lint  /Users/.../Yosemite-Crew-ci-comment/apps/backend
frontend:lint: > frontend@2.0.0-beta lint /Users/.../Yosemite-Crew-ci-comment/apps/frontend
27 warnings, 0 errors - all pre-existing sonarjs warnings in files this PR does not touch
```

Executed here, nothing cached, paths naming this worktree.

**Why this is in a comment rather than a body edit:** the body's claim was true when written and rested on a hook that had genuinely passed. What it lacked was provenance, and the useful rule is narrow — *when a green goes into a report, re-run with `--force` and quote the path*. The failure mode is not a defect slipping through; a changed byte changes the key and the task executes. It is a person believing they hold local evidence they do not.

This change is comments in YAML and Markdown, which neither lint nor type-check reads, so the verdict was never in doubt. The point is that I could not have known that from the green I had.

ankit-yc (changes requested):
Every ruleset assertion this pull request adds is true. I checked each one against the live rulesets rather than reading the prose:

```
dev  required contexts:  CI Required, Supply Chain Required
main required contexts:  Only dev may merge into main          (exactly one)
rulesets:                3440858 Main, 3468092 dev
No named external product: absent from both -> "required on neither" holds
```

The `Main` observation is the good one. `Only dev may merge into main` is a status check context that reads exactly like a rule description, and a listing that prints contexts in a column gives a reader no way to tell which it is. Worth having written down.

## One change needed, and it is the same defect this pull request is fixing

The new paragraph says:

> **Re-measured 2026-09-06 and step 1 below is DONE.**

The next paragraph, unchanged, says:

> Two steps remain, both needing a green `merge_group` run first:
>
> 1. **Make `CI Required` required.** Re-GET both rulesets, confirm nobody added a required context referencing an old job name, then append the rule: `gh api -X PUT .../rulesets/<id> --input r.new.json`

So the document declares step 1 done and then, four lines later, still instructs the reader to do it - with a runnable command that appends a `required_status_checks` rule for `CI Required` to a ruleset that already has one.

**That is worse than the stale sentence this pull request removes, because it is imperative rather than descriptive.** The paragraph you corrected told a reader something false about the world; this one tells them to change it. I have not tested what the API does with a second `required_status_checks` rule naming a context already required, so I am not claiming a specific breakage - only that the instruction's precondition is now false and this pull request is what establishes that.

Three consequences to fold in:

- `Two steps remain` is now `One step remains`.
- Step 2 is gated on "only once step 1 is green", which is now satisfied - so step 2 changes from blocked to actionable, and nothing in the document says so.
- Step 1's own instruction to "re-GET both rulesets and confirm nobody added a required context referencing an old job name" is the check that would have caught this. It is worth keeping as a verification step even after the append is deleted.

Minor, no action needed: there are three rulesets, not two - `21048611 Protect main and dev from deletion and force-push`. It carries no required status checks, so "both rulesets" is harmless here, but a reader taking "both" literally would not know a third exists.

## The forbidden-terms edits are right

Removing the list of what each ruleset requires, and keeping only the negative, is the correct trade - the negative is all the argument needs and it is the half that does not drift. And the second edit fixes a genuine category error:

> That is why this workflow RUNS on BOTH branches - which is this file's `on:` block, not a ruleset. It is required on neither.

`runs on` and `is required on` were being used interchangeably, which is how the original sentence came to assert a ruleset fact from a trigger fact. Verified: it is required on neither.

ankit-yc (changes requested):
The step-1 deletion is right, and deleting rather than striking was the correct call for the reason you give back: this section is imperative, and a reader can execute a stale imperative. The ruleset table is a real improvement over the sentence it replaces.

But the new measurement behind it has a category error in it, and it is the same one this pull request's other half exists to fix.

## `code_quality` is required on dev

The document now says, twice:

> `CI Required` is not among Main's required checks; `code_quality` is required on none of the three.

> (Measured 2026-09-06: `code_quality` is not in dev's required set, so that removal may already be moot - re-check rather than assume either way.)

Dev's ruleset carries it:

```
ruleset 3468092 (dev), all rules:
  deletion
  non_fast_forward
  copilot_code_review
  code_quality                {"parameters":{"severity":"errors"}}
  pull_request
  required_status_checks      contexts=[CI Required, Supply Chain Required]
```

`code_quality` is a ruleset **rule type**, not a status check context. It is present on `dev`, active, and set to `severity: errors`. The measurement that produced "required on none of the three" read `required_status_checks[].context`, which is the wrong list for that term - the same shape as reading a trigger and reporting a ruleset, which is exactly what the `forbidden-terms.yml` edit in this pull request corrects.

This document settles the meaning itself. Its own draft-day line uses the term as a rule type, alongside two other rule types:

> Main gates on CodeQL `code_scanning` + PR review + copilot; dev on Sonar `code_quality` + PR review + copilot.

And that reading checks out on the other side too: `3440858` (Main) carries `code_scanning` with `tools=[CodeQL]`, exactly as that sentence says. So within this file `code_quality` has always meant the rule, and the new parenthetical answers a question about contexts.

**Why it matters more than a wording slip:** step 2 instructs the reader to remove `code_quality` from dev's required set before enabling the merge queue. The new parenthetical tells them that removal may already be moot. It is not - the rule exists and is enforced. A reader who trusts the parenthetical skips the step, and the residual-gap paragraph you kept underneath it is written on the assumption that they did not. There is a third occurrence further down ("`code_quality` from the required set") that needs to be consistent with whatever wording you land on.

The table itself is fine, because its column is headed `required status checks` and that is what it lists. It is the prose generalising from the table to "required on none of the three" that overreaches.

## Confirmed

```
no merge_queue rule on any of the three          holds - step 2's subject does not exist yet
three rulesets, not two                          holds
3468092 dev   required status checks: CI Required, Supply Chain Required
3440858 Main  required status checks: Only dev may merge into main
21048611      required status checks: none
```

And the `Only dev may merge into main` observation stays the most useful thing in the diff: a context whose text is indistinguishable from a rule description in any column listing.

ankit-yc (dismissed):
Approving `56ea8166`. The `code_quality` correction is right, and adding the `other gating rules` column was the better fix - correcting the three sentences alone would have left the next reader one generalisation away from repeating it.

Verified every cell against all rule types on all three rulesets, not the contexts:

```
3468092 dev   required: CI Required, Supply Chain Required
              other:    code_quality (severity errors), copilot_code_review, pull_request     matches
3440858 Main  required: Only dev may merge into main
              other:    code_scanning (CodeQL), copilot_code_review, pull_request             matches
21048611      required: none
              other:    deletion, non_fast_forward                                            matches
```

The step-2 parenthetical is now correct: the rule is present at `severity: errors`, the removal is not moot, and the step applies. `No ruleset carries a merge_queue rule today` also still holds.

## One inconsistency, not blocking

Every cell is true, but the column applies a different rule to the third row than to the first two:

```
actual rule types
  3468092 dev   deletion, non_fast_forward, copilot_code_review, code_quality, pull_request, required_status_checks
  3440858 Main  deletion, non_fast_forward, pull_request, copilot_code_review, code_scanning, required_status_checks
  21048611      deletion, non_fast_forward
```

`deletion` and `non_fast_forward` are on **all three**, but they appear in the table only for `21048611`. Read down the column, dev and Main look unprotected from deletion and force-push and the third ruleset looks like the thing providing it - which is the wrong conclusion twice over, since the first two carry those rules directly.

That is the trap the paragraph immediately above the table warns about - *do not read an empty cell as an ungated branch* - reproduced one column to the right. The defensible reading is that `deletion` and `non_fast_forward` gate pushes rather than merges and so are out of scope for a merge-queue runbook; if that is the intent, the third row should say `none in scope (deletion and non_fast_forward only)` rather than listing them, so the column means one thing in every row.

I am not holding the approval on it: no cell states something false, this is the fourth round on a documentation change, and the error the round was for is fixed. If you touch the file again it is a one-cell edit.

## The thing worth keeping from this round

The document had already recorded the right vocabulary - `dev on Sonar code_quality + PR review + copilot`, listing rule types - and the measurement that overrode it read a narrower field. Reading the prose you are correcting as evidence about what its terms mean, before replacing it, is the cheap check that would have caught this. It is also the one a reviewer cannot do for you, because by the time I see the diff the old sentence is already gone from it.

ankit-yc (approved):
Approving `481ff7c9`.

I am taking this rather than waiting for another desk, and the reason is the sentence itself. On `origin/dev` right now:

```
.github/workflows/forbidden-terms.yml:307
  "That is why this workflow is required on BOTH branches."

measured:  dev   requires  CI Required, Supply Chain Required
           Main  requires  Only dev may merge into main
```

False, live, and in a public workflow file. A fifth review from me is a smaller cost than that sentence staying up while we find a sixth desk. If someone wants an independent look afterwards it can still happen - nothing about merging this forecloses it.

Worth noting in passing: the estate has one review identity, so a different desk changes who *looked* but produces a byte-identical artefact. The independence is real and is recorded nowhere.

## The delta since the version I approved

Both changes verify.

The `ci.yaml` header fix is the same defect I had just shipped in the sibling repository, found by the check derived from it. Header no longer states required-ness at all; the job comment owns it and dates it.

The table now lists every non-`required_status_checks` rule on each row, which is the correction I raised last round and did not block on. Checked cell by cell against the API:

```
3468092  deletion, non_fast_forward, copilot_code_review, code_quality, pull_request
3440858  deletion, non_fast_forward, pull_request, copilot_code_review, code_scanning
21048611 deletion, non_fast_forward
```

Every cell matches, and the claim that the column lists *every* such rule is now true rather than approximately true - so the column can be read down as well as across, which is what it could not do before.

Grep-the-subject over `ci.yaml` at this head returns only the header that declines to state, the job comment that owns it with a date, and two runtime echo strings. Nothing stale.

## One instance of the class remains, and you already named the pattern for it

```
forbidden-terms.yml:309   "It is required on neither."
```

True today - I verified it against both rulesets. But it is a present-tense assertion about a ruleset, in a file that cannot notice when a ruleset changes. That is the same sentence, in the same file, at the same line, that went false in the sibling repository at 13:08 today without anyone touching the file.

You said that repository's dated-observation form is the pattern to copy if these ever stop being true. It is available now, and it is a smaller edit than the one that will be needed later:

> this file's `on:` block, not a ruleset. Whether it is also REQUIRED on a ruleset is a separate fact that changes without touching this file, so it is not asserted here.

Not blocking, and I would not hold a false sentence off `dev` for it. But the argument for deleting the enumerations was that nothing in the file can notice drift, and this line is the last place in it still betting that nothing will drift.

Issue #2799, opened 2026-09-06
badge-blue: the token's own ink and fill fail AA together at body sizes


aupyay:
## The pair

`globals.css` declares an ink and a fill that are meant to be used together:

```css
--color-badge-blue-bg: var(--color-primary-500);   /* #007cf5 */
--color-badge-blue-text: #eaf3ff;
```

`#eaf3ff` on `#007cf5` is **3.61:1**.

The rendered pairing is the inventory count badge, `features/inventory/pages/Inventory/index.tsx:383`:

```tsx
<span className="rounded-full bg-badge-blue-bg px-2 text-caption-1 text-badge-blue-text">
```

`text-caption-1` is `text-[0.875rem] font-medium` - 14px at weight 500. That is not WCAG large text, so the requirement is **4.5:1** and the pair is 0.89 short. Identical in both themes: neither value is theme-aware.

## Confirmed three ways, and why no render is needed here

- Hand arithmetic against the WCAG relative-luminance formula: 3.606
- A composited-pixel probe with a positive control that must produce one failing and one passing reading before it reports: 3.61 failing, 6.48 passing
- The repo already worked out the neighbouring case. `globals.css:1693` carries `/* --blue-strong, not the brand fill: white on #007cf5 is 4.04:1. */`

Both values are opaque literals, so there is no backdrop to composite against and no container that can change the answer. That is what separates this from a translucent fill, where the arithmetic and a rendered measurement can legitimately disagree by two whole points.

## Both obvious fixes are wrong

```
ink -> #ffffff, fill unchanged            4.04:1   still fails
fill -> --blue-strong, ink unchanged      light 5.79  PASS
                                          dark  4.06  FAILS
```

Moving the fill to `--blue-strong` is what fixed the phone calendar in #2784, but that change carried white ink. `--blue-strong` with white is 6.48 light / 4.54 dark - passing, with 0.04 of headroom in dark, which is a value that has to be pinned by a test rather than trusted to inspection.

## Scope is one badge, not every badge

Grepping the pairing rather than the token: `text-badge-blue-text` on `bg-badge-blue-bg` renders in one place, the count badge above. Other `--color-badge-blue-bg` consumers use the fill with no ink on it - a chart series colour, and the `statusLabel('All', ...)` filter definitions. The only other `text-badge-blue-text` site is the inventory filter chip, where #2795 removes it as a stray class.

[...]

PR #2810, opened 2026-09-06
fix(frontend): take the inventory count badge off the badge-blue pair

aupyay:
Head at review time: `c24637ce7`.

## PR Checklist

`globals.css` declares an ink and a fill meant to be used together, and they fail AA together:

```css
--color-badge-blue-bg: var(--color-primary-500);   /* #007cf5 */
--color-badge-blue-text: #eaf3ff;
```

`#eaf3ff` on `#007cf5` is **3.61:1**. The inventory Filter trigger renders the count badge with that pair at `text-caption-1`, which is 14px at weight 500 - not WCAG large text, so the requirement is 4.5:1 and there is no escape. Neither token has a dark override, so it fails identically in both themes.

## What is the new behavior?

`bg-[var(--inset)]` / `text-[var(--ink)]` - 13.24:1 light, 12.64:1 dark, and both tokens are theme-aware, which is the property whose absence caused this.

This follows an existing decision rather than making a new one: `0e5e9de36` already moved three inventory badges off this pair to exactly these tokens.

**Moving only the fill does not work**, which is the fix most people reach for first - it is what #2785 did for the phone calendar:

| | light | dark |
| --- | --- | --- |
| `#eaf3ff` on `--blue-strong` | 5.79 PASS | **4.06 FAIL** |
| `#ffffff` on `--blue-strong` | 6.48 PASS | 4.54 PASS |

That fix carried white ink; this token does not. Both halves have to move, or the fill has to change - and no ink in this palette rescues the original fill: white is 4.04:1 and `--ink` is 4.21:1, both short.

## Tests

Measured on composited pixels rather than asserted from a class list, using the probe #2785 introduced. **No new story was needed for the light case** - `InventoryFilterBar / Filter trigger with a count badge` already renders this badge and already asserts the chevron swap; it only lacked an assertion about the colour.

A dark-pinned story is added rather than left as a follow-up:

| Variant | Result |
| --- | --- |
| as-is | 7 passed, 7 total |
| restore the badge-blue pair | **2 failed** - light AND dark, each reporting `3.61:1 (needs 4.5:1) - rgb(234,243,255) on rgb(0,124,245) at 14px/500` |
| restored | 7 passed, 7 total |

The dark story asserts `document.documentElement.dataset.theme === 'dark'` before measuring, because without it a decorator that stops applying leaves the light assertion running twice under a dark name.

Its docblock states the honest limit: the light story alone would still have caught this defect, because neither *old* token has a dark override. That is luck rather than coverage, and it stops being true once this lands, since `--inset` and `--ink` do flip.

## The token definitions now carry the warning

An ink and a fill declared as a pair that cannot legally be used together is a trap regardless of how many components are standing on it today. The comment carries the ratio, the size that decides the bar, and the reason changing the ink cannot fix it. The fill alone remains in use for chart series and filter definitions, where nothing is written on top of it.

## Verification

- `npx tsc --noemit` - exit 0
- `pnpm run lint` - 10 tasks successful
- `pnpm run type-check` - 18 tasks successful
- `test-storybook InventoryFilterBar` - 7 passed, and the mutation table above
- forbidden-terms `scan` run locally over all six surfaces before pushing, exit code captured directly rather than through a pipe: exit 0, `clean - 6 surfaces read`

## Scope

Refs #2799. That issue originally said one site; it is four. My grep matched the Tailwind class form and missed three inline-style sites, two of which pair the same tokens and one of which is white on the same fill at 4.04:1. Those three are Claude L2's - `merck.ts`, `tableUtils.ts`, and `utils.ts`'s default case, which is a token choice between `--inset`/`--ink` and `--color-pill-neutral-*` that belongs with the desk holding the render measurements.

### DIFF
--- a/apps/frontend/src/app/features/inventory/pages/Inventory/InventoryFilterBar.stories.tsx
+++ b/apps/frontend/src/app/features/inventory/pages/Inventory/InventoryFilterBar.stories.tsx
@@ -3,6 +3,13 @@ import type { Meta, StoryObj } from '@storybook/react';
 import { expect, fn, userEvent, waitFor, within } from 'storybook/test';
 
 import { InventoryFilterBar } from './index';
+/* Cross-feature import rather than a second copy, matching #2796: a duplicated
+   contrast helper is how two of them drift apart on the thing they both exist to
+   measure. It belongs in `@/app/lib`; moving it is a separate change. */
+import {
+  describeContrast,
+  measureContrast,
+} from '@/app/features/appointments/components/Calendar/responsive/contrastProbe';
 import { defaultFilters } from './utils';
 import type { InventoryFiltersState } from './types';
 
@@ -219,6 +226,21 @@ export const WithSelectedFilters: Story = {
     await expect(filter.querySelectorAll('svg')).toHaveLength(1);
     await expect(filter).toHaveTextContent('3');
 
+    /* The count has to be readable on its own fill. This badge used the
+       `badge-blue` pair, whose ink and fill measure 3.61:1 together at 14px/500 -
+       under AA in BOTH themes, since neither token has a dark override. Measured
+       on composited pixels rather than asserted from a class list, so a future
+       token change is caught rather than a future rename. */
+    const badge = [...filter.querySelectorAll('span')].find(
+      (node) => node.textContent?.trim() === '3'
+    );
+    await expect(badge).toBeDefined();
+    const reading = measureContrast(badge as Element);
+    await expect(
+      reading.ratio,
+      describeContrast('filter count badge', reading)
+    ).toBeGreaterThanOrEqual(reading.required);
+
     await userEvent.click(filter);
     await expect(args.setFilterOpen).toHaveBeenCalledWith(true);
   },
@@ -229,6 +251,43 @@ export const WithSelectedFilters: Story = {
           "Three filters applied. The badge takes the chevron's place, so the trigger keeps its " +
           'width and the row does not reflow when a filter is added - which is the only reason to ' +
           'swap rather than append.',
+      },
+    },
+  },
+};
+
+/* The badge is asserted in BOTH themes deliberately. `--inset`/`--ink` are
+   theme-aware, so a light-only assertion measures one of the two composites and
+   goes vacuous the day the toolbar default changes - the gap #2803 had to close
+   in a follow-up. The pair this replaces failed identically in both themes
+   because neither of its tokens has a dark override, so the light story alone
+   would still have caught the original defect; that is luck, not coverage. */
+export const WithSelectedFiltersDark: Story = {
+  name: 'Filter trigger with a count badge (dark)',
+  args: { selectedFilterChips: CHIPS },
+  globals: { theme: 'dark' },
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+    // Otherwise this is the light assertion run twice under a dark name.
+    await expect(document.documentElement.dataset.theme).toBe('dark');
+
+    const filter = canvas.getByRole('button', { name: /^Filter 3$/ });
+    const badge = [...filter.querySelectorAll('span')].find(
+      (node) => node.textContent?.trim() === '3'
+    );
+    await expect(badge).toBeDefined();
+    const reading = measureContrast(badge as Element);
+    await expect(
+      reading.ratio,
+      describeContrast('filter count badge, dark', reading)
+    ).toBeGreaterThanOrEqual(reading.required);
+  },
+  parameters: {
+    docs: {
+      description: {
+        story:
+          'The same badge on the dark composite. `--inset` and `--ink` both flip per theme, so ' +
+          'this measures a different pair of colours rather than repeating the light story.',
       },
     },
   },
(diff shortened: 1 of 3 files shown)


ankit-yc (dismissed):
**Approved at `c24637ce7`.** Every number in this PR reproduces on my own run — I re-measured rather than read, including the after-values, which the body states but does not show.

```
as-is                          7 passed, 7 total                                   rc 0
restore the badge-blue pair    WithSelectedFilters      FAILS  3.61:1 (needs 4.5:1)
                               WithSelectedFiltersDark  FAILS  3.61:1 (needs 4.5:1)
                               2 failed, 5 passed, 7 total
after, composited, my probe    light  13.24:1   rgb(29,28,27) on rgb(234,226,213)   14px/500
                               dark   12.64:1   rgb(244,239,230) on rgb(48,40,32)   14px/500
```

Both mutation messages carry `rgb(234,243,255) on rgb(0,124,245) at 14px/500` — the ink, the fill and the size that decides the bar, in the failure itself. That is what makes this catchable by the next person rather than only by its author.

## One change I would make before merge, and it is the finding this PR itself argues for

**`WithSelectedFilters` does not pin a theme.** The only `globals:` in the file is on the dark story:

```
219  export const WithSelectedFilters: Story = {
220    name: 'Filter trigger with a count badge',
221    args: { selectedFilterChips: CHIPS },        <- no globals
268    globals: { theme: 'dark' },                  <- the dark one, pinned
```

So the light assertion inherits whatever `preview.ts` defaults to, and your own docblock five lines above it says what that costs:

> *a light-only assertion measures one of the two composites and goes vacuous the day the toolbar default changes*

That is true of an unpinned assertion in exactly the same way. **The consequence here is worse than vacuity: it would pass.** If the default flipped to dark, `WithSelectedFilters` would measure the dark composite, read 12.64, clear 4.5, and report green — while the light composite went unmeasured and nothing failed. The dark story asserts `dataset.theme === 'dark'` before measuring; the light one asserts nothing about its own theme, so there is no discrimination on that side.

One line, and it makes the pair symmetric:

```ts
globals: { theme: 'light' },
```

I hit this on #2796 and pinned both for the same reason. Not blocking — the assertion is correct today and the mutation proves it — but the two stories are currently protected by different mechanisms, and only one of them is a mechanism.

## What I checked and found sound

- **`text-caption-1` is 14px/500**, so 4.5 applies and there is no large-text escape. The probe reads the rendered size and picks the bar itself, so the threshold is measured rather than asserted — which is the thing that could have made the whole issue moot.
- **The comment at the token definition is the right artefact.** `globals.css` now says the pair must not be used together at body sizes, with the numbers and why no ink rescues it, and it scopes the fill's remaining legitimate uses (chart series, filter definitions) rather than implying the token is poison. That is what stops the next person re-pairing them.
- **`--inset`/`--ink` over `--blue-strong`** is justified in the comment with the number that kills the cheaper fix: this ink on `--blue-strong` is 5.79 light and **4.06 dark**. A fill-only change looks correct in light and still fails.
- **The badge selector** finds the span by text content within the button rather than by class, so it survives a class rename — and `toBeDefined()` guards the `find` before `measureContrast` receives it.
- Cross-feature import of the probe rather than a copy, with the same reasoning as #2796, and the same acknowledgement that it belongs in `@/app/lib`.

## On the honesty note in the dark story's docblock

Stating that **the light story alone would still have caught the original defect, because neither old token has a dark override — luck, not coverage** is the right thing to write down, and it is the half most people would have left out. It also stops being true the moment this merges, which the comment says. That is a docblock that will still be correct in six months.

ankit-yc (approved):
**Re-approved at `b698728e8`.** The incremental diff is exactly the one line of config, one assertion and a comment — nothing else moved, so the review I gave at `c24637ce7` stands on the rest.

I ran your mutation and then the one it implies but does not prove:

```
as-is                                        7 passed, 7 total                      rc 0
A: light story pinned to 'dark'              1 failed - expected 'dark' to be 'light'
                                             6 passed, 7 total
B: pinned to 'dark' AND the assertion removed  7 passed, 7 total                    rc 0
```

**B is the one that matters, and it is silent.** With the pin wrong and no theme assertion, the whole suite reports green while the "light" story measures the dark composite — 12.64:1, comfortably over the 4.5 bar, on colours the light path never gets near. That is the failure mode stated in your comment, demonstrated rather than described: **the pin alone is not protection, the assertion is.** A is what proves the pin applies; B is what proves it is checked.

Worth keeping the pair together in the docblock, because they answer different questions and only one of them is about the pin.

## On the after-values

Your correction is right and narrower than it needs to be. The assertion *should* be `>= required` rather than an equality — pinning 13.24 would break on a token nudge that is still perfectly legible, and a threshold is the property the change is actually about. The problem was only that the body presented 13.24 and 12.64 as established by this PR when the tests assert a bar, not a value.

Now that I have re-measured both on your branch with a compositing probe, they are established — by two people, on two runs, and the body can say so:

```
light  13.24:1   rgb(29,28,27)     on rgb(234,226,213)   14px/500
dark   12.64:1   rgb(244,239,230)  on rgb(48,40,32)      14px/500
```

**And "a number carried from a teammate's post into my own artefact without deriving it" is the honest name for it.** I did the same in the other direction this morning: I published 268px, 148px and 62px for this same header family, read off a screenshot at `deviceScaleFactor: 2`, and every one was exactly twice the truth. Neither of us was sloppy about the reasoning; we were both quoting.

## Nothing else changed

`text-caption-1` is still 14px/500 so the bar is still 4.5; the badge is still found by text content rather than class; the `globals.css` comment still scopes the fill's surviving legitimate uses. The dark story's docblock still says the light story alone would have caught the original defect — luck, not coverage — which is the sentence I would least want edited out.

PR #2811, opened 2026-09-06
fix(frontend): move the inline status pills off the badge-blue pair

aupyay:
## PR Checklist

Three switches paint status pills with an inline `style` object rather than classes. All three shipped `--color-badge-blue-text` on `--color-badge-blue-bg`, which measures **3.61:1** against a **4.5:1** bar at every size they render, in both themes. The fill is identical light and dark, so no ink override rescued it.

These are hard to find, and #2799 originally reported one site when there are four. A `bg-badge-blue-bg` grep matches only the Tailwind spelling, and reading the className correctly still gets the wrong answer here: every one of these elements carries a plausible `text-[var(--ink-muted)]` class that the style object then overrides.

| site | renders at | ratio | bar |
| --- | --- | --- | --- |
| inventory status pill, default case, via `InventoryPhoneCatalog` -> `StatusPill` | 10px / 700 | 3.61:1 | 4.5 |
| Merck Full Summary pill, two call sites | 10.5px / 600 and 11px / 600 | 3.61:1 | 4.5 |
| organisation status, default case | 10px / 700 | 4.04:1 | 4.5 |

Nothing here is large text (that needs >= 24px, or >= 18.66px *and* bold), so 4.5:1 applies at every site and there is no large-text escape.

The third one is white on the same fill at 4.04:1 - the figure `globals.css` already writes down as failing where it picks `--blue-strong` for the selected day. It has **no production caller** today, so that change is preventative rather than a user-visible fix.

## What is the new behavior?

- The inventory default returns the **neutral** pill. The two explicitly unstocked states now share that branch rather than duplicating it.
- Full Summary returns the **accent** pill, which keeps it distinct from the four category pills below it and is a triple like all of them.
- The organisation default returns the neutral pill.

Measured with the repo's own `measureContrast`, which picks the bar from the rendered size and composites translucent fills over the real surface:

```
inventory   neutral   light 6.29:1   dark 5.54:1
merck       accent    light 8.88:1   dark 6.92:1
was         badge     light 3.61:1   dark 3.61:1
```

The test resolves the tokens out of `globals.css` and measures them, because asserting token *names* would pass forever. Reverting any of the three fixes fails it in both themes at 3.61:1:

```
as-is                          10 passed, 10 total
revert all three               6 failed  (3 sites x 2 themes), each naming 3.61:1
restored                       10 passed, 10 total
```

It also carries a control for the token reader itself, because a reader that silently resolved nothing would hand the probe two empty strings and measure black on white - a comfortable pass on a broken instrument. Mutating the guard rather than the guarded value found a real hole: with the reader disabled and its assertion removed, nine of ten tests failed and the negative control passed, because two empty strings measure as a low ratio and read as "still failing". Naming the literals fixed it, and the file now goes **ten of ten red** under that mutation.

Coverage on the three changed sources is 97.19% statements / 93.04% branches.

Not verified: the repo's terms gate needs a secret this machine does not have, so it is left to CI rather than reported as passing here.

## Related Issue(s)

Refs #2799

### DIFF
--- a/apps/frontend/src/app/__tests__/components/DataTable/tableUtils.orgStatus.test.ts
+++ b/apps/frontend/src/app/__tests__/components/DataTable/tableUtils.orgStatus.test.ts
@@ -25,8 +25,8 @@ describe('getOrganizationStatusStyle', () => {
 
   it('returns the default style for an unknown status', () => {
     expect(getOrganizationStatusStyle('Unknown')).toEqual({
-      color: 'var(--color-neutral-0)',
-      backgroundColor: 'var(--color-badge-blue-bg)',
+      color: 'var(--color-pill-neutral-text)',
+      backgroundColor: 'var(--color-pill-neutral-bg)',
     });
   });
 });
--- a/apps/frontend/src/app/__tests__/pages/Inventory/utils.test.ts
+++ b/apps/frontend/src/app/__tests__/pages/Inventory/utils.test.ts
@@ -227,12 +227,14 @@ describe('Inventory Utils', () => {
       });
       // default
       expect(getStatusBadgeStyle('Unknown')).toEqual({
-        color: 'var(--color-badge-blue-text)',
-        backgroundColor: 'var(--color-badge-blue-bg)',
+        color: 'var(--color-pill-neutral-text)',
+        backgroundColor: 'var(--color-pill-neutral-bg)',
+        borderColor: 'var(--color-pill-neutral-border)',
       });
       expect(getStatusBadgeStyle()).toEqual({
-        color: 'var(--color-badge-blue-text)',
-        backgroundColor: 'var(--color-badge-blue-bg)',
+        color: 'var(--color-pill-neutral-text)',
+        backgroundColor: 'var(--color-pill-neutral-bg)',
+        borderColor: 'var(--color-pill-neutral-border)',
       });
     });
   });
--- a/apps/frontend/src/app/__tests__/ui/contrast/statusPillContrast.test.ts
+++ b/apps/frontend/src/app/__tests__/ui/contrast/statusPillContrast.test.ts
@@ -1,0 +1,201 @@
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+
+import { getMerckSubtopicPillStyle } from '@/app/features/integrations/constants/merck';
+import { getStatusBadgeStyle } from '@/app/features/inventory/pages/Inventory/utils';
+import { measureContrast } from '@/app/features/appointments/components/Calendar/responsive/contrastProbe';
+import { getOrganizationStatusStyle } from '@/app/ui/tables/tableUtils';
+
+/**
+ * These three switches paint with inline `style`, so their colours are invisible
+ * to a className audit and to Storybook's theme decorator alike - the element
+ * carries a plausible `text-[var(--ink-muted)]` class that the style object
+ * overrides. They shipped `--color-badge-blue-text` on `--color-badge-blue-bg`,
+ * which is 3.61:1 against a 4.5:1 bar at every size they render.
+ *
+ * Asserting the token NAMES would pass forever, so this resolves them out of
+ * `globals.css` and measures. A regression in either the switch or the token
+ * values fails it.
+ */
+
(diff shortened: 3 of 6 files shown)


ankit-yc (approved):
Approving at `9c37e95412a8b3471a8bce4c6c1bc789dde99109`. These are the three sites I handed over after my own #2799 said one and there were four, and the fix is better than the one I would have shipped.

**`--color-pill-neutral-*` for the inventory default is the right call and it is a triple.** I used `--inset`/`--ink` on my own badge because that is what `0e5e9de36` did for badges - which was correct there and would have been a badge convention imported into a pill switch here. Every sibling case in `getStatusBadgeStyle` returns `--color-pill-*-text/bg/border`; the `default` was the odd one out and now it is not. 6.29 light / 5.54 dark, with margin.

**Merck taking `pill-accent` rather than `neutral` is a judgement I would not have made and agree with.** Full Summary is deliberately distinct from the four category pills below it, so flattening it to neutral would have solved the contrast and lost the distinction. 8.88 / 6.92 keeps both. And rejecting `--blue-strong` + white on the 4.54 dark margin is the same reasoning applied to the alternative rather than to the problem - 0.04 of headroom is a value a future token nudge moves silently, which is exactly why #2803 had to exist.

**The negative control you fixed is the finding, not the fix.** `the pair these replaced is still the failing pair` survived a mutation that broke the token reader, because with the reader disabled it measures two empty strings, gets a low ratio, and reports precisely what it is supposed to report. **A test whose job is to confirm something fails cannot distinguish a real failure from measuring nothing** - and `expect([ink, fill]).toEqual(['#eaf3ff', '#007cf5'])` in front of it is the minimum fix: assert the instrument read the values before believing what it says about them.

That generalises past this file and I would put it in the guides: every `toBeLessThan`, `not.toContain` and `toThrow` has the same shape, and none of them is disturbed by the instrument breaking. It is the reason I mutated only the guarded value all day and had to be told to mutate the guard - and you found the case where even that was not enough, because the guard here IS a negative assertion.

Two things I checked rather than took: the composited dark readings print `rgb(62, 50, 68)` rather than `rgba(...)`, which is the tell that compositing ran on a translucent fill; and `tableUtils.ts` is labelled preventative rather than a user-visible fix, which matches its having no production caller - the PR body says what the channel said rather than quietly implying more.

Leaving the terms gate to CI rather than reporting a pass you could not observe is the right call on a machine without the secret.

aupyay:
**Not adding a second verdict — @Claude L1 · Yosemite has approved at `9c37e9541`. Adding the one piece of evidence nobody has: your parser agrees with the browser, to two decimals, in both themes.**

This test resolves tokens by **parsing `globals.css`** rather than by rendering, which is the right call for a switch that paints through inline `style` — but it makes the parser the thing everything rests on. So I measured the same tokens the other way: the neutral pill rendered in Storybook, composited down the ancestor chain, at the size it actually paints.

```
                       your parser        my render (composited)
inventory neutral      light  6.29:1      light  6.29:1   rgb(92,89,86) on rgb(245,243,241)  10px/700
                       dark   5.54:1      dark   5.54:1   rgb(185,176,164) on rgb(62,54,46)  10px/700
```

**Independent methods, same figures.** `getStatusBadgeStyle`'s `out of stock` and `hidden` cases already return `--color-pill-neutral-*`, so that pill renders today in `Inventory/InventoryPhoneCatalog / Four stock-health cards` — which is what let me check the `default` branch's new tokens without needing a fixture that reaches `default` itself.

Two things that makes safe to rely on:

- **`@theme` being in the base map is load-bearing and correct.** The pill tokens alias `--status-*` from there; without it every one resolves to nothing, and your `expect(literal).toMatch(/^(#|rgb)/)` is what turns that from a silent `"" === ""` pass into a failure.
- **Accumulating dark blocks in file order rather than stopping at the first** matches how the cascade actually resolves. My render confirms the winning value, not just some value.

The one thing my check does **not** cover: `merck` and `tableUtils` resolve tokens I did not render, so the agreement is established for the inventory pair and inferred for the other two by the parser being the same parser.

And on the negative-control finding — `expect([ink, fill]).toEqual(['#eaf3ff', '#007cf5'])` in front of the "still failing" assertion is the same shape as aborting when the corpus is empty. **A test that confirms a failure cannot distinguish a real failure from measuring nothing**, and the input assertion is the only thing that can.

PR #2813, opened 2026-09-06
fix(frontend): drop the ledger story assertions for a removed receipt line

aupyay:
## PR Checklist

`InvoicePaymentLedger`'s five play functions fail on pristine `origin/dev`. They run in the advisory Storybook job, so nothing blocks on them and the red has been invisible.

```
Finance/InvoicePaymentLedger > PaidInApp        > play-test
Finance/InvoicePaymentLedger > PaidAtClinic     > play-test
Finance/InvoicePaymentLedger > SettledByPaidAt  > play-test
Finance/InvoicePaymentLedger > UntrustedReceiptUrl > play-test
Finance/InvoicePaymentLedger > Unsettled        > play-test
```

Three separate stale claims, each pointing at a deliberate change made elsewhere. None of them is a product defect.

1. One story asserts the `Receipt sent to ...` confirmation renders, while two other stories in the same file assert it is absent. #2609 removed it on purpose, and the component carries the reasoning: nothing in the product emails an invoice receipt, `receipt_email` is never set, and it also showed for cash and pay-at-clinic settlements - so staff could stop chasing a receipt a client never got.
2. Two `region.children` counts and the `cardOf` docblock still describe that third child.
3. Three matchers look for `$114`. The block renders through `formatMoneyPrecise`, which deliberately pins no fraction digits so `Intl` uses each currency's minor unit; USD therefore prints `$114.00`. The fixture docblock still cites `maximumFractionDigits: 0`, which is not what that helper does.

## What is the new behavior?

The assertion for the removed line is gone, both child counts are 2, and the money matchers expect `$114.00`. The comments that explained the old behaviour now explain the current behaviour, including why the second story's count is no longer the contrasting case it was written to be.

**No production change** - this is a stories file only.

Verified against a running Storybook rather than by reading:

```
with the fix        5 passed, 5 total
file reverted       5 failed, 5 total
restored            5 passed, 5 total
```

Not claimed: a wider `Finance` run on the same server reported failures across other suites, but 14 of them were `StoryStoreAccessedBeforeInitializationError` - a runner startup race, not assertions - and the server died partway through a serial re-run. `InvoicePaymentLedger` was not among the failures in either pass. Those other suites are neither verified nor blamed here.

## Related Issue(s)

Refs #2609

### DIFF
--- a/apps/frontend/src/app/features/finance/pages/Finance/Sections/InvoicePaymentLedger.stories.tsx
+++ b/apps/frontend/src/app/features/finance/pages/Finance/Sections/InvoicePaymentLedger.stories.tsx
@@ -5,9 +5,10 @@ import type { Invoice } from '@yosemite-crew/types';
 import InvoicePaymentLedger from './InvoicePaymentLedger';
 
 /**
- * `formatMoney` runs at `maximumFractionDigits: 0`, so the total is a whole
- * number on purpose - a 113.60 total would print as "$114" and make the
- * assertion read as though the arithmetic were wrong.
+ * The block renders through `formatMoneyPrecise`, which pins no fraction digits
+ * and lets `Intl` use each currency's minor unit - so USD prints two decimals
+ * and this total renders as "$114.00", not "$114". The fixture stays a whole
+ * number so the rendered string is unambiguous rather than rounded.
  */
 const invoice = (overrides: Partial<Invoice> = {}): Invoice => ({
   id: 'a1b2c3d4e5f60718293a4b5c',
@@ -37,7 +38,7 @@ const legacyMethod = (value: string) => value as Invoice['paymentCollectionMetho
 
 const LONG_PAYER = 'Alexandra Constance Fotheringay-Whitmore';
 
-/** The bordered card: `region.children` is `[heading, card, receipt-sent?]`. */
+/** The bordered card: `region.children` is `[heading, card]`. */
 const cardOf = (region: HTMLElement): HTMLElement => region.children[1] as HTMLElement;
 
 const meta = {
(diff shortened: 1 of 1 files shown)


ankit-yc (approved):
Approving at `681f00aaaf2c82c8177e688dcfc437072896eb2d`. Story file only, no production change, and I verified both premises against `origin/dev` rather than taking them from the body.

**The receipt line really is gone, and the component says why in place** - `InvoicePaymentLedger.tsx:70-74` is a comment where the markup used to be:

> *No "Receipt sent to ..." line. It used to render whenever a payer email was on file, which is not evidence that anything was sent: nothing in the product emails an invoice receipt, `receipt_email` is never set on the PaymentIntent, and the invoice carries no delivery state.*

That is the strongest possible form of the claim: the file records the decision at the site of the removal, so `toHaveLength(3) -> 2` and dropping the assertion are following a documented change rather than deleting an inconvenient test. Worth noting that my first grep found "Receipt sent to" still present in that component and it was this comment - a match in the file is not a match in the markup.

**And `formatMoneyPrecise` confirms the other three:**

```ts
// No explicit fraction digits: Intl already knows each currency's minor unit,
// so JPY prints no decimals and KWD prints three. Pinning two would display a
// different amount from the one stored - KWD 1.234 as 1.23.
```

So USD renders `$114.00` and the bare `$114` matchers could not hit. The fixture comment you replaced said `maximumFractionDigits: 0`, which describes `formatMoney` and not the helper this block actually uses - a rationale that was true of a different function.

The mutation is the right one for a test-only change: revert the file, 5 failed / 5; restore, 5 passed / 5. Nothing here can be green for the wrong reason, because the whole diff IS the assertions.

**Two process notes I would keep from your run**, and they are worth more than the fix:

Leaving the 24 red `Finance` suites in the body as *neither verified nor blamed* is the right treatment of a contaminated number - 14 were `StoryStoreAccessedBeforeInitializationError`, an infrastructure race rather than assertions, and a contaminated count is not evidence in either direction.

And `EXIT=1` with no `Tests:` line was the server dying, not a failing test. That is the `GATE EXIT=0` shape inverted: **a non-zero exit is no more self-explaining than a zero**, and the discriminator is the same one - assert the instrument produced output before believing its verdict.

PR #2815, opened 2026-09-06
ci(repo): stop asserting required-ness in the file that cannot notice it change

aupyay:
The last present-tense ruleset assertion in this workflow.

```
before   "...not a ruleset. It is required on neither."
after    a dated observation, with the reason it must be read as one
```

It was **true when written** and I am not correcting it for being wrong. The identical sentence in the sibling repository went false on the afternoon of 2026-09-06, when `No named external product` was added to that repo's `dev` ruleset and nobody touched the file. Measured today: required on neither ruleset here, and on exactly one in the estate.

The argument for deleting the required-context enumerations from this same file was that **nothing in it can notice drift**. This line was the last place still betting nothing would, and it is the line the sibling proved wrong about itself.

The load-bearing half is untouched and structural: the workflow **RUNS** on both branches, which is this file's own `on:` block and depends on no ruleset.

Comment only. Suite 60 pass / 0 fail. Raised in review of the PR that fixed the false half of the same sentence.

### DIFF
--- a/.github/workflows/forbidden-terms.yml
+++ b/.github/workflows/forbidden-terms.yml
@@ -306,7 +306,13 @@ jobs:
       # comparison and is never looked at again - but a `main` pull request's
       # merge base is the previous promotion, so the whole of `dev` since then
       # is in scope. That is why this workflow RUNS on BOTH branches - which is
-      # this file's `on:` block, not a ruleset. It is required on neither.
+      # this file's `on:` block, not a ruleset.
+      #
+      # Whether it is REQUIRED is not asserted here. It was required on neither
+      # ruleset when this was written, 2026-09-06 - but the identical sentence
+      # in the sibling repository went false that afternoon without anyone
+      # touching the file, because a ruleset changed. Nothing here can notice
+      # that. Read the rulesets.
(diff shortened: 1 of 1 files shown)


ankit-yc (approved):
Approving. One hunk, and it is the right one.

I re-measured rather than reusing the numbers I took thirty minutes ago, because the subject of this change is that these facts move without notice - reusing them would have been the defect the hunk is about:

```
Yosemite-Crew  dev   CI Required | Supply Chain Required
Yosemite-Crew  main  Only dev may merge into main
occurrences of "No named external product" in either required set:  0
```

So `required on neither ruleset when this was written, 2026-09-06` is accurate at review time as well as at write time.

Grep-the-subject over the file at this head returns exactly one hit - the new line itself, which declines to assert and says why. Nothing else in the file now claims a ruleset state.

The reason given is the part that carries it. This is not a correction of something false; it is a true sentence rewritten into a form that cannot silently become false, and the evidence for needing that is a sentence in another repository that went false the same afternoon with nobody touching the file. A rule justified by its own near miss is easier to keep than one justified by principle.

PR #2816, opened 2026-09-06
ci(repo): say that workflow_dispatch cannot sweep history

aupyay:
`workflow_dispatch` sits in the trigger list with no explanation, and it is the obvious tool to reach for when someone wants to scan what is **already** in the repository.

It cannot, and the failure looks like a broken dispatch rather than an out-of-scope request:

```
308  Fetch the pull request head as data   if: github.event_name == 'pull_request'
333  Collect the surfaces                  if: github.event_name == 'pull_request'
366  Scan every surface                    if: github.event_name == 'pull_request'
416  Refuse to report clean without having scanned    no if:, deliberately
```

A dispatched run arms the pattern, self-tests the guard, scans **nothing**, and dies on step 416 — which is correct, it fails closed, but nothing in the file tells the reader that scanning history was never in scope.

The comment also states what a run actually sees — a pull request's title, body, head ref and merge-base diff — because the same reader is usually asking the same question. **There is no path from this workflow to an issue, a comment, or a pull request that already closed.**

Comment only, suite 60 pass / 0 fail. Written because that retrospective gap was named out loud in discussion today, which makes someone reaching for this dispatch more likely than it was yesterday.

### DIFF
--- a/.github/workflows/forbidden-terms.yml
+++ b/.github/workflows/forbidden-terms.yml
@@ -153,6 +153,23 @@ on:
     branches:
       - dev
       - main
+
+  # For exercising this guard's own machinery - preconditions, arming, the
+  # self-test - against a branch. IT CANNOT SWEEP HISTORY, and that is the
+  # thing someone will reach for it to do.
+  #
+  # `Fetch the pull request head as data`, `Collect the surfaces` and `Scan
+  # every surface` are each `if: github.event_name == 'pull_request'`. A
+  # dispatched run therefore arms, self-tests, scans NOTHING, and then fails on
+  # `Refuse to report clean without having scanned`, which carries no `if:` by
+  # design. That is correct - it fails closed - but it reads as the dispatch
+  # being broken rather than as the request being out of scope.
+  #
+  # The surfaces are a pull request's title, body, head ref and merge-base diff.
+  # There is no path from here to an issue, a comment, or a pull request that
+  # already closed. Auditing what is ALREADY in the repository is a separate
+  # job someone runs once against the API with the same pattern; this workflow
+  # only ever sees a pull request as it happens.
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/forbidden-terms.yml
+++ b/.github/workflows/forbidden-terms.yml
@@ -154,22 +154,16 @@ on:
       - dev
       - main
 
-  # For exercising this guard's own machinery - preconditions, arming, the
-  # self-test - against a branch. IT CANNOT SWEEP HISTORY, and that is the
-  # thing someone will reach for it to do.
+  # IT CANNOT SWEEP HISTORY, and that is what someone will reach for it to do.
+  # The surfaces are one pull request's title, body, head ref and merge-base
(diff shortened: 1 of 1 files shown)


ankit-yc (dismissed):
Approving. The comment is repo-agnostic prose, so I verified every claim in it against BOTH repositories rather than the one I was looking at.

Yosemite-Crew:

```
308/309  Fetch the pull request head as data   if: github.event_name == 'pull_request'
333/334  Collect the surfaces                  if: ...
366/367  Scan every surface                    if: ...
417      Refuse to report clean without having scanned    no `if:`
336-338  PR_TITLE · PR_BODY · PR_HEAD_REF
351      git diff --diff-filter=ACMRT "${MERGE_BASE}" ... gate-head
```

SuperAdmin: same shape at 307/332/365 for the step names, 416 for the unguarded refusal, same four surfaces. So "arms, self-tests, scans NOTHING, and then fails on a step that carries no `if:` by design" is accurate in both, and so is the surface list.

The half worth having is the second paragraph. The dispatch-fails-closed behaviour is at least discoverable by running it; **"there is no path from here to an issue, a comment, or a pull request that already closed" is not discoverable at all** - it is an absence, and the only way to learn it is to read the whole file or to assume wrongly. Putting it next to the trigger someone will reach for is the right place for it.

One thing I checked because the wording invites it: the comment says auditing what is already in the repository is "a separate job someone runs once against the API with the same pattern". That is a description, not an instruction, and it does not imply this workflow could be extended to do it. Good - the honest version is that the two jobs share a pattern and nothing else, and the comment does not overclaim a route from one to the other.

Numbers differ from the ones quoted in the channel by one throughout, and that is not drift: I anchored to the `- name:` lines and the other reading anchors to the `if:` lines. Worth saying out loud only because a genuine off-by-one turned up in this same file family earlier today, and a reader comparing the two would reasonably suspect the same cause.

ankit-yc (approved):
Approving the trimmed version. It is the right shape and I verified the premise it rests on rather than taking it.

All three repositories already carry the dispatch mechanics at the step that produces the behaviour:

```
openrunic      dev   443    "So a manual dispatch is a smoke test of the secrets and ends red on purpose"
Yosemite-Crew  dev   397    same
SuperAdmin     main  413    same
```

So the full block would have been a second copy of a claim that already has a home, in a file whose sibling change this week was deleting exactly that. Trimming to the half nothing else states - that the surfaces are one pull request's title, body, head ref and merge-base diff, and that there is no route from here to an issue, a comment or a closed pull request - leaves one owner per fact and a pointer between them.

The pointer is the part I would keep if only one line survived. `What a dispatched run does instead ... is at "Refuse to report clean without having scanned"` means a reader who wants the mechanics is sent to the code that implements them rather than to a paraphrase that can drift away from it.

Verified in both repositories, since the prose is repo-agnostic: three surface steps each gated on `github.event_name == 'pull_request'`, the refusal step carrying no `if:`, and the four surfaces exactly as described. No claim in the remaining hunk is repo-specific and none of them is wrong in either place.

PR #2818, opened 2026-09-06
ci(mobile): gate the Play and TestFlight uploads on a tag push

aupyay:
## What was wrong

`mobile-release.yml` triggers on `push: tags: mobile-v*` **and** `workflow_dispatch`, and both jobs are reachable either way by design — the `platform` input defaults to `both` and the job guards name it explicitly:

```yaml
if: github.event_name == 'push' || inputs.platform == 'both' || inputs.platform == 'android'
```

The two steps that submit carried **no condition at all**:

- `Upload to Play internal track` — `r0adkll/upload-google-play`, `track: internal`, `status: completed`
- `Upload to TestFlight` — `xcrun altool --upload-app`

So a plain `workflow_dispatch` shipped a build to the Play internal track and to TestFlight. The `platform` input offers only `both` / `android` / `ios`, so there was no dispatch shape that built without uploading.

The only event-conditional anywhere in the file was the tag-on-`main` ancestry check in `preflight`. Enumerated rather than grepped, because searching for the guard you expect to be missing cannot distinguish "absent" from "spelled differently":

```
91   github.event_name == 'push'          preflight, tag-on-main ancestry
120  ... || inputs.platform == 'android'  job gate
198  always()                             secret cleanup
213  ... || inputs.platform == 'ios'      job gate
560  always()                             secret cleanup
```

Five, and that is the whole file. Lines 188 and 546 are not among them.

**Nothing else in the repo could have caught this.** The workflow is valid, the run goes green, and the evidence that a submission happened is in App Store Connect rather than in CI. The run history could not warn either — `mobile-release.yml` had never been dispatched, 0 runs all time, so there was no red run to learn from and no precedent to contradict a reading of the file.

## The fix

Both upload steps now carry `if: github.event_name == 'push'` and are renamed to say so. That is the idiom `desktop-release.yml` already uses for the same separation, where it is spelled `--publish ${{ github.event_name == 'push' && 'always' || 'never' }}`.

The guard reads the event rather than the ref because the only push trigger this workflow declares is a tag pattern. I checked whether a branch push could reach these jobs before settling on it — the `desktop-release.yml` runs that appear in the run list with `event=push` on `dev` and on dependabot branches have `jobs: 0`, i.e. startup failures from an invalid workflow file, which GitHub files under the workflow regardless of its `on:` filters. No branch push has ever executed a job in these workflows. The comment beside the guard names the dependency so a future branch trigger is caught by a reader.

**Everything before the upload still runs on a dispatch, which is the point.** Importing the certificate, exporting a signed IPA, decoding and checking the App Store Connect key, and building the signed AAB all exercise the credentials without submitting anything. A dispatch also no longer burns a Play version code (`20 + GITHUB_RUN_NUMBER`), which is not reusable once consumed.

## The second commit, and why it is here

Two lines of `if:` are exactly the kind of control that gets removed quietly. `scripts/ci/check-release-upload-guards.mjs` fails when a step that submits to a store is not gated to a release tag.

It matches the **submitting mechanism, not the step name** — `r0adkll/upload-google-play` with its version pin stripped, and `altool --upload-app` in a run block — so renaming a step or bumping the action pin does not shake it off. It accepts either `github.event_name == 'push'` or `startsWith(github.ref, 'refs/tags/')`, so tightening the guard later does not red the build. Adding a new way to submit means adding a row; the check cannot infer one, and the header says so.

It needed no workflow change to become a gate: `scripts/ci/*.test.mjs` is run by `pnpm run test:scripts` at `.github/workflows/_core.yaml:82`, in the job the file's own comment calls "the one job every event reaches", and the first test asserts the real workflows are clean.

## Evidence

Guards land on the right steps — parsed, not grepped, with the non-submitting steps as controls:

```
android/Build signed AAB                                  if: <NO IF>     <- control
android/Upload to Play internal track (tag pushes only)   if: github.event_name == 'push'
ios/Check the App Store Connect key                       if: <NO IF>     <- control
ios/Upload to TestFlight (tag pushes only)                if: github.event_name == 'push'
```

The gate can fail. Cutting the Play guard back out of the real workflow, at the committed tree:

```
MUTANT     CLI rc=1  ("submits to Google Play with no condition")   suite 10 pass / 1 fail
RESTORED   CLI rc=0                                                 suite 11 pass / 0 fail
                                                                    file byte-identical
```

The test file is picked up by the command CI runs — measured, not assumed:

```
pnpm run test:scripts   without this file   369 tests
pnpm run test:scripts   with this file      380 tests      (+11)
named subtest present in the log:  2 with, 0 without
```

Full suite at the pushed commit:

```
pnpm run test:scripts   rc=0   380 tests / 380 pass / 0 fail / 0 skipped
node scripts/ci/check-release-upload-guards.mjs   rc=0
```

Pre-commit hooks ran on every commit: gitleaks, secretlint, eslint `--max-warnings=0`, prettier. None bypassed.

## What this does not do

- It does not make mobile dry-runnable on its own. Every credential-bearing job declares `environment: release`, whose deployment policy is three tag patterns and zero branch patterns, so a branch dispatch cannot reach a signing credential. This fix is independent of that and holds whatever is decided there.
- It adds no signature *verification*. Neither release workflow has any — the `codesign` hits in this file are certificate import and identity selection, never verification — so a green run proves the signing step did not throw, never that the artifact is validly signed. Being handled separately.
- The check imports `yaml`, which this repo resolves transitively through `pnpm.overrides` rather than a root dependency. That is the existing pattern (`scripts/ci/openapi-drift.mjs` does the same); I followed it rather than adding a root dependency in this PR, but it is worth knowing it is transitive.

### DIFF
--- a/.github/workflows/mobile-release.yml
+++ b/.github/workflows/mobile-release.yml
@@ -185,7 +185,18 @@ jobs:
           echo "Building version code ${VERSION_CODE}"
           ./gradlew bundleRelease --no-daemon "-PversionCodeOverride=${VERSION_CODE}"
 
-      - name: Upload to Play internal track
+      # Only a `mobile-v*` tag push publishes. A workflow_dispatch builds and
+      # signs, then stops here, so the Android signing chain can be exercised
+      # without a Play submission and without burning a version code (Play
+      # rejects a repeat, and the run number that produced it is not reusable).
+      # This mirrors desktop-release.yml, where the same separation is spelled
+      # `--publish ${{ github.event_name == 'push' && 'always' || 'never' }}`.
+      # The guard reads `push` rather than the ref because the only push trigger
+      # this workflow has is `tags: mobile-v*`; adding a branch trigger above
+      # would make it fire on a branch, so add `startsWith(github.ref,
+      # 'refs/tags/')` here if that trigger ever changes.
+      - name: Upload to Play internal track (tag pushes only)
+        if: github.event_name == 'push'
         uses: r0adkll/upload-google-play@e738b9dd8f2476ea806d921b64aacd24f34515a5 # v1.1.5
         with:
           serviceAccountJsonPlainText: ${{ secrets.PLAY_SERVICE_ACCOUNT_JSON }}
@@ -543,7 +554,12 @@ jobs:
           ISSUER_ID: ${{ secrets.APP_STORE_CONNECT_ISSUER_ID }}
         run: ruby scripts/ci/asc-key-check.rb
 
-      - name: Upload to TestFlight
+      # Same separation as Android above. The steps before this one still run on
+      # a dispatch on purpose: importing the certificate, exporting a signed IPA
+      # and checking the App Store Connect key all exercise the iOS credentials
+      # without submitting anything. Only the upload is a submission.
+      - name: Upload to TestFlight (tag pushes only)
+        if: github.event_name == 'push'
         env:
           ISSUER_ID: ${{ secrets.APP_STORE_CONNECT_ISSUER_ID }}
         run: |
--- a/scripts/ci/check-release-upload-guards.mjs
+++ b/scripts/ci/check-release-upload-guards.mjs
@@ -33,6 +33,7 @@ import { readFileSync } from 'node:fs';
 import { fileURLToPath } from 'node:url';
 import path from 'node:path';
 import { parse } from 'yaml';
+import { resolveWithin } from './safe-path.mjs';
 
 const REPO_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../..');
 
@@ -69,10 +70,30 @@ export function classifyStep(step) {
   return null;
 }
 
-/** True when `condition` restricts the step to a release tag. */
+/**
+ * True when `condition` restricts the step to a release tag.
+ *
+ * A substring test is not enough and the counter-example is 78 lines above the
+ * step it would wave through. The `android` job is gated on
+ *
+ *   github.event_name == 'push' || inputs.platform == 'both' || inputs.platform == 'android'
+ *
+ * and copying that condition down onto the upload step is the most natural edit
+ * anyone will make to this file. It contains an accepted guard and submits on a
+ * plain dispatch anyway, because a disjunction can only widen: `A || B` runs
+ * whenever B alone is true, so no guard to the left of a `||` restricts
+ * anything. Any `||` is therefore rejected outright.
+ *
+ * `&&` is the opposite and is accepted, so tightening the shipped guard to
+ * `github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')` does not
+ * red the build. Every other spelling fails closed - a red build and a human
+ * reading this comment is the outcome we want from an expression we cannot
+ * classify.
+ */
 export function isTagOnly(condition) {
   if (typeof condition !== 'string') return false;
-  return TAG_ONLY_GUARDS.some((guard) => condition.includes(guard));
+  if (condition.includes('||')) return false;
+  return condition.split('&&').some((operand) => TAG_ONLY_GUARDS.includes(operand.trim()));
 }
 
 /**
@@ -99,11 +120,21 @@ export function findUnguardedSubmissions(source, file) {
   return findings;
 }
 
-/** Runs the check over the real workflow files. */
+/**
+ * Runs the check over the real workflow files.
+ *
+ * `files` is a module constant and nothing reaches this script from a request,
+ * so the containment assertion is not load-bearing today. It is here because a
+ * path built from a parameter is worth pinning inside the tree it claims to
+ * describe before someone passes one in, and `resolveWithin` is what the other
+ * scripts in this directory use for it.
+ */
 export function checkRepo(root = REPO_ROOT, files = RELEASE_WORKFLOWS) {
-  return files.flatMap((file) =>
-    findUnguardedSubmissions(readFileSync(path.join(root, file), 'utf8'), file)
-  );
+  return files.flatMap((file) => {
+    const resolved = resolveWithin(root, file);
+    if (resolved === null) throw new Error(`Workflow path escapes the repository: ${file}`);
+    return findUnguardedSubmissions(readFileSync(resolved, 'utf8'), file);
+  });
 }
 
 if (process.argv[1] === fileURLToPath(import.meta.url)) {
@@ -118,7 +149,8 @@ if (process.argv[1] === fileURLToPath(import.meta.url)) {
   if (findings.length > 0) {
     console.error(
       `\n${findings.length} store submission(s) reachable from a workflow_dispatch. ` +
-        `Gate each on one of: ${TAG_ONLY_GUARDS.join(' | ')}`
+        `Gate each on ${TAG_ONLY_GUARDS.join(' or ')}. ` +
+        `A condition containing '||' is never accepted: a disjunction can only widen.`
     );
     process.exit(1);
   }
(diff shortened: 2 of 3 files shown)


ankit-yc (changes requested):
Reviewed at `8fd3bfa1147586bfcc4fba49333673be89e607ac`. The design is right — pinning the submitting mechanism rather than the step name, and landing it in `scripts/ci/*.test.mjs` so it runs in a job every event reaches, is the correct shape for this gate.

One blocking finding, and Aikido found it first: **`isTagOnly` uses a substring match, so the most likely wrong guard passes.**

`isTagOnly` is `TAG_ONLY_GUARDS.some((guard) => condition.includes(guard))`. Any condition that *contains* an accepted guard is treated as tag-only, including conditions that add a dispatch path with `||`. That is not a hypothetical shape — it is the idiom this workflow already uses at the job level:

```yaml
# line 120, unchanged by this PR
if: github.event_name == 'push' || inputs.platform == 'both' || inputs.platform == 'android'
```

Copying the job's condition down onto the upload step is the single most natural edit, and the check does not object. End-to-end against the real workflow at this head:

```
step guard                                                    check rc   suite
github.event_name == 'push'                        (shipped)      0      11/11
github.event_name == 'push' || inputs.platform == 'android'       0      11/11   <- runs on dispatch, MISSED
```

The second row uploads to the Play internal track on a plain `workflow_dispatch` — the exact behaviour this PR exists to prevent — and the gate reports "No unguarded store submissions" and exits 0.

Aikido's suggested patch is correct here (unusually — this is `AIK_AI_logic_bugs`, not the boilerplate `prefer_simplicity` rewrite). Verified all four cells:

```
                                  includes()        trim() === guard
mutant (|| dispatch path)         rc=0  MISSED      rc=1  CAUGHT
shipped guards, unmodified        rc=0  green       rc=0  green
existing suite                    11/11             11/11
```

Exact match also fails closed: an unusual-but-safe spelling reds the build and a human looks, which is the right direction for this gate.

**The fix needs its own test, or it will be reverted silently.** Note the third row above — the suite is 11/11 under *both* implementations, so nothing currently distinguishes them. Every negative case in the file (`always()`, `undefined`) fails the substring test for the trivial reason that it shares no text with a guard. The comment on `a condition that is not tag-only does not count as a guard` says a copy-paste is the realistic mistake, which is exactly right — it just picked the copy-paste the check happens to catch. Suggest adding to that test:

```js
// The job-level condition 60 lines above is `push || inputs.platform == ...`.
// Copying it onto the upload step keeps a dispatch path open.
assert.equal(isTagOnly("github.event_name == 'push' || inputs.platform == 'android'"), false);
```

Second Aikido comment, on `checkRepo`'s `readFileSync`: not exploitable — `root` and `files` default to a module constant and the repo root, and nothing reaches this script from a request. I would still take the containment check rather than leave the comment open; it is four lines, it fails closed, and arguing a scanner finding down costs more than satisfying it.

Everything else holds up. Matching `uses` on the action name with the version stripped, and rejecting `altool --validate-app` while accepting `--upload-app`, are both good calls. Accepting `startsWith(github.ref, 'refs/tags/')` alongside the event check so a later tightening does not red the build is the right kind of forward accommodation.

Requesting changes on the `isTagOnly` predicate plus a test that pins it. Nothing else blocks.

aupyay:
Both review findings fixed in `4ead83090`.

**`isTagOnly` was a substring test**, so any condition *containing* an accepted guard passed — including the `android` job's own condition, 78 lines above the step it would wave through. Copying that down onto the upload step is the most natural edit anyone makes to this file, and it submits to Play internal on a plain dispatch.

A disjunction can only widen — `A || B` runs whenever `B` alone is true, so no guard to the left of a `||` restricts anything. Any `||` is now rejected outright. `&&` is accepted, so tightening to `github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')` still passes and the property the header promised is kept. Everything else fails closed.

**The larger half of the finding was that the suite was 11/11 under both implementations** — the fix would have been revertible in silence, which is this file's own failure mode reproduced inside the tests written to prevent it. Every negative case cleared the substring test for a trivial reason: it shared no text with a guard. Five new tests exercise a condition that *contains* a guard and is unsafe anyway.

All arms re-run at the pushed commit `4ead83090`, after the pre-commit hooks reformatted the files:

```
A  suite under the substring implementation            13 pass / 3 fail
   suite as shipped                                    16 pass / 0 fail
B  real workflow, job condition copied onto the step    CLI rc=1   (was rc=0)
C  real workflow, guard cut entirely                    CLI rc=1
D  restored                                             CLI rc=0, file byte-identical

pnpm run test:scripts   rc=0   385 tests / 385 pass / 0 fail / 0 skipped
```

The second comment is taken rather than argued down: `checkRepo` builds a path from a parameter, so it now goes through `resolveWithin` from `scripts/ci/safe-path.mjs` — the helper four sibling scripts already use — and throws when a path escapes the tree. Not reachable today; pinned before it is.

The CLI's failure message now states the rule it enforces, so a red build explains itself: *"A condition containing '||' is never accepted: a disjunction can only widen."*

ankit-yc (approved):
Re-reviewed at `4ead83090714d5935f040e92920351e82f2fcdf7`. Both findings fixed, and the predicate you shipped is better than the one I suggested.

**You were right to reject `trim() === guard`.** Verified the case that decides it:

```
condition                                                      ===guard   ||-reject (shipped)
github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')   REJECT     accept
```

Exact match reds the build on a strict tightening, which contradicts the header's own promise. `A || B` runs whenever `B` alone is true, so no guard left of a `||` restricts anything; `&&` is the opposite. That is the invariant, and `=== guard` was a proxy for it that also excluded the safe direction.

Probed the predicate across the shapes that matter:

```
true    github.event_name == 'push'                                   shipped guard
true    startsWith(github.ref, 'refs/tags/')                          ref guard
true    ... == 'push' && startsWith(github.ref, 'refs/tags/')         tightening, stays green
true    success() && github.event_name == 'push'                      conjunction
false   ... == 'push' || inputs.platform == 'android'                 the bug, now caught
false   inputs.platform == 'android' || ... == 'push'                 guard on the right of ||
false   !(github.event_name == 'push')                                negation
false   github.event_name != 'push'                                   inverted operator
false   always()  /  inputs.platform == 'android'                     no guard
```

No safety hole in that table. Two shapes fail closed on a *safe* guard — `${{ github.event_name == 'push' }}` (the expression-wrapper form, which is legal Actions syntax) and a guard with extra inner whitespace. Both produce a red build carrying the rule in its message, so a human reads the header and fixes the spelling. Correct direction, and not worth widening the parser for. Worth knowing before someone hits it and thinks the gate is broken.

Row three of my last review is what I care about, and it moved:

```
suite, old substring implementation    16 tests   13 pass / 3 FAIL
suite as shipped                       16 tests   16 pass / 0 fail
```

The fix is pinned. Reverting `isTagOnly` to the old body now reds three tests, where before nothing in the file distinguished the two implementations.

End-to-end against the real workflow, mutation applied to the committed tree and restored:

```
C  Play guard -> `push || inputs.platform == 'android'`   CLI rc=1 (was rc=0 at 8fd3bfa11)
D  restored                                                CLI rc=0, byte-identical
   pnpm run test:scripts                                   rc=0, 385 tests / 385 pass / 0 fail / 0 skipped
```

The failure message now states the rule rather than just the verdict, which is what makes a red build self-explaining.

`checkRepo` going through the existing `resolveWithin` in `scripts/ci/safe-path.mjs` is the right call — I checked, it is pre-existing on `dev` (added in `7774b3865`) and four sibling scripts already use it, so this is reuse rather than a fifth copy. Not reachable today, pinned before it is.

Scanners at this head: Aikido `SUCCESS`, GitGuardian `SUCCESS`, Gitleaks `SUCCESS`, CodeQL neutral, two still running.

Approving.

PR #2819, opened 2026-09-06
ci(desktop): verify desktop signatures instead of assuming them

aupyay:
## PR Checklist

`desktop-release.yml` never verifies a signature. Grep for the tools across both release workflows returns two hits, and both are certificate import rather than verification:

```
$ grep -nE 'codesign|spctl|signtool|stapler' .github/workflows/desktop-release.yml .github/workflows/mobile-release.yml
mobile-release.yml:307:  security import ... -T /usr/bin/codesign
mobile-release.yml:329:  IDENTITY=$(security find-identity -v -p codesigning ...)
```

Zero hits in `desktop-release.yml`. So a green Windows or macOS job has only ever proved that signing did not throw.

That matters on macOS because the build has a path that produces an unsigned-but-plausible artifact. `apps/desktop/scripts/notarize.js` falls back to `codesign --sign -` whenever `APPLE_ID`, `APPLE_APP_SPECIFIC_PASSWORD` or `APPLE_TEAM_ID` is absent or rejected. A release run with a revoked Apple credential produces an ad-hoc signed `.app` and a green job.

## What is the new behavior?

Both jobs verify their output before the upload step.

**The check is a conjunction, not `codesign --verify`.** Measured on macOS 15 (arm64) against a real Developer ID signed and notarized app, against the ad-hoc bundle that `notarize.js` produces, and against an unsigned one:

| check | Developer ID | ad-hoc | unsigned |
|---|---|---|---|
| `codesign --verify --deep --strict` | rc=0 | **rc=0** | rc=1 |
| `spctl -a -t exec` | rc=0 | rc=3 | rc=3 |
| `stapler validate` | rc=0 | rc=65 | rc=65 |
| `Authority=Developer ID Application` | present | absent | absent |
| `CodeDirectory` flags | `runtime` | `adhoc,runtime` | n/a |
| `TeamIdentifier` | set | `not set` | n/a |

`codesign --verify` exits 0 on the ad-hoc bundle. It is the check most reached for and it is blind to the only signing failure this build can actually produce. `@electron/notarize` already runs that same call (`lib/check-signature.js` runs `codesign -vvv --deep --strict`) before notarizing, which is why nothing has ever caught this - and it only runs on the path where the credentials are present.

So the verdict asserts all of: Developer ID authority in the chain, team identifier set, not ad-hoc, Gatekeeper accepts, notarization ticket stapled.

The `.app` is the target rather than the `.dmg`: `@electron/notarize` staples the ticket to the bundle (`lib/index.js` calls `stapleApp` after `notarizeAndWaitForNotaryTool`), and the disk image is built from the already-stapled app. Asserting a stapled ticket on the `.dmg` would false-fail a good build.

Windows asserts Authenticode `Status -eq 'Valid'` and a signer subject matching `build.win.signtoolOptions.publisherName`, read out of `package.json` so the check cannot drift from what is actually signed.

Both steps fail on an empty artifact set. "Nothing to check" and "everything checked out" must not produce the same exit code.

## Verification

Full desktop suite at `142521474`:

```
Test Suites: 71 passed, 71 total
Tests:       1093 passed, 1093 total
Statements   : 98.37%   Branches : 90.87%   Functions : 97.68%   Lines : 98.37%
```

Above the desktop floors in `_test.yaml` (statements 95, branches 88, functions 95, lines 95). `lint` and `type-check` both exit 0.

New module measured alone: 36 tests, statements 99.05%, functions 100%.

**End to end against real artifacts on macOS 15**, one Developer ID signed and notarized app and the ad-hoc bundle:

```
GOOD  (Developer ID + notarized)   exit 0
ADHOC (what notarize.js produces)  exit 1
  - The bundle is ad-hoc signed. ...
  - No "Developer ID Application" authority in the certificate chain.
  - TeamIdentifier is not set on the signature.
  - spctl assessment rejected the artifact (exit 3).
  - stapler validate found no stapled notarization ticket (exit 65).
```

Note the positive control: the check is not simply always-failing.

**Mutation table.** A verification step that cannot fail is worse than none, because it converts "we never checked" into "we checked and it was fine". Every mutation below was applied to the shipped tree and reverted:

```
M1  drop ad-hoc rejection                     1 failed, 35 passed
M2  drop Developer ID authority check         2 failed, 34 passed
M3  drop stapler check                        3 failed, 33 passed
M4  drop spctl check                          3 failed, 33 passed
M5  main returns 0 on failure                 2 failed, 34 passed
M6  runner ignores spawn error                1 failed, 35 passed
M7  adhoc flag name typo                      2 failed, 34 passed
M8  authority startsWith -> includes          1 failed, 35 passed
M9  team check uses includes                  1 failed, 35 passed
M10 parse Authority by substring              1 failed, 35 passed
M11 adhoc read from wrong field               2 failed, 34 passed
M12 drop null-status guard                    1 failed, 35 passed
BASELINE                                     36 passed
```

M8 and M10 survived the first pass and are the reason for the second commit. Both are the substring class - a check that accepts anything CONTAINING the accepted text - which is the same defect a scanner flagged on the sibling mobile-guards PR. The added cases are an authority that contains the Developer ID prefix without starting with it, and a bundle whose executable path and identifier carry the word `Authority`.

## What this PR does not establish

- **The Windows step has not been executed.** It cannot be: every signing job sits behind the `release` environment, whose deployment branch policy is three tag patterns and zero branch patterns, so no branch dispatch reaches it. The macOS logic was exercised against real artifacts locally; the PowerShell step is reasoned from the `Get-AuthenticodeSignature` contract and reviewed, not run. Worth a reviewer's eye on that specifically.
- It does not make an unsigned build fail earlier. `forceCodeSigning` is deliberately not set, because local unsigned builds are a documented workflow in `apps/desktop/docs/RELEASE-TESTING.md`. The CI step is the gate, and CI is where it matters.
- It changes no credential, environment, or branch-protection setting.

## Related Issue(s)

No tracking issue. Came out of a release dry-run readiness review that established the release path had three signing chains and no verification on any of them.

### DIFF
--- a/.github/workflows/desktop-release.yml
+++ b/.github/workflows/desktop-release.yml
@@ -174,6 +174,41 @@ jobs:
           AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      # Nothing downstream of the build re-reads the signature, so without this a
+      # green job means only that signing did not throw. Runs before the upload so
+      # an unsigned installer cannot become an artifact other jobs consume.
+      - name: Verify the Windows Authenticode signatures
+        working-directory: apps/desktop
+        shell: pwsh
+        run: |
+          $ErrorActionPreference = 'Stop'
+          # Read the expected publisher from the electron-builder config rather
+          # than repeating it, so the check cannot drift from what is signed.
+          $expected = (Get-Content package.json -Raw | ConvertFrom-Json).build.win.signtoolOptions.publisherName
+          if ([string]::IsNullOrWhiteSpace($expected)) {
+            throw 'build.win.signtoolOptions.publisherName is not set in package.json; cannot verify.'
+          }
+          $files = @(Get-ChildItem -Path dist -Filter *.exe -File -ErrorAction SilentlyContinue)
+          # An empty artifact set must fail. "Nothing to check" and "everything
+          # checked out" are not allowed to look the same from outside.
+          if ($files.Count -eq 0) { throw 'No .exe found in apps/desktop/dist; nothing was verified.' }
+          $failures = @()
+          foreach ($file in $files) {
+            $sig = Get-AuthenticodeSignature -FilePath $file.FullName
+            $subject = if ($sig.SignerCertificate) { $sig.SignerCertificate.Subject } else { '<unsigned>' }
+            Write-Host "$($file.Name): status=$($sig.Status) subject=$subject"
+            if ($sig.Status -ne 'Valid') {
+              $failures += "$($file.Name): Authenticode status is $($sig.Status), expected Valid."
+            } elseif ($subject -notlike "*$expected*") {
+              $failures += "$($file.Name): signed by $subject, expected $expected."
+            }
+          }
+          if ($failures.Count -gt 0) {
+            foreach ($failure in $failures) { Write-Host "::error::$failure" }
+            exit 1
+          }
+          Write-Host "All $($files.Count) Windows artifacts carry a valid $expected signature."
+
       - name: Upload signed Windows artifacts
         if: always()
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
@@ -218,6 +253,32 @@ jobs:
           APPLE_APP_SPECIFIC_PASSWORD: ${{ secrets.APPLE_APP_SPECIFIC_PASSWORD }}
           APPLE_TEAM_ID: ${{ secrets.APPLE_TEAM_ID }}
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      # scripts/notarize.js falls back to `codesign --sign -` when APPLE_ID,
+      # APPLE_APP_SPECIFIC_PASSWORD or APPLE_TEAM_ID is absent or rejected, and an
+      # ad-hoc signed bundle passes `codesign --verify --deep --strict` with exit
+      # 0 - measured, macOS 15. So the obvious check is blind to the one failure
+      # this build can actually produce, and the script asserts the Developer ID
+      # authority, the team identifier, Gatekeeper acceptance and the stapled
+      # ticket instead. See apps/desktop/scripts/verify-macos-signature.js.
+      #
+      # The .app is the target rather than the .dmg: @electron/notarize staples
+      # the ticket to the bundle (lib/index.js calls stapleApp after notarizing),
+      # and the disk image is built from the already-stapled app.
+      - name: Verify the macOS signature and notarization
+        working-directory: apps/desktop
+        run: |
+          set -euo pipefail
+          # `-print -quit` rather than `| head -1`: under `pipefail` a closed pipe
+          # makes find exit 141 and fails the step for the wrong reason.
+          APP=$(find dist -maxdepth 2 -type d -name '*.app' -print -quit)
+          # An empty artifact set must fail. "Nothing to check" and "everything
+          # checked out" are not allowed to look the same from outside.
+          if [ -z "$APP" ]; then
+            echo "::error::No .app bundle under apps/desktop/dist; nothing was verified."
+            exit 1
+          fi
+          node scripts/verify-macos-signature.js "$APP"
 
       - name: Upload signed macOS artifacts
         if: always()
--- a/.github/workflows/desktop-release.yml
+++ b/.github/workflows/desktop-release.yml
@@ -175,8 +175,17 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       # Nothing downstream of the build re-reads the signature, so without this a
-      # green job means only that signing did not throw. Runs before the upload so
-      # an unsigned installer cannot become an artifact other jobs consume.
+      # green job means only that signing did not throw.
+      #
+      # Step position does NOT contain a bad installer, and it must not be read
+      # that way: the upload step below is `if: always()` so it runs even when
+      # this fails, and on a tag the build step itself already passed
+      # `--publish always`, attaching the installers to the draft release before
+      # this step runs. What actually holds the line is job-level: `publish`
+      # needs verify + windows + macos and `release` needs windows + macos, so a
+      # failure here means the draft is never made public and no manual release
+      # is created. The artifact and the draft asset are deliberately still
+      # produced, because a rejected installer is the thing you need to inspect.
       - name: Verify the Windows Authenticode signatures
         working-directory: apps/desktop
         shell: pwsh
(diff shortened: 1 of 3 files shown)


ankit-yc (dismissed):
Approving. Every load-bearing claim in the description verifies, and the mutation result below is stronger than what the body claims for itself.

## The claims, checked

```
desktop-release.yml @ origin/dev   codesign|spctl|signtool|stapler   0 hits
mobile-release.yml  @ origin/dev   same grep, CONTROL                2 hits
```

Zero against a control that fires, so the "never verified a signature" premise is an absence rather than a broken query.

The ad-hoc fallback is exactly as described — `notarize.js:39` gates on `!APPLE_ID || !APPLE_APP_SPECIFIC_PASSWORD || !APPLE_TEAM_ID`, logs *"Ad-hoc signing macOS app for local launch verification"*, and signs with `codesign --sign -`. **A revoked Apple credential produces a plausible bundle and a green job**, which is the failure this exists for.

## Mutation: all six conjuncts, none survives

The body says the check is a conjunction. I neutralised each condition in turn (`if (false && …)`, anchor confirmed before each substitution) and ran the suite:

| conjunct dropped | tests failing |
|---|---|
| `codesign --verify` exit code | 2 |
| adhoc flag | 1 |
| Developer ID authority | 2 |
| TeamIdentifier | 2 |
| spctl gatekeeper | 3 |
| stapler ticket | 3 |
| reverted | **36 pass / 36** |

**Every conjunct is load-bearing and separately covered.** `rejects the ad-hoc bundle even though codesign --verify exits 0` is the test that justifies the whole design, and `fails each condition independently while the others pass` is the one that makes the table above predictable rather than lucky.

Fail-closed default confirmed: `requireNotarization = true`, `deps.requireNotarization !== false`, and the workflow passes only a path — so the stapling condition cannot be skipped by anything CI does. That was the place I expected to find the fall-open reintroduced and it is not there.

Also confirmed: both empty-artifact-set guards (`"Nothing to check" and "everything checked out" must not produce the same exit code`), and Windows reading `build.win.signtoolOptions.publisherName` out of `package.json` with an explicit throw when unset.

## One precision on the description, not a change request

> Both jobs verify their output before the upload step.

True about ordering, but `Upload signed macOS artifacts` carries `if: always()` — **pre-existing, not introduced here** — so a failed verification still uploads the artifact. The real guarantee is the stronger one and worth stating as it is: verification failing fails the job, and `publish` has `needs:` on the build jobs, so **an unverified artifact cannot reach a release**. It can still land in the workflow's artifact list, which is the right behaviour for debugging a failed signing run.

Worth a word in the body only because "verify before upload" reads as a gate on the upload, and the gate is actually on the release.

## Boundary worth naming

The 36 tests exercise the parser and the conjunction over captured `codesign`/`spctl`/`stapler` output. They do not invoke those tools. The comparison table in the header says it was measured on macOS 15 against real Developer ID, ad-hoc and unsigned artifacts — that measurement is the evidence for the fixtures, and it is the part no unit test can re-derive. Fine as it stands; it just means a change to the real tools' output format would be caught by CI running on a real artifact, not by this suite.

ankit-yc (changes requested):
Reviewed at `142521474`. The macOS module is right and the ad-hoc finding is the important thing in this PR — `codesign --verify` returning 0 on the bundle `notarize.js` actually produces is the difference between a check and a decoration. `hasTeamIdentifier` excluding the literal `not set`, `isAdHoc` testing membership in the split flag list rather than a substring, and `hasDeveloperIdAuthority` anchoring with `startsWith` on both the line and the value all hold up under reading.

One blocking item, one worth hardening, one note.

---

## Blocking — the step comment asserts a safety property the file does not have

```yaml
# Runs before the upload so an unsigned installer cannot become an artifact
# other jobs consume.
```

Both halves are false as written, and someone will rely on this.

**1. The upload happens anyway.** Both upload steps are unconditional:

```
212  - name: Upload signed Windows artifacts     213  if: always()
283  - name: Upload signed macOS artifacts       284  if: always()
```

`if: always()` runs the step when the verify step above it has failed, so a failed verification still produces `desktop-windows-signed` / `desktop-macos-signed`.

**2. On the path that matters, the installers are already published before verify runs.** The build step is

```yaml
run: pnpm run desktop:dist:win -- --publish ${{ github.event_name == 'push' && 'always' || 'never' }}
```

so on a tag push electron-builder attaches the `.exe`, `latest.yml` and the mac assets to the draft release **inside the build step**, before the new verify step executes. Step order cannot protect that.

**The protection you actually have is job-level, and it is sound:**

```
publish:  if: github.event_name == 'push'         needs: verify, windows, macos
release:  if: github.event_name == 'workflow_dispatch'   needs: windows, macos
```

A failed verify reds `windows`, so the draft is never flipped public and the dispatch-path draft is never created. That is the real guarantee and it is a good one — it just is not the one the comment describes. Please say that instead: the artifacts and the draft assets are produced regardless, and what verification gates is the release becoming visible.

I am blocking on a comment because this thread has spent the day on controls that read as gates and are not, and a comment that tells the next reader an unsigned installer cannot escape is exactly that shape.

---

## Worth hardening — the publisher match is a substring, over a non-ASCII literal

```powershell
} elseif ($subject -notlike "*$expected*") {
```

Two things, on a step you have said has never executed:

**It is the same class Aikido flagged on #2818 and that your own M8/M10 fixed in this PR.** `-like "*X*"` accepts any Subject *containing* the publisher name, so `CN=Not DuneXploration UG (haftungsbeschränkt) Ltd` passes. Asserting the `CN=` value rather than searching the whole Subject makes it exact, and the Subject is the one field here you can anchor on.

**And the comparand is non-ASCII:**

```
build.win.signtoolOptions.publisherName = "DuneXploration UG (haftungsbeschränkt)"
isascii = False        U+00E4 at index 33        utf-8: ...beschr\xc3\xa4nkt)
```

A mismatch here does not fail open, it **fails a good release** — the one direction that is expensive on release day, on the step nobody can exercise before then. `pwsh` and a UTF-8 workflow file should handle it, and I could not test it either. Matching on an ASCII-safe anchor such as `CN=DuneXploration UG` sidesteps the question rather than betting on it.

---

## Note, not a change request — macOS verifies one bundle, Windows verifies all of them

```bash
APP=$(find dist -maxdepth 2 -type d -name '*.app' -print -quit)
```

versus the Windows step's `foreach ($file in $files)` over every `.exe`. Today `build.mac.target` is `["dmg","zip"]` with no `arch`, so `macos-14` produces a single arm64 bundle and `-print -quit` finds the only one. If an `arch` list is ever added the check silently covers half the output while still printing success. Worth a line in the comment saying the single bundle is a property of the current config, or iterating like the Windows step does.

(The `-print -quit` rather than `| head -1` reasoning under `pipefail` is right and worth keeping.)

---

Everything else stands: reading `publisherName` from the config rather than repeating it, failing when the artifact set is empty so "nothing to check" cannot look like "everything checked out", targeting the `.app` rather than the `.dmg` with `stapleApp` cited for why, and declining `forceCodeSigning` because it would break a documented local workflow. The M1–M12 table with each mutation applied to the shipped tree and reverted is the standard, and `MUTATION DID NOT APPLY` after a prettier rewrite silently un-applied one is the same hazard from the other side — worth carrying into anything else that mutates a formatted file.

ankit-yc (approved):
Re-reviewed at `c90a1dbf3`. Read `reviews` immediately before submitting, per the rule we adopted an hour ago: the dismissed approval and my own changes-requested both sit at `142521474`, so this approval is on a sha neither earlier review saw. Approving — all four items are fixed, and two are better than what I asked for.

## Verified at the sha, not taken from the description

**Comment** — the old text is gone (`grep -cF 'cannot become an artifact other jobs consume'` → 0) and lines 177–188 now name the real guarantee: `if: always()` on the upload, `--publish always` inside the build on a tag, and the job-level `needs` as what actually holds the line. The added sentence — *"the artifact and the draft asset are deliberately still produced, because a rejected installer is the thing you need to inspect"* — is the part I did not think to ask for and is the reason the design is right rather than merely documented.

**Publisher** — `GetNameInfo(SimpleName)` + `-cne` is better than the `CN=DuneXploration UG` anchor I suggested. Mine still pattern-matched the DN and so was still a bet on how .NET quotes a value containing parentheses; reading the CN value removes the question instead of narrowing it. Printing both lengths in the failure message is a good touch — a trailing-space or normalisation mismatch is otherwise invisible in the log.

**Arch** — I filed this as a note and you fixed it. `find -print0 | xargs -0` with an explicit `COUNT` guard, and I checked the receiving end rather than assuming: `main` does `argv.slice(2)` and `verifyTargets(targets)`, usage string `<artifact> [artifact...]`. Multiple bundles in one invocation are genuinely handled, so the `xargs` batching cannot silently drop one.

---

## The timestamp line — approving with it, and saying plainly why that is a judgement and not a verification

This is the one thing in the PR I cannot check, and I think you are right to ship it. Recording the reasoning so the decision is auditable rather than implied:

- **It fails closed.** If `TimeStamperCertificate` comes back null on a correctly timestamped binary, the `windows` job reds and the release does not publish. It cannot ship an unsigned or untimestamped installer. The cost is a blocked release, not a bad one.
- **The fallback is one line**, and it is the line itself.
- **An untimestamped signature is a real latent failure** — the installers stop validating when the certificate rolls, and nothing else in the pipeline would notice.

Where I differ slightly from your framing: I think null-on-RFC3161 is **more likely than not**, not merely possible. `Get-AuthenticodeSignature` surfaces the legacy Authenticode counter-signature; an RFC3161 timestamp is carried as a separate unauthenticated attribute, and `windows-trusted-sign.js` configures `timestampRfc3161` exclusively. I could not confirm that on any source I was willing to cite — the PowerShell issues I found are about `Set-AuthenticodeSignature` and catalog signing, not this — so it stays a suspicion and it is not grounds to block.

**It is answerable before release day, cheaply, and without the release gate.** The question is a property of the runner image, not of our certificate: a `windows-latest` job with **no `environment:`** — so a branch dispatch reaches it, unlike everything else we looked at today — running `Get-AuthenticodeSignature` over already-signed binaries present on the image and reporting which of them populate `TimeStamperCertificate`. If none do, the property does not surface RFC3161 there and the line needs replacing before it blocks a release. If some do, the suspicion dies.

I will run that and post the result if you want it — say the word and I will not touch this branch. It is the same move as your two macOS arms: stop reasoning about the artifact and go look at one.

---

The extraction probe against the shipped YAML body is the right instrument, and the near-miss you reported is the more valuable half. An empty `step.sh` exiting 0 across all three cases is a third instance of today's pattern — a command that never ran, reporting as the success you wanted — and the "zero bundles must fail" case catching it is exactly why a negative case has to be in the table. Re-extracting after prettier and diffing against the body you tested is the discipline that makes the numbers attributable, and I would not have thought to diff it.

PR #2820, opened 2026-09-06
ci(repo): read job-level guards, and cover desktop's own publish steps

aupyay:
## The finding

Review of the gate merged in #2818 found that stripping `if: github.event_name == 'push'` off `desktop-release.yml`'s `publish` job — **the job that flips a draft release public** — left the check green. Reproduced here before fixing anything:

```
mutation: remove the publish job's if:      gate rc=0   MISSED
control:  plant an upload-google-play step
          into that same desktop job        gate rc=1   fires
```

The control and the mutation are in the same file and the same job, so it is not file scope — the gate reads `desktop-release.yml` correctly.

## Two independent causes, and fixing either alone is wrong

**1. `SUBMISSIONS` listed only mobile mechanisms.** Desktop does not reach a store, it reaches the public: `gh release edit --draft=false` in `publish`, and `softprops/action-gh-release` in the dispatch-only `release` job.

**2. The check read `step.if` and nothing else.** Desktop guards at the **job** (`publish` carries the condition, its step carries none); mobile guards at the **step**. So adding the mechanism on its own reds the *correct* tree — measured, not predicted:

```
mechanism added, no job-level reading, shipped tree:
  "job 'publish' step 'Publish the release' submits to a public GitHub Release
   with no condition."                      rc=1   <- false positive on correct code
```

A step runs only when its job runs **and** its own condition passes, so a tag-only condition at either level restricts it. Both are now consulted.

**This cannot weaken the mobile side, and that is the regression this change could have caused.** `isTagOnly` rejects any `||`, and the `android`/`ios` job conditions are disjunctions (`... || inputs.platform == 'android'`), so a mobile step whose own guard is deleted is still a finding. Pinned by its own test.

**`action-gh-release` is a submission unless `draft: true`.** The dispatch-only `release` job sets it, which is why it is not a finding today — but flipping that one word publishes from a branch, and nothing else would notice.

## Evidence

Every arm applied to the real workflow files and reverted, at the pushed commit `fb7c84b33` after the hooks reformatted:

```
M0  baseline                                   rc=0
M1  publish job guard deleted                  rc=1    was rc=0
M2  publish job guard widened with ||          rc=1
M3  action-gh-release draft: true -> false     rc=1
M4  mobile step guard deleted, job if has ||   rc=1    regression guard
M5  restored                                   rc=0    files byte-identical
```

The tests discriminate rather than decorate. Run against the implementation currently on `dev`:

```
new tests vs dev's implementation    18 pass / 3 fail
new tests vs this implementation     21 pass / 0 fail

pnpm run test:scripts   rc=0   390 tests / 390 pass / 0 fail / 0 skipped
```

**Two of the five new tests pass under both implementations, deliberately, and neither is evidence for this change:** the mobile regression guard *must* pass under both — that is its job — and the desktop false-positive guard has nothing to false-positive on until the mechanism exists. Saying which tests carry the argument and which are guards seemed better than a bare 21/21.

## Severity — latent, not reachable

`publish` needs `verify`, `windows` and `macos`, and those sit behind the `release` environment, whose deployment policy is three tag patterns and zero branch patterns. The job cannot run from a branch whatever its own `if:` says.

It becomes load-bearing the moment that policy changes, or if the pattern is copied into a new environment. **That is the same reasoning that made the unguarded Play upload look safe this morning: safe because of a different control, with nothing pinning its own.**

## Credit

The mutation that found this was run by a reviewer on my file after I had planted a *mobile* mechanism into a desktop workflow and concluded the gate covered desktop. It read the file; it did not know desktop's own way of publishing. **A planted hit proves the scanner reaches the file, not that it recognises what lives there.**

### DIFF
--- a/scripts/ci/check-release-upload-guards.mjs
+++ b/scripts/ci/check-release-upload-guards.mjs
@@ -50,6 +50,15 @@ export const RELEASE_WORKFLOWS = [
 export const SUBMISSIONS = [
   { store: 'Google Play', uses: 'r0adkll/upload-google-play' },
   { store: 'App Store Connect', run: 'altool --upload-app' },
+  // Desktop does not reach a store, it reaches the public. `gh release edit
+  // --draft=false` is the step that takes a draft release public, and a public
+  // release is what electron-updater's clients poll.
+  { store: 'a public GitHub Release', run: '--draft=false' },
+  // action-gh-release publishes unless the release is explicitly a draft. The
+  // dispatch-only `release` job sets `draft: true`, which is why it is not a
+  // finding today - but flipping that one word publishes from a branch, and
+  // that is the edit this entry exists to catch.
+  { store: 'a GitHub Release', uses: 'softprops/action-gh-release', exemptWhenDraft: true },
 ];
 
 /** Conditions that restrict a step to a release tag. Either one is enough. */
@@ -63,7 +72,9 @@ export function classifyStep(step) {
   if (!step || typeof step !== 'object') return null;
   const uses = typeof step.uses === 'string' ? step.uses : '';
   const run = typeof step.run === 'string' ? step.run : '';
+  const isDraft = Boolean(step.with && step.with.draft === true);
   for (const submission of SUBMISSIONS) {
+    if (submission.exemptWhenDraft && isDraft) continue;
     if (submission.uses && uses.split('@')[0] === submission.uses) return submission;
     if (submission.run && run.includes(submission.run)) return submission;
   }
@@ -107,7 +118,12 @@ export function findUnguardedSubmissions(source, file) {
     for (const step of job?.steps ?? []) {
       const submission = classifyStep(step);
       if (!submission) continue;
-      if (isTagOnly(step.if)) continue;
+      // A step runs only when its job runs AND its own condition passes, so a
+      // tag-only condition at EITHER level restricts it. Desktop guards at the
+      // job (`publish` carries the condition, its step does not) and mobile at
+      // the step, and reading only one level misses one of them: reading only
+      // steps waves through a desktop `publish` job whose guard was deleted.
+      if (isTagOnly(step.if) || isTagOnly(job?.if)) continue;
       findings.push({
         file,
         job: jobId,
@@ -148,11 +164,11 @@ if (process.argv[1] === fileURLToPath(import.meta.url)) {
   }
   if (findings.length > 0) {
     console.error(
-      `\n${findings.length} store submission(s) reachable from a workflow_dispatch. ` +
+      `\n${findings.length} outward-facing publish step(s) reachable from a workflow_dispatch. ` +
         `Gate each on ${TAG_ONLY_GUARDS.join(' or ')}. ` +
         `A condition containing '||' is never accepted: a disjunction can only widen.`
     );
     process.exit(1);
   }
-  console.log(`No unguarded store submissions in ${RELEASE_WORKFLOWS.length} release workflows.`);
+  console.log(`No unguarded publish steps in ${RELEASE_WORKFLOWS.length} release workflows.`);
 }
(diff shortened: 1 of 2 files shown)


ankit-yc (dismissed):
Approving at fb7c84b33. Reviewed as Claude L2 Yosemite Crew - byline because our review account carries none, and this PR came out of a finding I filed, so the chain should be readable.

Ran it rather than read it, in a detached worktree at the PR head, deps installed, tree clean and restored after every mutation. Baseline: gate rc=0, `pnpm run test:scripts` 390/390.

Both causes are fixed and I confirmed they are independent:

```
A: stop reading job-level if   rc=1  desktop 'publish' becomes a false positive
B: stop reading step-level if  rc=1  mobile 'android' Play upload uncovered
C: release job draft:true->false rc=1  'Create draft GitHub Release'
D: delete desktop publish job guard rc=1
E: delete mobile Play step guard (job if carries ||) rc=1  <- no job-level rescue
```

E is the regression that mattered: reading job-level `if` could have let the `android` job's disjunction wave through a stripped step guard. It does not, and that is now measured rather than argued.

Mutations the CLI could not see, re-measured against the suite, since the CLI on a correct tree is the wrong instrument for them:

```
F: isDraft forced true            suite 389/1 fail
H: drop --draft=false mechanism   suite 388/2 fail
I: drop action-gh-release mech    suite 389/1 fail
```

Non-blocking, one line with no test behind it.

`if (condition.includes('||')) return false;` survives removal under both instruments - CLI rc=0 and suite 390/390. It is not dead code, but its effect is narrower than the comment implies:

```
condition                                              with line   without
github.event_name == 'push' || inputs.platform == ...    false      false
github.event_name == 'push' && (inputs.x || inputs.y)    false      TRUE
```

The plain disjunction is already rejected by the exact-match on `&&`-split operands, so the early return changes nothing there. Its only observable effect is on `A && (x || y)`, where it keeps the checker conservative. That shape is genuinely safe - the `&&` still requires push - so this guards conservatism rather than correctness, which is why it is not blocking.

But by the rule you applied on 2818, a guard with no test is a guard that gets deleted quietly, and this one currently reads as redundant to anyone who tries. One assertion pins it:

```js
assert.equal(isTagOnly("github.event_name == 'push' && (inputs.x || inputs.y)"), false);
```

Your restraint on the two non-discriminating tests is the right call and worth saying: labelling them guards rather than counting 21/21 as coverage is the distinction most reports lose.

ankit-yc (approved):
Re-approving at `ab485b672` after my `fb7c84b33` approval was dismissed by the push. Reviewed by Claude L2 Yosemite Crew.

**The `||` early return is load-bearing and I re-measured the claim rather than taking it.** I was the one who called it conservatism; that was wrong and the precedence case inverts it.

Fresh detached worktree at `ab485b672`, tree clean, restored and re-verified clean after every mutation.

Implementation is unchanged from the sha I already approved — the `.mjs` delta is entirely inside the JSDoc block. The behavioural change is the two new tests.

```
BASELINE  node --test check-release-upload-guards.test.mjs   23 pass / 0 fail
          node scripts/ci/check-release-upload-guards.mjs    rc=0
          full test:scripts (392 tests)                      392 pass / 0 fail

MUTATION  delete `if (condition.includes('||')) return false;`  (one line, diff confirmed 1 deletion)
          suite   21 pass / 2 FAIL   -> tests 14 and 15, both new in this PR
          CLI     rc=0               <- blind, exactly as measured pre-push
restored  git status clean
```

Both new tests are mutation-sensitive; neither is a non-discriminating guard.

**I took the claim one level further, to the consumer.** `isTagOnly` is the unit, but `findUnguardedSubmissions` is what the gate calls, so a hole is only real if it survives there. Synthetic workflow, one `r0adkll/upload-google-play` step, job guarded by the precedence condition:

```
                                                 isTagOnly   findUnguardedSubmissions
shipped module                                     false      1 finding
`||` line removed                                  TRUE       0 findings   <- the hole, end to end
```

So a real Play upload inside `inputs.force || inputs.x && github.event_name == 'push'` is waved through without that line. Order-dependence confirmed too: the same logical shape with the guard on the left of the `&&` is rejected either way, which is why pinning both spellings matters.

With the line in place `isTagOnly` is sound: any `||` fails closed, and with no `||` present every `&&` operand is required, so an operand that is exactly a guard genuinely restricts.

PR #2821, opened 2026-09-06
ci(repo): cover gh release create, the third way desktop reaches the public

aupyay:
Follow-up to #2820, found in review of it.

## The gap

`desktop-release.yml` has **three** ways to reach the public, and the third is in the one job with no protection:

```
JOB verify   if: <none>   environment: <none>   permissions: contents: write   <- runs on a branch dispatch today
  step Open a draft release for the build jobs
       if: github.event_name == 'push'     <- safety 1, unpinned
       gh release create "$TAG" ... --draft <- safety 2, unpinned
```

`publish` was latent because it `needs` two environment-gated jobs. **`verify` has no such protection**, so this is a lower bar than the gap #2820 closed.

## The entry is bare, and that is the design decision

No draft exemption, and nothing reads the command text for flags. The shipped step already carries a step-level tag guard, so there is nothing to exempt.

Two proposals were tried and dropped before this one:

- **A `--draft` word-boundary regex.** `--draft=false` is a superstring of `--draft`, so the naive form disarms the mechanism #2820 added; and a *comment* naming the flag exempts a command that no longer passes it.
- **A command splitter** (drop comment lines, join continuations, split on `;`/`&&`/`||`). Better, and still defeated by a trailing inline comment — stripping those correctly means knowing whether a `#` is inside quotes, which is a shell, not a gate.

Both stop being necessary once the question is *"does this need expressing at all"* rather than *"how do I express `--draft` safely"*.

## Two costs, in the header rather than discovered later

**Deleting `--draft` while keeping the tag guard is not a finding.** That is a real defect — the tag path would open a published, asset-less release, which the comment above `publish` explains is visible to beta updaters immediately — but it is a different property. This check asks whether something outward-facing is reachable from a dispatch, not whether the release starts as a draft. That wants its own check.

**A drafted `gh release create` on a deliberately dispatch-only path is a false positive**, and the finding reads `no condition`. Measured on a fixture respelling the `release` job's step with the CLI. It fails closed, but the repair is an exemption like `action-gh-release`'s, **not** a tag guard on a job that is dispatch-only by design — and the message will not say that, so the header does.

## Evidence

Real workflow, every arm applied and reverted:

```
shipped                            rc=0
M-B  step tag guard deleted        rc=1   the reachable hole
M-A  --draft deleted, guard kept   rc=0   deliberate, documented
restored                           rc=0
```

```
entry removed:  suite 25 pass / 2 FAIL
as shipped:     suite 27 pass / 0 fail
pnpm run test:scripts   rc=0   396 tests / 396 pass / 0 fail / 0 skipped
```

Two of the four new tests are the documented costs — they assert current behaviour so the omission and the false positive are decisions rather than surprises. They are labelled as such and not counted as coverage; **the entry is defended by the other two.**

### DIFF
--- a/scripts/ci/check-release-upload-guards.mjs
+++ b/scripts/ci/check-release-upload-guards.mjs
@@ -54,6 +54,24 @@ export const SUBMISSIONS = [
   // --draft=false` is the step that takes a draft release public, and a public
   // release is what electron-updater's clients poll.
   { store: 'a public GitHub Release', run: '--draft=false' },
+  // `gh release create` opens a release, and the only shipped one is guarded at
+  // the step (`verify`'s `Open a draft release for the build jobs` carries
+  // `if: github.event_name == 'push'`), so there is NO draft exemption here and
+  // nothing reads the command text for flags. Two consequences, both deliberate:
+  //
+  //   * Deleting `--draft` while KEEPING the tag guard is not a finding. That is
+  //     a real defect - the tag path would open a published, asset-less release,
+  //     which the comment above the `publish` job explains is visible to beta
+  //     updaters immediately - but it is a different property. This check asks
+  //     whether something outward-facing is reachable from a dispatch, not
+  //     whether the release it opens starts as a draft. That wants its own check.
+  //
+  //   * A DRAFTED `gh release create` on a deliberately dispatch-only path IS a
+  //     false positive here, and the finding will read `no condition`. Measured:
+  //     respelling the `release` job's step with the CLI reds it. The repair in
+  //     that case is an exemption like the one below, NOT a tag guard on a job
+  //     that is dispatch-only by design. Do not follow the message off a cliff.
+  { store: 'a GitHub Release', run: 'gh release create' },
   // action-gh-release publishes unless the release is explicitly a draft. The
   // dispatch-only `release` job sets `draft: true`, which is why it is not a
   // finding today - but flipping that one word publishes from a branch, and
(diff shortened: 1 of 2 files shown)


ankit-yc (approved):
Approving. I drove every arm myself in a detached worktree at `96b45ae51` rather than reading the evidence block, and added one the description does not claim.

```
as shipped, real workflows                                  rc=0
M-B  step-level tag guard deleted on the release-opening step
       -> job 'verify' step 'Open a draft release for the build jobs'
          submits to a GitHub Release with no condition        rc=1
M-A  --draft deleted, guard kept                             rc=0   (documented cost)
M-C  new SUBMISSIONS entry removed AND guard deleted          rc=0
restored after each arm, tree clean                          0 diff lines
pnpm run test:scripts                                        rc=0, 396/396, 0 fail
```

**M-C is the arm that matters and it was only exercised against the test suite before.** With the `gh release create` row removed, deleting the real guard goes silent on the real workflow. So the entry is what catches the reachable hole, not an incidental effect of another row. Apply-checks printed for each mutation — M-A initially looked inert to me because `--draft=false` in `publish` also matches a bare `--draft` grep; the diff shows the `verify` flag did come out.

One non-blocking observation, and it is inherited rather than introduced here. `classifyStep` matches with `run.includes(...)`, so a step whose script merely *mentions* the command in a shell comment is a finding:

```
run: block containing "# opened by verify with gh release create"   -> 1 finding
run: block containing a real invocation                             -> 1 finding
CONTROL neither                                                     -> 0
CONTROL unrelated comment                                           -> 0
```

The same is true of `altool --upload-app` and `--draft=false` today, so this is a property of the matcher and not a regression. It fails closed, and the header already warns against following the message off a cliff, so it is noise rather than a hole — worth knowing if someone later writes a comment about release mechanics inside a `run:` block.

The two cost-recording tests are labelled as such and the entry is defended by the other two; that is the right split and I agree they should not be counted toward coverage.

PR #2823, opened 2026-09-06
fix(frontend): declare each dark token once, and read the AA bar off the file

### LINKED ISSUE #2822: The dark theme declares 80 tokens twice, and no test sees --blue-strong move off its 0.04 AA margin
## What

`html[data-theme='dark']` in `apps/frontend/src/app/globals.css` spans lines 820-1148 and declares
**174 custom properties, 80 of them more than once**. Every duplicate pair currently holds the same
value, so nothing renders wrong today. The hazard is that **an edit to the first copy of a
duplicated token is silently inert** — the later declaration in the same rule wins at equal
specificity — and **no test in the repo notices a change to the value at all**.

Both halves are measured below, each with a control.

## Editing the first copy has no effect

`--blue-strong` is declared at `globals.css:866` and again at `globals.css:993`, both at brace
depth 1 inside the single rule opened at line 820 (verified by a brace-depth scan with comments
stripped, not by eye).

```
mutation   line 866 only:  --blue-strong: #ff0000
render     token resolves to #2f74d9, cell renders rgb(47, 116, 217)   NO EFFECT

control    line 993 only:  --blue-strong: #ff0000
render     token resolves to #ff0000, cell renders rgb(255, 0, 0)      effect
```

Measured in a browser on `appointments-calendar-phonedaystrip--dark-selected-cell` at 390px dark.
The control is what makes the first result meaningful: the render is demonstrably capable of
changing, so "no effect" is the cascade and not a stale page.

## No test catches a change to the token

`--blue-strong` in dark is `#2f74d9`, which is **4.54:1 against white — 0.04 above the 4.5 bar**.
It is consumed by 22 non-test, non-story files.

```
mutation   BOTH declarations -> #5b93e8   (white on it = 3.10:1, a real AA failure)
suite      pnpm --filter frontend test:ci
           Test Suites: 1042 passed, 1042 total
           Tests:       13201 passed, 13201 total       rc=0
```

The nearest guard, `__tests__/components/Calendar/responsive/contrastProbe.test.ts:73`, asserts

```js
expect(measureContrast(mount('rgb(255, 255, 255)', 'rgb(47, 116, 217)')).ratio).toBe(4.54);
```

`rgb(47, 116, 217)` is a **literal carried by the test**, not a value read from `globals.css`. It
pins the arithmetic of `measureContrast`, which is worth pinning, but it cannot see the token move.
That is the whole mechanism: the guard's input is a copy of the artefact rather than the artefact.

## Why it matters now

The 0.04 headroom is not confined to one surface. Claude L3 measured the same `4.54` on the
inventory filter chips' dark count badge (#2796) and I measured it on the phone day strip's
selected cell (#2803), independently and on different components — it is one shared token, so any
nudge moves every consumer at once, below AA, with a green suite and a green PR.

## Suggested direction, not a prescription

1. Delete the shadowed declarations so each token is declared once in the rule. Purely mechanical
   — all 80 pairs are value-identical, so this cannot change a rendered colour, and that invariant
   is worth asserting in the same change.
2. Give the contrast guard the real value: read `--blue-strong` out of `globals.css` (or out of a
   rendered element) instead of restating it, and assert `>= 4.5` rather than `=== 4.54`. Then
   mutate the token and watch it fail, because a guard that has never failed is not yet a guard.

Not urgent and not a live defect: the shipped value clears AA in both themes. Filing it so the
next person who nudges a token finds this rather than a green build.

Found during nightly QA on `dev @ 69ea5e339`; tree verified byte-identical to `HEAD` afterwards.

aupyay:
## PR Checklist

Two halves of #2822, both measured there.

`html[data-theme='dark']` in `apps/frontend/src/app/globals.css` declares 174 custom properties across 254 declarations - 80 tokens appear twice. Every pair holds the same value, so nothing renders wrong and nothing can. The symptom is that at equal specificity the later declaration wins, so **an edit to the first copy is silently inert**. `:root` had one more of the same shape, `--blue-text`, whose own surviving comment says the token is "aliased, never re-stated".

And no test could see the value move. `--blue-strong` is `#2f74d9` in dark, **4.54:1 against white - 0.04 above the bar** - and it is the fill under white text at 30 sites in 21 files, so one nudge moves every consumer at once. The nearest guard asserted a literal the test carried:

```js
// contrastProbe.test.ts:73
expect(measureContrast(mount('rgb(255, 255, 255)', 'rgb(47, 116, 217)')).ratio).toBe(4.54);
```

That pins the arithmetic of `measureContrast`, which is worth pinning and is untouched here. It cannot see the token move, because its input is a copy of the artefact rather than the artefact.

## What is the new behavior?

**Each token is declared once.** 81 shadowed declarations removed (80 in the dark rule, 1 in `:root`). Two comment blocks went with them because they were byte-duplicated at the surviving declaration; leaving them would have parked prose above tokens it does not describe. The merge banner in the dark rule now records the shape instead of warning against a reorder that no longer matters.

**This cannot change a rendered colour, and that is checked rather than asserted.** Every rule's winning value and every resolved literal in both themes, before and after:

```
8 rules, 370 resolved tokens per theme
diff before.json after.json   ->   rc=0, identical
```

Section A and section B were not a paste of one another - A has 36 properties B lacks, B has 58 A lacks, and the overlap is 80 with zero differing values. So neither section could simply be deleted; only the shadowed lines could go.

**Three test changes.**

- `__tests__/support/globalsTokens.ts` - one reader for the file: rules, per-theme maps (`@theme` and `:root` into light, every dark rule accumulating in file order so the later wins), `var()` chain resolution, and a resolver that **throws** on a token resolving to nothing rather than handing a probe an empty string it reads as black on white. `statusPillContrast.test.ts` now shares it instead of carrying a second copy - two copies of a reader is the same drift this issue is about.
- `blueStrongContrast.test.ts` - resolves `--blue-strong` and `--white-text` out of `globals.css` and asserts **the bar, not the number**, in both themes, at the smallest size any consumer renders (10px/700, so the strict 4.5 and never the 3.0 large-text escape). Includes a control that the two themes resolve to *different* literals, without which the dark arm would be a second copy of the light one.
- `customPropertyShadowing.test.ts` - fails on any custom property declared twice inside one rule. Duplication *across* rules is untouched: that is how theming and scoping work.

## Evidence

Full frontend suite at `6c1f9903f15ffecc9fc37a311cb795b7d38e1afb`:

```
npx jest --ci      rc=0    Test Suites: 1044 passed    Tests: 13210 passed
pnpm --filter frontend run type-check   rc=0
pnpm --filter frontend run lint         0 errors, 2 warnings (both pre-existing, in files this PR does not touch)
```

Baseline on `dev` was 1042 suites / 13201 tests. The delta is exactly the two new files and their 9 tests.

**Every new guard was mutation-checked. A guard that has never failed is not yet a guard.**

```
ARM A  --blue-strong -> #5b93e8 (white on it = 3.10:1)
       blueStrongContrast.test.ts    rc=1   1 failed, 3 passed   <- the dark arm
       contrastProbe.test.ts         rc=0   9 passed             <- the literal guard, still blind, as filed

ARM B  reintroduce one duplicate declaration in the dark rule
       customPropertyShadowing       rc=1   "--blue-strong at 914 | 915 - all but the last are inert"

ARM C  restore the pre-PR globals.css
       customPropertyShadowing       rc=1   81 offenders - it would have caught what shipped

ARM D  break the reader so resolve() returns ""
       blueStrong + statusPill       rc=1   14 failed, "did not resolve to a colour in light"
                                            - loudly, not as a comfortable pass on a dead instrument

ARM E  give --white-text a translucent dark override: rgba(255, 255, 255, 0.75)
       blueStrongContrast            rc=1   2 failed
                                            "--blue-strong dark: 3.27:1 (needs 4.5:1) -
                                             rgba(255,255,255,0.75) on rgb(47,116,217) at 10px/700"
```

Arm D is the one that matters most: an empty string measures as black on white, which is a very comfortable 21:1.

## What this does not cover

- The consumers are not driven individually. The pairing is asserted at the token level - white on `--blue-strong` - because that is what all 30 sites use; a per-site render assertion would be 30 copies of one measurement. The predicate for that count is in the test header.
- No Aikido scan ran locally (no MCP or scan skill available in this session); CI's scan is the check.

## Related Issue(s)

Fixes #2822

### DIFF
--- a/apps/frontend/jest.config.ts
+++ b/apps/frontend/jest.config.ts
@@ -341,6 +341,9 @@ const ESM_PACKAGES = [
   'direction',
   'bcp-47-match',
   'css-selector-parser',
+  // postcss reaches for nanoid at load; the reader-agreement guard parses
+  // globals.css with it. ESM-only, no CJS build.
+  'nanoid',
   '@ungap/.*',
   'entities',
   'highlight\\.js',
--- a/apps/frontend/src/app/__tests__/support/globalsTokens.ts
+++ b/apps/frontend/src/app/__tests__/support/globalsTokens.ts
@@ -1,0 +1,152 @@
+/**
+ * One reader for `globals.css`, shared by every guard that needs the artefact
+ * rather than a copy of it.
+ *
+ * The failure this exists to prevent is the one recorded on #2822: a contrast
+ * guard that restates a token's value as a literal keeps passing after the
+ * token moves, because its input is a transcript of the file rather than the
+ * file. Anything asserting on a token colour should resolve it through here.
+ *
+ * Kept in `__tests__/support/`, which `jest.config.ts` lists in
+ * `testPathIgnorePatterns` - a module of helpers under `__tests__` is otherwise
+ * collected as a suite and fails for having no tests.
+ */
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+
+export type Declaration = { line: number; prop: string; value: string };
+
+export type Rule = {
+  selector: string;
+  /** 1-based line of the `{` that opens the rule. */
+  open: number;
+  /** 1-based line of the `}` that closes it. */
+  close: number;
+  /** Custom properties declared directly in this rule, in file order. */
+  declarations: Declaration[];
+};
+
+const DECLARATION = /^\s*(--[\w-]+)\s*:\s*([^;]+);/;
+const VAR_CHAIN = /^var\(\s*(--[\w-]+)\s*(?:,\s*([^)]*))?\)$/;
+
+/**
+ * Top-level rules and the custom properties declared directly inside them.
+ *
+ * Line comments are stripped before parsing so a token named inside prose is
+ * not read as a declaration, and brace depth is tracked so a declaration
+ * nested one level further down does not attach to the enclosing rule.
+ */
+export const parseRules = (css: string): Rule[] => {
+  const rules: Rule[] = [];
+  const lines = css.split('\n');
+  let depth = 0;
+  let buffer = '';
+  let current: Rule | null = null;
+
+  for (let i = 0; i < lines.length; i += 1) {
+    const line = lines[i].replace(/\/\*.*?\*\//g, '');
+    const decl = DECLARATION.exec(line);
+    if (depth === 1 && decl && current) {
+      current.declarations.push({ line: i + 1, prop: decl[1], value: decl[2].trim() });
+    }
+    for (const ch of line) {
+      if (ch === '{') {
+        if (depth === 0)
+          current = { selector: buffer.trim(), open: i + 1, close: -1, declarations: [] };
+        depth += 1;
+        buffer = '';
+      } else if (ch === '}') {
+        depth -= 1;
+        if (depth === 0 && current) {
+          current.close = i + 1;
+          rules.push(current);
+          current = null;
+        }
+        buffer = '';
+      } else {
+        buffer += ch;
+      }
+    }
+    if (depth === 0) buffer = '';
+  }
+  return rules;
+};
+
+/**
+ * Properties this rule declares more than once, with every declaration.
+ *
+ * At equal specificity the last one wins, so every earlier copy is dead text
+ * that an editor will nonetheless read, edit, and watch do nothing.
+ */
+export const duplicatesWithin = (rule: Rule): { prop: string; declarations: Declaration[] }[] => {
+  const byProp = new Map<string, Declaration[]>();
+  for (const d of rule.declarations) {
+    const seen = byProp.get(d.prop);
+    if (seen) seen.push(d);
+    else byProp.set(d.prop, [d]);
+  }
+  return [...byProp.entries()]
+    .filter(([, declarations]) => declarations.length > 1)
+    .map(([prop, declarations]) => ({ prop, declarations }));
+};
+
+const isDarkRule = (selector: string) => /data-theme\s*=\s*['"]?dark/.test(selector);
+
+/**
+ * `@theme` is Tailwind 4's base layer and is theme-independent, so it belongs
+ * in the light map or tokens that only alias through it resolve to nothing -
+ * and a missing value would compare "" against "" and read as a pass.
+ * More than one rule matches the dark selector, so these accumulate in file
+ * order and let the later one win, exactly as the cascade does.
+ */
+export const themeMaps = (rules: Rule[]) => {
+  const light = new Map<string, string>();
+  const dark = new Map<string, string>();
+  for (const rule of rules) {
+    const isBase =
+      rule.selector.split(',').some((part) => part.trim() === ':root') ||
+      rule.selector.trim().startsWith('@theme');
+    if (!isDarkRule(rule.selector) && !isBase) continue;
+    for (const d of rule.declarations) {
+      if (isDarkRule(rule.selector)) dark.set(d.prop, d.value);
+      else light.set(d.prop, d.value);
+    }
+  }
+  return { light, dark };
+};
+
+export const GLOBALS_CSS_PATH = join(process.cwd(), 'src/app/globals.css');
+export const GLOBALS_CSS = readFileSync(GLOBALS_CSS_PATH, 'utf8');
+export const RULES = parseRules(GLOBALS_CSS);
+
+const { light: LIGHT, dark: DARK } = themeMaps(RULES);
+export { LIGHT, DARK };
+
+/** Follows a `var(--a, fallback)` chain down to a literal colour. */
+export const resolve = (value: string, dark: boolean): string => {
+  let current = value;
+  for (let i = 0; i < 12; i += 1) {
+    const m = VAR_CHAIN.exec(current.trim());
+    if (!m) return current.trim();
+    const next = (dark ? DARK.get(m[1]) : undefined) ?? LIGHT.get(m[1]) ?? m[2];
+    if (next === undefined) return current.trim();
+    current = next;
+  }
+  return current.trim();
+};
+
+/**
+ * `resolve`, with the silent miss made loud.
+ *
+ * A token that resolves to nothing hands a contrast probe an empty string,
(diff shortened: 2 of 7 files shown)


ankit-yc (commented):
Reviewed at `6c1f9903f15ffecc9fc37a311cb795b7d38e1afb`, after re-running everything at the new head — the arms below were first taken at `4d879b84b` and re-taken after the ink-assertion commit.

## The safety claim holds on a second instrument

You asked for the before/after comparison through anything that is not your script. I ran `postcss` (8.5.26, from the repo's own store) over `git show origin/dev:...globals.css` and the PR's copy, collecting every custom-property declaration at any nesting depth, keeping the last per rule — which is what the cascade applies at equal specificity:

```
                   before   after
rules with tokens      8        8      same rule set: true
effective decls      635      635
duplicate decls       81        0
effective-value diffs between the two:  0
CONTROL  --white-text #ffffff -> #fefefe in the after file  ->  1 diff, named
```

81 is your number, reached by a real CSS parser. The two maps are identical and the comparator is shown able to say otherwise.

## Finding: the reader misses a declaration whose value is wrapped, so the duplicate scan's negative is narrower than it reads

`DECLARATION` requires the terminating `;` on the line that names the property. Prettier already wraps two declarations past the print width — `--glass-pill-shadow` at `globals.css:742` (`:root`) and `:844` (dark). Both are invisible to `parseRules`:

```
postcss     635 declarations
parseRules  633 declarations      <- the two wrapped ones
```

That is not a wrong colour today. It is that `customPropertyShadowing` cannot see a duplicate of a wrapped property, which is the one defect the file exists to catch:

```
plant a second --glass-pill-shadow in html[data-theme='dark']  ->  5 passed, rc=0    MISS
CONTROL: plant a second --glass-95 (single line), same rule, same place
                                                               ->  1 failed, named   caught
```

The control is the point — the plant is well-formed and in scope; only the wrap decides whether the guard sees it.

Two smaller consequences of the same line-oriented assumption, both latent rather than live:

- A token wrapped **only** in the dark rule would drop out of `DARK`, and `resolve` would fall back to the light literal — a dark assertion silently measuring the light colour. Your own control catches it for this token: wrapping `--blue-strong: #2f74d9` reds *"resolves --blue-strong to a different literal in each theme"*. The containment is per-token and lives in `blueStrongContrast`, not in the reader.
- A multi-line selector keeps only its last line, because `buffer` resets at every line end while `depth === 0`. `parseRules(":root,\n.yc-scope {\n --x: #111111;\n}")` returns the selector `.yc-scope` and `themeMaps` drops `--x` from the light map entirely. No token-bearing rule in `globals.css` is written that way today — postcss and `parseRules` agree on all 8 — but the file already contains selectors the parse reports as `)`.

## Suggested fix, run rather than described

Join forward to the `;` before matching. 12 lines in `globalsTokens.ts`, no change to any caller:

```diff
-    const decl = DECLARATION.exec(line);
+    let joined = line;
+    for (let j = i + 1; j < lines.length && !joined.includes(';') && /^\s*--[\w-]+\s*:/.test(line); j += 1) {
+      joined += ` ${lines[j].replace(/\/\*.*?\*\//g, '').trim()}`;
+    }
+    const decl = DECLARATION.exec(joined);
     if (depth === 1 && decl && current) {
-      current.declarations.push({ line: i + 1, prop: decl[1], value: decl[2].trim() });
+      current.declarations.push({ line: i + 1, prop: decl[1], value: decl[2].replace(/\s+/g, ' ').trim() });
     }
```

With it applied, on this branch:

```
the three suites unchanged            19 passed, 3 suites
wrapped duplicate planted             1 failed  - "--glass-pill-shadow at 844=... | 848=... - all but the last are inert"
declarations the reader now sees      635       - equal to postcss
```

Worth a fixture alongside the existing comment fixture, since a wrapped declaration is now something the parser claims to handle.

## The rest reproduces

```
ARM A  --blue-strong dark -> #5b93e8    blueStrongContrast   1 failed / 3 passed
                                        contrastProbe:73     9 passed        (still blind, as you filed)
ARM C  pre-PR globals.css restored      customPropertyShadowing  1 failed, 81 offenders
       sha of that file == git show origin/dev:apps/frontend/src/app/globals.css
baseline at 6c1f990                     19 passed, 3 suites
```

Arm C is the one that decides this, and it does what you said it does.

Holding the approval only for the wrapped-declaration fix — the guard's whole value is its negative, and it is a smaller change than the follow-up PR that would carry it. Push it and I will approve at the new head.

ankit-yc (approved):
Approving on two mutations run here, not on reading the patch.

Worktree detached at 81d91e6a3f1f4a469a749c1e6eedea582d149b4b, `git status --short` empty before and after each arm.

**Arm A - my plant.** A second, WRAPPED `--glass-pill-shadow` inserted into `html[data-theme='dark']`:

```
customPropertyShadowing   1 failed, 5 passed   rc=1
"html[data-theme='dark'] [819-1064]: --glass-pill-shadow at 844=... | 853=...
 - all but the last are inert"
```

844 and 853 are both wrapped declarations, and the message names both lines with their values. That is the case that returned 5 passed / rc=0 before.

**Arm B - the arm I most wanted re-taken, since the reader is now asserted equal to postcss.** Reverting the forward-join (`DECLARATION.exec(joinToTerminator(lines, i))` -> `DECLARATION.exec(line)`):

```
globalsReader   2 failed, 4 passed   rc=1
```

So the postcss equality is a live oracle and not a decorative assertion: the reader can no longer drift from a real parser without a named failure. That is the part that outlives this bug, and it is worth more than the dedupe.

Restored, `src/app/__tests__/ui/tokens/` 5 suites / 38 tests pass, rc=0. Checks at head: 43 SUCCESS, 5 SKIPPED, 0 failed.

One observation, not a blocker: the sixteen wrapped selector lists you found are the sharper half of this. A `themeMaps` that picks light from dark off a selector string is one re-wrap away from not recognising the dark rule, and the failure mode is dark assertions passing while measuring light - green, and wrong. The postcss equality now covers it, which is the right place for it to be covered.

PR #2824, opened 2026-09-06
chore(repo): remove unused design tokens package

aupyay:
The design-tokens workspace package had no runtime consumers and its published entrypoints were never produced. This removes the package, its frontend dependency, its lockfile importer, and updates the mobile token mapping documentation so it no longer refers to the deleted source.\n\nValidation:\n- staged secret scan, Prettier, ESLint, and secretlint passed\n- repository lint completed with 0 errors and existing warnings\n- repository type-check completed: 17 successful tasks\n- Storybook CI remains advisory pending owner approval for required-gate policy changes


ankit-yc (approved):
Verified inert before approving. No import, require, or dynamic import of @yosemite-crew/design-tokens anywhere outside the package (control: @yosemite-crew/types = 669 sites, so the query can match); no relative reaches into packages/design-tokens; no tsconfig path aliases. The one grep hit outside the package was a doc comment in apps/mobileAppYC/src/theme/semanticTokens.ts whose own text states the app does not depend on it.

core / Build shared packages and frontend passed at 6m47s, Build frontend (shared) at 6m31s, CodeQL javascript-typescript at 7m12s. That build is what confirms inertness with the real toolchain rather than my grep, which is why this approval waited for it.

The doc fix is right: both the header block and the inline design-tokens/src/color.ts citation were updated, and the drift note (twelve contract keys absent, six extra) was preserved by rewording to "former contract" rather than dropped. That drift is the fact worth keeping.

Recorded before it disappears: packages/design-tokens/src/radius.ts encoded a third card-radius contract (radiusRole.card = xl = 1.5rem = 24px) against the PIMS design system (--radius-card = --radius-xl = 20px) and the shipped app (18px). Both sources name the token xl and map card to xl, so the role structure matches and only the scale value differs. Deleting this removes a stale competitor, not evidence.

PR #2825, opened 2026-09-06
refactor(frontend): declare the card surface once and migrate the copies onto it

aupyay:
**Stacked on #2823** (`fix/2822-dark-token-shadowing`), which adds the postcss-backed globals.css reader this PR's tests use. GitHub retargets this to `dev` when #2823 merges. Read the diff against that base — it is one commit.

Closes the "44 hand-copied surface recipes" line of the UI freeze (criterion 2, surfaces).

## What was wrong

The warm-bone card frame — `--screen` ground, hairline border, 18px radius, the two-stop depth shadow — existed as exactly one CSS class (`.TableShell`) and as a hand-retyped Tailwind string in 39 more files and a hand-written CSS block in 12. 51 files, one recipe, nothing able to enforce it: a four-utility string spread across a `className` is not greppable, and a reviewer cannot see that two cards agree.

`.yc-card-surface` in `globals.css` is now the only declaration, beside `.yc-card-elevated`. `.TableShell` keeps its behaviour (`overflow`, `min-height`, `isolation`) and drops the four surface properties it used to restate — the file that owned the recipe was itself one of the fifty-one copies of it.

## The no-op is the claim worth checking, because the copies were not written the same way

Only the shadow was byte-identical across all 51. Background was spelled three ways and the border three ways:

| spelled | resolves to | light | dark |
| --- | --- | --- | --- |
| `bg-[var(--screen)]` | `--screen` | ✓ | ✓ |
| `bg-neutral-0` | `--color-neutral-0` → `--screen` | ✓ | ✓ |
| `var(--color-surface-card)` | → `--color-neutral-0` → `--screen` | ✓ | ✓ |
| `border-[var(--hairline)]` | `--hairline` | ✓ | ✓ |
| `border-card-border` | → `--color-neutral-200` → `--hairline` | ✓ | ✓ |
| `border-hairline` | `--color-hairline` | ✓ | ✓ |

So the migration changed no colour. That is a fact about `globals.css` that a retint can break silently, so `cardSurface.test.ts` asserts it in both themes rather than assuming it — with a control that the comparator can still tell `--band` from `--screen`, so the six passes are not a resolver returning a constant.

## Radius is a ladder, and reconciling it is NOT this PR

~20 page-level containers and tables at 18px, ~14 entity tiles at 16px, 6 in-panel sub-cards at 14px. `--tile` and `--inset` preserve those exactly.

Three sources currently disagree about what a card radius is — the shipped app says 18px, the design system's `--radius-card` says 20px, and the now-deleted `packages/design-tokens` said 24px. Collapsing 51 cards onto one value is a visible change to every card in the product and needs its own before/after, so this PR deliberately changes no pixel.

## Two call sites are named in the test rather than hidden

- **`ExploreCard`** steps its radius at `xl` (`rounded-[14px] xl:rounded-2xl`). `.yc-card-surface` is unlayered so it beats Tailwind's layered bg/border/shadow utilities — which also means a layered `xl:rounded-2xl` can no longer override it, and the tile would silently freeze at 14px on desktop. It resolves in the radius change, which removes the step.
- **The 11 CSS-declared copies.** Three carry the shadow over a transparent ground with no background of their own, and two have `:hover` / `--current` rules that override the frame, so applying an unlayered class there is a cascade question with its own screenshots, not a rename.

Both are enumerated in the test, so a *new* copy fails rather than hiding among the known ones.

## The gate collapses whitespace before matching

Prettier hard-wraps a `box-shadow` past the print width and **9 of the 12 CSS occurrences are wrapped**. A line-oriented scan finds 3 of 12 and reports a confident near-zero. There is a test arm that proves the flattened matcher sees a wrapped declaration and the raw one does not.

## Why unlayered, and why `--flat` exists

`.yc-card-surface` sits outside `@layer`, exactly as `.yc-glass-overlay` already does, so it beats Tailwind's layered utilities. That is what lets `.TableShell` drop the `!important` it carried on all four properties without changing which declaration wins. It is also why the speciality accordion's collapsed state needed `.yc-card-surface--flat`: its reduced `shadow-[0_1px_2px_var(--sh03)]` was a layered utility and would now lose silently.

## Verification

Each mutation reds a **named** test:

| mutation | test that failed |
| --- | --- |
| `.yc-card-surface` radius 18px → 20px | declares the four surface properties and nothing else |
| a `background` put back on `.TableShell` | leaves .TableShell holding behaviour only |
| a Tailwind copy planted in `TaskCard` | no hand-typed Tailwind copies beyond the one responsive tile |
| a **hard-wrapped** CSS copy planted in `DataTable.css` | confines the CSS-declared copies to the named files |
| `--color-neutral-0` repointed at `--band` | background spellings agree (both themes) |

```
frontend suite   1046/1046 files   13230/13230 tests   rc=0
npx tsc --noemit                                       rc=0
pnpm --filter frontend run lint                        rc=0  (2 pre-existing warnings, neither in a touched file)
```

`tsc` and the suite were run in this worktree at this commit. The pre-push hook's type-check was a turbo **cache replay** from another worktree, so it is not the evidence above — it is a different run and it is not being quoted here.

## Visual evidence: 466 before/after pairs, 465 of them pixel-identical

Two static Storybook builds — a clean checkout of this PR's base (`81d91e6a3`) and this head — captured
with Playwright at 1280x800, `deviceScaleFactor: 1`, reduced motion, and an injected stylesheet zeroing
every transition. Corpus is **every story in every migrated component**: all 40 have a colocated
`.stories.tsx`, 236 stories, both themes, both arms.

```
pairs compared      466        222,078,400 pixels
                    465  IDENTICAL, zero channels differ
                      1  44 px differ by 1/255, perceptible(>8) = 0
```

The single non-identical pair is `CompanionHistoryTimeline / Record drawer` in light. Its 44 pixels sit
in a 939x35 band at the top of the frame — the rounded ends of the sort pill and the search field,
neither of which is a `.yc-card-surface`. Antialiasing, not a surface change.

### Controls, because "no visual change" is what a broken capture also reports

```
the two arms really are different builds   'yc-card-surface' in built assets: after 44 files, before 0
the server stayed up                       index.json probed 200 at the START and END of each arm
the comparator can see a difference        light-vs-dark, same arm, same code path -> 100.0% of pixels
the harness is silent when nothing changes a no-op stylesheet moves 0 pixels
```

### The class is doing the work — read off the DOM, not inferred from the pixels

Identical pixels cannot by themselves distinguish "the class applied and reproduces the old frame" from
"the class is inert". 223 of 236 stories render at least one `.yc-card-surface`, 626 elements in total,
and their computed frames collapse to exactly the declared ladder and nothing else:

| variant | radius | light ground / border | dark ground / border | count |
|---|---|---|---|---|
| `.yc-card-surface` | 18px | `rgb(247,243,236)` / `rgb(229,220,207)` | `rgb(47,39,30)` / `rgb(64,54,43)` | 341 |
| `--tile` | 16px | same | same | 206 |
| `--inset` | 14px | same | same | 74 |
| `--flat` | 18px | same, contact shadow only | same | 4 |

### Sensitivity, measured rather than assumed

| mutation of `.yc-card-surface` | pixels differing | max delta |
|---|---|---|
| none (no-op stylesheet) | 0 | 0 |
| `border-radius` 18 -> 17 | 255 | 18 |
| `border-radius` 18 -> 14 | 362 | 29 |
| shadow alpha `.05` -> `.049` | 105 | 1 |
| ground `247,243,236` -> `247,243,235` | 109,082 | 1 |

A one-unit ground change moves ~110k pixels; a no-op moves none. That is what makes 465 pairs at
exactly zero a statement rather than an absence.

## The `!important` removal, answered in a real browser on both arms

`.TableShell` used to mark all four surface properties `!important` so it would beat Tailwind's
layered utilities. It now declares none of them and the frame comes from the unlayered
`.yc-card-surface`. That is a **cascade** change, and neither a stylesheet-parsing test nor a pixel
diff can answer it — specificity and `@layer` order decide it at compute time. So it was measured:
`getComputedStyle` in Chromium, on both arms, against the same six competitor probes, on
`RoomTable/Default` and two `TableHead` stories in both themes (6 subjects).

**The computed frame is identical on every subject**: `18px`, `rgb(247,243,236)` / `rgb(47,39,30)`
ground, `1px solid rgb(229,220,207)` / `rgb(64,54,43)`, same two-stop shadow.

| competitor applied to the same element | before | after | subjects changed |
|---|---|---|---|
| inert sheet, class not applied | shell holds | shell holds | 0/6 |
| layered (`@layer utilities`) utility, normal | shell holds | shell holds | 0/6 |
| layered utility, `!important` | probe wins | probe wins | 0/6 |
| **unlayered author class, normal** | **shell holds** | **probe wins** | **6/6** |
| unlayered author class, `!important` | probe wins | probe wins | 0/6 |
| **inline `style`, normal** | **shell holds** | **probe wins** | **6/6** |
| inline `style`, `!important` (control) | probe wins | probe wins | 0/6 |

Two rows change, and they are the honest cost of dropping `!important`: a consumer's plain unlayered
class or a normal inline style now wins where it used to lose. Everything the class was carrying
`!important` *for* — beating layered Tailwind utilities — is unchanged, which is what
`.yc-glass-overlay` already relies on.

**Nothing in the tree exercises either row.** Every rule matching a rendered `.TableShell` that
declares any of `background`, `border`, `border-radius`, `box-shadow` was enumerated on both arms:

```
before   [base]        *, ::after, ::before, ::backdrop   border:0px solid   (Tailwind's reset)
         [(unlayered)] .TableShell                        all four, !important
after    [base]        *, ::after, ::before, ::backdrop   border:0px solid
         [(unlayered)] .yc-card-surface                   all four, no !important
control  2585 / 2526 / 2522 selector rules walked per subject, 5-7 matching the element
```

No third rule on either arm, and **no `.TableShell` element carries a `style` attribute at all**.
`.table-list`, the one other author class on a shell, declares only `display` inside a media query.

Pixel coverage for `.TableShell` specifically: `RoomTable`'s 10 stories x 2 themes are in the 466
above, and 3 of the 4 `TableHead` stories x 2 themes were captured separately — **6 more pairs, all
identical**, with a cross-theme control on the same arm at 100%.

### Not covered

`GenericTable`'s three stories time out before `#storybook-root` gains a child, in both themes and on
**both arms** — pre-existing, and the cause is in the story rather than the component: `tableUtils.ts:13`
does `itemNoun.charAt(0)` on an `itemNoun` those stories do not pass, so the story throws before mounting.
Identical console error on the base commit. `GenericTable.tsx` therefore has no pixel evidence, but it is
covered by the computed-frame table above through its sibling shells. `TableHead/Consistency` fails the
same way on both arms. 10 further stories mount no card at all (empty, loading and permission-denied).

### DIFF
--- a/apps/frontend/src/app/__tests__/components/Stats/StatCardShell.test.tsx
+++ b/apps/frontend/src/app/__tests__/components/Stats/StatCardShell.test.tsx
@@ -110,7 +110,7 @@ describe('StatCardShell', () => {
     );
 
     const surface = container.querySelector('.min-h-89') as HTMLElement;
-    expect(surface).toHaveClass('gap-3.5', 'overflow-hidden', 'rounded-[18px]');
+    expect(surface).toHaveClass('gap-3.5', 'overflow-hidden', 'yc-card-surface');
     expect(surface).not.toHaveClass('min-h-75');
     expect(surface).not.toHaveClass('gap-2.5');
   });
@@ -122,12 +122,12 @@ describe('StatCardShell', () => {
       </StatCardShell>
     );
 
-    expect(container.querySelector('.min-h-75')).toHaveClass(
-      'bg-[var(--screen)]',
-      'border',
-      'border-[var(--hairline)]',
-      'rounded-[18px]',
-      'min-h-75'
-    );
+    /* The frame is `.yc-card-surface` in globals.css - one declaration of the
+       --screen ground, hairline border, 18px radius and two-stop shadow that
+       this file used to spell out as four utilities, and that 39 other files
+       spelled out identically. Asserting the class is what makes a silent
+       divergence impossible; asserting the utilities only proved this one
+       component still agreed with itself. */
+    expect(container.querySelector('.min-h-75')).toHaveClass('yc-card-surface', 'min-h-75');
   });
 });
--- a/apps/frontend/src/app/__tests__/components/chat/SharedEntityCard.test.tsx
+++ b/apps/frontend/src/app/__tests__/components/chat/SharedEntityCard.test.tsx
@@ -146,7 +146,7 @@ describe('SharedEntityCard', () => {
   it('renders the tokenized hairline card chrome when mine is true', () => {
     const { container } = render(<SharedEntityCard entity={makeEntity()} mine />);
     const card = container.firstChild as HTMLElement;
-    expect(card.className).toContain('border-[var(--hairline)]');
+    expect(card.className).toContain('yc-card-surface');
     // The legacy cool-slate accent border is gone (dark-mode safe).
     expect(container.querySelector('.border-primary-300')).not.toBeInTheDocument();
   });
@@ -154,8 +154,8 @@ describe('SharedEntityCard', () => {
   it('renders the same card chrome whether or not the message is mine', () => {
     const { container: a } = render(<SharedEntityCard entity={makeEntity()} mine />);
     const { container: b } = render(<SharedEntityCard entity={makeEntity()} mine={false} />);
-    expect((a.firstChild as HTMLElement).className).toContain('border-[var(--hairline)]');
-    expect((b.firstChild as HTMLElement).className).toContain('border-[var(--hairline)]');
+    expect((a.firstChild as HTMLElement).className).toContain('yc-card-surface');
+    expect((b.firstChild as HTMLElement).className).toContain('yc-card-surface');
   });
 
   it('renders an icon glyph for the entity', () => {
--- a/apps/frontend/src/app/__tests__/features/appointments/components/Calendar/responsive/__snapshots__/PhoneMonthOverview.test.tsx.snap
+++ b/apps/frontend/src/app/__tests__/features/appointments/components/Calendar/responsive/__snapshots__/PhoneMonthOverview.test.tsx.snap
@@ -82,7 +82,7 @@ exports[`PhoneMonthOverview — misc matches the month overview layout 1`] = `
     </div>
   </div>
   <div
-    class="rounded-2xl border border-[var(--hairline)] bg-[var(--screen)] px-2.5 pb-2.5 pt-3 shadow-[0_1px_2px_var(--sh03),0_8px_22px_var(--sh05)]"
+    class="yc-card-surface yc-card-surface--tile px-2.5 pb-2.5 pt-3"
   >
     <div
       class="mb-1.5 grid grid-cols-7"
--- a/apps/frontend/src/app/__tests__/ui/tokens/cardSurface.test.ts
+++ b/apps/frontend/src/app/__tests__/ui/tokens/cardSurface.test.ts
@@ -1,0 +1,189 @@
+/**
+ * The card surface is declared once, and the migration onto it was a no-op.
+ *
+ * The warm-bone frame - `--screen` ground, hairline border, the two-stop depth
+ * shadow - existed as exactly one CSS class (`.TableShell`) and as a hand-typed
+ * Tailwind string in 39 more files and a hand-typed CSS block in 12. Nothing
+ * could enforce it, because a four-utility string spread across a `className`
+ * is not greppable and a reviewer cannot see that two cards agree.
+ *
+ * Two claims are worth a test rather than a sentence:
+ *
+ *   1. Replacing `bg-neutral-0` / `border-card-border` with a class that writes
+ *      `var(--screen)` / `var(--hairline)` changed no colour. That is only true
+ *      because those utilities alias those tokens, in BOTH themes - a fact
+ *      about globals.css that a future retint can break silently.
+ *   2. No copy came back. The gate is the point of the whole exercise; without
+ *      it the 39 files return over a month and we do this again.
+ *
+ * The CSS-literal check collapses whitespace before matching. Prettier
+ * hard-wraps a `box-shadow` past the print width, and 9 of the 12 CSS
+ * occurrences are wrapped - a line-oriented scan finds 3 of 12 and reports a
+ * confident near-zero.
+ */
+import { readdirSync, readFileSync } from 'node:fs';
+import { join } from 'node:path';
+
+import postcss from 'postcss';
+
+import { GLOBALS_CSS, resolveColour } from '@/app/__tests__/support/globalsTokens';
+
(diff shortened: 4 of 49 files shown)


ankit-yc (approved):
Reviewed at 5b7e9d165 in a fresh worktree with deps installed, not from the PR page.

I checked the cascade claim by a different method than the runtime probe, so the two results are independent. Statically, across the tree:

  TableShell application sites          2  (GenericTable.tsx, RoomTable.tsx)
  inline style= on a TableShell element 0
  .table-list declares any of the four  0  (only display:none, inside a media query)
  control, same query on .yc-card-surface  7 matches -- the query can find them

That agrees with the browser walk: both cascade rows that flip are reachable in principle and unexercised in fact.

I also verified the hinge of the whole argument rather than taking it. .yc-card-surface at globals.css:1899 is NOT inside any @layer -- confirmed by brace depth, alongside .yc-glass-overlay (1855) and .yc-card-elevated (1867), the two unlayered companions it is modelled on. My first naive check said the opposite because it matched an '@layer components {' that had already closed; the depth-tracking version, with a control that correctly finds @layer base opening at 8 and closing at 23, is what settles it. Worth stating because if that class were layered, the whole competitor table would read differently.

Guard, run and mutated here, each restored, git status clean after every arm:

  cardSurface.test.ts                       13 passed  rc=0
  radius 18 -> 20 in .yc-card-surface       rc=1  reds '.yc-card-surface is the only declaration
                                                  of the card frame > declares the four surface
                                                  properties and nothing else'
  background restored to .TableShell        rc=1
  planted Tailwind copy of the recipe       rc=1
  all restored                              rc=0

The mutation reds the correctly named test, not merely something.

Separately, #2826 landed a mandatory colour ratchet on dev after this branch was cut, so I merged origin/dev into this head in a scratch worktree and ran the new gate against the result: rc=0, 575 across 115 files, with the gate's own selftest green in the same tree as the liveness control. This PR does not trip it.

The 466-pair capture and the two instrument failures disclosed against it -- a reducedMotion transition making a synchronous getComputedStyle return the pre-mutation value, and CSSStyleRule.cssRules existing but empty so a recursive walk skips every style rule and reports a confident zero -- are the reason the evidence is credible. Both were caught by a printed control rather than by care.

Full frontend suite is CI's; I ran the guard this PR adds and drove its mutations.

PR #2826, opened 2026-09-06
ci(frontend): a colour gate that can tell drift from a deliberate literal

aupyay:
Freeze criterion 3: the lint/CI gate that fails the build on a new hardcoded colour, plus the literal cleanup that ships with it.

## Why this is not a ban on hex

Aiming at literal zero would have caused a regression. Five files hold 15 literals that are *correct*, each verified at the call site rather than assumed:

| file | n | why it must stay a literal |
|---|---|---|
| `WalletPassPreview.tsx` | 9 | reproduces the wallet-pass service's fixed pass colours; Apple/Google Wallet render the real pass outside this app and do not follow its theme |
| `StripeOnboarding/index.tsx` | 1 | Stripe's brand purple, handed to Stripe Connect's own iframe, which cannot read this document's custom properties |
| `Availability.tsx` | 1 | toggle knob deliberately fixed white — the call site already records why: `--screen` flips, so the off knob was 1.15:1 in espresso and vanished |
| `InvoiceStep.tsx` | 2 | a `<style>` string injected into a print popup that loads none of our CSS |
| `themeCore.ts` | 2 | the `<meta name="theme-color">` fallback; browser chrome does not resolve `var()` |

So the baseline has two halves. `files` is debt, ratcheted per file. `justified` carries a written reason per entry — and it is not an exemption:

- an entry with no usable `why` exits **2** (could-not-run), not 0
- a justified file is pinned in **both** directions, because a ceiling would let a new literal hide behind a reason written for a different one
- `--update` can never move a file into `justified`; a reason is written by a person

## What the gate found, and this PR fixes

**`--scrim` is declared nowhere.** Both chat modals painted their backdrop from `var(--scrim, rgba(29,28,27,0.44))`, so the fallback was what painted — in *both* themes. The canonical PIMS modal backdrop is `--sh55`: `rgba(29,28,27,0.55)` light -> `rgba(0,0,0,0.8)` dark, with 25 other consumers. These two were the only overlays in the app whose scrim did not respond to the theme at all.

Two more of the same shape: `var(--success-soft, #e7f4ec)` under an ink that *did* flip (3.65:1 light, 2.13:1 dark, under the 3:1 a glyph needs — already fixed in the component, so the story's prose is corrected rather than left describing it as live), and `var(--on-cta, #fff)` painted on `var(--cta)`, which in dark is the bone `#f2ece1`.

`tokenFallbackClosure.test.ts` makes the class non-recurring. It plants the offending shape so the empty result is not a matcher that cannot match, plants the same text *inside a comment* so the prose recording these decisions is not what gets deleted to make it pass, and pins both directions of the vocabulary so an empty `declared` set cannot read as clean.

## Evidence

```
node scripts/ci/check-hardcoded-colours.mjs
  no new hardcoded colours: 575 to migrate across 115 files,
                            15 justified across 5 files, 2044 files scanned   rc=0
node scripts/ci/check-hardcoded-colours.mjs selftest      15 cases pass       rc=0
pnpm run test:scripts                    416/416          rc=0
pnpm --filter frontend run test          1043 suites / 13204 tests   rc=0
pnpm --filter frontend run lint          0 errors, 2 pre-existing warnings   rc=0
```

Every arm mutation-checked, each restored, `git status` clean after:

```
add a literal to a justified file       1 -> 2   rc=1, and it prints the reason on record
remove the justified literal            1 -> 0   rc=1   (pinned both ways, not a ceiling)
blank a "why" to "because"                       rc=2   could-not-run, not a pass
make justified a ceiling                         reds "pinned in BOTH directions"
drop the reason from the drift payload           reds "carries the reason on record"
let a justified file also sit in debt            reds "not also carried as debt"
revert the whole-file match to per-line          reds both wrapped-colour selftest cases, rc=2
restore the --scrim fallback to ShareEntityModal reds the closure test at that file and line
```

The ratchet also fired three times unprompted during this work, on my own removals — that is the arm that stops the baseline becoming a document.

## A hole closed while it was still latent

The scanner matched line by line. Prettier hard-wraps a long declaration inside the parens, and a line-oriented scan sees an opening with no closing paren on either line and reports neither — the gate keeps returning a clean, plausible number while the corpus grows. Matching over the whole stripped source closes it.

Latent here, not live: the tree holds exactly one wrapped functional colour today and it is inside a block comment, so the count did not move. Worth closing now rather than when a number does.

## Corpus, stated so it is not confused with the scorecard's

575 counts `.ts`/`.tsx`/`.css` under `apps/frontend/src`, **including stories**, excluding `globals.css` and tests, and counts `rgba()`/`hsl()` as well as hex. A count that excludes stories is a different, equally valid number over a different corpus — this one is the gate's, and it is the one CI holds.


ankit-yc (approved):
Reviewed at cf7cc756e in a fresh worktree, not from the PR page. Reproduced the gate's own numbers and then drove four mutations, each restored, git status clean after:

  clean tree                    rc=0  no new hardcoded colours: 575 across 115 files, 15 justified across 5, 2044 scanned
  selftest                      rc=0  15 cases pass
  planted #ff00ff               rc=1  New hardcoded colours. Use a token from globals.css instead
  justified literal REMOVED     rc=1  pinned in both directions, not a ceiling
  HEX matcher broken            rc=2  could not run: the scanner failed its own cases, so its count means nothing
  baseline missing              rc=2  could not run: no baseline at scripts/ci/...

The third is the one that earns the approval. A scanner that cannot recognise a colour would otherwise print a perfectly plausible 'no new hardcoded colours' and rc=0; because the selftest runs inside the same invocation as the measurement, a dead instrument is a could-not-run rather than a clean pass. No failure mode drains to green.

Also verified the shipped behaviour change independently: --scrim is declared 0 times in globals.css (control: --hairline, 5 declarations), so both chat modals were painting their inline fallback in both themes. --sh55 is declared at globals.css:771 light and :921/:1030 dark, so the replacement is theme-responsive where the fallback was not.

The justified list is the right shape: 15 entries across 5 files, each carrying a written reason tied to its call site, and pinned in both directions so a new literal cannot hide behind a reason written for a different one.

The _core.yaml placement is correct and now documented in the file: the frontend matrix only runs for workspaces the diff touched, and a colour literal arrives in exactly the change that does not look like a frontend change. I withdrew my request to split this into its own job -- that would force a needs edit on CI Required, and a required-check aggregate silently ceasing to cover this gate is worse than a check name that is imprecise.

Findings: 12 review comments, all at superseded shas (9112b4dcf 10, ef4b46dfa 1, 118f1a5ec 1), 0 at cf7cc756e. 39 checks green, 0 failing.

PR #2827, opened 2026-09-06
test(ci): PROOF ONLY - watch the colour gate fail, then close

aupyay:
Not for merge. Stacked on #2826 to demonstrate that the colour gate fails a real pull request, per the request to plant a colour and watch the build go red before calling 3b done. Will be closed and the branch deleted once the failing run is recorded.


aupyay:
Closing: the proof is recorded. The gate failed `core / Detect affected workspaces` on this PR with `New hardcoded colours ... LaunchGrowTab.tsx: 3 -> 4 ... 192: #ff00ff`, exit 1, and #2826 now emits a titled annotation so the cause is legible without opening the log. Branch deleted.

PR #2828, opened 2026-09-06
ci(repo): version UI adoption scorecard

aupyay:
Moves UI adoption measurement into the repository as a reviewable scorecard command, on one ref and one frontend source corpus, so the figures are comparable between commits.

**This body previously carried a figure set measured before the corpus exclusion landed, and claimed the empty-oracle control travelled "the real guard path". Both are corrected below.** The second commit is the review's three findings; the description is rewritten rather than appended to, because it lands in the merge commit as the record of what the tool measures.

## What it reports, on `dev` at `57e0db282`

```
UI freeze adoption @ 57e0db282
  Buttons:       170/358 files migrated (47%), 188 files remaining
  SegmentedPill:  16/23  files migrated (70%),   7 files remaining
  PanelStates:     4/9   files migrated (44%),   5 files remaining
  Overlays:      127/147 files migrated (86%),  20 files remaining
  StatusPill:     75/104 files migrated (72%),  29 files remaining
  [unit: files; corpus 762 .tsx in apps/frontend/src/app/, excl. ui/primitives, stories, __tests__]
```

The unit and the corpus size are printed on **every** line, not once in a header. Two separate errors during this PR's review were a numerator and a denominator drawn from different populations, and no line of output said which population it meant.

`PanelStates 4/17 (24%)`, `Overlays 127/195 (65%)` and `StatusPill 75/128 (59%)` were the old body's figures. They differ because of the anchoring fix below, not because the codebase moved.

## The three findings this PR was blocked on

**1. Anchoring — this is what moves the numbers.** Three of the five raw patterns were bare identifier alternations. Measured in-corpus:

```
matches of Modal|Dialog|Sheet|Popover                  2819
  standalone word                                       286
  inside a longer identifier                           2533   (90%)
top identifiers: setShowModal 406 · showModal 399 · Modal 278 · ModalHeader 177
```

39 of Overlays' 68 "remaining" files matched only inside a prop name or a util import path — `appointmentCentralModalUtils`, `activePopoverKey` — and contained no overlay of any kind. The patterns now anchor to a JSX tag name.

The obvious anchor `<(Modal|…)` over-corrects: it drops `<AppointmentCentralModal`, a hand-rolled overlay that belongs in the miss set, and shrinks the miss sets back toward empty-by-construction. The shipped form keeps those and excludes generics, because a generic argument is also a `<` followed by a name:

```
(^|[^A-Za-z0-9_])<[A-Za-z0-9]*(Modal|Dialog|Sheet|Popover)

                        bare      tag-name     shipped
Overlays  raw files      159         108          101      7 generic-only files drop out
CONTROL  anchored set not a subset of bare set:     0 files
```

**2. Units.** See the output block above.

**3. The controls could not fail — and nothing ran them.** `--selftest` printed `5 fixtures + 4 negative controls passed`. The empty-oracle arm asserted on an error it threw itself:

```js
assertThrows(() => { throw new Error('PanelStates: adoption oracle is unmeasured'); }, …)
```

Demonstrated at the artefact rather than argued, by deleting the guard that arm is named for from the old script:

```
old script, unmutated                          "4 negative controls passed"   rc=0
old script, oracle emptiness guard DELETED     "4 negative controls passed"   rc=0   <- unchanged
```

Separately, `git grep -rn 'ui-freeze-scorecard|selftest' .github/` shows **no workflow ever invoked that flag**, so the four controls were not running in CI at all.

The logic is now exported and covered by `scripts/ci/ui-freeze-scorecard.test.mjs` — the convention every other script in `scripts/ci/` already follows, and one `_core.yaml` actually runs via `pnpm test:scripts`. The oracle control uses the real git runner against a marker directory that does not exist, so it travels the same call path as the measurement, with the marker that *does* exist as its discriminator.

## Evidence

Ten single-line mutations of the source, each checked against the suite:

```
M1  oracle emptiness guard deleted           caught   <- the one the old selftest missed
M2  stale-fixture guard neutered             caught
M3  tautology guard neutered                 caught
M4  zero-total guard deleted                 caught
M5  generic anchor removed from tagName      caught
M6  tag anchor removed entirely              caught (2 tests)
M7  stories/tests corpus exclusion dropped   caught
M8  git rc=1 "no match" branch deleted       caught
M9  unit+corpus suffix dropped from report   caught
M10 bypassing counts every raw file          caught (2 tests)
survivors: 0
```

```
node --test scripts/ci/ui-freeze-scorecard.test.mjs      13 pass  0 fail  0 skipped   rc=0
pnpm test:scripts (whole suite)                         440 pass  0 fail  0 skipped   rc=0
prettier --check "scripts/ci/*.mjs"                     all matched files pass
pre-commit hooks (gitleaks, eslint --max-warnings=0, prettier, secretlint)   pass
```

The `440` needed a `pnpm install` in the worktree first: without it four unrelated suites abort on `Cannot find module 'semver'` and the run reports 292. A dependency-resolution failure and a logic failure are not the same result, and the totals differ by 148 cases.

## Known limits

- The unit is **files**, and the adoption criterion is stated in **usages**. They are different quantities: at ~2.3 `<button` per file, a file metric cannot register migration inside an already-matching file. Moving to usages is not a mechanical `-lE` → `-oE` swap — `git grep` emits `ref:path:matchtext` under `-o`, and the corpus filter tests the whole string, so the using-side numerator silently becomes `0` (measured: `0` vs `194`). That is a follow-up, and the unit is now printed so the ambiguity is visible on every line.
- Anchoring is syntactic. A tag rendered through a variable (`const El = Modal; <El />`) is not counted.
- The scorecard command itself is still not wired into any workflow; only its tests are. Wiring it is a separate decision about what should block a merge.


ankit-yc (approved):
Re-reviewed at a3a453cb0 in a fresh worktree. All four requested fixes are in, and I verified each rather than reading the diff.

  1. role= restored to role=['"]group['"] with fixture '<div role="group">'  -- SegmentedPill is 16/23 (70%) again, not the 16/113 the widened pattern produced
  2. the dead index-3 field is gone from the PanelStates tuple; fixture now binds to the intended element
  3. the '?? raw' fallback is gone -- assert(fixture && new RegExp(raw).test(fixture)) means a family WITHOUT a fixture now fails instead of self-certifying against its own pattern
  4. negative selftests added via an assertThrows helper

Run here:

  node scripts/ci/ui-freeze-scorecard.mjs --selftest   rc=0  '5 fixtures + 4 negative controls passed'
  node scripts/ci/ui-freeze-scorecard.mjs              rc=0  measured against dev 9d3674e07

The negative controls were the whole point of the fourth item, so I mutated the guards they protect, each restored, git status clean after:

  remove the zero-total throw                rc=1
  remove the tautology throw                 rc=1
  remove the stale-fixture assert            rc=1
  restored                                   rc=0

Each guard is caught by its own selftest. That is the difference between a control and a claim, and it is what the earlier revisions were missing.

Every family also reproduces my own independent measurement from earlier tonight, including the Buttons figure I had flagged as a 7-file discrepancy -- it now reports 170/358, which is the number I measured, so that gap is closed:

  Buttons 170/358 (47%)   SegmentedPill 16/23 (70%)   PanelStates 4/17 (24%)
  Overlays 127/195 (65%)  StatusPill 75/128 (59%)

One gap, not blocking and not a defect in this PR: the scorecard measures no field/input family, so it does not yet produce the criterion-10 field-adoption number (7 routed / 35 bypassing on the merged #2831). Adding a Fields family is the natural follow-up now that the primitive has landed on dev.

Correction I owe on this PR: I described it publicly as stalled with four fixes outstanding, reading 'last updated 22:21Z' as gone-quiet. 22:21Z is when the fixes were pushed. The work was done and waiting on my re-review, not stopped.

ankit-yc (changes requested):
Superseding my approval (id 5127003842) at the same head a3a453cb0. The code is correct; the coverage CLAIM is one arm wider than the coverage, and that is the arm that matters in a file whose job is to be trusted.

Claude L3 found it and I reproduced it independently:

  line 70   throw new Error('PanelStates: adoption oracle is unmeasured');   <- hand-thrown, inside the selftest
  line 72   'adoption oracle is unmeasured'                                  <- asserted on the line above
  line 115  if (oracleFiles.length === 0) throw new Error(...)               <- the REAL guard, never reached

Mutation, one line, syntax verified so it is not a broken-file red:

  BASE, unmutated                       selftest rc=0
  line 115 guard neutered to if(false)  selftest rc=0   <- GREEN, still prints '4 negative controls passed'
  node -c                               rc=0            <- valid file, so the green is real
  restored                              rc=0, tree clean

So three of the four negative controls fire and the fourth cannot. My own approval said 'I mutated all three guards to confirm they fire' -- that was accurate and incomplete: I mutated the three that live inside measure() and never mutated the oracle check, because it is inline in the top-level adoption.map() and no control reaches it. I approved a four-control claim having tested three.

Why it is worth blocking rather than filing as a follow-up: the sentence '5 fixtures + 4 negative controls passed' is exactly what would license deleting line 115 in six months. A gate that overstates its own coverage is the one artefact where landing it is worse than waiting, because everything downstream trusts that line.

The remedy is Claude L3's and it is small: lift the ls-tree + emptiness check into a callable oracleFiles(name, marker), then point the control at a marker directory that does not exist so it travels the measurement's real path. Roughly six lines.

Nothing else changed -- the shipped guard is present and correct, and the scorecard's numbers are not in question. Buttons 170/358, SegmentedPill 16/23, PanelStates 4/17, Overlays 127/195, StatusPill 75/128 all still reproduce against dev 9d3674e07 on my desk.

Re-approve on request once the fourth control travels the production path.

aupyay:
## Consolidated review notes for #2828 — three rounds in one place

Since this PR was approved and then re-blocked, three separate rounds of findings landed across a long discussion thread. Collecting them here so nothing has to be reconstructed from the conversation, and ordering them so the published numbers move **once** rather than three times.

All figures below measured on `dev` at `a9f9db3f0`, over the scorecard's own corpus (`apps/frontend/src/app/`, excluding `/ui/primitives/`, `*.stories.*`, `__tests__`).

---

### 1. The blocker: one of the four negative controls cannot fail

`--selftest` prints `5 fixtures + 4 negative controls passed`. Three of the four fire. The fourth does not:

```
mutant (base ± exactly one line)        selftest rc
BASE, unmutated                         0  green
guard neutered: stale-fixture           1  red
guard neutered: tautology               1  red
guard neutered: zero-total              1  red
guard neutered: empty-oracle            0  GREEN
```

The empty-oracle arm stays green both under a condition flip (`oracleFiles.length === 0` → `=== -1`) and under deleting the guard outright, because it hand-throws the error it then asserts on:

```js
assertThrows(
  () => { throw new Error('PanelStates: adoption oracle is unmeasured'); },
  'adoption oracle is unmeasured', 'empty oracle');
```

The production check it is named for is reachable and does fire — running the script against an older `dev` commit where `ui/primitives/PanelStates/` does not yet exist exits 1 with exactly that message. So this is the only untested arm over the mechanism that separates *fully migrated* from *pattern is dead*.

**Suggested fix (~6 lines):** the other three guards live inside `measure()`, which is why they are callable. Lift the `ls-tree` + emptiness check into `oracleFiles(name, marker)` and point the control at a marker directory that does not exist, so the control travels the measurement's real path.

While that line is untested, `4 negative controls passed` is a coverage claim one arm wider than the coverage — which is the sentence that would license deleting the guard later.

---

### 2. Pattern anchoring — do this first, it moves the numbers most

Three of the five `raw` patterns are bare identifier alternations, so they match substrings of longer identifiers rather than rendered elements:

```
in-corpus matches of Modal|Dialog|Sheet|Popover      2819
  standalone word                                     286
  inside a longer identifier                         2533   (90%)
top identifiers: setShowModal 406 · showModal 399 · Modal 278 · ModalHeader 177 · CenterModal 137
```

The effect on the published figures: of Overlays' 68 "remaining" files, 39 match only inside a prop name, a util import path or a helper identifier — `appointmentCentralModalUtils`, `activePopoverKey`, `appointmentPopoverId`, `PopoverCompanion` — and contain no overlay of any kind.

**But the obvious fix over-corrects.** Three options, measured:

```
FAMILY         bare                <[A-Za-z0-9]*(Word)      <(Word)
PanelStates    4/17  = 23%         4/9   = 44%              4/5   = 80%
Overlays       127/195 = 65%       127/154 = 82%            127/131 = 96%
StatusPill     75/128 = 58%        75/104 = 72%             75/85  = 88%
Buttons        170/358 = 47%       unchanged (already anchored on `<button`)
SegmentedPill  16/23  = 69%        unchanged (already anchored on the attribute)
CONTROL  raw file counts must be ordered   159 >= 108 >= 67   OK
```

`<(Modal|...)` is too strict: it drops `<AppointmentCentralModal`, which is a hand-rolled overlay and belongs in the miss set. It also shrinks Overlays' miss set to 4 and PanelStates' to 1 — back toward the empty-by-construction failure this file has already had three rounds of. The middle column anchors to a JSX **tag name** containing the word rather than to the bare word, and is the one I would take.

**Correction to this section, added after I tested my own recommendation the way I tested the other one.** A generic type parameter is also a `<` followed by a name, so the middle column counts `useRef<HTMLDialogElement>`, `useState<ModalView>`, `Partial<GroupModalProps>` and `Omit<PetPassportModalProps>` as overlays:

```
tag-name pattern, occurrences                       274
  preceded by an identifier char (a generic)         29
files matching the tag-name pattern                 108
  matching ONLY via a generic position                7   <- false positives
CONTROL  non-generic set minus full set               0
```

Requiring the `<` not to follow an identifier character fixes it, and allowing start-of-line costs nothing here (101 files either way, so no column-0 JSX tag is being missed):

```
(^|[^A-Za-z0-9_])<[A-Za-z0-9]*(Modal|Dialog|Sheet|Popover)

FAMILY         corrected            was (tag-name)
PanelStates    4/9   = 44%          4/9   = 44%     unchanged
Overlays       127/147 = 86%        127/154 = 82%   7 generic-only files drop out
StatusPill     75/104 = 72%         75/104 = 72%    unchanged
```

**That is the pattern I would take — the corrected middle column, not the one in the table above.**

---

### 3. Units — and the trap if this becomes usage-based

The adoption criterion is stated in **usages** (raw usages vs primitive usages); the scorecard reports **files**, because `grep -lE` counts files with matches. They are different quantities, not the same one disagreeing: at ~2.3 `<button` per file, and with 1820 lines of frontend change moving every family by zero, a file metric cannot register migration inside an already-matching file.

Whichever unit is chosen, please **print the unit and the corpus size on the same line as the percentage**. Two separate errors in the review thread were a numerator and a denominator drawn from different populations, and no output said so.

If it goes usage-based, `-lE` → `-oE` is **not** a mechanical swap:

- Bare alternations must be anchored first (see §2) or the counts inflate 10–50×.
- `git grep` emits `ref:path` under `-l` but `ref:path:matchtext` under `-o`, and the corpus filter tests the whole string. The using-side pattern contains the literal `ui/primitives/`, so every row is excluded and the numerator silently becomes `0` — measured, `0` vs `194`. Filter on the path field only.

---

### 4. The PR description

It currently reports a third figure set — `Buttons 177/442, SegmentedPill 16/24, StateScreens 0/17, Overlays 127/286, StatusPill 75/140`. Those denominators predate the corpus exclusion, `StateScreens` is not a family name the code uses, and the body's `\n\n` is literal so it renders as one run-on line. Worth refreshing in the same push, since it lands in the merge commit as the record of what the tool measures.

---

**Suggested order:** §2 anchoring → §3 units → §1 control → §4 description, so the figures settle once. §1 is the only one blocking the review; §2 is the one that changes the numbers most.

Nothing here was in the original four-fix ask — all of it was found after that push landed. Happy to re-review as soon as it is pushed.

ankit-yc (commented):
Holding on one finding, and it is the same defect class this PR's previous review round was about — a numerator and a denominator drawn from different populations. Everything else here checks out and I verified the two claims I most wanted to be true rather than reading them.

## What I confirmed first, because the rest only matters if this is right

**The new tests are actually gated.** Not "`_core.yaml:82` runs `pnpm test:scripts`" — the whole chain:

```
ci.yaml            on: pull_request: {}          (no branch filter)
  job `core`       uses: ./.github/workflows/_core.yaml, no `if:`
    job `detect`   name: Detect affected workspaces, no `if:`
      step         run: pnpm run test:scripts
        glob       node --test scripts/ci/*.test.mjs ...   <- picks the new file up, 12 siblings
  job `ci-required`  needs: [core, test, migration, sonar, frontend-quality, mobile-version]
                     failure/cancelled in any need -> exit 1
rulesets/3468092 (dev, active)  required_status_checks: "CI Required", "Supply Chain Required"
```

So a red arm in this file fails `core`, fails `CI Required`, and blocks a merge into `dev`. Worth stating because `ci-required` treats `skipped` as a pass, so an aggregate can cover a job that never ran — `detect` carries no condition, and that is what makes this chain real rather than nominal.

**No CI caller for the scorecard script itself**, as you said: `git grep -nE 'ui-freeze-scorecard|ui:freeze-scorecard' -- .github/` is rc=1, with `check-hardcoded-colours` found by the same grep at rc=0 as a control. The script stays a manual reporting tool; what is gated is its logic. That is the honest arrangement and the PR body says so.

**Mutation spot-check**, three of your ten re-run independently, plus a base row:

```
                                        rc   tests  pass  fail  skipped
base, unmutated                          0     13     13     0      0
M1  oracle emptiness throw deleted       1     13     12     1      0   not ok 1  (the one that survived before)
M8  `error.status === 1` branch deleted  1     13     12     1      0   not ok 11
M4  zero-total throw deleted             1     13     12     1      0   not ok 6
```

My first run of this harness was dead — I pointed `node --test` at `…scorecard.mjs.test.mjs`, so all four rows returned `rc=1` with an empty summary and would have read as "every mutation caught" if I had only looked at the exit codes. The base row is what caught it.

---

## The finding: the numerator is not restricted to the corpus the report names

`corpusSize` filters `.endsWith('.tsx')`. `matchingFiles` filters `inCorpus` — primitives, stories, `__tests__` — and **never filters extension**. So `report()` prints

> `[unit: files; corpus 762 .tsx in apps/frontend/src/app/, excl. ui/primitives, stories, __tests__]`

beside a `using` count that contains files which are not `.tsx`. At this head:

```
family          raw files        using files
Buttons         {.tsx: 252}      {.tsx: 169, .ts: 1}
SegmentedPill   {.tsx: 7}        {.tsx: 16}
PanelStates     {.tsx: 8}        {.tsx: 4}
Overlays        {.tsx: 101}      {.tsx: 123, .ts: 4}
StatusPill      {.tsx: 33}       {.tsx: 69, .ts: 6}
```

Eleven `.ts` files inside the migrated count, across three families. Several are barrel re-exports — `ui/index.ts`, `ui/overlays/index.ts`, `ui/overlays/Loader/index.ts` — which match `(from|import)[[:space:]][^[:space:]]*ui/overlays` and render nothing. A barrel is counted as a component that adopted the primitive.

Effect on the figures in the description:

```
              as published      numerator restricted to .tsx
StatusPill    75/104   72%      69/98    70%
Overlays     127/147   86%     123/143   86%
PanelStates    4/9     44%       4/9     44%   (unaffected)
```

**And no arm can see it.** I added the filter the report already claims — one `.filter((file) => file.endsWith('.tsx'))` in `matchingFiles` — and re-ran:

```
MINE  .tsx filter ADDED    rc=0   13 tests, 13 pass, 0 fail
```

Thirteen green with the filter and thirteen green without it. The suite is blind on the axis the output annotation is about, which is why the unit line reads as a fix for the population problem without being one.

Suggested: filter in `matchingFiles` (or fold `.tsx` into `inCorpus`), and give it an arm that flips — drop the filter and watch a named barrel reappear in the count. Please don't take my `.tsx` line as the fix on its word; whether the intended population is "renders JSX" or "any module in the corpus" is your call, and if it is the latter then the *report line* is what changes.

---

## Second, latent, no change strictly required — but it should be on the record

`measure()` validates the bypass pattern with `new RegExp(raw)`. `matchingFiles` applies the same `raw` through `git grep -lE`, POSIX ERE. Those are different dialects, and `SegmentedPill`'s `role=[\x27"]group[\x27"]` is read differently by each. Three-file probe repo, one file per case:

```
input                  git grep -lE (selects files)   new RegExp (validates the fixture)
role="group"           match                          true
role='group'           MISS                           true
role=xgroup2           MATCH                          false
control: /div/         all three                      -
```

In ERE the bracket is the literal set `{ \ x 2 7 " }` — so the pattern drops the single-quoted form it was written to catch, and accepts `role=` followed by any of those characters. **Zero effect on this tree**: `role='group'` occurs 0 times in the corpus, and both spellings give SegmentedPill the identical `16/23, 70%, 7 raw files`. Single-quoted roles do exist though — `role='button'`, `role='tab'`, `role='alert'` in 3 files — so this is a property of which values happen to be double-quoted today, not of the codebase's style.

The structural half is what makes it worth writing down: the staleness guard validates in the dialect that does not select the files, and the fixture it validates against (`<div role="group">`) is precisely the case where the two dialects agree. Any future pattern using a JS-only construct — `\d`, `\s`, `\b`, `(?:…)` — passes the guard while git grep reads something else.

---

Fix the population question and I will approve. The oracle control travelling the real `ls-tree` path, the `git rc=1` branch having a `128` sibling so it cannot pass by accident, and moving off a selftest flag nothing ran are all the right calls, and `# skipped 0` on every row of my table matters here — a suite that reports its cases as skipped is the other way this could have looked green.

ankit-yc (dismissed):
Approving at `cf0d709b5`. Both findings are closed at the mechanism, and the sign argument is better than the fix I suggested — it turned my "your call" into something that was not a taste question at all.

I had framed the population as a choice between "renders JSX" and "any module in the corpus". It was not: the raw patterns are JSX and cannot match a `.ts` file, so `bypassing` was already `.tsx`-only by construction and only `using` disagreed. Verified rather than accepted — across every counted population at this head, **782 files, 0 non-`.tsx`**, and the leak added to numerator and total while adding nothing to `bypassing`, which on a fraction below 1 can only push the figure up. The scorecard could overstate adoption and could never understate it. That is a stronger statement than "the numbers were off by a bit" and it belongs in the record.

Output at this head matches the corrected column exactly:

```
Buttons        169/357  47%      (was 170/358)
SegmentedPill   16/23   70%      (unchanged)
PanelStates      4/9    44%      (unchanged)
Overlays       123/143  86%      (was 127/147)
StatusPill      69/98   70%      (was  75/104)
corpus 762 .tsx
```

Folding the whole population definition into `inCorpus` and having `corpusSize` call it rather than re-spell `.endsWith('.tsx')` is the right shape — the defect was two places that had to agree, so the fix is one place.

## Mutations, re-run here against a snapshot taken before any edit

```
                                        rc   tests  pass  fail  skipped   failing test
base                                     0     16    16     0      0      -
M-A  .tsx filter deleted                 1     16    14     2      0      not ok 12, not ok 13
M-B  pattern back to the JS spelling     1     16    15     1      0      not ok 4
M-C  staleness back to new RegExp        1     16    15     1      0      not ok 5
M-D  comment reworded, must NOT fire     0     16    16     0      0      -
```

**M-B is the row that earns the new arm, and I nearly published it as a survivor.** My first attempt lost the pattern to shell quoting, never applied, and printed `rc=0, 16 pass` — the exact signature of a mutation that walked through the guard. What caught it was asserting the anchor was unique and counting occurrences on both sides of the edit (`1 -> 0` for the ERE spelling, `0 -> 1` for the JS one) before running anything. Re-run properly it takes down `not ok 4` and only that.

And your reading of it is right and is the difficult part: converting the existing fixture test to ERE is **necessary and not sufficient**, because `<div role="group">` is double-quoted and the two dialects agree there. The only arm that discriminates is the one with a single-quoted subject. A guard whose fixture the two candidate engines both accept cannot tell them apart — which is the same sentence as "its one fixture sits exactly where they agree", now with a test that would fail if someone reverted it.

## On the residual you stated

You flagged that `grep -E` is the same POSIX family as `git grep -lE` but not literally its engine. I measured that rather than leaving it as reasoning. Both patterns, four cells each, with a `CONTROL-MARKER` column that had to hit:

```
engine                                    role='   role="   xgroup2   CONTROL
-- shipped  role=['"]group['"]
/usr/bin/grep (BSD grep 2.6.0-FreeBSD)      Y        Y         n         Y
git grep -E  (git 2.50.1)                   Y        Y         n         Y
-- old      role=[\x27"]group[\x27"]
/usr/bin/grep                               n        Y         Y         Y
git grep -E                                 n        Y         Y         Y
```

Cell for cell, on both spellings. Two caveats on that table, because they are the limits of what I actually ran: those are the only two engines on this machine, and CI runs GNU grep on ubuntu, which I did **not** test. Also worth knowing rather than acting on — `matchesRaw` execs `grep` off `PATH`, so the engine it validates in is whatever the host resolves; node got `/usr/bin/grep` here, and a host with a different `grep` first would validate somewhere else. Neither of those changes my verdict; both are the kind of thing that is cheap to know now and expensive to rediscover.

The `# skipped 0` on every row of my table matters as much as the failures. Nothing in this file is a case reporting itself as skipped.

ankit-yc (approved):
Re-approving at `e068efc2c`. First the part that is mine.

**I approved `cf0d709b5` with `Aikido Security: check code` sitting at `failure`, and I never read the checks on this PR at all.** Confirmed rather than conceded — `commits/cf0d709b5/check-runs` returns 58 entries and `Aikido Security: check code failure` is among them. I ran `gh pr checks` on a different PR early in the session and then, on this one, went straight from "the mutations reproduce" to approval. Two rounds of review on this file, each producing a finding neither of us had, while a red security gate sat above both of them. Reproducing someone's evidence table is not the same act as reading the PR's state, and I substituted the first for the second.

That it cannot block a merge here is the aggravating fact, not the mitigating one. Required contexts on the `dev` ruleset are exactly `CI Required` and `Supply Chain Required`; Aikido is neither, so nothing but a reviewer was ever going to stop it — which is precisely the argument I made one gate over about `storybook-interactions.yml`, applied to a check I then walked past. Second one tonight, and I made the general version of the point before missing the specific one.

## The fix

```
scorecard output at cf0d709b5   Buttons 169/357 47%  SegmentedPill 16/23 70%  PanelStates 4/9 44%
                                Overlays 123/143 86%  StatusPill 69/98 70%
scorecard output at e068efc2c   identical, line for line
```

Those are two runs from two different commits in one session, so the no-op claim is a before/after rather than an assertion. `node --test`: **16 tests, 16 pass, 0 fail, 0 skipped, rc=0**. And the check itself flipped, which is the measurement that outranks the comment body: `failure` / "1 new LOW issue was introduced" at `81572d1e1` and `cf0d709b5`, `success` / "No new issues were introduced" at `e068efc2c`, one deleted line between them. You are right that reading `AIK_AI_prefer_simplicity` off the review-comment endpoint told you what was flagged and not that it was the whole of what made the check red; the conclusion flipping is the count.

Full check enumeration at this head, since that is the thing I failed to do last time: `total_count=24`, returned 24 — **0 failure, 0 cancelled**, 15 success, 3 skipped, 1 neutral, 5 still running (grype, detect, CodeQL javascript-typescript, Chromatic, build). Nothing red.

The `new Set(usingFiles)` at the call site was genuinely redundant — `measure`'s first line is `const using = new Set(usingFiles)` — and deleting the outer one is the right direction, since the inner conversion is the one that has to stay for the stub-fed tests.

## On the four empty surfaces

Your enumeration failure is the more useful half of that message and it is not the same error as mine. Four empty surfaces do feel like more support than one, and the arithmetic that makes them feel that way is wrong: they are not independent samples of "does the text exist", they are four draws from a set that never contained it. The bound you nearly published — *"the finding text exists only behind the dashboard login"* — would have been load-bearing for me, because I would have stopped looking on your authority. An absence is as wide as the surfaces queried and no wider, and the surface you did not query is not visible from the ones you did.

PR #2830, opened 2026-09-06
fix(ci): the colour baseline records the commit it describes

aupyay:
Follow-up to #2826, found by a reviewer within minutes of it merging.

## The bug

The baseline is a per-file count **of a specific tree**. Point it at a tree that predates it and every literal that commit *removed* reads as a literal this branch *added*. The counts cannot tell the two apart — both are "the tree has more than the baseline says".

Not hypothetical. Reproduced here in a detached worktree at `69ea5e339` (the PR base before #2826) carrying dev's new gate and baseline, which is exactly the pairing a reviewer hit:

```
New hardcoded colours. Use a token from globals.css instead:
  .../AddAppointmentCentralModal/index.tsx: 0 -> 2    208: #d32f2f   987: #d32f2f
  .../ForgotPassword.stories.tsx:           0 -> 2    237: #e7f4ec   238: #2f9e63
  .../NetworkDirectoryModal.tsx:            0 -> 1     92: rgba(29,28,27,0.44)
  .../ShareEntityModal.tsx:                 0 -> 1    149: rgba(29,28,27,0.44)
rc=1
```

Every one of those is a literal **#2826 deleted**. A real gate, a real red, real file names and line numbers, and a completely invented finding — one keystroke from telling another desk their PR introduced new colours.

## The fix: stop inferring, print the state

`--update` records `generated_at`. Any **failing** run then reports which of three cases it is:

| state | meaning | printed |
|---|---|---|
| `current` | the baseline's commit is in this history | nothing extra |
| `stale` | it is not | "the findings above may be its deletions — merge origin/dev and re-run" |
| `unknown` | git would not answer | said out loud, never treated as current |

Only `merge-base --is-ancestor` **rc=1** means "no". rc>1 is "could not answer" and maps to `unknown` — collapsing it into `stale` prints a confident "you are behind" every time git fails, and collapsing it into `current` restores the silence this exists to remove.

**Context only, never a new failure.** A branch merely being behind is not an error, and making it one would red every open PR the moment the baseline moves.

On the stale tree above, the same run now ends:

```
READ THIS BEFORE ACTING ON THE ABOVE. The baseline describes commit 1007378ea,
which is NOT in this tree's history.
Literals that commit REMOVED will appear above as literals this branch ADDED -
real file, real line, real colour, wrong conclusion.
Merge origin/dev into this branch and re-run before believing any of it.
```

## Evidence

The git runner is injectable so all four states are tested without depending on which commits a CI clone happens to have fetched — otherwise the test grades the checkout rather than the code.

```
gate                 575 to migrate / 115 files, 15 justified / 5, 2044 scanned   rc=0
selftest             15 cases pass                                                rc=0
scanner tests        30 pass, 0 fail   (24 -> 30)
pnpm test:scripts    422 pass, 0 fail
```

Three mutations, each reds a named test:

```
collapse rc>1 into stale        -> "merge-base failing for any OTHER reason is unknown, not stale"
treat an unknown commit as current -> "a commit this clone does not have is unknown, and says so"
stop recording generated_at     -> "the shipped baseline records the commit it describes"
```

And the three live states, run end to end: a genuine new literal on an up-to-date tree prints **no** staleness note; the pre-#2826 tree prints the stale note; a `generated_at` this clone does not have prints the "could not check" note.

Counts are untouched — 575 across 115 files, 15 justified across 5, before and after. The only diff to the baseline JSON is the one new field.

### DIFF
--- a/apps/frontend/src/app/ui/README.md
+++ b/apps/frontend/src/app/ui/README.md
@@ -63,6 +63,21 @@ Those live in the baseline's `justified` map, each with a written `why`. The
 list is not an exemption:
 
 - an entry with no usable reason fails the gate with exit 2, not a pass
-- a justified file is pinned in **both** directions, so a *new* literal cannot
+- a justified file is pinned in **both** directions, so a _new_ literal cannot
   hide behind a reason written for a different one
 - `--update` never moves a file into `justified`; a reason is written by a person
+
+### If the gate reports colours you did not write, read the last line first
+
+The baseline is a per-file count of a specific commit, recorded as `generated_at`.
+Run it on a branch cut before that commit and every literal the baseline's commit
+_removed_ appears as a literal your branch _added_ — real file, real line, real
+colour, wrong conclusion. On any failure the gate now says which case it is:
+
+```
+READ THIS BEFORE ACTING ON THE ABOVE. The baseline describes commit 1007378ea,
+which is NOT in this tree's history.
+```
+
+Merge `origin/dev` and re-run before believing the findings. If git cannot answer,
+it says that too rather than assuming the baseline is current.
--- a/scripts/ci/check-hardcoded-colours.mjs
+++ b/scripts/ci/check-hardcoded-colours.mjs
@@ -48,6 +48,7 @@ import { readFileSync, writeFileSync, readdirSync, statSync, existsSync } from '
 import { join, dirname, relative, resolve, isAbsolute, sep } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import process from 'node:process';
+import { spawnSync } from 'node:child_process';
 
 const repoRoot = join(dirname(fileURLToPath(import.meta.url)), '../..');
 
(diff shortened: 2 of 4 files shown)


ankit-yc (dismissed):
Reviewed at b9f0c217b in two fresh worktrees, driving all three states rather than reading the diff.

  clean tree                                   rc=0  575 / 115 files, 15 justified / 5, 2044 scanned
  non-justified DECREASE (removed a literal)   rc=1  'Colours were removed without retightening the baseline'
  STALE, reproduced end to end: worktree at 69ea5e339 with dev's gate and baseline copied in,
    merge-base --is-ancestor 1007378ea HEAD -> rc=1, correctly stale
                                               rc=1  + 'READ THIS BEFORE ACTING ON THE ABOVE ... NOT in this tree's history'
  UNKNOWN, generated_at is a sha not in the clone
                                               rc=1  + 'Could not check ... commit deadbeefd is not in this clone'
  UNKNOWN, generated_at absent entirely        rc=1  + 'Could not check ... the baseline records no commit'

The three-way split is the right one and unknown is never collapsed: contains.ok -> current, contains.status === 1 -> stale, everything else (rc>1, spawn error, null status, unresolvable HEAD, missing field) -> unknown. Collapsing rc>1 into stale would print a confident 'you are behind' every time git itself fails, and into current would restore exactly the silence this exists to remove.

One correctness coupling worth recording, because it is not local to the change: printing the note ONLY on a failing run is sufficient only because the ratchet is bidirectional. A stale baseline shifts counts in either direction, and both a gain and a loss fail -- I verified the loss arm above on an ordinary non-justified file, not just a justified one. If this gate ever became a ceiling that ignored decreases, a stale baseline could produce a silent false PASS and the note would then need to print on success too. Worth a comment beside the freshness call so a later ceiling-only change does not quietly break it.

Context-only rather than a new failure is the right call: a branch merely being behind is not an error, and making it one would red every open PR the moment the baseline moves.

On the motivation -- my own case was self-inflicted, I paired dev's baseline with a tree predating it while testing. The case this actually serves is the one that is nobody's mistake: any branch cut before 1007378ea gets the identical red without the developer doing anything wrong, and the counts genuinely cannot distinguish 'you added a colour' from 'the baseline removed one'. Both are 'the tree has more than the baseline says'.

ankit-yc (approved):
Re-reviewed at a70400c54 after the dismissal. Delta from b9f0c217b is a comment expansion in the scanner plus the new test -- no behavioural change to compare() or baselineFreshness(), confirmed by diff.

Driven in a fresh worktree:

  full scanner suite                          31 tests, 31 pass, 0 fail
  gate on clean tree                          rc=0  575 / 115 files, 15 justified / 5, 2044 scanned
  selftest                                    rc=0  15 cases pass

  MUTATION -- remove the decrease arm from compare(), the exact future change the
  coupling was about:
    not ok 12 - compare reports an improvement so the ratchet has to tighten
    not ok 31 - the staleness note may live under the failure arm only while a LOSS also fails
    29 pass, 2 fail;  restored, git status clean

Test 31 fires for the right reason and its NAME tells a future reader which dependency they broke, not merely that something broke. That is strictly better than the comment I proposed: a comment is read by everyone except the person about to make this gate a ceiling. Converting my suggestion into an assertion was worth the dismissal and the second pass.

The prose comment is there too, at the call site, and it names the failure direction explicitly -- a stale baseline becomes a SILENT FALSE PASS if decreases stop failing. Both halves are right: the comment explains, the test enforces.

I also re-confirmed the three freshness states still behave at this head via the suite; the earlier end-to-end runs at b9f0c217b (stale reproduced on a 69ea5e339 worktree, both unknown arms) are unchanged by this delta since neither function was touched.

PR #2832, opened 2026-09-07
refactor(frontend): complete textarea adoption

aupyay:
## PR Checklist

- [x] The PR title follows our guidelines: https://github.com/YosemiteCrew/Yosemite-Crew/blob/main/CONTRIBUTING.md#commit-message-convention.
- [x] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [x] All existing tests and lints pass.

## What is the current behavior?

Eighteen production files bypass the canonical form-field contract by rendering native textareas directly. The shared `Textarea` also requires a placeholder even though native textareas do not.

## What is the new behavior?

- Routes every production textarea through the canonical `Textarea` primitive.
- Preserves existing labels, values, handlers, disabled states, and sizing overrides.
- Makes the shared placeholder prop optional to match native textarea behavior.
- Replaces the touched invoice print fallback's HTML-string construction with DOM nodes and `textContent`.
- Adds regression coverage for the shared textarea contract and safe invoice print output.

Validation at `d21932ca8`:
- Full frontend suite: 1,049 suites and 13,241 tests pass.
- Affected-surface suite: 24 suites and 564 tests pass.
- Focused changed-file coverage: 100% statements and lines, 94.54% branches, 95.83% functions.
- Production build, TypeScript, pre-commit, and full pre-push lint/type-check hooks pass.
- Aikido: 21 changed TypeScript files, zero findings.
- React Doctor: 92/100, no issues.
- Desktop and phone Storybook checks pass in light and dark themes.
- Both new regression tests were mutation-checked.

## Related Issue(s)

Fixes #2829

PR #2846, opened 2026-09-08
ci(frontend): ratchet raw textarea out of production tsx

aupyay:
## What this closes

`#2832` migrated every production raw `<textarea>` to the canonical `Textarea` and the review closed with one gap on the record: *"the count is 19 → 1 and the survivor is the primitive — nothing stops it going back up."* This is that stop. It is 23 additive lines in `apps/frontend/eslint.config.mjs`; no product code changes.

Baseline on `dev` at `57e0db282`, measured rather than quoted:

```
/usr/bin/grep -rlE '<textarea' src --include='*.tsx'                       20 files
  ... minus __tests__ / *.test.tsx / *.stories.tsx                          1 file
  the survivor                                            src/app/ui/Input.tsx
```

## Why eslint and not a jest guard

The review suggested ~15 lines of jest on the `cardSurface.test.ts` template. `react/forbid-elements` already exists in the plugin set `eslint-config-next/core-web-vitals` brings in, so the rule is 8 lines of configuration plus its exemption and comments — and it reports at `line:col` in the editor before the file is ever committed, which a test that greps the tree cannot do.

## Evidence — four arms, driven by mutation, not by one passing run

A ratchet that has only ever been observed passing has told nobody anything, so each arm below had to be able to come out the other way.

```
A  planted <textarea> in a production file      rc=1   error at 2:11    FIRES
B  the same plant under __tests__/              rc=0   0 hits           correctly silent
B2 arm B with the `ignores` line deleted        rc=1   error at 2:11    <- so the ignores
                                                                          line is what makes
                                                                          B pass, not a
                                                                          global ignore
C  src/app/ui/Input.tsx as shipped              rc=0   0 hits           exempt
D  arm C with its exemption block deleted       rc=1   error at 27:6    <- so the rule does
                                                                          reach the primitive;
                                                                          the exemption is
                                                                          load-bearing, the
                                                                          rule is not dead
```

`B2` and `D` are there because `B` and `C` are both passes, and a pass is also what a rule that never runs produces. Each was checked by deleting exactly one thing and watching the same file change verdict. Probe files were removed afterwards; `find` for the probe paths returns empty and the worktree carries only the config change (`git diff --numstat` → `23 0 apps/frontend/eslint.config.mjs`).

## The whole-suite arm, with a base to compare against

```
                        rc   errors   warnings
base   57e0db282 tree    0        0          1
head   this change       0        0          1
```

The base arm was taken with the config file restored to its `HEAD` content in a tree `git diff --quiet HEAD` confirmed was otherwise identical to `57e0db282` — so the only difference between the two rows is this diff. The single warning is the pre-existing `sonarjs/no-nested-functions` that the config's own comment documents; this change neither adds nor clears it.

## Scope and known limits

- The rule keys on the **JSX element**, so it cannot see `React.createElement('textarea', …)` or a `dangerouslySetInnerHTML` string. It raises the cost of the regression, it does not make it unreachable.
- Stories and tests are deliberately exempt: several assert against the raw element, and forbidding it there would force test rewrites that have nothing to do with the field contract.
- Any future exemption has to be added as its own `files` entry with a reason on the line, which is why the primitive's exemption is written that way rather than as an inline disable.

### DIFF
--- a/apps/frontend/eslint.config.mjs
+++ b/apps/frontend/eslint.config.mjs
@@ -142,7 +142,7 @@ const eslintConfig = [
        (label, help, error, required, disabled, 40px control). #2832 took the
        raw-element count from 19 production files to 1; this keeps it there. */
     files: ['**/*.tsx'],
-    ignores: ['**/__tests__/**', '**/*.test.tsx', '**/*.stories.tsx'],
+    ignores: ['**/__tests__/**', '**/*.test.tsx'],
     rules: {
       'react/forbid-elements': [
         'error',
@@ -155,9 +155,11 @@ const eslintConfig = [
     },
   },
   {
-    // ui/Input.tsx is the primitive implementation and the one place the raw
-    // element may appear. Any other exemption needs a reason on this line.
-    files: ['**/ui/Input.tsx'],
+    // The primitive implementation, and the one place the raw element may
+    // appear. Pinned to the exact path so a future file whose name happens to
+    // end ui/Input.tsx does not inherit the exemption. Any other exemption
+    // needs its own entry with a reason on this line.
+    files: ['src/app/ui/Input.tsx'],
     rules: { 'react/forbid-elements': 'off' },
   },
 ];
--- a/apps/frontend/eslint.config.mjs
+++ b/apps/frontend/eslint.config.mjs
@@ -137,6 +137,29 @@ const eslintConfig = [
       '@next/next/no-img-element': 'off',
     },
   },
+  {
+    /* Criterion 10 ratchet: the canonical Textarea owns the field contract
+       (label, help, error, required, disabled, 40px control). #2832 took the
+       raw-element count from 19 production files to 1; this keeps it there. */
+    files: ['**/*.tsx'],
+    ignores: ['**/__tests__/**', '**/*.test.tsx', '**/*.stories.tsx'],
+    rules: {
+      'react/forbid-elements': [
+        'error',
+        {
+          forbid: [
+            { element: 'textarea', message: 'Use Textarea from ui/primitives; it owns the shared field contract.' },
+          ],
+        },
+      ],
+    },
+  },
+  {
+    // ui/Input.tsx is the primitive implementation and the one place the raw
+    // element may appear. Any other exemption needs a reason on this line.
+    files: ['**/ui/Input.tsx'],
+    rules: { 'react/forbid-elements': 'off' },
+  },
 ];
 
 export default eslintConfig;


ankit-yc (dismissed):
Approving. I re-ran the four arms myself on a clean checkout of `55722dd89` rather than reading them off the description, because arms B and C are exactly the pair a config block that never matched anything would also produce.

**The first run of my own harness was dead, which is the reason this is worth writing down.** I passed `--format unix`; that formatter is no longer bundled with core ESLint, so every arm exited `rc=2` with an empty result and *all five printed "0 forbid-elements hits"* — including the two that must fire. A table of five identical zeros is what a working exemption and a completely inert rule both look like. Re-run with `--format json` and a per-arm `files analysed` count:

```
arm                                  rc   files   result
A  <textarea> planted in src/app/ui/RatchetProbe.tsx    1     1    react/forbid-elements at 2:11
B  same plant under src/app/__tests__/                  0     1    0 errors
B2 arm B with the `ignores` line deleted                1     1    react/forbid-elements at 2:11
C  src/app/ui/Input.tsx as shipped                      0     1    0 errors
D  arm C with the exemption block deleted               1     1    react/forbid-elements at 27:6
```

`B2` and `D` are the load-bearing rows: each removes one line and the *same file* changes verdict, so B's silence is the `ignores` entry and C's silence is the exemption — not an unregistered rule. Both mutations were confirmed applied (`diff` against a snapshot taken before any edit, rc=1) and reverted from that snapshot; `git diff --quiet -- apps/frontend/eslint.config.mjs` is clean afterwards. For D I also loaded the mutated config through `import()` first, so the "it fires" row is not a config that failed to parse.

Whole-app lint at this head:

```
eslint . --ext .js,.jsx,.ts,.tsx     rc=0    3053 files    0 errors    1 warning
  src/app/hooks/useLazyAuthStore.ts:76:31  sonarjs/no-nested-functions
```

That warning is the pre-existing one the config's own comment documents, in a file this PR does not touch.

Two notes, neither blocking:

1. `files: ['**/ui/Input.tsx']` is a path suffix, not a pinned file. `git ls-files` has exactly one match today, so the exemption is tight as shipped — but a future `.../ui/Input.tsx` anywhere under the app inherits it silently. The comment already requires a reason per exemption line, which is the right shape; worth knowing the glob is broader than the one file it names.
2. `**/*.stories.tsx` is excluded alongside the test globs. Stories are not production, but they are the surface Chromatic renders, so a raw `<textarea>` added in a story is both unratcheted and visible. Fine as a deliberate line; just not the same argument as the test exclusion.

The stated limit is accurate and I checked it rather than taking it: the rule keys on the JSX element name, so `React.createElement('textarea', …)` is not covered.

aupyay:
Both review notes taken at `b3bdc17d4`. Each has a before/after arm on the same planted file rather than an argument.

**Note 1 — `**/ui/Input.tsx` is a path suffix.** Confirmed: a decoy at `src/app/components/ratchetprobe/ui/Input.tsx` containing a raw `<textarea>` linted clean (`rc=0`) under the suffix pattern — it inherited the exemption exactly as described. Pinned to `src/app/ui/Input.tsx` it errors (`rc=1`, `1:29`), and the real primitive still lints clean (`rc=0`), so the pin is not a silent kill.

**Note 2 — stories are not tests.** Checked the cost first: `0` of `671` `*.stories.tsx` files carry a raw `<textarea>` today, so the exclusion had no beneficiary and dropping it puts 671 Chromatic-rendered files in scope for free. A planted `<textarea>` in a story was silent before (`rc=0`) and errors now (`rc=1`, `1:27`).

Whole app after: **3053 files, 0 errors, 1 warning** — the pre-existing `sonarjs/no-nested-functions` in `useLazyAuthStore.ts`, unchanged.

Not changed, noted rather than quietly kept: `**/*.test.tsx` in that `ignores` list matches nothing today (every `*.test.tsx` is under `__tests__/`). Left as belt-and-braces because tests legitimately render raw markup; it is dead weight, not a hazard.

ankit-yc (approved):
Re-approving at `b3bdc17d4`. Both notes were taken, and both are now carried by an arm that flips rather than by an argument — re-run here from a clean checkout, not read off the description.

```
arm                                                    rc   files   result
A  <textarea> planted in src/app/ui/                    1     1     forbid-elements 2:11
B  same plant under src/app/__tests__/                  0     1     0 errors      (still ignored)
E  same plant in a .stories.tsx                         1     1     forbid-elements 2:11   <- was silent before
F  decoy at src/app/components/ratchetprobe/ui/Input.tsx 1     1    forbid-elements 2:11   <- was exempt before
C  the real src/app/ui/Input.tsx                        0     1     0 errors      (still exempt)
```

`F` and `C` are the pair that matter for the pin: the decoy is a path that merely *ends* `ui/Input.tsx`, and under the old `**/ui/Input.tsx` it inherited the exemption silently. Pinned, it errors — and `C` is what stops the pin from being a silent kill of the exemption it replaced. I built the decoy at a different path from yours and got the same verdict.

Whole app at this head:

```
eslint . --ext .js,.jsx,.ts,.tsx    rc=0   3053 files   0 errors   1 warning
  of those 3053, 671 are .stories.tsx      <- the newly in-scope set, analysed, clean
  src/app/hooks/useLazyAuthStore.ts:76:31  sonarjs/no-nested-functions   (pre-existing)
```

I re-derived both premises rather than accepting them, each with a control that had to be non-zero:

```
find src -name '*.stories.tsx'                       671    grep -rlE '<textarea' … *.stories.tsx    0
                                                            CONTROL '<button' in stories            81
find src -name '*.test.tsx' ! -path '*__tests__*'      0    CONTROL same find without the exclusion 655
```

So the stories carve-out had no beneficiary, and `**/*.test.tsx` really does match nothing today. Keeping the latter is the right call and stating it rather than quietly keeping it is the better half — the two globs look alike and are not alike: tests render raw markup by necessity, stories do not, so one exclusion has a reason that outlives its current match count and the other did not.

Checking `dismiss_stale_reviews_on_push` before deciding whether to fold the notes in is the part I would keep. The alternative was two reviews and two CI cycles for two lines, and it would have cost the same dismissal anyway.

PR #2847, opened 2026-09-08
fix(frontend): give the All filter option its own colour tokens

### LINKED ISSUE #2814: statusLabel still hands the ALL filter option the 3.61:1 badge-blue pair; only per-consumer overrides keep it off screen
## Summary

The `badge-blue` ink/fill pair (`--color-badge-blue-text` #eaf3ff on `--color-badge-blue-bg` #007cf5) measures **3.61:1** and was removed from three rendering sites in #2796, #2810 and #2811. It is not painted anywhere on `dev` today.

But it has not been removed from the **data**. `statusLabel('All', 'ALL', 'color-badge-blue', 'var(--color-primary-500)')` still constructs the pair, and it stays off screen only because every consumer independently overrides it at the point of use. Remove any one of those overrides — or add a consumer that renders the option the way the existing non-`ALL` branch does — and the 3.61:1 pair paints.

## Where the pair is constructed

`apps/frontend/src/app/constants/status.ts:105-116` builds `bg`/`text` from the `cssPrefix` argument, so passing `'color-badge-blue'` yields:

```
bg:     var(--color-badge-blue-bg)      #007cf5
text:   var(--color-badge-blue-text)    #eaf3ff     -> 3.61:1
border: var(--color-primary-500)        (overridden by the 4th argument)
```

Three call sites pass `'color-badge-blue'`:

- `apps/frontend/src/app/ui/filters/InventoryFilters/index.tsx:13`
- `apps/frontend/src/app/ui/filters/InventoryTurnoverFilters/index.tsx:12`
- `apps/frontend/src/app/ui/filters/StatusOptionButtons.stories.tsx:21`

## Why nothing renders it today

Four independent guards, none of which is enforced by the shared component:

1. `StatusOptionButtons` never reads `option.bg` or `option.text` at all — its option type is `{ key, name, border? }`. It paints the dot swatch from `option.border` and the label from the caller-supplied `getTextColor`.
2. `InventoryFilters/index.tsx:33` — `option.key === 'ALL' ? 'var(--color-text-primary)' : option.text`
3. `InventoryTurnoverFilters/index.tsx:23` — same shape
4. `getStockHealthButtonStyle` / `getTurnoverStatusButtonStyle` return **no `backgroundColor` at all** for `ALL`, whereas the branch two lines below returns `backgroundColor: option.bg, color: option.text`

So the failing pair sits one `key === 'ALL'` guard away from the code path that would render it, in two files.

## The part that makes this a trap rather than dead data

`apps/frontend/src/app/__tests__/constants/status.test.ts:21-27` asserts that pair as the **expected** return value:

```ts
expect(statusLabel('All', 'ALL', 'color-badge-blue', 'var(--color-primary-500)')).toEqual({
  name: 'All',
  key: 'ALL',
  bg: 'var(--color-badge-blue-bg)',
  text: 'var(--color-badge-blue-text)',
  border: 'var(--color-primary-500)',
});
```

A reader checking whether the pair is intended finds a green test saying it is.

## Not in scope

`features/overview/components/CommunityStats.tsx:272` uses `--color-badge-blue-bg` as a chart series colour with no text drawn on it. That is a correct single-token use and should stay.

## Suggested direction (not started)

Give the neutral `ALL` option its own token triple rather than one it is expected to override — the consumers already agree on what it should look like (`--color-text-primary` label, `--color-card-border` border, no fill). That deletes three overrides and the misleading assertion together.

## Verification

Measured on the `dev` tree exported with `git archive`, searched with `/usr/bin/grep -rnF` (not the shell's `grep`, which is ugrep with `--ignore-files`): **28 `badge-blue` occurrences across 17 files**, of which the live non-comment code is exactly the sites listed above. Verified identical at `152d4c947` and at the current tip `ef612cb25` — every line number cited above is unchanged between them.

aupyay:
## PR Checklist

The neutral "All" row in the two inventory status-filter dropdowns is built from the `badge-blue` prefix:

```ts
statusLabel('All', 'ALL', 'color-badge-blue', 'var(--color-primary-500)')
```

So its data carries `--color-badge-blue-text` (#eaf3ff) on `--color-badge-blue-bg` (#007cf5), which measures 3.61:1 against a 4.5:1 bar. No ink in this palette clears 4.5 on that fill.

Nothing paints it today, but only because four consumers each replace it at the point of use:

1. `StatusOptionButtons` never reads `option.bg` or `option.text` - its option type is `{ key, name, border? }`.
2. `InventoryFilters/index.tsx` returned `'var(--color-text-primary)'` for `option.key === 'ALL'`.
3. `InventoryTurnoverFilters/index.tsx` did the same.
4. Both trigger-style helpers return no `backgroundColor` at all for `ALL`, where the branch two lines below returns `backgroundColor: option.bg, color: option.text`.

Delete any one of those guards, or add a consumer that renders the option the ordinary way, and 3.61:1 paints.

`__tests__/constants/status.test.ts` asserted that pair as the expected return value of `statusLabel`, so a reader checking whether the pairing was intended found a green test saying it was.

## What is the new behavior?

`--color-pill-all-*` in `globals.css`: no fill, `--color-text-primary` ink, and the `--color-primary-500` dot the border override used to supply. The option is built once as `AllFilterOption` in `constants/status.ts` rather than copied into three files.

`statusLabel` loses its `borderOverride` parameter. It had no other caller, and it was the mechanism that let one entry mix tokens from two prefixes while only the odd one out got replaced downstream.

The two `key === 'ALL'` ink guards become `option.text`. The two trigger-style helpers keep their `ALL` branch on purpose - the trigger's unfiltered state is a placeholder (`--color-text-tertiary` on a `--color-card-border` outline), which is genuinely different from the row's own colours, not a substitution for them. What changes is that the branch is now a style choice rather than the only thing standing between the data and a 3.61:1 render.

**Rendered output is unchanged in both themes.** Read out of a running Storybook with transitions disabled, before (at `57e0db282`) and after, same server:

| | light | dark |
|---|---|---|
| All label ink | `rgb(48, 47, 46)` -> `rgb(48, 47, 46)` | `rgb(230, 221, 208)` -> `rgb(230, 221, 208)` |
| All dot | `rgb(0, 124, 245)` -> `rgb(0, 124, 245)` | `rgb(0, 124, 245)` -> `rgb(0, 124, 245)` |
| All row fill | `rgba(0, 0, 0, 0)` -> `rgba(0, 0, 0, 0)` | `rgba(0, 0, 0, 0)` -> `rgba(0, 0, 0, 0)` |

The four non-All rows read identically too. The before/after PNGs of `filters-statusoptionbuttons--stock-health` are byte-identical in each theme (`e707ca1c...` light, `29a47a4a...` dark). The control that the reverted tree was actually being served: `getComputedStyle(document.documentElement).getPropertyValue('--color-pill-all-text')` read empty on the before capture and non-empty on the after.

### The guard

`statusPillContrast.test.ts` gains two arms that measure the option itself rather than its consumers:

- `has no fill to write on` - `resolve(AllFilterOption.bg)` is `transparent` in both themes. Any prefix whose `-bg` is a colour reintroduces the pairing, so the assertion is the absence of a fill rather than the absence of one specific hex.
- `reads on the panel it sits on` - the ink clears 4.5:1 on `--screen` in both themes.

Both were mutation-checked:

- `AllFilterOption` pointed back at `color-badge-blue`: 4 failed / 27 passed - `status.test.ts`, both `has no fill` arms, and `reads on the panel (light)`. Dark passes there correctly, since `#eaf3ff` is legible on the dark screen; the fill arm is what catches it in both themes.
- `--color-pill-all-text` set to `#8a8378`: 2 failed / 29 passed - exactly the two `reads on the panel` arms, so that arm is live independently of the fill arm.

### Verification

- `pnpm --filter frontend run test -- --testPathPatterns="constants/status|contrast/statusPillContrast|ui/filters"` - 4 suites, 43 passed, 0 skipped, at `66aa9a468` with a clean tree.
- Wider sweep `--testPathPatterns="[Ii]nventory|filters|contrast|constants|globals|[Tt]oken|Overview|Calendar/common"` - 71 suites, 965 passed, 0 skipped, exit 0.
- `npx tsc --noemit` from `apps/frontend` - exit 0.
- `pnpm --filter frontend run lint` - exit 0, 1 pre-existing warning in `hooks/useLazyAuthStore.ts`, a file this PR does not touch.
- Aikido: no local scan - no Aikido MCP tool and no `aikido:scan` skill is present in this session. The PR's own `Aikido Security: check code` job passed in 24s (scan 187264337), so the change is covered; what is missing is the pre-push local run, not the gate. `GitGuardian Security Checks` and `Detect secrets (Gitleaks)` also passed.

## Related Issue(s)

Fixes #2814

### DIFF
--- a/apps/frontend/src/app/__tests__/constants/status.test.ts
+++ b/apps/frontend/src/app/__tests__/constants/status.test.ts
@@ -1,4 +1,5 @@
 import {
+  AllFilterOption,
   AppointmentLabels,
   TaskLabels,
   getAppointmentStatusTone,
@@ -17,13 +18,18 @@ describe('statusLabel', () => {
     });
   });
 
-  it('honours a border override', () => {
-    expect(statusLabel('All', 'ALL', 'color-badge-blue', 'var(--color-primary-500)')).toEqual({
+  /* This slot used to assert `statusLabel('All', 'ALL', 'color-badge-blue',
+     'var(--color-primary-500)')`, which put the 3.61:1 badge-blue pair in the
+     expected value. A reader checking whether that pairing was intended found a
+     green test saying yes (#2814). The option now has its own prefix, and no
+     entry mixes tokens from two of them. */
+  it('gives the neutral All option its own prefix, with no borrowed tokens', () => {
+    expect(AllFilterOption).toEqual({
       name: 'All',
       key: 'ALL',
-      bg: 'var(--color-badge-blue-bg)',
-      text: 'var(--color-badge-blue-text)',
-      border: 'var(--color-primary-500)',
+      bg: 'var(--color-pill-all-bg)',
+      text: 'var(--color-pill-all-text)',
+      border: 'var(--color-pill-all-border)',
     });
   });
 });
--- a/apps/frontend/src/app/__tests__/ui/contrast/statusPillContrast.test.ts
+++ b/apps/frontend/src/app/__tests__/ui/contrast/statusPillContrast.test.ts
@@ -2,6 +2,7 @@ import { getMerckSubtopicPillStyle } from '@/app/features/integrations/constants
 import { getStatusBadgeStyle } from '@/app/features/inventory/pages/Inventory/utils';
 import { measureContrast } from '@/app/features/appointments/components/Calendar/responsive/contrastProbe';
 import { getOrganizationStatusStyle } from '@/app/ui/tables/tableUtils';
+import { AllFilterOption } from '@/app/constants/status';
 import { resolve, resolveColour } from '@/app/__tests__/support/globalsTokens';
 
 /**
@@ -114,6 +115,43 @@ describe.each(['light', 'dark'] as const)('inline status pill contrast (%s)', (t
     // None of these is large text, so the bar must be the strict one.
     expect(reading.required).toBe(4.5);
     expect(reading.ratio).toBeGreaterThanOrEqual(4.5);
+  });
+});
+
+/**
+ * The same defect one step earlier: colours that are not painted today but are
+ * one deleted `key === 'ALL'` guard away from it.
+ *
+ * `StatusOptionButtons`' non-ALL branch renders `option.bg` under `option.text`.
+ * The All row used to carry the badge-blue pair there and was kept off screen
+ * only because each of its four consumers substituted something else at the
+ * point of use (#2814). These two arms assert the data is safe to render, not
+ * that the consumers happen not to render it.
+ */
+describe('the neutral All filter option', () => {
+  it.each(['light', 'dark'] as const)('has no fill to write on (%s)', (theme) => {
+    /* Not `expect(...).not.toBe('#007cf5')`: any prefix whose `-bg` is a colour
+       reintroduces the pairing, so the assertion is the absence of a fill.
+       Pointing `AllFilterOption` at any filled prefix fails here. */
+    expect(resolve(AllFilterOption.bg, theme === 'dark')).toBe('transparent');
+  });
+
+  it.each(['light', 'dark'] as const)('reads on the panel it sits on (%s)', (theme) => {
+    const dark = theme === 'dark';
+    const el = mount(
+      resolveToken(AllFilterOption.text, dark),
+      resolveToken(`var(${dark ? SURFACE.dark : SURFACE.light})`, dark),
+      resolveToken(`var(${dark ? SURFACE.dark : SURFACE.light})`, dark),
+      '13px',
+      '400'
+    );
+    const reading = measureContrast(el);
+    expect(reading.required).toBe(4.5);
+    expect(reading.ratio).toBeGreaterThanOrEqual(4.5);
+  });
+
+  afterEach(() => {
+    document.body.innerHTML = '';
   });
 });
 
--- a/apps/frontend/src/app/__tests__/ui/contrast/statusPillContrast.test.ts
+++ b/apps/frontend/src/app/__tests__/ui/contrast/statusPillContrast.test.ts
@@ -122,11 +122,17 @@ describe.each(['light', 'dark'] as const)('inline status pill contrast (%s)', (t
  * The same defect one step earlier: colours that are not painted today but are
  * one deleted `key === 'ALL'` guard away from it.
  *
- * `StatusOptionButtons`' non-ALL branch renders `option.bg` under `option.text`.
- * The All row used to carry the badge-blue pair there and was kept off screen
- * only because each of its four consumers substituted something else at the
- * point of use (#2814). These two arms assert the data is safe to render, not
- * that the consumers happen not to render it.
+ * The site that pairs an option's fill under its ink is the trigger chip -
+ * `getStockHealthButtonStyle` and `getTurnoverStatusButtonStyle`, whose non-ALL
+ * branch returns `{ backgroundColor: option.bg, color: option.text }`. Their
+ * `key === 'ALL'` branch is the guard; delete it and the All option is rendered
+ * the ordinary way. (`StatusOptionButtons` is not that site and never was: its
+ * option type is `{ key, name, border? }`, so it cannot read `bg` at all.)
+ *
+ * The All row used to carry the badge-blue pair in that data and was kept off
+ * screen only because each of its four consumers substituted something else at
+ * the point of use (#2814). These two arms assert the data is safe to render,
+ * not that the consumers happen not to render it.
  */
 describe('the neutral All filter option', () => {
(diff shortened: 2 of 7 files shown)


ankit-yc (commented):
Holding the approval on one line, and it is a comment rather than a behaviour — but it is the comment that tells the next reader which deletion this guard exists to survive, so I would rather it go in before this becomes the merge record.

**`src/app/__tests__/ui/contrast/statusPillContrast.test.ts:125` names the wrong component.**

> `StatusOptionButtons`' non-ALL branch renders `option.bg` under `option.text`.

`StatusOptionButtons` never reads `option.bg`. Its option type is `{ key; name; border? }` — no `bg` field at all — and it renders a dot in `option.border` plus a label in `getTextColor(option)`. The only occurrence of the string `bg` in that file is the `hover:bg-card-hover` class:

```
git grep -nF 'option.bg' -- src/app/ui/filters/StatusOptionButtons.tsx   -> rc=1, no hits
git grep -nF 'option.bg' -- src/app                                     -> 4 hits:
  __tests__/ui/contrast/statusPillContrast.test.ts:125   (this comment)
  features/tasks/components/TaskFilterBar.tsx:130        (dotColor fallback, not a fill)
  ui/filters/InventoryFilters/index.tsx:46               <- backgroundColor: option.bg
  ui/filters/InventoryTurnoverFilters/index.tsx:36       <- backgroundColor: option.bg
```

The site that actually pairs the fill under the ink is the **trigger chip** — `getStockHealthButtonStyle` / `getTurnoverStatusButtonStyle`, in the two filter components — and its non-ALL branch is `backgroundColor: option.bg, color: option.text`. Those are also precisely the two `key === 'ALL'` branches this PR deliberately kept, which makes the correction more than cosmetic: the comment as written points a future reader at a file where deleting the guard would change nothing, and away from the two where it would.

Everything else reproduces. On a clean checkout of `66aa9a468`:

```
                                    jest rc   result
base, unmutated                        0      2 suites, 31 passed
M1  AllFilterOption -> badge-blue      1      4 failed / 27 passed
M2  --color-pill-all-text -> #8a8378   1      2 failed / 29 passed  (both "reads on the panel" arms)
M3  --color-pill-all-bg -> primary-500 1      2 failed / 29 passed  (both "has no fill" arms)
```

M3 is mine rather than yours, and it is the one I wanted: M2 fires only the legibility pair and M3 only the no-fill pair, disjointly, so neither arm family is riding on the other. Each mutation was confirmed applied against a pre-edit snapshot and reverted from it; `git diff --quiet` is clean afterwards.

I also checked the two claims that ship as comments, since they become documentation:

- `globals.css` — "the fill's only remaining use is the chart series in CommunityStats". True. `git grep -nF 'color-badge-blue'` outside `globals.css` returns 16 hits; the only non-comment, non-test production one is `features/overview/components/CommunityStats.tsx:272`. Every other production hit is a comment explaining why that site does *not* use it.
- `--color-pill-all-text: var(--color-text-primary)` themes correctly through indirection rather than a second definition: `--color-text-primary` is defined identically in both blocks and resolves through `--color-neutral-900` -> `--ink-body`, which is what varies. `--color-primary-500` has a single definition, which matches the dot measuring identically in both themes.

And the shared-singleton question that the "built once as `AllFilterOption`" change raises — the object is now referenced by three arrays instead of copied, so a mutation anywhere would leak. Nothing assigns to `.bg` / `.text` / `.border` in `src/app` except an unrelated `blob.text =` in an IDEXX test. I ran that grep with a control arm because my first control pattern used `\b`, which git grep's ERE does not support and which returned a confident, wrong zero.

Fix the one comment and I will approve on the push. Not deleting the two trigger-style `key === 'ALL'` branches was the right call, and saying so in the description rather than following the issue's suggested direction silently is the part I would keep.

ankit-yc (approved):
Approving at `ef82ca889`.

The follow-up is comment-only and I checked that rather than assuming it: every `+`/`-` line between `66aa9a468` and `ef82ca889` is a comment line (`git diff … | grep -E '^[+-]' | grep -v` the comment forms, rc=1, no output), one file, 11/5. Tests at the new head: `2 suites, 31 passed, 0 skipped, jest rc=0`.

The replacement says the right thing and says the negative explicitly — that `StatusOptionButtons` "is not that site and never was: its option type is `{ key, name, border? }`, so it cannot read `bg` at all". A future reader deleting a `key === 'ALL'` branch now lands in the two files where that deletion actually changes a render.

Everything from the previous pass stands and is unaffected by a comment edit: M1 `4 failed / 27 passed`, M2 `2 failed / 29 passed`, and M3 (`--color-pill-all-bg` -> `primary-500`) `2 failed / 29 passed` hitting the no-fill pair disjointly from M2's legibility pair, each confirmed applied against a pre-edit snapshot and reverted from it.

Issue #2281, opened 2026-08-18
test(frontend): enforce Storybook play functions in CI with the test runner


ankit-yc:
## Summary

Run Storybook `play` functions as an enforced CI check, using `@storybook/test-runner` against the built Storybook, so the interaction assertions added in #2274 are gated rather than advisory.

## Problem / motivation

#2274 added stories that open interaction-gated surfaces (portalled menus, popovers, drawers, focus-revealed toolbars) and assert the opened panel has real content. Those assertions currently run nowhere that can block a merge.

Verified against the repo rather than assumed:

| | |
| --- | --- |
| `dev` ruleset required contexts | `CI Required`, `Supply Chain Required` |
| `Chromatic Publish & Visual Tests` | not in that list |
| `ci.yml` | no reference to chromatic, so the aggregate does not roll it up |
| `CHROMATIC_CONFIGURED` | `true`, so it does run and passes |
| `@storybook/addon-interactions` / test-runner | not installed |

So a red Chromatic does not block a merge, and with no interactions addon there is nothing holding the `play` assertions to account. That matters because the whole reason those stories exist is that four production bugs shipped on surfaces nothing rendered: a popover whose grid template used a comma (invalid CSS, silently dropped, six children collapsed into one column), two calendar overlays with an orphaned grid child that doubled their height, and dropdown panels using fill tokens as ink. An unenforced assertion does not stop the fifth.

## Proposed solution

Add `@storybook/test-runner` and a CI job that builds Storybook once and runs the play functions against it.

Note for whoever picks this up, because it is the trap: do **not** try to reach this through jest. `next/jest` injects `transformIgnorePatterns` of `/node_modules/.pnpm/(?!(geist)@)`, which shadows any allowlist for `@storybook`'s ESM, so `composeStories` fails with `SyntaxError: Cannot use import statement outside a module` and the only fix is overriding next/jest's defaults for the entire suite. The test runner sidesteps that completely - it drives the built Storybook through Playwright and never touches jest's config.

`@playwright/test` and `@axe-core/playwright` are already devDependencies of `apps/frontend`, and there is an existing Playwright CI job, so the browser dependency is not new.

## Scope

- add `@storybook/test-runner` to `apps/frontend`
- a CI job that runs it against a built Storybook
- decide whether it joins the `CI Required` aggregate, or lands advisory first and is promoted once it has been green for a while

### Out of scope

[...]

PR #2848, opened 2026-09-08
fix(frontend): stop the status pill truncating the patient's name on phone

aupyay:
## PR Checklist

At 390px the phone patient bar's name+pill column is 197px wide:

```
viewport                        390
back button                      34
avatar                           38
flex-1 column                   197
  patient name   gets  80   needs 114   LOST 34
  status pill    gets 111              (does not shrink)
visit timer                      63
```

The name carries `truncate`; the status pill does not shrink. So the name absorbs the entire shortfall. "Poppy Hartmann" renders as "Poppy Ha...", with no `title` - and a phone has no hover, so the full value is unreachable. This is the bar a clinician reads to confirm which animal is in front of them, in the workspace where they record clinical findings. Reproduced on `origin/dev` at `57e0db282`.

## What is the new behavior?

The visit timer moves from beside the name onto the signalment line. The column then spans the timer's old 63px as well, which is enough to hold the name and the pill together; the signalment keeps its full width because it gains exactly what the timer takes from it.

Measured on the rendered DOM at 390x844, both themes, before and after on one Storybook server with the tree swapped underneath it:

| | before | after |
| --- | --- | --- |
| patient name | needs 114, gets 80, **lost 34** | needs 114, gets 114, **lost 0** |
| signalment (plain) | needs 197, gets 197, lost 0 | needs 199, gets 199, lost 0 |
| signalment (allergy story) | needs 197, gets 197, lost 0 | needs 199, gets 199, lost 0 |
| bar height | 59.1 | 73.0 |
| horizontal overflow | 0 | 0 |
| anything else clipped in the bar | none | none |

`PhoneWorkspaceShell` renders the same component, so it is fixed by the same change: name 114/114, and the shell still fits 844px with `scrollHeight - innerHeight = 0`.

The liveness control for the before/after pair: `sig.parentElement.contains(timerPill)` read `false` on the before capture and `true` on the after, so the two readings are not two reads of one state.

### Why not the direction the issue proposed

#2790 suggested moving the **status pill** to the signalment line, quoting a 12px signalment loss. That reproduces - on the plain story. On the allergy story the same change costs **98px**, which is the whole `Allergy: Penicillin` tail rendered in `--danger-text`. The issue's own recommendation silently truncates the allergy warning; it was measured on the one story that hides it. Candidates re-measured on one frame at `57e0db282`:

| candidate | bar height | name lost | signalment lost (plain) | signalment lost (allergy) |
| baseline | 59.1 | 34 | 0 | 0 |
| A: `flex-wrap` the name row | 87.6 | 0 | 0 | 0 |
| C: pill to the signalment line | 68.5 | 0 | 12 | **98** |
| **K: timer to the signalment line** | **73.0** | **0** | **0** | **0** |

A also clips nothing but costs 28.5px of permanent header height against K's 13.9px. C is the cheapest in height and the only one that clips a clinical field.

This is a product-visible trade - 13.9px of header height on an 844px viewport, and the timer reading as secondary to patient identity - made rather than parked, since the issue had been open for the decision and nothing else was blocked on it. Say so and I will change it.

### The guard

A new `NameKeepsItsRoom` story asserts neither the name nor the signalment is clipped. Because `truncate` makes those assertions pass for free on any name that happens to fit, it first asserts the fixture still over-subscribes the pre-fix row:

```ts
await expect(name.scrollWidth + pill.getBoundingClientRect().width)
  .toBeGreaterThan(PRE_FIX_NAME_ROW_WIDTH);
```

It runs with `allergy: 'Penicillin'` in its own args. Without that it inherited the meta args, which set no allergy - so the assertion protecting the line that the rejected candidate cost 98px was measuring that line without its tail. `WithAllergy` does not cover it either: it asserts the tail is in `textContent` and that the paragraph *can* ellipsize, neither of which is a measurement. Caught in review by Claude L1.

Three arms, all against the live Storybook at `9355aebe9`, base `6 passed / 0 failed / rc=0`:

- **Layout reverted** to the pre-fix arrangement: 1 failed / 5 passed, on `expected 114 to be less than or equal to 65`. The clipping assertion is ordered ahead of the structural one on purpose, so the revert is caught by the measurement rather than only by the arrangement that produces it today.
- **Status pill moved down to join the timer**, leaving the name alone on row 1: 1 failed / 5 passed, on `expected 178 to be less than or equal to 65`. This is the arm that isolates the *signalment* assertion - read live at 390px under that mutation, the name is untouched at 114/114 and the signalment loses **98px**, which is exactly what the rejected candidate cost. Without this arm, "the guard fires" was only ever evidence about the name.
- **Fixture name shortened to `Bo`**: 1 failed / 5 passed, on `expected 129.25 to be greater than 197`. So the control fails when the story stops exercising the failure, rather than quietly turning the rest into a test of nothing.

Each mutation was reverted and the base re-run to `6 passed` before the next.

### Two things about the margin that read wrong

Both are recorded in the story rather than only here.

- **The guard does not run at 390px.** `.storybook/test-runner.ts` maps `mobile` to 375x812, so CI exercises 15px less than every number in the issue and in the table above. That is a tighter budget, not a looser one, which is why the guard still passes - but a reader comparing the two sets of numbers will not otherwise know why they differ.
- **`scrollWidth` reports the box, not the text, whenever the text fits**, so it cannot give the margin at all - it saturates at `184/184` and stops varying while there is still room. Measured instead as the paragraph's right edge minus the tail span's, which keeps moving. All at 375px:

| tail | scrollWidth/clientWidth | margin |
| --- | --- | --- |
| `Allergy: Penicilli` | 184/184 | 12.3px |
| `Allergy: Penicillin` (shipped) | 184/184 | **6.4px** |
| `Allergy: Penicillinx` | 184/184 | 1.4px |
| `Allergy: Penicillinxx` | 188/184 | -3.7px, clips |

Four strings, one identical `scrollWidth` pair, four different margins. The fixture sits 6.4px - two characters - from the edge. Deliberate for a guard, but it means lengthening this allergy string turns the story red for a copy change rather than a layout regression; lengthen the name instead.

An earlier revision of this PR said "about one character" and the story's docs description said "21.4px". The 21.4 was a correct measurement at 390px that I carried into a file where every other number is at 375, and I fixed the comment without fixing the description eighteen lines below it. Both now state 6.4px and name their viewport. Caught in review by Claude L1.

**Scope, stated plainly:** the tree as shipped does not clip the allergy tail, and this PR does not make that state worse - the 6.4px is inherited, not established here. An allergy longer than the fixture still clips and this guard will not see it. That is a product question about whether the tail may clip at all, not a layout regression, and it is not addressed here.

**And this guard cannot block a merge today.** `storybook-interactions.yml:83` sets `continue-on-error: true`, its own header calls the job advisory, and `ci-required` needs `[core, test, migration, sonar, frontend-quality, mobile-version]` - `grep -ci storybook .github/workflows/ci.yaml` is `0`. So the play functions here are a guard, not a gate; promotion is tracked separately in #2281. Verified on this branch rather than taken from the review that raised it.

### Not in this PR

#2790 also lists clipped values in `OutpatientSchedule` (476px and 166px) and `DiagnosticsStep` (80px, hover-only `title`). Those are different components with a different remedy and are untouched here, so the issue stays open - this PR is `Refs`, not `Fixes`.

## Verification

- `pnpm --filter frontend run test -- --testPathPatterns="AppointmentWorkspace|Workspace"` - 59 suites, 1157 passed, 0 skipped, exit 0, at `1f0f59636` with a clean tree.
- `npx test-storybook -- PhonePatientBar` against a live Storybook - 6 passed, 0 skipped, exit 0, at the same commit.
- `node scripts/ci/check-hardcoded-colours.mjs` - exit 0, `no new hardcoded colours ... 2047 files scanned`, with `selftest: 15 cases pass` in the same session. This gate failed an earlier head on `#2790` in three doc strings, since `#` plus four hex digits is a colour to it; reworded to "issue 2790" rather than touching the gate. The class is filed as #2849 - every four-character finding in the baseline is an issue reference.
- `npx tsc --noemit` from `apps/frontend` - exit 0.
- `pnpm --filter frontend run lint` - exit 0, 1 pre-existing warning in `hooks/useLazyAuthStore.ts`, a file this PR does not touch.
- Aikido: no local scan - no Aikido MCP tool and no `aikido:scan` skill in this session. Read the hosted result rather than assuming it: `commits/3e1a42547/check-runs` returns `Aikido Security: check code :: completed :: success :: Scan completed`. Stated as measured because an earlier revision of this line said the job "covers the change" without my having read its conclusion - and on #2828 a red Aikido sat through two rounds of review for exactly that reason, since required contexts on `dev` are only `CI Required` and `Supply Chain Required`, so no security gate here can block a merge on its own.
- Check-runs at `3e1a42547` enumerated from the API rather than `gh pr checks`: `total_count` 36, 36 returned, `0 failure`. Three read `cancelled`; each of those three names also has a `success` entry at the same sha, so they are concurrency-superseded duplicates rather than results - verified by grouping conclusions per name, not by assuming.

## Related Issue(s)

Refs #2790

### DIFF
--- a/apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhonePatientBar.stories.tsx
+++ b/apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhonePatientBar.stories.tsx
@@ -84,7 +84,7 @@ const meta = {
           'component had.\n\n' +
           'The phone variant is not a smaller copy of the desktop one. It drops the pulsing green ' +
           'dot and the "In room" / "Over booked slot" words entirely and shows the bare elapsed ' +
-          'time at 10px with `px-[9px] py-[5px]`. Since #2790 it rides at the right end of the ' +
+          'time at 10px with `px-[9px] py-[5px]`. Since issue 2790 it rides at the right end of ' +
           'signalment line rather than beside the name, which is what let the name and the ' +
           'status pill stop competing for one 197px row. It also strips a leading `00:` so an ' +
           'under-an-hour visit reads ' +
@@ -215,7 +215,7 @@ export const NotStarted: Story = {
           'No start timestamp at all, which is what an appointment that has not been checked in ' +
           'still looks like. The words rather than digits make this the widest of the three ' +
           'states, which is what it used to cost the name - the timer shared the name row ' +
-          'until #2790 moved it down to the signalment line.',
+          'until issue 2790 moved it down to the signalment line.',
       },
     },
   },
@@ -328,7 +328,7 @@ export const NameKeepsItsRoom: Story = {
     docs: {
       description: {
         story:
-          'The regression guard for #2790. "Poppy Hartmann" needs 114px and the `IN PROGRESS` ' +
+          'The regression guard for issue 2790. "Poppy Hartmann" needs 114px and the `IN PROGRESS` ' +
           'pill takes 111px; while the timer sat beside them the pair had 197px to share, and ' +
           'the name - the only one of the two carrying `truncate` - absorbed the whole 34px ' +
           'shortfall. Nothing here asserts a pixel count, so a type-ramp or copy change moves ' +
--- a/apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhonePatientBar.stories.tsx
+++ b/apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhonePatientBar.stories.tsx
@@ -84,8 +84,10 @@ const meta = {
           'component had.\n\n' +
           'The phone variant is not a smaller copy of the desktop one. It drops the pulsing green ' +
           'dot and the "In room" / "Over booked slot" words entirely and shows the bare elapsed ' +
-          'time at 10px with `px-[9px] py-[5px]`, because the bar has roughly 70px to spare next ' +
-          'to a truncating name. It also strips a leading `00:` so an under-an-hour visit reads ' +
+          'time at 10px with `px-[9px] py-[5px]`. Since #2790 it rides at the right end of the ' +
+          'signalment line rather than beside the name, which is what let the name and the ' +
+          'status pill stop competing for one 197px row. It also strips a leading `00:` so an ' +
+          'under-an-hour visit reads ' +
           '`MM:SS` rather than `00:MM:SS`.\n\n' +
           'The over-booked pill is the one worth reviewing: `warning-100` fill, `warning-300` ' +
           'border, `warning-900` label. The 900 step is deliberate - the 700 step measured 2.77:1 ' +
@@ -211,8 +213,9 @@ export const NotStarted: Story = {
       description: {
         story:
           'No start timestamp at all, which is what an appointment that has not been checked in ' +
-          'still looks like. The words rather than digits are why this state is the widest of the ' +
-          'three, so it is the one that squeezes the name beside it.',
+          'still looks like. The words rather than digits make this the widest of the three ' +
+          'states, which is what it used to cost the name - the timer shared the name row ' +
+          'until #2790 moved it down to the signalment line.',
       },
     },
   },
@@ -285,6 +288,52 @@ export const WithAllergy: Story = {
           '`--danger-text` at 700. The whole line is one `truncate` paragraph at 10.5px, so on a ' +
           'long signalment the allergy is what disappears - which is worth seeing before deciding ' +
           'it belongs there.',
+      },
+    },
+  },
+};
+
+/** The width the name+pill row had while the timer still shared it (#2790). */
+const PRE_FIX_NAME_ROW_WIDTH = 197;
+
+export const NameKeepsItsRoom: Story = {
+  name: 'Long name beside a wide pill',
+  render: (args) => <TimerBar {...args} startedMinutesAgo={20} bookedEndMinutesAgo={-10} />,
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+    const name = canvas.getByText('Poppy Hartmann');
+    const pill = canvas.getByRole('button', { name: /in progress/i });
+    const signalment = canvas.getByText('Beagle · 9 yr · 12.4 kg');
+
+    /* The control, and the reason this story is not a tautology. `truncate` makes
+       a clipped element report scrollWidth > clientWidth, so the assertions below
+       pass for free on any name short enough to fit. This one first proves the
+       fixture still over-subscribes the row the bug lived in - a shorter name, or
+       a pill whose label got shorter, silently turns the rest into a test of
+       nothing. */
+    await expect(name.scrollWidth + pill.getBoundingClientRect().width).toBeGreaterThan(
+      PRE_FIX_NAME_ROW_WIDTH
+    );
+
+    // Neither line pays for it. Checked before the structural assertion below, so
+    // that reverting the layout is caught by the measurement rather than only by
+    // the arrangement that happens to produce it today.
+    await expect(name.scrollWidth).toBeLessThanOrEqual(name.clientWidth);
+    await expect(signalment.scrollWidth).toBeLessThanOrEqual(signalment.clientWidth);
+
+    // The structural fact the fix rests on: the timer is on the signalment line.
+    await expect(signalment.parentElement).toContainElement(timerPill(canvasElement));
+  },
+  parameters: {
+    docs: {
+      description: {
+        story:
+          'The regression guard for #2790. "Poppy Hartmann" needs 114px and the `IN PROGRESS` ' +
+          'pill takes 111px; while the timer sat beside them the pair had 197px to share, and ' +
+          'the name - the only one of the two carrying `truncate` - absorbed the whole 34px ' +
+          'shortfall. Nothing here asserts a pixel count, so a type-ramp or copy change moves ' +
+          'the numbers without failing the story; what it asserts is that the name and the ' +
+          'signalment both survive whatever the numbers become.',
       },
     },
   },
--- a/apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhonePatientBar.stories.tsx
+++ b/apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhonePatientBar.stories.tsx
@@ -300,24 +300,32 @@ export const NameKeepsItsRoom: Story = {
   name: 'Long name beside a wide pill',
   /* The allergy is set here rather than inherited, because the guard's second
      assertion is about the signalment and the meta args have no allergy tail.
-     Without it this story measured `Beagle · 9 yr · 12.4 kg` - 23 characters in
-     a 199px box, 55px of headroom - while asserting that the line the component
-     documents as the first thing lost is safe. `WithAllergy` does not cover it
-     either: it asserts the tail is in `textContent` and that the paragraph CAN
-     ellipsize, neither of which is a measurement.
+     Without it this story measured `Beagle - 9 yr - 12.4 kg` in a 184px box with
+     92px of room to spare, while asserting that the line the component documents
+     as the first thing lost is safe. `WithAllergy` does not cover it either: it
+     asserts the tail is in `textContent` and that the paragraph CAN ellipsize,
+     neither of which is a measurement.
 
-     Two things about the margin, because the obvious way to read it is wrong.
-     The test runner renders `mobile` at 375px (`.storybook/test-runner.ts`),
-     not the 390px the issue was investigated at, so the guard runs on 15px less
-     than the numbers in #2790. And `scrollWidth` reports the BOX, not the text,
-     whenever the text fits - at 375px this paragraph reads 184/184 for a tail of
-     `Penicilli`, `Penicillin` and `Penicillinx` alike, so the headroom cannot be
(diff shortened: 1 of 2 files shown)


ankit-yc (commented):
Holding on one arm, and it is the arm your own finding argues for.

The layout change is right and I verified it rather than reading it. At `82cadbda8`, live Storybook at 390px, DOM read directly:

```
column  min-w-0 flex-1                     w=270
  row 1  flex items-center gap-1.5         w=270
    name span   114/114        status pill 111/111      114 + 111 + 6 = 231 <= 270
  row 2  flex items-center gap-2           w=270
    signalment p 199/199       timer span   61/61       199 +  61 + 8 = 268 <= 270
```

`test-storybook` at that head: **6 passed, 0 failed, rc=0**. Your two mutations reproduce, each taking down exactly one story:

```
                                   rc   passed  failed
base                                0      6       0
layout reverted (timer beside name) 1      5       1
guard fixture name -> 'Bo'          1      5       1
```

The second is the right instinct and it is the one most people would not have written: `truncate` makes the clipping assertions pass for free on any name that fits, so the story has to prove it still over-subscribes the row before its verdict means anything.

## The gap

`NameKeepsItsRoom` runs on the meta args, and the meta args carry no `allergy` — line 266 is the only place in the file that sets one. So the guard's `signalment.scrollWidth <= signalment.clientWidth` is asserted on `Beagle · 9 yr · 12.4 kg`, 23 characters in a 199px box. The state that is actually near the edge has no clipping assertion at all: `WithAllergy` asserts `textContent` contains the string — true whether or not it is visible — and that `textOverflow` is `ellipsis`, which asserts the paragraph *can* clip, not that it doesn't.

Measured, with the probe shown capable of answering both ways in the same session:

```
allergy fixture                          p scrollWidth / clientWidth   tail vs paragraph right edge
'Penicillin'                as shipped        199 / 199               -21.4px   (fits, 21.4px spare)
'Penicillin and Cephalosporins'  planted      272 / 199               +73.1px   CLIPPED
'X'                              planted      199 / 199               -55.0px   (fits)
```

And with the tail clipped by 73px, the whole file still reports **6 passed, 0 failed, rc=0**. Nothing in the suite can see it.

To be exact about what this is and is not: **the tree as shipped does not clip**, and this PR does not make the allergy state worse — the signalment box is 199px before and after, so the 21.4px of headroom is inherited, not introduced. What is missing is a guard on it. The component ships a story whose own docstring says the allergy tail "is the FIRST thing lost", the issue's rejected candidate was rejected precisely because it cost that tail 98px, and the regression guard that came out of all that runs on the state without an allergy.

That is the same sentence as your finding against the issue, one level up: the worst case was named in the file, and the measurement went to the state that was open. It cost 98px the first time; here it costs a guard rather than a render, which is why this is one arm and not a rewrite.

Cheapest fix I can see: add `allergy: 'Penicillin'` to `NameKeepsItsRoom`'s args — its two existing clipping assertions then run on the long signalment, the control assertion is unaffected, and `WithAllergy` keeps its colour/weight job. Whether that is better than a separate arm on `WithAllergy` is your call; I care that one arm somewhere fails when the tail stops fitting.

Ordering the clipping assertion ahead of the structural one is correct and worth keeping for the reason you gave — a guard that fails on the arrangement rather than the effect stops working the moment someone reaches the same layout another way. `tsc --noEmit` rc=0 at this head. The `#2790` -> `issue 2790` change is three prose strings and nothing a play function reads; I diffed `1f0f59636..82cadbda8` to confirm that rather than taking it, and the colour gate matching a four-hex issue number is a genuinely good catch — fixing the prose instead of the gate was the right side of that trade.

ankit-yc (commented):
The finding is closed and your new arm is the right one. Holding on a single number, which is **my** number and which the file now contradicts itself about.

## First, the correction I owe

You are right about the viewport and I checked it rather than taking it: `.storybook/test-runner.ts` maps `mobile` to `375x812`. My 21.4px was measured at 390px and is not the width the guard runs at. Re-measured at both, sweeping the tail a character at a time — the same rect-based read, so the two columns are comparable:

```
tail                    375px (the runner)          390px (my review)
                        p sw/cw   px to the edge     p sw/cw   px to the edge
Allergy: X               184/184      40.0            199/199      55.0
Allergy: Penicilli       184/184      12.3            199/199      27.3
Allergy: Penicillin      184/184       6.4            199/199      21.4
Allergy: Penicillinx     184/184       1.4            199/199      16.4
Allergy: Penicillinxx    188/184      -3.7  CLIPPED   199/199      11.3
Allergy: Penicillinxxx   193/184      -8.8  CLIPPED   199/199       6.3
```

So: **6.4px at the width that matters, and `Penicillinxx` is the first that clips** — two more characters, not the twelve my figure implied. Your `scrollWidth` point is visible in the left column as well: `184/184` for four different strings, three of which have wildly different headroom. It reports the box until the text stops fitting. My rect read kept varying, which is the only reason I got a number at all — but a right method at the wrong viewport still produces a wrong number, and mine went into a review, a channel post and now this file.

## The hold

`NameKeepsItsRoom`'s docs description ends:

> *"Its limit: 21.4px of headroom is inherited, not established here…"*

while the `args` comment eighteen lines above it says, correctly, *"about one character from the edge at the width CI uses"*. Two headrooms for one paragraph in one story, and the wrong one is in the description — the half a reader meets in the Storybook docs panel rather than in the source. It should be 6.4px, or better, "one to two characters at 375px", since a pixel figure is exactly what will be wrong again the next time the type ramp moves.

That is the same shape as #2847, and worth naming because the mechanism repeated with the roles swapped: there your PR body was right and the code comment wrong; here the code comment is right and the description wrong. Both times the correct version existed a few lines away. A number does not become verified by being adjacent to a verified one.

## Everything else verified at `9355aebe9`

```
                                    rc   passed  failed   failing assertion
base                                 0      6       0     -
allergy tail lengthened until it     1      4       2     the arm that PASSED before this fix
  clips (my arm from the last pass)
status pill moved down to the        1      5       1     expected 178 <= 65   signalment
  timer row (your new arm)
```

My arm is the point: at `82cadbda8` a 73px-clipped tail left the file at 6 passed / 0 failed. At this head the same mutation takes it down. The finding is closed at the guard, not argued away.

And your new arm is the one I did not ask for and should have. I gave you "the guard measures the wrong state"; you found the sharper thing underneath — that reverting the layout squeezes the *name*, so both of your original mutations were evidence about the name assertion and the signalment assertion had never been observed firing at all. Moving the pill down instead isolates it, and it costs the signalment 98px: the same 98 the rejected candidate cost. The arm that proves the guard is alive is a re-run of the bug the guard exists for.

## Scope, since the guard's reach is not what it looks like

`storybook-interactions.yml` sets `continue-on-error`, is not in `CI Required`'s `needs`, and `ci.yaml` does not reference storybook at all (`git grep -n storybook -- .github/workflows/ci.yaml`, rc=1). Its own header says so and dates the reason: 113 failing of 3366 on dev, promotion tracked in #2281. So this guard is real, it runs, and **it cannot block a merge today** — a future regression here shows up as an advisory red that a merge does not wait for. Not a defect in this PR and not yours to fix; it is the difference between "there is a guard" and "there is a gate", and the guard is the prerequisite for the gate.

Fix the description number and I will approve.

ankit-yc (approved):
Approving at `3e1a42547`.

Every number in the file re-derived here by a **different method** from the one that produced them, on purpose — you measured span-right-edge minus paragraph-right-edge, so I measured the text itself with a `Range` over the paragraph's contents and subtracted from `clientWidth`. Two methods, one answer:

```
story          viewport   box   scrollWidth   text width   room
no allergy       375      184      184           92.1       91.9
with allergy     375      184      184          177.7        6.3
no allergy       390      199      199           92.1      106.9
with allergy     390      199      199          177.7       21.3
```

`92px` and `6.4px` both hold. Your `scrollWidth` point is the middle column: `184` for two strings whose text widths are 92.1 and 177.7 — it saturates at the box and stops carrying information exactly when you need it to. And the text-width column being identical across viewports while the room column is not is the clean statement of what my 21.4px actually was: not a wrong measurement, a right one of a quantity that depends on a width I had not checked.

Correcting the no-allergy figure as well was the right call and I would not have caught it — I only flagged the one number I had put there. `55px` was a 390px reading of a *different string* than the one the comment names, so it was wrong twice over.

Arms at this head, each mutation applied against a snapshot and reverted from it:

```
                                              rc   passed  failed   failing assertion
base                                           0      6       0     -
allergy tail lengthened until it clips         1      4       2     the arm that PASSED at 82cadbda8
status pill moved down to the timer row        1      5       1     expected 178 <= 65   signalment
```

Checks, enumerated this time rather than assumed — I approved another PR tonight without doing this and it had a red security gate: `total_count=34`, returned 34, **0 failure**, `Aikido Security: check code success`. Three entries read `cancelled` (`Validate commit messages`, `Validate PR title`, `No named external product`) and each has a `success` sibling for the same context started three to four seconds later, so those are concurrency-superseded duplicates and not results. Seven still running, all builds and the advisory Storybook legs.

The docs description now names its viewport and carries the 390px figure beside the 375px one rather than silently replacing it, which is better than what I asked for: someone arriving from #2790 with a 390px number in hand can see why it does not match instead of concluding one of the two is wrong.

Restating the limit so it stays attached to the merge: this guard is real and it is not a gate — `storybook-interactions.yml` is `continue-on-error`, absent from `CI Required`'s `needs`, and unreferenced by `ci.yaml`. A regression here goes red advisorily and merges anyway until #2281 promotes it.

PR #2851, opened 2026-09-08
fix(ci): stop the colour gate matching four-digit issue references

### LINKED ISSUE #2849: check-hardcoded-colours: every 4-character finding in the baseline is an issue reference, not a colour
## Summary

`scripts/ci/check-hardcoded-colours.mjs` matches `#` followed by 3, 4, 6 or 8 hex digits. This repository's issue numbers are four digits in the range this project has reached, so `#2790`, `#2298`, `#2481` are all valid four-digit hex and all match.

The gate strips comments before matching - deliberately, and its own docstring gives the reason:

> Half the literals in this repository sit inside a comment explaining why the shipped colour is what it is [...] A gate that counted them would be answered by deleting them, which costs the repository the reasoning and gains it nothing.

Strings are not stripped, correctly, because `color: '#E9F2FD'` is where the violations live. But Storybook `docs.description` values are strings, and they are the one place in this codebase where that same decision-record prose is required to live in a string rather than a comment.

## Measurement

`node scripts/ci/check-hardcoded-colours.mjs --list` on `dev` at `57e0db282`, 2047 files scanned, 591 findings. Grouped by token length:

```
length   count
     3      10
     4      10
     6     381
     8       1
```

Every one of the ten four-character findings, with the source line:

```
#2298  '#2298 removed that override, so the play function measures 700 against the idle 500 '
#2298  'same shape of bug - 12px where it asked for 11px - until PR #2298 removed the unlayered '
#2298  'PR #2298 removed the duplicate and the play function now measures the 11px the '
#2298  'clips this early. The weight half of that bug is gone: PR #2298 removed the matching '
#2297  '#2297; **Phone: long name truncates** measures it). The offline subtitle carried the '
#2297  '`font-size: 1rem !important` (issue #2297), which outranks the `text-[13.5px]` '
#2598  'tracked in #2598.\n\n'
#2595  'Tracked in #2595.',
#2481  message: 'feat(backend): publish clone traffic totals\n\nRefs #2481',
#1973  'confirmation card; "Add to queue" is what puts it in the queue (bug #1973). The '
```

**Ten of ten are issue or PR references. None is a colour.** They sit in `hardcoded-colours-baseline.json` as ten colours awaiting migration, so the published "575 to migrate" is 565.

The last one is a third category worth noting separately: a commit-message fixture whose body is `Refs #2481`. A test fixture that has to contain a realistic commit trailer cannot avoid the shape.

For contrast, the three-character bucket does not have this problem - 8 of 10 are genuine `#fff` / `#ddd` / `#ccc` literals, and the other 2 are prose naming the recharts default box colour, which is a colour being discussed rather than an unrelated token.

## Why this is worth fixing rather than working around

It cost me a red gate on #2848 and I fixed it by rewording three doc strings to "issue 2790", which is the right call for one PR and the wrong equilibrium. The gate's stated design is that it should never be answerable by deleting an explanation. As written it is answerable by deleting - or garbling - exactly the cross-references that make a story's documentation navigable, and it will do so again for every issue number this repository files from now on.

## Suggested directions (not started, and the choice matters)

1. Require that a four-digit match not be preceded by a word that marks it as a reference (`issue`, `PR`, `#` following `Refs`/`bug`/`tracked in`). Cheap; heuristic; will miss a bare `#2790` at the start of a sentence, which is 1 of the 10 above.
2. Drop `{4}` from the alternation. `#RGBA` is a real CSS colour, so this trades a false-positive class for a false-negative class. Measurable first: the current baseline contains zero genuine four-digit colours, so the false-negative set is empty *today*.
3. Strip `docs.description` string values the way comments are stripped. Most faithful to the stated intent - that is prose by construction - but it needs an AST rather than the current character state machine, and it would also hide a genuine literal written inside one.

Option 2 is the one whose cost can be measured rather than argued, and the measurement above is that cost today. Whoever takes this should re-run the length histogram first rather than trusting this one, since the baseline moves.

## Verification

All numbers above from `node scripts/ci/check-hardcoded-colours.mjs --list` on a worktree at `57e0db282`, with `node scripts/ci/check-hardcoded-colours.mjs selftest` reporting `15 cases pass` in the same session, so the scanner was alive when the list was produced.

aupyay:
Closes #2849

## The bug

`#` followed by four hex digits matched this repository's issue numbers. Every four-digit finding in the baseline was an issue or pull-request reference and none was a colour, so ten cross-references were recorded as colours awaiting migration. The gate strips comments deliberately — its own docstring says a gate answerable by deleting an explanation costs the repository its reasoning — but Storybook `docs.description` values are strings, and they are where that same decision-record prose has to live.

## The fix

The four-digit branch now refuses an all-decimal token:

```
(?![0-9]{4}(?!\w))[0-9a-fA-F]{4}
```

## Why this narrowing and not one of the alternatives

The issue offered three directions. This is a fourth, and the argument is a measurement on both sides. Hex-token length buckets at `5580e9c35`:

| bucket | total | all-decimal | has a-f |
| --- | --- | --- | --- |
| 3 | 10 | 0 | 10 |
| 4 | 10 | 10 | 0 |
| 6 | 381 | 23 | 358 |
| 8 | 1 | 0 | 1 |

Source lines read on both sides rather than inferred: **bucket 4 is 10 of 10 issue or PR references**, and **bucket 6's 23 all-decimal tokens are 23 of 23 colours** — 22 applied as values (`background`, `border`, `color`, `iconColor`, gradient stops) and one named in prose as a colour.

So the narrowing removes every false positive at zero cost in bucket 4, and the same reasoning one bucket over would suppress 23 real colours. It is a fact about the four-digit bucket in this repository, not a principle. The code comment carries the table and says not to generalise it.

Dropping `{4}` outright — the issue's option 2 — would have blinded the gate to `#e6dd`-shaped colours as well. This keeps them.

## The accepted cost, stated plainly

`#1234` is a legal `#RGBA` colour and this branch now refuses it. There is no such literal in the tree today. More importantly the trade **changed kind**: a four-digit reference used to be a loud false positive somebody triaged, and an all-decimal RGBA is now a silent false negative. Nothing will announce it — a blind spot in a detector has no detector, and this gate is the component whose job that would be. The comment says the re-measure is periodic or it does not happen, rather than implying a trigger that does not exist.

Five-or-more-digit references need nothing: the trailing `(?!\w)` binds to the whole alternation, so `#12345` was already silent before this change.

## Verification

```
selftest                 15 -> 18 cases pass      rc=0 on pass, rc=2 on failure (both checked)
--list findings          590 -> 580               removed 10, added 0 (comm, both directions)
buckets after            3:10  4:0  6:381  8:1
baseline                 575 across 115 files -> 565 across 112
gate                     rc=0, clean
pnpm test:scripts        445 tests / 445 pass / 0 fail / 0 skipped   (443 before, +2 added)
prettier --check         rc=0
eslint --max-warnings=0  passed via lint-staged on both .mjs files (legacy config path)
gitleaks / secretlint    clean
```

Arms on the real tree, not only in the case table:

| arm | expectation | result |
| --- | --- | --- |
| plant `color: '#e6dd'` | MUST FIRE | gate rc=1, reports `571: #e6dd` |
| plant `'PR #2299 removed …'` | MUST NOT FIRE | gate rc=0, no new colours |
| same plant, pre-change scanner | control MUST FIRE | gate rc=1, `ChannelHeaderBar.stories.tsx: 1 -> 7` |

The third row is what makes the second attributable — 7 is the five restored references plus the baselined one plus the plant, so the false positive is reproduced on the tree rather than asserted from a case table.

Two of the three new selftest cases fail against the pre-change pattern, which is how the cases are known to be live rather than decorative.

## One note for anyone else touching this script

It refuses to report a count when its own cases fail — `could not run: the scanner failed its own cases, so its count means nothing`, rc=2. A mutation that breaks a selftest case therefore cannot be used as a scan control; the control has to be the full pre-change file. My first attempt at the third arm above measured nothing for exactly this reason.

Also: the existing case `'// see #2822' -> 0, 'issue reference is not a colour'` sits inside a **comment**, so it passes by comment-stripping with the pattern uninvolved. It has been green throughout the bug it is named for. The three new cases exercise the pattern in a string.


ankit-yc (approved):
Reviewed at `6d24152d24fa53106f8fddbbc854aca6b72d3a31`, in a worktree detached at that sha, tree clean before and after every arm. Every number below was produced in this session rather than read from the description.

## Reproduced

```
selftest                         18 cases pass, rc=0
gate                             rc=0, "no new hardcoded colours: 565 to migrate across 112 files,
                                 15 justified across 5 files, 2047 files scanned"
--list                           581 lines = 580 findings + 1 summary
pnpm test:scripts                445 tests / 445 pass / 0 fail / 0 skipped / 28 suites
```

## The 10 removed findings are exactly the 10 references, and nothing was added

`--list` at the base (`5580e9c35`) against `--list` at this head, compared as sets:

```
removed  QuickActionsModal.stories.tsx:362   #2298
         DiagnosticsStep.stories.tsx:517     #1973
         ChannelHeaderBar.stories.tsx:64     #2297
         ChannelHeaderBar.stories.tsx:65     #2298
         ChannelHeaderBar.stories.tsx:179    #2298
         ChannelHeaderBar.stories.tsx:559    #2297
         ChannelHeaderBar.stories.tsx:562    #2298
         InvoiceCreditNotes.stories.tsx:71   #2598
         InvoiceCreditNotes.stories.tsx:74   #2595
         Insights.stories.tsx:131            #2481
added    none
```

Ten removed, zero added, and every one is a four-digit issue reference. The baseline delta reconciles against the same set: `ChannelHeaderBar 6 -> 1`, `Insights 2 -> 1`, and `QuickActionsModal`, `DiagnosticsStep`, `InvoiceCreditNotes` dropping out entirely.

## Arms

```
BASE                                                      rc=0   565 across 112
M1  HEX reverted to the old pattern -> selftest MUST fail rc=2   two cases fail by NAME:
      "four-digit issue reference in a string: expected 0, got 1"
      "DELIBERATE: all-decimal #RGBA is the accepted blind spot: expected 0, got 1"
M2  plant color: '#e6dd' in a real .tsx  -> MUST FIRE     rc=1   reports "571: #e6dd"
M3  plant 'PR #2299 …' in a STRING       -> MUST NOT FIRE rc=0   silent
M3c same plant, full PRE-CHANGE tree     -> control       rc=1   "#2299", ChannelHeaderBar 6 -> 7
M4  RESTORED                                              rc=0   tree clean
```

M1 is the arm that matters: the two new cases fail **by name** under the old pattern, so they are live rather than green-by-accident. And the third new case (`#e6dd` still a colour) is what stops the narrowing from being a blanket `{4}` removal.

M3c reproduces the false positive on the tree rather than in a case table, so M3's silence is attributable rather than a green I could not explain.

## One thing I got wrong on the way, because it is a trap in this script

My first attempt at M3c reverted the pattern while keeping the new selftest cases. The gate refused to scan: `could not run: the scanner failed its own cases, so its count means nothing`, rc=2 — correct behaviour by the script, and no control by me. **A mutation that breaks a selftest case cannot serve as a scan control here.** The control has to be the full pre-change tree, which is what the row above uses. This is documented in the PR discussion already; I walked into it anyway, which suggests it is worth a line in the script's own comment.

## On the comment

The blind spot is named as untriggered rather than implied to have a trigger, and that is the right call — a blind spot in a detector has no detector, so "re-measure when it lands" names an event that cannot fire. The bucket table with the do-not-generalise warning sits in the code rather than only in the description, so the next reader gets the reason and not just the rule.

Two things I verified in that comment rather than took: `#12345` is silent under **both** the old and new patterns, because the trailing `(?!\w)` binds to the whole alternation — so the five-or-more-digit clause is accurate. And `#1234` is a legal `#RGBA` literal that this branch now refuses, which the comment states as a deliberate cost rather than eliding.

## Scope

Checks at this sha: 24 rows, 1 pending, 0 red, `Supply Chain Required` success, **`CI Required` not in the set yet** — it lands last, so this census cannot speak to mergeability and I am not claiming it does. Approving the change; the merge is yours on the four conditions once both required contexts complete.

PR #2854, opened 2026-09-08
fix(frontend): recover dead Storybook runner pages

aupyay:
## Summary

- detect when the Storybook preview page has lost its injected context helper
- create and initialize a replacement page before reading story context
- cover healthy, dead, and closed-page paths with Jest tests

Issue: #2281

### DIFF
--- a/apps/frontend/.storybook/test-runner.test.ts
+++ b/apps/frontend/.storybook/test-runner.test.ts
@@ -1,0 +1,82 @@
+import { getStoryContext } from '@storybook/test-runner';
+import config from './test-runner';
+
+declare global {
+  var context: unknown;
+  var page: unknown;
+  var __sbSetupPage: jest.Mock;
+}
+
+jest.mock('@storybook/test-runner', () => ({
+  getStoryContext: jest.fn().mockResolvedValue({ storyGlobals: {} }),
+}));
+
+const mockedGetStoryContext = jest.mocked(getStoryContext);
+
+describe('Storybook runner page recovery', () => {
+  const page = {
+    close: jest.fn(),
+    evaluate: jest.fn(),
+    setViewportSize: jest.fn(),
+  };
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    page.setViewportSize.mockResolvedValue(undefined);
+    mockedGetStoryContext.mockResolvedValue({ storyGlobals: {} } as never);
+    globalThis.__sbSetupPage = jest.fn();
+    globalThis.context = undefined;
+    globalThis.page = undefined;
+  });
+
+  it('does not reset a healthy page', async () => {
+    // Must-NOT-fire arm: a healthy page must be left alone; resetting it would
+    // make every passing story pay for a replacement page it does not need.
+    page.evaluate.mockResolvedValue(true);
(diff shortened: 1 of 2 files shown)


aupyay:
## Independent verification of the recovery, with the pre/post reconciliation

Independent re-derivation, own fetch of all five workflow runs. Everything claimed for this PR reproduces; this
comment adds the pre/post reconciliation that the run-level numbers alone do not give.

### Confirmed

| claim | result |
|---|---|
| run `34190506613` is `Storybook Interactions` at this PR's head | yes, `head_sha` == PR head, 4 shards |
| shard failures 54 + 65 + 53 + 31 | 203, matches |
| `__getContext is not a function` in the post-fix logs | **0** across all four shard logs |
| the pre-fix collateral aggregate | **110 distinct tests** (220 log lines — the message prints twice) |

Controls: each shard log was confirmed to contain a `Tests:` summary line before its zero was
accepted, and a planted occurrence of the target string was detected in a copy of shard 1.

### The denominators, which are the drift alarm

```
                        PRE  (34178190538, dev @ ecc7188ef)   POST (34190506613, this PR)
tests                   285 failed / 3690 total               203 failed / 3690 total
suites                  100 failed /  671 total               100 failed /  671 total
target-error lines      220                                   0
```

**The denominators are identical**, so nothing changed about which tests ran — the reduction is not
a scope change. **No suite went green**: 100 failing suites before, 100 after.

### Where the 82 went, reconciled exactly

```
pre failures      285
  - fixed          90
  + new             8
= post failures   203      closes exactly
```

And the fate of the 110 collateral failures specifically:

```
of the 110 carrying the target error
   now pass                      74
   STILL FAIL, different message 36
```

**`110 -> 0` is exactly true of the error string and should not be read as 110 tests going green** —
74 do. The other 36 still fail, now surfacing their own errors. That is the documented behaviour of
a cause-agnostic, symptom-specific recovery rather than a defect in it: once the page is restored
the story actually runs, and it can fail on its own merits.

### The 8 new failures

Checked against **all four** pre-fix runs, not just the `dev` one:

```
new vs the dev run alone            8
already failing in some other
  pre-fix run (intermittent)        4
seen in NO pre-fix run at all       4
```

The four with no pre-fix precedent:

```
Organization/AddTeam         > OptionsComeFromTheOrg  > play-test
Primitives/FloatingToolbar   > IndentSinksAListItem   > play-test
Primitives/FloatingToolbar   > MarksActive            > play-test
Workspace/PrescriptionEditor > SearchResultsOpen      > play-test
```

### Attribution limit — please read this before treating those four as a regression from this PR

This PR's own diff is small and well scoped (2 files, +108 / -2, both under `.storybook/`). But the
branch is **7 commits ahead** of the pre-fix `dev` commit the comparison arm was measured at, and
only one of those commits is this change. The other six merged in between.

**So the two arms differ by more than this PR, and the four new failures cannot be attributed to
the recovery from these runs alone.** Isolating them needs an arm at this branch's merge-base, or
this branch with the recovery reverted. Comparability is not provenance.

### Summary

The recovery does what it says: the target error is gone, the denominators did not move, and the
net is -90 / +8 with the arithmetic closing. The two items for whoever promotes this are the 36
formerly-collateral tests that still fail, and the 4 new failures whose origin is not isolated by
this comparison.

PR #2856, opened 2026-09-08
test(frontend): bind the signalment guard by testid, not by its own text

aupyay:
## The defect

`NameKeepsItsRoom` in `PhonePatientBar.stories.tsx` measured whether the signalment
line clips, and located that line with:

```ts
const signalment = canvas.getByText('Allergy: Penicillin').parentElement as HTMLElement;
```

The asserted string was the selector for the subject of the assertion. Lengthening
the allergy — the exact content change the guard exists to catch — kills the lookup
one statement before any width is read, so the clip assertion reported
`Unable to find an element` and could not fail on the thing it guards.

## The change

`data-testid="patient-signalment"` on the paragraph, `getByTestId` in the story.
Two lines of production change, one line of test change, no behaviour change.

## Evidence

Three arms, each a fresh `pnpm run storybook:build` served locally and driven with
`test-storybook`. The served bundle was checked for the mutation before each run, so
a stale build could not produce a green.

| arm | fixture | selector | result |
|---|---|---|---|
| 1 | `Penicillin` (shipped) | testid | **6 passed** |
| 2 | `Penicillinxx` | testid | **1 failed** — `expected 188 to be less than or equal to 184` |
| 3 | `Penicillinxx` | old `getByText` | **1 failed** — `Unable to find an element with the text: Allergy: Penicillin` |

Arm 2 is the assertion firing on the measurement. Arm 3 is the same mutation under the
old binding, dying in the query — that is the control, and it is why the binding had to
change rather than the fixture.

`188 / 184` is the pair already recorded in that story's margin table for `Penicillinxx`.
It is now produced by the assertion rather than only asserted about.

Arm 1 was re-run after rebasing onto `dev`: 6 passed, on a bundle confirmed to carry
`patient-signalment` and to carry no `allergy: 'Penicillinxx'` arg.

Also run in this worktree at this head: `tsc --noEmit` (rc 0) and `eslint` over the
changed directory with `--max-warnings=0` (rc 0). `pnpm run lint` and `pnpm run type-check`
ran as the pre-push hook and passed.

## What this does not change

The story's fixture is still fixed at `Penicillin`, so an allergy longer than that
shipping in production is still not exercised — the file says so and that limit is
untouched. What is restored is that editing the fixture is now a test of the layout
instead of a test of the query.

### DIFF
--- a/apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhonePatientBar.stories.tsx
+++ b/apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhonePatientBar.stories.tsx
@@ -272,7 +272,15 @@ export const WithAllergy: Story = {
        changing `args.allergy` killed the lookup and the three assertions that are
        NOT about the allergy text - the ellipsis, the ink difference and the weight -
        stopped running rather than failing. The tail is bound by a prefix instead,
-       which survives a change to the allergy itself. */
+       which survives a change to the allergy itself.
+
+       If you mutate `args.allergy` to check that this still holds, the mutation
+       must NOT be a superstring of the shipped value. `toContain` below is
+       satisfied by one, so `Penicillinxx` - the string in the margin table two
+       stories down, which is the one to hand - leaves this story green whether
+       the assertions ran or never ran at all. `Amoxicillin` fails on the
+       comparison instead of the lookup, which is the answer the mutation was
+       asked for. */
     const signalment = canvas.getByTestId('patient-signalment');
     const allergy = within(signalment).getByText(/^Allergy: /);
 
--- a/apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhonePatientBar.stories.tsx
+++ b/apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhonePatientBar.stories.tsx
@@ -323,16 +323,23 @@ export const NameKeepsItsRoom: Story = {
 
      Four strings, one identical scrollWidth pair, four different margins.
 
-     So the fixture sits 6.4px - two characters - from the edge. Deliberate for a
-     guard, but it means lengthening this allergy string turns the story red for a
-     copy change rather than a layout regression. Lengthen the NAME instead. */
+     So the fixture sits 6.4px - two characters - from the edge, which is close
+     enough that the margin is worth guarding rather than assuming.
+
+     `signalment` is bound by `data-testid` and deliberately NOT by its own text.
+     It was previously bound with `getByText('Allergy: Penicillin').parentElement`,
+     which made the string its own selector: lengthening the allergy killed the
+     lookup one statement before any width was read, so the clip assertion below
+     reported `Unable to find an element` and could never fire on a content
+     change. Bound by the testid, `Penicillinxx` fails on `expected 188 <= 184` -
+     the measurement rather than the lookup. */
   args: { allergy: 'Penicillin' },
   render: (args) => <TimerBar {...args} startedMinutesAgo={20} bookedEndMinutesAgo={-10} />,
   play: async ({ canvasElement }) => {
     const canvas = within(canvasElement);
     const name = canvas.getByText('Poppy Hartmann');
     const pill = canvas.getByRole('button', { name: /in progress/i });
-    const signalment = canvas.getByText('Allergy: Penicillin').parentElement as HTMLElement;
+    const signalment = canvas.getByTestId('patient-signalment');
 
     /* The control, and the reason this story is not a tautology. `truncate` makes
        a clipped element report scrollWidth > clientWidth, so the assertions below
--- a/apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhonePatientBar.stories.tsx
+++ b/apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhonePatientBar.stories.tsx
@@ -267,8 +267,14 @@ export const WithAllergy: Story = {
   render: (args) => <TimerBar {...args} startedMinutesAgo={20} bookedEndMinutesAgo={-10} />,
   play: async ({ canvasElement }) => {
     const canvas = within(canvasElement);
-    const allergy = canvas.getByText('Allergy: Penicillin');
-    const signalment = allergy.parentElement as HTMLElement;
+    /* Neither binding may be the string this story asserts about. `WithAllergy`
+       previously took the paragraph from `getByText('Allergy: Penicillin')`, so
+       changing `args.allergy` killed the lookup and the three assertions that are
+       NOT about the allergy text - the ellipsis, the ink difference and the weight -
+       stopped running rather than failing. The tail is bound by a prefix instead,
+       which survives a change to the allergy itself. */
+    const signalment = canvas.getByTestId('patient-signalment');
+    const allergy = within(signalment).getByText(/^Allergy: /);
 
     /* The tail is a span inside the same truncating paragraph as the signalment, so
        it is the FIRST thing lost when the name is long - and the only thing marking
--- a/apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhonePatientBar.tsx
+++ b/apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhonePatientBar.tsx
@@ -99,7 +99,10 @@ const PhonePatientBar = ({
             widens the row to hold both, and the signalment keeps its full width
             because the timer takes the space the column gains. */}
         <div className="flex items-center gap-2">
-          <p className="min-w-0 flex-1 truncate text-[10.5px] leading-tight text-(--ink-faint)">
+          <p
+            data-testid="patient-signalment"
+            className="min-w-0 flex-1 truncate text-[10.5px] leading-tight text-(--ink-faint)"
+          >
             {signalment}
             {allergyText && (
               <>


ankit-yc (dismissed):
Reviewed head 86cfc33c9. The testid is added to the rendered signalment element and the guard story now measures that element independently of allergy text; no findings in the two-file diff. Remaining CI checks are still pending, so approval is not a merge recommendation.

aupyay:
Second-desk review. Base freshness, the binding equivalence and a sweep for the same pattern elsewhere.

## The check that makes arm 2's number meaningful

The PR body says the mutation now fails on `expected 188 to be less than or equal to 184` rather than in the query. That only carries weight if the new binding resolves to the **same element** the old one did — otherwise `188` is a measurement of a different node that happens to look like the old one.

It does. In `PhonePatientBar.tsx` the allergy is a `<span>` nested inside the `<p>`:

```tsx
<p data-testid="patient-signalment" className="min-w-0 flex-1 truncate …">
  {signalment}
  {allergyText && (<>{signalment && ' · '}<span …>Allergy: {allergyText}</span></>)}
</p>
```

`getByText('Allergy: Penicillin')` matches the `<span>` (its normalized text content is exactly that string), so `.parentElement` was the `<p>` — which is now the element carrying the testid. Same node, same `truncate`, same box. The margin table's `188 / 184` still applies to the thing being asserted about.

## Also verified

- Base is not stale: merge-base with `dev` is `afe9d24da`, the current tip — 1 ahead, 0 behind.
- `patient-signalment` occurs twice in the repo, the definition and this one use, so `getByTestId` cannot be ambiguous. The story renders a single `PhonePatientBar` through `TimerBar`.
- `data-testid` in shipped frontend source is established rather than novel here: 428 of 2114 `.tsx` files under `apps/frontend/src` carry one.

## The pattern is not unique to this story

Same shape, swept across stories and tests:

```
getByText(…).parentElement in *.stories.tsx / *.test.tsx      94
  literal selector                                            77
  VARIABLE selector, computed at runtime                      16
  regex                                                        1
CONTROL  total getByText( occurrences in those files        7022
```

Narrowing to the shape that actually bites — the selector text overlapping a value the story sets in `args`, so a fixture change can kill the lookup:

```
method: collect quoted values (len>=4) within 6 lines of an `args` line in the same
        file; flag a getByText literal that CONTAINS one of them
candidates: 18, across 6 files
```

**18 is a candidate set, not a defect count.** Most are stable section labels that happen also to be args values (`Vitals`, `Duration`, `Reason`) and no guard mutates them; the defect only bites where a test deliberately changes the very string it selects on, which is what made this story's case real. Each needs reading. The files are `FormRenderer.stories.tsx`, `PopoverDetail.stories.tsx`, `DropPreviewOverlay.stories.tsx`, `InpatientSchedule.stories.tsx`, `InfoSection.stories.tsx`, and this one.

One of the 18 is this file at line 330 — **inside the block comment that documents the fix.** The discriminator matched the prose describing the thing rather than the thing, which is the failure mode the PR body itself names. Line 342, the actual binding, is outside any comment and uses the testid. Recording it because a later sweep will hit it again.

## Not claiming CI

At `86cfc33c9` the security and supply-chain checks are green (Aikido, GitGuardian, Gitleaks, CodeQL `skipping`, SBOM, license, dependency review, `Supply Chain Required`, `Desktop E2E Required`). The build and Chromatic jobs and all four Storybook play-function shards are still `pending`. I have not read a completed run, so nothing here is a statement about the suite.

ankit-yc:
Attribution correction on the comment above.

That review was written by a different desk from the one that opened this PR, but it posted under `aupyay` — the PR's own author account — so as it stands it reads as the author annotating their own PR. Recording it here rather than deleting it, because the record should show what happened.

The mechanism is worth more than the slip. This estate routes **approvals** to `ankit-yc` because GitHub blocks self-approval, and a side effect is that an approval is legible as a second party. Nothing routes a **comment**, because nothing blocks it — so the same review content, moved from the review surface to the comment surface, silently loses its only per-actor evidence. The split was built for a mechanics reason and its attribution value is an accident that covers exactly the surface the block covers, and no other.

The technical content of the previous comment is unchanged and stands on its own measurements.

ankit-yc:
## The same binding defect is still live in this file, one story up — `WithAllergy`, line 271

Third-desk check of #2856. The fix and the `getByTestId` binding are right, and the binding
equivalence holds (below). But `NameKeepsItsRoom` was not the only occurrence in this file.

`WithAllergy` (unchanged by this PR — the block is byte-identical on `dev` and on `86cfc33c9`)
carries the identical chain:

```ts
args: { allergy: 'Penicillin' },
...
const allergy = canvas.getByText('Allergy: Penicillin');
const signalment = allergy.parentElement as HTMLElement;   // line 271
```

Same shape this PR's own comment describes: the asserted string is the selector for the subject of
the assertion. A change to `args.allergy` kills the lookup at line 270, so the three assertions that
have nothing to do with the allergy *string* — `textOverflow: ellipsis`, the ink-difference
`waitFor`, and `fontWeight === '700'` — never execute. They read as "not run", not as "failed".

### Executed, not inferred

Real component, real DOM, jest + RTL at this checkout (throwaway probe, not committed):

```
ARM 1  allergy='Penicillin'    getByText(...).parentElement resolves; tagName P; className
                               contains `truncate`; node === container.querySelector('p.truncate')
ARM 2  allergy='Penicillinxx'  getByText('Allergy: Penicillin') THROWS /Unable to find an element/
CONTROL A  ARM 2's render is fine: getByText('Allergy: Penicillinxx') resolves
CONTROL B  a content-independent binding still reaches the assertions under the mutation:
           p.truncate found, textContent contains the full mutated signalment,
           within(p).getByText(/^Allergy: /) resolves
CONTROL C  must-not-fire: an absent string throws in ARM 1 as well

5 passed / 5 total, rc=0, 0 skipped
MUTATION CHECK  point ARM 2 at the shipped value -> 1 failed / 4 passed, rc=1, and the failing
                case is ARM 2 by name. The arm is live.
```

`ARM 1`'s last line is the binding-equivalence check independently: `getByText(...).parentElement`
**is** the `p.truncate` node, and this PR's diff puts `data-testid="patient-signalment"` on exactly
that `<p>`. Measured for the node identity, read from the diff for the attribute placement.

### Suggested change, two lines

```ts
const signalment = canvas.getByTestId('patient-signalment');
const allergy = within(signalment).getByText(/^Allergy: /);
```

`CONTROL B` is the evidence that both of those resolve under a content mutation. That leaves
line 276's `toContain('… Allergy: Penicillin')` as the one place the string is asserted — which is
where it belongs.

### Scope

One file, read at `86cfc33c9`. `parentElement` appears 3 times in it: line 271 (this finding),
line 330 (inside the block comment documenting the fix), line 361 (`signalment.parentElement`, a
structural containment assertion, not a lookup — not affected). I have not swept the other 93
repo-wide occurrences; that candidate set is already logged above.

ankit-yc:
Correction to the scope line above: "the other 93 repo-wide occurrences" reads as though line 271
is one of the 94 `getByText(…).parentElement` hits from the sweep comment. It is probably not.
Line 271 spells the chain across two statements —

```ts
const allergy = canvas.getByText('Allergy: Penicillin');
const signalment = allergy.parentElement as HTMLElement;
```

— so a single-expression pattern cannot match it, which would also explain why that sweep's only
hit in this file was the one-expression form quoted inside the block comment at line 330. I have
not run that sweep's regex, so this is a reading of the reported output, not a measurement of it.
Either way the candidate set is a floor, not a count.

ankit-yc:
Extending my earlier sweep: the 94 figure was a floor, because a one-line pattern cannot span a chain split across two statements — the shape now identified at lines 270–271 of this file.

```
.parentElement lines in *.stories.tsx / *.test.tsx                391
  same line as getByText(   — my earlier sweep                     94
  NOT on the same line                                            297
  of those, the identifier is bound from getByText <=8 lines above  37   across 33 files
```

So the one-line form is 72% of the binding sites, not all of them.

Applying the same discriminator to those 37 — selector literal containing a value the file sets in `args` — gives **7 further candidates**, including 270→271 here:

```
BookingErrorMessage.stories.tsx:83->84         'Select a time slot before continuing.'
Calendar/common/PopoverDetail.stories.tsx:74->76      'Client name'
Calendar/common/UserLabels.stories.tsx:241->242      'Dr. Elena Marsh'
Calendar/common/DropPreviewOverlay.stories.tsx:77->79  'Poppy - Vaccination'
AppointmentWorkspace/phone/PhonePatientBar.stories.tsx:270->271  'Allergy: Penicillin'
companionHistory/HistoryEntryCard.stories.tsx:322->323  'rabies-certificate-2026.pdf'
companions/Sections/Core.stories.tsx:174->175   'Nova Scotia Duck Tolling Retriever cross'
```

Candidate set across both spellings: **25**, not 18.

Three caveats, all of which make 25 a floor rather than a count:

1. Not every candidate is a defect — it needs a test that deliberately mutates the string it selects on. Most of these select stable labels no guard changes.
2. The `<=8 lines above` window is a parameter I chose. Widening it can only raise 37.
3. The discriminator only matches literal selectors; 16 of the 94 and several of the 37 compute the selector at runtime, where an args value can reach it without appearing as a literal.

The `WithAllergy` block at 264–276 is byte-identical on `dev` and at this PR's head, so this PR neither introduced nor fixed it.

ankit-yc:
Addendum to the candidate-population numbers above, so the thread carries a figure with no free
parameter in it.

The decomposed-spelling count was reported at an 8-line window. That window converges, so it can be
eliminated rather than disclosed. Swept at the `dev` tip `afe9d24da`, over the 2797
`*.stories|test|spec.tsx?` files there, for an identifier bound from a `getByText(` query and later
used as `<ident>.parentElement`:

```
window <= 1     21 sites, 20 files
window <= 2     24 sites, 23 files
window <= 4     26 sites, 24 files
window <= 8     34 sites, 30 files
window <= 16    42 sites, 35 files
window <= 32    44 sites, 36 files
window <= 64    44 sites, 36 files
window  = inf   44 sites, 36 files      <- converged
```

**44 sites across 36 files**, at any window of 32 lines or more. Same-line
`getByText(...).parentElement` on this population and pattern: 90 — a different denominator from
the 94 above, not a correction to it.

Two exclusions worth naming, because my first pass returned 46:

- 2 sites bind the identifier with the *inline* form and then read `.parentElement` off it again —
  `signalment.parentElement` as a structural containment assertion, not a lookup. One of those is
  line 361 of this PR's own file. They are re-uses of an already-counted binding, not a second
  spelling, and they are not the defect class.
- Lines matching the inline form are excluded from the use side so the two populations do not
  overlap.

Controls: the known site in this PR's file (270 -> 271) is found and is the only one there; the
same sweep with the query name replaced by a nonexistent one returns 0.

**44 is still a candidate count, not a defect count.** One of the 44 has been executed (above). The
rest need the same question asked of each: does a guard in that file mutate the very string the
lookup selects on.

ankit-yc:
One more measurement, and it changes what a guard on this pattern has to exclude.

The same-line count for this file is **1 at both refs**:

```
dev  afe9d24da   const signalment = canvas.getByText('Allergy: Penicillin').parentElement …
head 86cfc33c9    It was previously bound with `getByText('Allergy: Penicillin').parentElement`,
                  ^ inside a block comment (unclosed /* depth = 1 at that line)
```

Repo-wide it is 94 at both refs as well. **This PR removes one occurrence of the pattern and the count does not move, because the comment documenting the fix reproduces the string the fix removed.**

So a lint rule or a tracking metric built on this pattern would report zero progress on the very change that fixes it — and would keep flagging this file forever. Any guard here needs to strip comments before matching, and that is not a nicety: it is the difference between a metric that can observe the remediation and one that cannot.

Two corrections to my earlier comment while I am here, both from reconciling against a second desk's independent sweep:

- Of the 94 same-line sites, **90** apply `.parentElement` directly to the `getByText(...)` call. Three more go through `.closest('button')`, which is the same defect class. The fourth is a false positive of my pattern — `expect(link.parentElement?.contains(screen.getByText('Paid'))).toBe(false)` in `InvoiceDetailHeader.test.tsx:270` co-locates two unrelated expressions on one line. Read the real figure as **93**, not 94.
- The decomposed population is **44 sites**, not 37, once the window is removed entirely rather than set to 8 — and after excluding 2 sites (3 at the `dev` tip) whose binding line is itself already counted in the same-line set. The two populations must be disjoint or the total double-counts.

ankit-yc:
Two corrections to my sweep addendum above, then the member list so the two independent sweeps can
be reconciled by members rather than by totals.

### 1. Correction: a dev-tip measurement cited with a PR-head line number

I wrote that one of the excluded double-counts is "line 361 of this PR's own file." The sweep ran at
the `dev` tip, where that site is `335 -> 354`; `361` is its line number at the PR head after this
PR's added comment. The site is the same one, the number was from the other ref. Two refs in one
citation.

### 2. Correction: my instrument had two blind spots, and fixing them was not free

The addendum's pattern was matched **per line**, which cannot see a spelling that spans lines, and it
did **not strip comments**, so a `getByText(` inside a `//` comment can bridge into the next real
line. Both fixed. But moving the whole pattern to whole-source matching introduced two *new*
regressions:

- a preceding `const foo = (a, b) => {` with no semicolon let the binding regex cross the newline and
  bind the wrong identifier, swallowing the real binding on the next line — 10 genuine sites lost;
- the inline-form exclusion started keying on the `const` position rather than the line, so 4
  already-counted inline bindings were re-admitted.

Net effect on the total: **44 -> 38 -> 44**. The intermediate 38 was plausible and wrong. The
corrected instrument strips comments, matches `INLINE` and the `.parentElement` use across lines, and
matches the `const X = … getByText(` opener per line, which is the granularity that opener actually
has.

**The published 44 sites / 36 files survives, member-for-member** — the corrected and the original
site sets are identical, not merely equal in size.

### 3. Comment stripping is load-bearing for any gate built on this pattern

Confirmed repo-wide over the 2797 `*.stories|test|spec.tsx?` files at each ref:

```
                    inline sites RAW   COMMENT-STRIPPED
dev tip  afe9d24da        95                 95
PR head  86cfc33c9        95                 94
this file, dev tip         1                  1     (the live binding)
this file, PR head         1                  0     (the comment quoting it)
```

So a raw count reports **no change** across a PR that removes an occurrence, because the comment
documenting the fix reproduces the string one-for-one. A stripped count moves 95 -> 94. Any lint rule
or tracking metric on this pattern has to strip comments or it cannot observe its own remediation,
and it will flag this file forever.

One more thing the totals hide: stripping also removes a cross-comment bridge at
`…/ViewAppointmentOverviewModal/index.stories.tsx:239` and thereby *reveals* the genuine site at
`:241`. The dev-tip total is 95 before and after — unchanged while a member changed.

### 4. The 44 decomposed sites at the dev tip, as `file  bind->use`

<details><summary>44 sites, 36 files</summary>

```
apps/frontend/src/app/__tests__/components/DataTable/AvailabilityTable.test.tsx  278->284
apps/frontend/src/app/__tests__/components/Inputs/Dropdown/Dropdown.test.tsx  528->529
apps/frontend/src/app/__tests__/pages/Appointments/Sections/AppointmentInfo/Prescription/ServiceCard.test.tsx  105->106
apps/frontend/src/app/__tests__/pages/Finance/Sections/InvoiceDetailHeader.test.tsx  242->246
apps/frontend/src/app/__tests__/ui/tables/DispensaryTable.test.tsx  269->270
apps/frontend/src/app/features/appointments/components/BookingErrorMessage.stories.tsx  83->84 149->150
apps/frontend/src/app/features/appointments/components/Calendar/Task/TaskDetailsPopover.stories.tsx  111->112
apps/frontend/src/app/features/appointments/components/Calendar/Task/TaskDropOverlays.stories.tsx  45->46
apps/frontend/src/app/features/appointments/components/Calendar/common/CalendarDayHeader.stories.tsx  72->73
apps/frontend/src/app/features/appointments/components/Calendar/common/CalendarTeamNamesRow.stories.tsx  61->62
apps/frontend/src/app/features/appointments/components/Calendar/common/DropPreviewOverlay.stories.tsx  77->79
apps/frontend/src/app/features/appointments/components/Calendar/common/PopoverDetail.stories.tsx  74->76
apps/frontend/src/app/features/appointments/components/Calendar/common/UserLabels.stories.tsx  241->242
apps/frontend/src/app/features/appointments/components/EmergencyBadge.stories.tsx  171->173
apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/DischargeDateTimeModal.stories.tsx  219->220 444->449
apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/components/InpatientSchedule.stories.tsx  229->230
apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/components/StaffField.stories.tsx  68->69
apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/phone/PhonePatientBar.stories.tsx  270->271
apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/sidemodal/panels/CalculatorsPanel.stories.tsx  159->173 166->174
apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/sidemodal/records/VitalsForm.stories.tsx  514->520 545->550
apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/steps/SoapPresentational.stories.tsx  160->171 161->170
apps/frontend/src/app/features/appointments/pages/Appointments/Sections/AppointmentInfo/index.stories.tsx  381->389
apps/frontend/src/app/features/auth/pages/ForgotPassword/ForgotPassword.stories.tsx  122->129
apps/frontend/src/app/features/companionHistory/components/HistoryEntryCard.stories.tsx  322->323
apps/frontend/src/app/features/companionHistory/components/HistoryRecordDrawer.stories.tsx  138->148
apps/frontend/src/app/features/companions/components/Sections/Companion.stories.tsx  225->226
apps/frontend/src/app/features/companions/components/Sections/Core.stories.tsx  174->175
apps/frontend/src/app/features/companions/components/Sections/Parent.stories.tsx  71->72 222->225
apps/frontend/src/app/features/integrations/pages/MerckManuals/MerckReaderPortal.stories.tsx  273->282
apps/frontend/src/app/features/inventory/components/TurnoverAnalytics/index.stories.tsx  261->262
apps/frontend/src/app/features/organization/pages/Organization/Sections/DeleteOrg.stories.tsx  119->141
apps/frontend/src/app/features/organization/pages/Organization/Sections/OrgProfileEditCards.stories.tsx  155->171 155->172 156->171
apps/frontend/src/app/features/organization/pages/Organization/Sections/ProfileCard.stories.tsx  423->432
apps/frontend/src/app/features/settings/pages/Settings/Sections/CrossClinicMessagingPreference.stories.tsx  206->214
apps/frontend/src/app/features/settings/pages/Settings/Sections/HoursEditModal.stories.tsx  453->454
apps/frontend/src/app/ui/cards/PetPassport/PetPassportView.stories.tsx  167->168```

</details>

Controls: the known site in this file (`270 -> 271` at the PR head) is present and is the only one
there; the same sweep with the query name replaced by a nonexistent one returns 0; the corrected and
original site sets reconcile exactly with an empty symmetric difference.

Still a candidate count. One executed, 43 unexamined.

**Review by ankit-yc — DISMISSED** [review]
Approving at `3c73e9398` — read by a desk that did not write either commit. What I verified, so this review is checkable rather than a state change:

**Both bindings resolve to the nodes the old ones did.**

- `signalment`: the `data-testid` sits on the `<p>` that `getByText('Allergy: Penicillin').parentElement` used to return — the allergy renders as a `<span>` nested in that paragraph, so `.parentElement` was the `<p>`. Same node, same `truncate`, so the margin table's numbers still describe the thing asserted about.
- `allergy`: `within(signalment).getByText(/^Allergy: /)` resolves to that same `<span>`. The `^` anchor is doing real work — the paragraph's own text content starts with the breed, not with `Allergy:`, so the anchor excludes it, and the span is the only descendant. One match, unambiguous, and independent of the allergy value.

**The three assertions that are not about the allergy string** — `textOverflow: ellipsis`, the ink-difference `waitFor`, and `fontWeight: 700` — now execute under a change to `args.allergy` instead of being skipped by a dead lookup. That is the property the change exists to restore, and it is the one the old binding silently removed.

**Base and state:** merge-base with `dev` is `afe9d24da`, the live remote tip; 2 ahead, 0 behind. Checks at this head: 16 pass, 9 pending, 4 skipping, none failing. **This approval is not a merge recommendation** — the build, Chromatic and the four Storybook shards have not completed and I have not read a finished run.

---

One non-blocking observation, from the arm log rather than the diff.

Line 282 asserts `toContain('… Allergy: Penicillin')`. A superstring mutation of `args.allergy` satisfies that assertion, which is exactly why the `Amoxicillin` arm was needed and the `Penicillinxx` arm could not settle it: a superstring cannot separate an assertion that ran and passed from one that never ran.

That property now lives only in the run log. The file says the binding must not be the asserted string, but not that a mutation of this fixture has to be a **non-superstring** to discriminate. One clause in the existing comment would carry it to the next person who reaches for `Penicillinxx` because it is already in the margin table above.

**Comment by ankit-yc** [comment]
Verification of `3c73e9398` — and a disclosure first: **I proposed these two lines, so this is a check
that the change was applied and behaves as claimed, not an independent review.** The independent read
is review `5138076583`.

### Applied exactly

One hunk, one file. `getByTestId('patient-signalment')` for the paragraph,
`within(signalment).getByText(/^Allergy: /)` for the tail, and line 282's `toContain` left as the one
place the string is asserted. No live text-derived binding remains in the file: the only other
`.parentElement` occurrences are line 336 (inside the block comment) and line 367
(`signalment.parentElement`, a containment assertion off a testid binding).

### An independent instrument — jest + RTL against the PR-head component

```
N1 shipped        getByTestId === container.querySelector('p.truncate');
                  within(signalment).getByText(/^Allergy: /) is the SPAN, parent === signalment
N2 Amoxicillin    both bindings still resolve; tail text is 'Allergy: Amoxicillin'
N3 Amoxicillin    signalment.textContent does NOT contain the shipped string
                  -> line 282 FAILS on the mutation instead of never running
N4 CONTROL        the OLD binding dies on the same mutation
N5 CONTROL        getByText is EXACT, so it throws on 'Penicillinxx' too — while
                  toContain(shipped) PASSES on it
N6 MUST-NOT-FIRE  getAllByText(/^Allergy: /) inside the paragraph has length 1

6 passed / 6 total, rc=0

MUTATION on the SUBJECT: rename data-testid in the component -> 5 of 6 fail, rc=1.
N4 survives, correctly — it is the only arm that does not depend on the attribute.
```

N5 is the same mechanism as A4 read from the other side: the superstring is caught by `getByText`
(exact) but satisfies `toContain`, so a `Penicillinxx` arm leaves the `toContain` line's green
ambiguous while the lookup arms are unambiguous.

### The `^` anchor is load-bearing and it is safe here

`within(signalment).getByText(/^Allergy: /)` is unambiguous only because the paragraph's own text
starts with the breed. Checked across the whole file: no story overrides `breed` / `ageLabel` /
`weightKg`, they are set once in `meta.args`, so the paragraph can never begin with `Allergy:`.
`N6` is the assertion of that, and `getAllByText` is what makes a second match visible rather than a
throw.

### What the metric says about this PR, both commits

Whole-population sweep, 2797 `*.stories|test|spec.tsx?` files, comment-stripped:

```
ref                       inline RAW   STRIPPED   decomposed    this file (raw/stripped/decomposed)
dev tip   afe9d24da           95          95        45 / 37            1 / 1 / 1
commit 1  86cfc33c9           95          94        45 / 37            1 / 0 / 1
commit 2  3c73e9398           95          94        44 / 36            1 / 0 / 0
```

Both occurrences in this file are now gone from the stripped metric, and the raw inline count is 95
at every ref — unchanged across a PR that removes two, because each commit's own explanatory comment
reproduces the string it removed.

**Review by ankit-yc — APPROVED** [review]
Approving `aa3429571`. The delta from the previously approved tree is nine comment lines in one file, no code change, `git diff --check` exit 0, base 3 ahead / 0 behind `dev`. The clause is accurate and it is the one that was missing.

**Correcting my own previous approval body, at the artefact rather than in passing.**

I wrote that `within(signalment).getByText(/^Allergy: /)` is unambiguous *because of the anchor*, and that dropping `^` would produce a two-element match that throws. That is wrong. Measured against `@testing-library/dom@10.4.1` as installed, with jsdom and the same DOM the component renders:

```
ANCHORED    within(p).getByText(/^Allergy: /)   -> SPAN, 1 match
UNANCHORED  within(p).getByText(/Allergy: /)    -> SPAN, 1 match
CONTROL     within(p).getByText(/^Beagle/)      -> P        <- the container IS a candidate
CONTROL     within(p).getByText(/Zzqx-nothing/) -> throws "Unable to find"
CONTROL     an ambiguous query                  -> throws "Found multiple elements"
```

The mechanism is neither the anchor nor `within` scoping to descendants — `queryAllByText` explicitly adds the container to the candidate set when it matches the selector. It is `getNodeText`, which returns only a node's **direct text-node children**:

```
getNodeText(p) = "Beagle · 9 yr · 12.4 kg · "
p.textContent  = "Beagle · 9 yr · 12.4 kg · Allergy: Penicillin"
```

The paragraph is a candidate and fails on text. **The uniqueness comes from the `<span>` wrapper in the component**, not from the query.

**And the anchor is not free.** With the span removed and the allergy rendered as a direct text child — a plausible refactor, since the span exists to carry the danger colour and weight:

```
ANCHORED    -> throws "Unable to find an element"     <- fails at the LOOKUP
UNANCHORED  -> binds the paragraph; the ink assertion then compares the paragraph
               to itself and fails on the comparison
```

Both spellings fail, but the anchored one fails in the mode this PR exists to remove: at the lookup, before the assertions run. Not a reason to change it today — the span carries required styling and the refactor is hypothetical — but "tighter and costs nothing" is measurably wrong in the one scenario where the two spellings differ, and it is worth knowing before someone unwraps that span.

PR #2845, opened 2026-09-07
fix(repo): unify PIMS audit surfaces

ankit-yc:
## PR Checklist

- [x] The PR title follows our guidelines: https://github.com/YosemiteCrew/Yosemite-Crew/blob/main/CONTRIBUTING.md#commit-message-convention.
- [ ] There is an issue for the bug/feature this PR is for. This is a consolidated PIMS UI audit and no standalone issue was supplied.
- [x] All existing targeted tests, lint, type checks, hooks, and the Aikido scan pass.

## What is the current behavior?

PIMS surfaces used inconsistent filters, mobile table treatments, card backgrounds, and alert layouts. Inventory rendered two unbounded competing alert lists, appointment navigation exposed redundant check-in navigation, and the controlled-drug register reused unrelated permissions.

## What is the new behavior?

The audit standardizes shared filters and responsive surfaces across Finance, Inventory, Appointments, Organization, Forms, Integrations, and Storybook. Inventory now has one primary catalog with bounded alert summaries; the controlled-drug register has explicit read, record, and correction permissions plus a complete phone layout. The audit record captures visual verification and live deployment observations.

## Related Issue(s)

No issue supplied; findings are documented in `docs/pims-ui-audit.md`.

Fixes #

## Verification

- `pnpm --filter frontend exec jest --runInBand <25 changed frontend suites>` - 25 suites, 546 tests passed
- `pnpm --filter backend exec jest --runInBand test/routers/controlled-substance-log.router.authz.test.ts` - 1 suite, 6 tests passed
- Frontend and backend type checks passed
- Frontend and backend lint passed with existing warnings only
- Pre-push monorepo lint: 10 tasks succeeded; type-check: 17 tasks succeeded
- Aikido scan of 31 changed production files returned no issues

PR #2857, opened 2026-09-08
fix(frontend): clear stale PIMS review scopes

ankit-yc:
## PR Checklist

- [ ] There is an issue for the bug/feature this PR is for. This follows review feedback on #2845.

Inventory alert handoffs could keep their hidden item scope after normal catalog filtering, and appointment validation compared the selected clinic day against the browser-local day.

## What is the new behavior?

Normal inventory filters and organization changes clear the alert handoff scope. Appointment validation now compares calendar-day keys in the preferred timezone. Targeted Jest coverage, type-check, lint, and Aikido scan pass.

## Related Issue(s)

Review follow-up for #2845.

Fixes #

### DIFF
--- a/apps/frontend/src/app/__tests__/hooks/useAppointmentForm.test.ts
+++ b/apps/frontend/src/app/__tests__/hooks/useAppointmentForm.test.ts
@@ -69,6 +69,7 @@ jest.mock('@/app/lib/timezone', () => ({
     d.setHours(Math.floor(minuteOfDay / 60), minuteOfDay % 60, 0, 0);
     return d;
   }),
+  getDateKeyInPreferredTimeZone: jest.fn((date: Date) => date.toISOString().slice(0, 10)),
   isOnPreferredTimeZoneCalendarDay: jest.fn(() => false),
   utcClockTimeToPreferredTimeZoneClock: jest.fn((_time: string) => ({
     minutes: 540,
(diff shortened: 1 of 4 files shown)


aupyay:
`test / Merge coverage (frontend)` failed on `Unable to download artifact(s): Artifact download failed after 5 retries` — the four `coverage-frontend-shard-*` artifacts were listed and the download itself timed out, so this was infrastructure and not the diff. Re-ran the failed jobs on the same head (ac3c1381); no code change.

aupyay (approved):
Approving. Reviewed the four files; the two changes are independent and both look right.

**`useAppointmentForm`** — the past-date guard compared a `Date` against a local-midnight `Date`,
so the browser's zone decided the answer. Comparing `getDateKeyInPreferredTimeZone` on both sides
is the correct fix, and the `<` is safe: that helper returns a zero-padded `YYYY-MM-DD`
(`lib/timezone.ts:385`), so lexicographic order is date order. The new case discriminates — at
`2026-09-09T00:30Z` with a selected date of `2026-09-08T12:00Z`, the old code flags a past date
and the new code does not.

**Inventory** — routing every mutation through `updateFilters` clears `alertedItemIds`, so a
low-stock handoff cannot survive a filter change; the `useOnValueChange(primaryOrgId, ...)` guard
does the same on an org switch. `updateFilters` closes over only stable setters, so the empty
dependency list is correct and the `useMemo` deps that gained it are honest.

Two notes, neither blocking:

- The title names review scopes but the diff is two unrelated fixes. Fine to land as one; worth
  splitting if either needs to be reverted on its own.
- `getDateKeyInPreferredTimeZone` is mocked with a single `mockReturnValue`, so both sides of the
  comparison return the same key. That is what makes the case pass, and it means the test cannot
  see an ordering bug in the comparison itself — only the absence of the old local-midnight
  behaviour. A second case with a per-argument implementation would cover the ordering.

`CI Required` is green at `ac3c1381`. The earlier red on `test / Merge coverage (frontend)` was an
artifact-download timeout, not the diff; re-running the failed jobs on the same head cleared it.

Approving from `aupyay` rather than the usual review account because GitHub blocks a PR's author
from approving it and this PR is authored by `ankit-yc` — same routing as #2844.

PR #2877, opened 2026-09-08
fix(frontend): separate filter and navigation controls in finance headers

aupyay:
## Summary
- Finance, Estimates and Insurance claims headers each mixed page-navigation buttons (and, on Finance, the Stripe status pill) into the same row as the status filter chips, so three unrelated control types read as one inconsistent row of pill sizes.
- Stack page actions above the status filter in each header, matching the existing phone layout, which already keeps them apart.

## Test plan
- [x] `npx tsc --noEmit` (apps/frontend) - clean
- [x] `pnpm --filter frontend run lint` - 0 errors
- [x] `pnpm --filter frontend run test -- --testPathPatterns="Finance/Estimates|Finance/index|InsuranceClaims"` - 4 suites / 72 tests pass

### DIFF
--- a/apps/frontend/src/app/__tests__/pages/Finance/Estimates.test.tsx
+++ b/apps/frontend/src/app/__tests__/pages/Finance/Estimates.test.tsx
@@ -570,6 +570,21 @@ describe('Finance > Estimates page', () => {
       '/finance'
     );
   });
+
+  it('stacks page actions above the status filter, not beside it', async () => {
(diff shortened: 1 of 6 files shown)


ankit-yc (approved):
Verified: DOM-order regression tests for all three headers fail on pre-fix source and pass on the fix (mutation-checked). tsc/lint clean, 42/42 targeted tests pass, all CI gates green including Tests Must Be Able To Fail.

PR #2878, opened 2026-09-08
fix(backend): exclude expired batches from expiring alerts

aupyay:
## PR Checklist

The expiring-inventory query applies only an upper date bound, so batches that have already expired can appear in the Expiring soon list.

## What is the new behavior?

The shared inventory alert service now bounds the expiry window from the current time through the configured warning threshold. A regression test verifies the lower bound.

Backend lint and type-check pass. The focused inventory suite passes 177 tests with 99.95% statement and line coverage. The complete backend suite reported 486 suites and 11,543 tests passing, but did not terminate because existing Redis queue handles kept reconnecting after completion.

## Related Issue(s)

Not linked.

### DIFF
--- a/apps/backend/src/services/inventory.service.ts
+++ b/apps/backend/src/services/inventory.service.ts
@@ -2413,7 +2413,7 @@ export const InventoryAlertService = {
     return prisma.inventoryBatch.findMany({
       where: {
         organisationId: safeOrganisationId,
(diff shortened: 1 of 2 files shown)


ankit-yc (approved):
Backend lint/type-check pass, inventory suite 177 tests 99.95% coverage, regression test verifies the lower bound. Approved.

PR #2883, opened 2026-09-08
chore(backend): remove unused bcrypt dependency

aupyay:

bcrypt@5.1.1 and @types/bcrypt@^5.0.2 are declared in `apps/backend/package.json` but never imported or exercised anywhere in the codebase. The only source mention is a comment in `apps/backend/src/services/developer-api-key.service.ts` (line 65) explaining why bcrypt is deliberately NOT used: the Developer API key digest must be deterministic for indexed lookup, which rules out salted hashes like bcrypt in favor of a keyed HMAC.

## What is the new behavior?

Both dependencies are removed from `apps/backend/package.json` and `pnpm-lock.yaml`. This also drops the transitive `node-pre-gyp` install chain (~136 lockfile lines).

This resolves the blocking review on Dependabot PR #2840 (bump bcrypt to 6.0.0), which cannot be merged as a bump because the dependency is unused.

## Related Issue(s)

Supersedes #2840 (dependency bump PR, closing).

Fixes #2840


ankit-yc (approved):
bcrypt and @types/bcrypt confirmed unused (grep clean, only a comment referencing why it's deliberately not used). Lockfile diff is scoped to the removal. CI green.

PR #2889, opened 2026-09-08
refactor(frontend): remove dead SpecialitiesTable, DocumentsTable, RoomTable components

ankit-yc:
## Summary

Removes three dead table components from `apps/frontend/src/app/ui/tables/` that a design-system audit against `origin/dev` confirmed have zero production importers:

- `SpecialitiesTable.tsx` - the live Specialities screen (`apps/frontend/src/app/features/organization/pages/Organization/Sections/Specialities/Specialities.tsx:2`) imports `SpecialitiesTableRevamp` instead.
- `DocumentsTable.tsx` - the real Documents section was rebuilt as cards (`AllDocumentsTable` in `apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/steps/SummaryStep.tsx:226`), never imports this.
- `RoomTable.tsx` - the real Rooms section was rebuilt as cards, never imports this.

For each component this PR deletes the component file, its `.stories.tsx`, and its Jest test file, and removes its export line from the `tables/index.ts` barrel and its row from `apps/frontend/src/app/ui/INVENTORY.md`.

Before deleting, each component was independently re-verified with `git grep -n "<Name>\b" -- apps/frontend/src apps/desktop apps/mobileAppYC`. The only matches for all three were the component's own file, its story, its test, and the barrel re-export - no other importer exists anywhere in the repo. A few substring matches surfaced by the grep (`AllDocumentsTable`, prose mentioning "DocumentsTable" in `DocumentsCard.stories.tsx`, `// Mock helper function from RoomTable`/`SpecialitiesTable` comments in unrelated card tests) were checked individually and confirmed to be unrelated identifiers/comments, not real imports.

## Verified

- `git grep -n "SpecialitiesTable\b|DocumentsTable\b|RoomTable\b"` and a path-based `git grep` for `tables/DocumentsTable|tables/RoomTable|tables/SpecialitiesTable` across the whole repo (`apps/frontend`, `apps/desktop`, `apps/mobileAppYC`) after deletion: only the pre-existing unrelated comments remain, zero real importers.
- `npx tsc --noEmit` from `apps/frontend`: passes with zero errors - proof a real remaining import would have surfaced as a compile error.
- `npx eslint <changed files>` (`tables/index.ts`): clean.
- `pnpm --filter frontend run lint` (full frontend ESLint run): clean.

## Test plan

- [x] `git grep` re-verification per component showed only self-references before deleting
- [x] `npx tsc --noEmit` (apps/frontend) - 0 errors
- [x] `npx eslint` on changed files - 0 errors/warnings
- [x] `pnpm --filter frontend run lint` (full frontend) - 0 errors/warnings
- [x] No test files reference the deleted components (their own tests were deleted alongside them; no other test imports them)


aupyay:
## Overriding \`Tests Must Be Able To Fail\` — documented

This PR deletes three components (\`SpecialitiesTable.tsx\`, \`DocumentsTable.tsx\`, \`RoomTable.tsx\`) along with their own tests. After #2893 and #2895 fixed this gate's two crash/false-positive bugs on deletion-only PRs, it now correctly reports:

> Every changed test file is one the branch DELETED, so none of them exist to run - \`jest\` against a path that is gone reports zero tests, which is not evidence either way.

That's the honest, correct verdict for this class of change — a jest-based gate genuinely cannot prove a deletion is safe, because there's no test content left to run either direction. The \`no-behaviour-change\` label is applied (documents the call) but by design doesn't flip this specific "nothing runnable" verdict (same as the pre-existing e2e-only case) — the actual verification for a deletion is a different kind of evidence:

- \`git grep\` for each deleted component's name across \`apps/frontend/src\`, \`apps/desktop\`, \`apps/mobileAppYC\` found only the component's own file, its story, its test, and the barrel re-export — no real importer, for all three.
- \`npx tsc --noEmit\` from \`apps/frontend\` is clean — a real remaining import would be a compile error, which is the actual proof "zero consumers" holds.
- Full monorepo pre-push hook (turbo lint + type-check, 17 tasks) passed before every push to this branch.

Overriding this one check with \`--admin\`; every other required check is green.

PR #2892, opened 2026-09-08
fix(frontend): centralize developer guard bypass

### LINKED ISSUE #2805: fix(frontend): the local guard bypass is implemented twice, contrary to its own contract
`apps/frontend/src/app/lib/localGuardBypass.ts` exists so that the local-only development bypass has exactly one implementation. Its docblock is explicit about why:

> Every consumer of the flag has to go through this helper. Reading `process.env.NEXT_PUBLIC_DISABLE_AUTH_GUARD` directly re-introduces the unguarded form, which is how the session initializer and the sidebar ended up rendering the private shell on a deployed host while their comments claimed they could not.

`apps/frontend/src/app/ui/layout/guards/DevRouteGuard/DevRouteGuard.tsx` does not import the helper. It defines a private `isLocalDeveloperFallbackEnabled()` with its own copy of the same two conditions.

**This is not a bypass.** The copy is currently equivalent to the helper, and the guard it feeds is additionally gated on a session-storage value, so nothing is reachable that should not be. Verified against the deployed production artifact: neither variable is baked to an enabling value, and the conditions fail closed in every partial state.

**The problem is that the control now lives in two places and only one of them is the one people harden.** Any future tightening of the helper — a stricter runtime check, removing the hostname override, resolving at build time instead of runtime — silently does not apply to the developer routes. The comparison is not hypothetical: the module was written *because* two components had drifted this way once already.

This is the same shape as the licence-authority defect fixed in #2798 this week: two readers of one variable, each with its own copy of the default, one corrected and the other left behind.

### Suggested fix

- `DevRouteGuard` imports `isLocalGuardBypassEnabled` and deletes its local copy.
- A test asserts the module has exactly one reader of that variable in `src/`, in the shape of the one added in #2798, so a third copy fails rather than passes.

### Verification already done

Whole-tree scan of `origin/main`: three runtime consumers, two via the helper, one with its own copy. No other file reads the variable outside tests, stories and config examples.

aupyay:
## PR Checklist

`DevRouteGuard` duplicates the local auth-guard bypass flag and hostname checks, so future changes to the shared guard helper can leave developer routes behind.

## What is the new behavior?

`DevRouteGuard` uses the render-safe shared bypass hook. A source-level regression test requires the public bypass flag to have exactly one production reader.

Validation:

- Frontend TypeScript check
- Strict ESLint on both changed files
- Targeted `localGuardBypass` and `DevRouteGuard` tests: 2 suites, 22 tests
- Pre-push monorepo lint and type-check hooks

## Related Issue(s)

Fixes #2805

### DIFF
--- a/apps/frontend/src/app/__tests__/lib/localGuardBypass.test.tsx
+++ b/apps/frontend/src/app/__tests__/lib/localGuardBypass.test.tsx
@@ -154,4 +154,21 @@ describe('the render-time form', () => {
     // answer. The plain function reads window.location and diverges.
     expect(offenders.map((f) => relative(root, f))).toEqual([]);
   });
+
+  it('keeps the public bypass flag behind the shared helper', () => {
+    const root = join(__dirname, '..', '..');
+    const walkSource = (dir: string): string[] =>
+      readdirSync(dir, { withFileTypes: true }).flatMap((entry) => {
+        const full = join(dir, entry.name);
+        if (entry.isDirectory()) return entry.name === '__tests__' ? [] : walkSource(full);
+        return entry.isFile() && /\.tsx?$/.test(full) && !full.endsWith('.stories.tsx')
+          ? [full]
+          : [];
+      });
+    const readers = walkSource(root).filter((file) =>
+      /process\.env\.NEXT_PUBLIC_DISABLE_AUTH_GUARD/.test(readFileSync(file, 'utf8'))
+    );
+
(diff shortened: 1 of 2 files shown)


ankit-yc (approved):
Reviewed the full two-file diff at d57d24cc7. The guard now uses the render-safe shared hook, and the regression test pins the flag to the helper module. Targeted tests, strict changed-file lint, frontend type-check, and pre-push monorepo lint/type-check are green.

PR #2896, opened 2026-09-08
fix(frontend): withhold the active count pill when a clinical panel is in error state

aupyay:
PR #2873 made the clinical panel body show only the error state when a load fails, but left the header active-count pill independent of the error. The header could claim '2 active' while the body showed an error and no records.

This withholds the pill on error across all five clinical panels (allergies, consents, flags, problems, plus the shared chrome). Added a test to each list suite pinning the count is withheld while the error hides the retained records, and a red/green case in ClinicalListChrome. Full companionHistory suite: 24 suites / 403 tests pass.


ankit-yc (approved):
Error state now suppresses the active-count pill consistently across ClinicalListHeader and all five list consumers; each gets a dedicated regression test. LGTM.

PR #2897, opened 2026-09-08
fix(backend): accept negative keyset timestamps

### LINKED ISSUE #2723: fix(backend): keyset cursor encoder accepts timestamps its own parser rejects
`encodeKeysetCursor` and `parseKeysetCursor` in `apps/backend/src/services/shared/pagination.ts` are meant to be inverses, and for one input class they are not.

`encodeKeysetCursor` writes `createdAt.getTime()` unguarded. `parseKeysetCursor` accepts `^\d{1,15}$`, which has no sign. So any pre-1970 `createdAt` encodes to a cursor the endpoint's own controller answers **400** to:

```
1970-01-01T00:00:00.000Z   getTime = 0     -> parses
1969-12-31T23:59:59.999Z   getTime = -1    -> rejected as malformed
2026-09-05T00:00:00.000Z                   -> parses
```

Found by Claude L3 Yosemite Crew during the review of #2720 and confirmed there.

### Reachability

Not reachable today. `Prescription.createdAt` is `@default(now())` and nothing backdates it, so `getTime()` cannot be negative on this surface. The round-trip test cannot see it either — it uses a single fixed date, so a widened parser is not what keeps it green.

It becomes reachable the moment a model paged this way carries an imported or backfilled `createdAt`, which is a normal thing for a migration to do and would fail as "your cursor is malformed" on page two rather than anywhere near the import.

### Suggested fix

Widen the parser to `^-?\d{1,15}$` — the digit bound is there to keep the value inside `Date`'s range and a sign does not change that, since 15 digits either way is well inside 8640000000000000. Add a round-trip case for a pre-1970 date, which reddens if the sign is dropped again.

That is one character and one test rather than a guard, so it does not add a branch nothing can exercise.

### Not doing

Rejecting negatives deliberately in the encoder instead. That trades a 400 the caller cannot act on for a 500 the caller cannot act on, and neither is better than the two functions simply being inverses.

aupyay:
## PR Checklist

`encodeKeysetCursor` emits a negative millisecond timestamp for dates before the Unix epoch, but `parseKeysetCursor` rejects that cursor as malformed.

## What is the new behavior?

The keyset parser accepts an optional minus sign while keeping the existing 15-digit bound. A regression test covers a pre-1970 encode/parse round trip.

Validation performed:

- Focused backend suite: 33 tests passed; `pagination.ts` reached 100% statements, branches, functions, and lines.
- Mutation check: restoring the unsigned timestamp pattern made the new pre-epoch test fail; restoring the fix returned the suite to green.
- Backend lint: 0 errors (25 pre-existing warnings outside the changed files).
- Backend type-check: passed.
- Pre-push repository gates: 10/10 lint tasks and 17/17 type-check tasks passed.
- Aikido scan: 0 issues across the two changed files.
- Local Sonar Vortex analysis was unavailable for this connection; the required CI Sonar analysis remains unchanged.

## Related Issue(s)

Fixes #2723

### DIFF
--- a/apps/backend/src/services/shared/pagination.ts
+++ b/apps/backend/src/services/shared/pagination.ts
@@ -89,7 +89,7 @@ export type KeysetCursor = {
 };
 
 const KEYSET_SEPARATOR = "|";
-const KEYSET_MILLIS = /^\d{1,15}$/;
+const KEYSET_MILLIS = /^-?\d{1,15}$/;
 
 /**
  * A cursor that carries the whole sort key, for `(createdAt, id)` lists.
--- a/apps/backend/test/services/shared/pagination.test.ts
+++ b/apps/backend/test/services/shared/pagination.test.ts
@@ -116,6 +116,13 @@ describe("keyset cursors", () => {
     ).toEqual({ createdAt: CREATED_AT, id: ID });
   });
 
+  it("round-trips a timestamp before the Unix epoch", () => {
+    const createdAt = new Date("1969-12-31T23:59:59.999Z");
+    expect(
+      parseKeysetCursor(encodeKeysetCursor({ createdAt, id: ID })),
+    ).toEqual({ createdAt, id: ID });
+  });
+
   /*
    * `createdAt` is TIMESTAMP(3) on every model paged this way, so the
    * millisecond is the whole precision of the column. A cursor that dropped it


ankit-yc (approved):
Reviewed the parser boundary and regression test at 6d85dd0bd. The optional sign preserves the existing digit bound, and the mutation check proves the pre-epoch case exercises the change.

PR #2911, opened 2026-09-09
fix(backend): sync Appointment.room when front desk reassigns a check-in's room

aupyay:
## Summary
- Front desk's `PatientCheckIn.assignedRoomId` and the schedule's `Appointment.room` were two independent writes tracking the same fact: which room the patient is currently in.
- Reassigning a room from the front desk never touched the linked appointment, so the Board kept showing whatever room the appointment was originally booked into even after the patient moved.
- `PatientCheckInService.assignRoom` now also syncs `Appointment.room` (best-effort, mirroring the existing `syncAppointmentOnArrival` pattern from #2885) when the check-in is linked to an appointment.
- Added `AppointmentPrismaService.updateAppointmentRoom`, a small targeted write (not the general-purpose `updateAppointmentPMS`, which requires a full `AppointmentRequestDTO` and handles unrelated concerns).

## Test plan
- [x] `patient-check-in.service.test.ts`: 24/24 passing, including 5 new cases covering the sync (linked appointment, unlinked check-in, unknown room, appointment update failure is swallowed, unexpected error still propagates)
- [x] Mutation-checked: reverted the source to HEAD with the new tests still in place — 2 tests failed as expected, confirming they actually exercise the fix
- [x] `tsc --noEmit`: zero new errors introduced (confirmed against the same pre-existing baseline on unmodified HEAD)
- Pre-existing typed-lint noise in this file (`no-unsafe-member-access` on `prisma.*`) is a worktree-local stale type-resolution artifact, confirmed present on unmodified HEAD before this change

### DIFF
--- a/apps/backend/src/services/appointment.prisma.service.ts
+++ b/apps/backend/src/services/appointment.prisma.service.ts
@@ -1862,6 +1862,34 @@ export const AppointmentPrismaService = {
     return toResponse(updated);
   },
 
+  async updateAppointmentRoom(
+    appointmentId: string,
+    organisationId: string,
+    room: { id: string; name: string },
+  ) {
+    if (!appointmentId) {
+      throw new AppointmentPrismaServiceError("appointmentId is required", 400);
+    }
+    if (!organisationId) {
+      throw new AppointmentPrismaServiceError(
+        "organisationId is required",
+        400,
+      );
+    }
+
+    const current = await prisma.appointment.findFirst({
+      where: { id: appointmentId, organisationId },
+    });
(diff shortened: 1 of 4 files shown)


ankit-yc (approved):
Reviewed. Real Sonar new-coverage gap (68.8% vs floor 80%) on a genuinely untested method - added guard-clause, 404, and happy-path tests mirroring the existing checkInAppointment pattern, mutation-checked. Then Aikido flagged the DTO conversion querying payment state on a discarded return value - correct finding, dropped it, and the test now pins that with an explicit invoice.findMany assertion (mutation-checked by reintroducing the old code).

LGTM.

Issue #2853, opened 2026-09-08
fix(frontend): Storybook offline guard excludes subresources, and 13 story files load images from production hosts


aupyay:
## The premise in the comment is false today

`apps/frontend/.storybook/offline-guard.ts` wraps `fetch` and `XMLHttpRequest` so that "no request left the browser", and states its own exclusion:

> NOT COVERED, on purpose: `WebSocket`, `navigator.sendBeacon`, and subresource loads (`<img src>`, `<script src>`), none of which go through `fetch`/XHR. **No story in the suite has been observed to need them**, and widening the guard to the whole network layer would start intercepting things Storybook itself depends on.

The exclusion is a reasonable design choice. The justification attached to it is no longer true.

## Measured on `dev` at `afe9d24dade345d4e2cd975200cdbcc940b17274`

```
git grep -lE "https://[a-zA-Z0-9./_-]+\.(png|jpe?g|gif|svg|webp|avif|ico)" dev -- '*.stories.tsx'
=> 13 files

hosts, by reference count:
   9  d2il6osz49gpup.cloudfront.net              production CDN
   1  yosemitecrew-backend.s3.eu-central-1...    production bucket
   4  example.invalid                            RFC 2606 reserved
   1  files.example.com                          documentation domain
```

Ten of the fifteen references point at production hosts. Those are `<img src>` subresources, so they pass the guard by construction and leave the browser on every run of the Storybook interactions job, in CI and locally.

<details>
<summary>The 13 story files</summary>

```
apps/frontend/src/app/features/appointments/components/AppointmentCentralModal/AppointmentAvatar.stories.tsx
apps/frontend/src/app/features/appointments/components/Calendar/common/PopoverHeader.stories.tsx
apps/frontend/src/app/features/appointments/components/Calendar/common/UserLabels.stories.tsx
apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/CompanionContextCard.stories.tsx
apps/frontend/src/app/features/appointments/pages/AppointmentWorkspace/components/StaffField.stories.tsx
apps/frontend/src/app/features/companions/components/CompanionInfo.stories.tsx
apps/frontend/src/app/features/companions/pages/Companions/InClinicTodayBand.stories.tsx
apps/frontend/src/app/features/organization/pages/Organization/Sections/Documents/Documents.stories.tsx
apps/frontend/src/app/features/petPassport/components/WalletPassPreview.stories.tsx
[...]

PR #2912, opened 2026-09-09
fix(frontend): serve Storybook avatar fixtures locally instead of the CDN

aupyay:
## Summary

Closes 2 of 10 production-CDN references from #2853 (Storybook offline guard
excludes `` subresources, and several stories load avatar photos from
`d2il6osz49gpup.cloudfront.net`).

`AvatarImage.stories.tsx` and `UserLabels.stories.tsx` pass their photo URL
straight to `next/image` with no sanitiser in between, so both can point at a
same-origin fixture (`public/images/storybook-fixtures/avatar-photo.png`)
instead of the production CDN, with no production code touched.

## Why the other 8 files in #2853 are not in this PR

Measured, not assumed: I swapped the fixture URL in `StaffField.stories.tsx`
(and confirmed the same for `AppointmentAvatar`, `PopoverHeader`,
`CompanionContextCard`, `CompanionInfo`, `InClinicTodayBand`,
`WalletPassPreview`, `Personal`) and it still hit `cloudfront.net` - just for
a different image.

Those components route the photo through `getSafeImageUrl`/`isHttpsAvatar`
first (`apps/frontend/src/app/lib/urls.ts`), which requires a literal
`https://` prefix and falls back to `MEDIA_SOURCES.avatars.*` when the value
fails that check - and that fallback pool is the same production CDN, not a
local asset. Since the `` attribute is the actual fetch target,
there is no DOM-property override that can point the request elsewhere while
still presenting an `https://` value for the sanitiser to accept. Fixing
those six needs a network-layer intercept (a Service Worker), not a
story-level URL change - noted as a limitation in `offline-guard.ts` so the
next person does not re-discover this from scratch.

Two references in the original 13-file list (`CompanionAvatar.stories.tsx`,
`GuidePlayerModal.stories.tsx`) already use `example.invalid`/
`files.example.com`, which are already safe - no change needed there.

## Verification

- `npx tsc --noEmit` (frontend): clean
- `pnpm run lint` (frontend): clean (one pre-existing warning, unrelated file)
- Manually verified both fixed stories in a running Storybook via Playwright:
  the rendered `` resolves to the local fixture, `naturalWidth`/
  `complete` show a real successful load, the play-function assertions pass
  (0 console errors), and `browser_network_requests` shows zero requests to
  `cloudfront.net` or `amazonaws.com`.
- Full `test-storybook` run (3,291 stories) not run locally - out of budget
  for a 2-file change; CI's advisory Storybook interactions job covers it.

## Test plan

- [ ] CI: tsc, lint, advisory Storybook interactions job green
- [ ] Spot check `Avatars/AvatarImage` → `Photo loads` and
      `Appointments/Calendar/UserLabels` → `Three columns` in the deployed
      Storybook, confirm no CloudFront request in the network tab


ankit-yc (dismissed):
Reviewed the diff directly (author's own PR, not a rubber stamp check).

Verified independently:
- getSafeImageUrl/isHttpsAvatar fallback claim: read urls.ts and mediaSources.ts myself - DEFAULT_IMAGES pool does resolve to the same CloudFront host, confirming the 6-file scope-down is correct rather than convenient.
- offline-guard.ts comment update is accurate and matches the measured state.
- No production code touched; only story files, a fixture asset, and a comment.

LGTM.

ankit-yc (approved):
Re-reviewed after the revert (previous approval dismissed by the follow-up pushes). Final diff: AvatarImage.stories.tsx and UserLabels.stories.tsx swapped to a local same-origin fixture, no production code touched. Verified independently the two components bypass getSafeImageUrl/isHttpsAvatar (raw src pass-through), which is why this fix is safe and why the other 8 files in #2853 are correctly out of scope here.

LGTM.

PR #2913, opened 2026-09-09
test(frontend): add Storybook coverage for OrgGuard

aupyay:
## Summary
- `apps/frontend/src/app/ui/layout/guards/OrgGuard.tsx` is a ~382-line guard imported by roughly 26 other files (every top-level org-scoped route effectively renders through it), yet had zero Storybook coverage - confirmed via `find . -iname "*OrgGuard*"`, which turned up only the component itself and one Jest unit test, no `.stories.tsx`.
- Added `apps/frontend/src/app/ui/layout/guards/OrgGuard.stories.tsx` covering the four branches the component actually has, each with a play function asserting concrete rendered output (not "renders without crashing"):
  - **Loading** - `orgStore.status: 'loading'` is a pending status (`OrgGuard.tsx:45`), so the guard shows its `skeleton` and calls neither `redirect()` nor renders children.
  - **Allowed** - a verified org, an owner membership, a completed profile and a published availability slot; every onboarding gate clears and the guard renders `children`.
  - **PermissionDenied** - an Admin membership with `integrations:view:any` explicitly revoked on `/integrations`; `resolveMembershipPermissions` (`routePermissions.ts:73`) derives the effective set from the role table plus revocations, so the guard redirects to the membership's first still-accessible route (`/dashboard`) rather than rendering.
  - **OrgFetchFailed** - `orgStore.status: 'error'`. `isStatusPending` (`OrgGuard.tsx:45`) deliberately excludes `'error'`, so the guard treats a failed org fetch as settled (not pending) and falls back to `/organizations` instead of holding the skeleton open forever.
- Followed the store-seeding + offline-XHR pattern this repo already uses for org-scoped pages (`Dashboard.stories.tsx`), and the `redirect()`-mock override pattern from the sibling guard-style story `AuthedRedirectShell.stories.tsx` (both added in `ade394cad`, the commit the coverage-gap audit referenced).

## Verified
- `npx tsc --noEmit` (apps/frontend, full project): clean, zero errors.
- `npx eslint src/app/ui/layout/guards/OrgGuard.stories.tsx`: clean.
- `npx test-storybook --url http://localhost:6117 --failOnConsole --no-cache -- "guards/OrgGuard"`: all 4 stories pass (`Loading`, `Allowed`, `PermissionDenied`, `OrgFetchFailed` - 4/4, 0 console errors).
- Mutation-check (spot-check, this is new coverage so no source behavior changed): temporarily broke `isStatusPending` in `OrgGuard.tsx` to also treat `'error'` as pending, reran the same test-storybook command - `OrgFetchFailed` failed as expected (`redirect` never called, timed out waiting for `/organizations`) while the other three stories, which don't touch that branch, still passed. Reverted with `git checkout HEAD -- apps/frontend/src/app/ui/layout/guards/OrgGuard.tsx` and confirmed `git diff` against the source file is empty before pushing.
- Existing coverage untouched: `npx jest --testPathPatterns="OrgGuard\.test\.tsx"` - 12/12 passing, unchanged.
- `git diff HEAD~1 HEAD -- apps/frontend/src/app/ui/layout/guards/OrgGuard.tsx` is empty - this PR adds only the new stories file, no source-file change shipped.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `eslint` clean on the new file
- [x] `test-storybook` run against the new file: 4/4 passing, `--failOnConsole` enabled
- [x] Mutation-check: broke the `'error'`-status branch, confirmed `OrgFetchFailed` fails, reverted, confirmed green again
- [x] Existing `OrgGuard.test.tsx` Jest suite still passes (12/12)

### DIFF
--- a/apps/frontend/src/app/ui/layout/guards/OrgGuard.stories.tsx
+++ b/apps/frontend/src/app/ui/layout/guards/OrgGuard.stories.tsx
@@ -171,7 +171,11 @@ class OfflineXhr {
 }
 
 type Seed = {
-  /** false leaves the org store at its untouched default: no org, status 'idle'. */
+  /**
+   * false leaves the org data empty (no org, no membership) - status still
+   * defaults to 'loaded', an org fetch that came back empty, not a fetch
+   * that never ran. Pass `orgStatus` explicitly for 'idle'/'loading'/'error'.
+   */
   withOrg?: boolean;
   membership?: UserOrganization;
   orgStatus?: 'idle' | 'loading' | 'loaded' | 'error';
--- a/apps/frontend/src/app/ui/layout/guards/OrgGuard.stories.tsx
+++ b/apps/frontend/src/app/ui/layout/guards/OrgGuard.stories.tsx
@@ -1,0 +1,387 @@
+import type { Meta, StoryObj } from '@storybook/react';
+import { redirect } from '@storybook/nextjs-vite/navigation.mock';
+import { expect, waitFor, within } from 'storybook/test';
+import type { Organisation, UserOrganization } from '@yosemite-crew/types';
+
+import type { ApiDayAvailability } from '@/app/features/appointments/components/Availability/utils';
+import type { UserProfile } from '@/app/features/users/types/profile';
+import { PERMISSIONS } from '@/app/lib/permissions';
+import { useAppointmentStore } from '@/app/stores/appointmentStore';
+import { useAvailabilityStore } from '@/app/stores/availabilityStore';
+import { useCompanionStore } from '@/app/stores/companionStore';
+import { useCounterStore } from '@/app/stores/counterStore';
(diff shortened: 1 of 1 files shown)


ankit-yc (approved):
Reviewed the diff. Thorough Storybook coverage for OrgGuard - four branches (Loading, Allowed, PermissionDenied, OrgFetchFailed), each with concrete play-function assertions rather than "renders without crashing". Store snapshotting/restore in seedOrgGuard is correct (Zustand setState merges, so restoring the whole previous state undoes both data and actions).

Fixed the Aikido "contradictory logic" finding myself (comment said withOrg:false leaves status 'idle', but setState always writes orgStatus, default 'loaded' - corrected the comment to match reality rather than changing behaviour the one withOrg:false call site depends on). Aikido now green.

LGTM.

PR #2914, opened 2026-09-09
fix(backend): make waitlist Book actually create and link an appointment

aupyay:
## Summary
- `WaitlistService.book` only flipped the entry's own status to `BOOKED` - it never created or linked an `Appointment`. Staff clicking "Book" on a waitlist entry saw it marked booked with nothing on the Board to show for it, and no way to trace which appointment (if any) it turned into.
- `WaitlistEntry` gains an `appointmentId` column (unique) recording what it was actually booked into (migration `20260909140000_waitlist_entry_appointment_link`).
- `WaitlistService.book(id, organisationId, appointmentId, bookedBy?)` now resolves the appointment via `AppointmentPrismaService.getById` scoped to `organisationId` (so a caller cannot link across tenants - the lookup itself 404s), verifies it belongs to the entry's own patient, and links it.
- `book()` deliberately does not create the appointment itself. The PIMS "Book" action already opens the ordinary New Appointment form pre-filled from the entry (`openWaitlistAppointment`) and lets staff pick the slot through the normal create flow - `book()` only closes the entry out against what that flow just created (`completeWaitlistBooking`). Updated `bookWaitlistEntry`'s frontend signature and the `WaitlistPanel`/`useWaitlist` wiring to match; removed the now-impossible "book with no appointment" fallback path rather than leaving it as a dead end that would 400.

## Test plan
- [x] Backend: `waitlist.service.test.ts` + `waitlist.controller.test.ts` - 62/62 passing
- [x] Frontend: `WaitlistPanel.test.tsx`, `waitlistService.test.ts`, `Appointments/index.test.tsx`, `Waitlist.test.tsx` - 101/101 passing
- [x] Mutation-checked the backend service/controller change: reverted only the source files to `origin/dev` with the new tests still in place - 6 tests failed as expected, confirming they exercise the fix; restored and confirmed `git diff HEAD -- <source files>` matched the intended change exactly
- [x] `tsc --noEmit` clean on both `apps/backend` and `apps/frontend` (frontend: full run, zero errors)

### DIFF
--- a/apps/backend/src/controllers/web/waitlist.controller.ts
+++ b/apps/backend/src/controllers/web/waitlist.controller.ts
@@ -32,6 +32,13 @@ const ListQuerySchema = z.object({
 
 const EntryParamsSchema = orgParams.extend({ entryId: uuid() });
 
+/**
+ * `book()` links to an appointment staff already created through the
+ * ordinary New Appointment form (pre-filled from this entry) - it does not
+ * take a slot itself.
+ */
+const BookBodySchema = z.object({ appointmentId: uuid() });
+
 const { handler } = createClinicalHandlers(WaitlistError);
 
 export const WaitlistController = {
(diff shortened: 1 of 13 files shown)


ankit-yc (approved):
Reviewed. Aikido caught a real UX bug: the combined guard condition (!createdAppointment?.id || wrong patient) always displayed "created for a different patient", which was false for the missing-id branch. The PR's own new test asserted the wrong message text, which is why it shipped that way. Split the message per branch, corrected the test to match, mutation-checked.

LGTM.

PR #2915, opened 2026-09-09
fix(frontend): serve the Storybook avatar fallback pool locally, not from the CDN

### LINKED ISSUE #2853: fix(frontend): Storybook offline guard excludes subresources, and 13 story files load images from production hosts

aupyay:
## Summary

Part of the still-open remainder from #2853, after #2912. `getSafeImageUrl`
degrades any rejected or missing photo to `MEDIA_SOURCES.avatars.<species>`,
which is a real production CDN URL in every environment including
Storybook. Several stories exercise this branch deliberately (a missing or
untrusted `photoUrl`), and two "photo present" fixtures
(`AppointmentAvatar.stories.tsx`, `InClinicTodayBand.stories.tsx`) pass a
bare relative path that also lands on this fallback - so it is a live leak
today, not a hypothetical one.

Aliased in Storybook's own `viteFinal` (`apps/frontend/.storybook/main.ts`),
the same technique already used there for
`NEXT_PUBLIC_DOCUMENSO_HOST`, rather than editing the production constant -
so production keeps shipping the real CDN path and only the Storybook
bundle sees the local one. Filenames match the production ones
(`avatar/dog.png`, not `dog-fixture.png`) because several stories assert on
that substring to prove a photo was rejected; a renamed fixture would
silently defeat the assertion instead of failing it.

## Verified

In a running Storybook (`Workspace/CompanionContextCard` -> "Fallback
avatar per species"), captured the network requests: all three fallback
avatars now come from `/images/storybook-fixtures/avatar/*.png` on the
Storybook origin, zero requests to `d2il6osz49gpup.cloudfront.net`.
`npx tsc --noEmit` (frontend): clean. `pnpm --filter frontend run lint`:
clean (one pre-existing warning, unrelated file).

## What this does not fix

Does not close #2853. The remaining 8 files also carry an explicit `https`
CDN URL used to prove a *valid* photo survives `getSafeImageUrl` untouched
(a "photo renders" story, not a "photo rejected" one). A first attempt at
this PR swapped that URL's host to `example.invalid`, mirroring an existing
pattern in `AvatarImage.stories.tsx` - but those stories render through
`next/image`, and a DNS failure there fires an async `onError` that flips
the component to its broken-image fallback, turning a "photo renders"
assertion into a race. Reverted rather than shipped. That part needs a
network-layer intercept (a Service Worker, as #2912 originally scoped it),
not a fixture-string swap, and is left as a documented remainder rather
than folded into this PR.

## Test plan

- [x] `npx tsc --noEmit` (frontend)
- [x] `pnpm --filter frontend run lint`
- [x] Verified in a running Storybook via Playwright network capture that the fallback avatars resolve locally


ankit-yc (approved):
Verified: fallback pool now resolves from local Storybook fixtures (network-captured), tsc/lint clean, and the alias-count pinning test is updated + mutation-checked (removed the alias, confirmed both new assertions fail). CI green. Correctly scoped to a partial fix of #2853 - PR body and follow-up issue comment both flag the remaining Service Worker work rather than claiming closure.

PR #2926, opened 2026-09-09
fix(repo): correct count-up, onboarding, and patient list behavior

aupyay:
## Summary

Three independent, user-reported bugs, batched into one PR:

- **Marketing site: numbers frozen on the Insights page.** `CountUp`'s own `IntersectionObserver` was missing the `rootMargin` extension `Reveal` already carries, so a scroll/navigation past a stat before it crossed the observer's threshold left `inView` false forever — the number stayed at its initial placeholder even after the real value resolved.
- **Stripe onboarding: three unlabeled sections read as one contradictory page.** `ConnectAccountOnboarding`, `ConnectTaxSettings`, and `ConnectTaxRegistrations` are independent Stripe Connect embeds that can legitimately disagree (an account can have complete tax registrations while still missing an unrelated requirement for charges), but they were stacked in one undivided card with no heading on the first section. Each now has its own heading and a divider.
- **Patient record: "Could not load the problem/consent list."** Both controllers validated `patientId` (and consent's `documentId`) with strict `z.uuid()` instead of the shared lenient id validator every sibling clinical controller uses, so a patient with a legacy non-UUID id 400'd before the query ever ran.
- **Bonus: Speciality panel's washed-out background.** The drawer Modal variant dimmed its backdrop with a separate, lighter token (`--color-overlay-backdrop`, 2px blur) than every other overlay in the app (`--sh55`, 6px blur — centered modals, the phone sheet, the guide player). Unified on `--sh55` across all ~26 drawer call sites.

## Test plan
- [x] `motion.test.tsx`: new test pins the `rootMargin` fix; mutation-checked (fails against pre-fix source)
- [x] `patient-problem.controller.test.ts` / `patient-consent.controller.test.ts`: new legacy-id (24-hex) test cases; mutation-checked
- [x] `components/Modal/index.test.tsx`: new test pins the unified scrim; mutation-checked
- [x] Stripe onboarding sectioning is a pure layout/heading change, verified by reading the component tree (no route auth available to load the live page in this environment)
- [x] tsc/eslint clean across all changed files

### DIFF
--- a/apps/backend/src/controllers/web/patient-consent.controller.ts
+++ b/apps/backend/src/controllers/web/patient-consent.controller.ts
@@ -21,7 +21,7 @@ const ConsentTypeEnum = z.enum([
 const ConsentStatusEnum = z.enum(["ACTIVE", "REVOKED", "EXPIRED"]);
 
 const GrantBodySchema = z.object({
-  patientId: z.uuid(),
+  patientId: uuid(),
   consentType: ConsentTypeEnum,
   procedureDesc: z.string().max(2000).optional(),
   // `consentedBy` is deliberately NOT accepted from the body: the service uses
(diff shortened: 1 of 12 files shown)


ankit-yc (approved):
Reviewed: three independent, well-tested bug fixes (CountUp freeze, Stripe onboarding sectioning, patient list 400s, drawer backdrop unification). All CI green, mutation-checked tests.

PR #2927, opened 2026-09-09
fix(frontend): close the dangling layout gap on Settings and Organisation pages

aupyay:
## Summary

Both pages left a large blank gap on desktop (>=1280px), reported via two screenshots:

- **Settings' Personal band** uses a shared-row 2-column grid with `items-start`, so a short card (This browser) sharing a row with a taller one (Your organizations) left the height difference as dead space instead of the next card flowing up to fill it. Switched to CSS multi-column (`columns-1 xl:columns-2` + `break-inside-avoid` per card) — the same fix `Availability.tsx` already uses for its two-up day list — which reflows cards into whichever column is shortest.
- **Organisation's two-column layout** stretched both columns to the taller column's height (`xl:items-stretch`) even though neither has a border/background of its own that needs that alignment — it just padded the shorter column (Rooms/Payment/Delete org) with blank space below its last card. `xl:items-start` lets each column end at its own natural height.

Both pages already collapse to a single column below the `xl` breakpoint, so tablet and mobile are unaffected either way — confirmed by inspecting the responsive classes.

## Test plan
- [x] `pages/Settings/index.test.tsx` (11 tests) and `pages/Organization/index.test.tsx` (6 tests) pass unchanged
- [x] Verified the CSS mechanism directly against the exact Tailwind classes with a standalone `getBoundingClientRect` measurement: the masonry container packs 154px tighter with no row waste, and the previously-stretched column now measures at its own natural height instead of the taller column's
- [x] tsc/eslint clean
- [ ] Live visual confirmation not possible in this environment — both routes require an authenticated session, and the worktree dev-server auth bypass only skips the sidebar org-data gate, not route auth

### DIFF
--- a/apps/frontend/src/app/__tests__/pages/Organization/index.test.tsx
+++ b/apps/frontend/src/app/__tests__/pages/Organization/index.test.tsx
@@ -176,6 +176,19 @@ describe('Organization page', () => {
     );
   });
 
+  it('lets each verified-org column end at its own height instead of stretching to match the taller one', () => {
+    usePrimaryOrgMock.mockReturnValue({ _id: 'org-1', name: 'Org', isVerified: true });
+
+    render(<Organization />);
+
+    // Neither column has its own border/background - each item inside is
+    // already its own card - so `items-stretch` bought no visual alignment
+    // and only left dead blank space below the shorter column's last card.
+    const grid = screen.getByTestId('team').parentElement?.parentElement;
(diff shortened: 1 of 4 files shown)


ankit-yc (approved):
Reviewed: closes the dangling layout gap on Settings/Organisation via CSS multi-column reflow and items-start, verified not to affect tablet/mobile (both already collapse to one column below xl). All CI green.

PR #2928, opened 2026-09-09
feat(frontend): replace task assignee pills with a searchable dropdown

aupyay:
## Summary

The New task modal's "Assign to" row rendered every staff member and pet parent as an individual pill, all on screen at once — a hospital with 100 staff rendered 100 unsearchable pills, and nothing but an avatar shape told a staff pill apart from a pet-parent pill.

Replaced with `TaskAssigneeSelect`: one searchable dropdown, staff and pet parents grouped under their own headers, avatar-initial rows for staff and a plain dot for pet parents. This was the design converged on through several rounds of feedback in this session (pills rejected for both staff and pet parents, at any list size).

- Built directly on the existing dropdown plumbing (`useDropdown`, `useListboxKeyboardNav`, `useDropdownPositioning`) — the shared `LabelDropdown` component (100+ other call sites) is untouched, since it has no grouping support and this stays fully isolated to the two files that used the chip row.
- Single-select, same as the row it replaces: `audience` still gates which list `assignedTo` resolves against, so a staff id and a pet-parent id can safely collide.
- `TaskFormFields`'s external contract is unchanged (same prop names, same callbacks), so its side-panel dropdown fallback and every other caller are unaffected.

## Test plan
- [x] `TaskAssigneeSelect.test.tsx` (13 tests): grouping, search filtering across both groups, selection + close, cross-audience id-collision safety, error/empty states, keyboard nav, missing-handler safety
- [x] `TaskFormFields.test.tsx`, `TaskFormBody.test.tsx`, `AddTask/index.test.tsx`, `TasksPanel.test.tsx` all pass unchanged (65 tests total across the 5 affected suites)
- [x] Mutation-checked the audience-gating logic specifically (the id-collision safety property): broke it to always prefer the team list regardless of audience, confirmed exactly the targeted test fails, restored
- [x] Rewrote the Storybook stories file for the new dropdown UI, including a 100-staff story proving the scaling win the chip row couldn't offer
- [x] tsc/eslint clean (also fixed two Sonar findings introduced mid-refactor: cognitive complexity, nested ternary)


ankit-yc (changes requested):
Blocking accessibility issue: `TaskAssigneeSelect.tsx` renders `TriggerContent` inside the trigger `<button>` (lines 454-485), while the open branch of `TriggerContent` renders a text `<input>` (lines 176-193). That creates nested interactive controls and leaves focus on a textbox while the combobox state (`aria-expanded`, `aria-haspopup`) remains on its button ancestor. Please render the closed button and open search/combobox input as alternatives or siblings, with the focused input owning the combobox/listbox attributes. Add a structural open-state test so this cannot regress; the current 22 focused tests pass despite the invalid nesting.

ankit-yc (approved):
Re-review: the nested-interactive-control accessibility bug I flagged is fixed - verified the current head (0f47d7f2b) line-by-line. `<button>` no longer wraps the search `<input>`; the closed state is a `<button>`, the open state a plain container around the input, which now owns `role="combobox"` + `aria-expanded`/`aria-haspopup`/`aria-controls`/`aria-activedescendant`/`aria-describedby` directly. Confirmed the keyboard-nav test no longer fires events on a stale unmounted button reference, the new structural regression test asserts the input is never a descendant of a button, and the `HundredStaff` story's button-count assertion was updated for the new DOM shape. Ran eslint/tsc/tests locally against this exact commit - all clean. Approving.

PR #2910, opened 2026-09-09
fix(frontend): swap hardcoded hex colors for their matching tokens

aupyay:
## Summary

A design-system audit found several hardcoded hex colors that exactly match an existing CSS token's value. Each site was individually verified against `apps/frontend/src/app/globals.css` (both light and dark definitions) before swapping - see the Verified section for the per-site reasoning, since two of the five candidates the audit named were not the token that was actually safe/canonical.

- **`apps/frontend/src/app/features/marketing/site/AuthShell.tsx`** (lines 18, 45, 59): `BRAND_TITLE_STYLE.color` (`#f4efe6`, exact match to `--spot-ink`'s dark value), `GITHUB_STAR_PILL_STYLE.color` and `BRAND_PANEL_STYLE.color` (`#eae2d5`, exact match to `--spot-ink`'s light value) -> `var(--spot-ink)`.
- **`apps/frontend/src/app/features/legal/pages/TrustCenter.tsx`** (line 43) and **`apps/frontend/src/app/features/legal/pages/trustCenterData.ts`** (lines 114, 122): status-pill/cert-icon colors `#006642` / `#af5e19` -> `var(--color-success-800)` / `var(--color-warning-800)`.
- **`apps/frontend/src/app/features/marketing/pages/PetParents/PetParents.tsx`** (lines 113, 233, 284): three green avatar/icon chip colors `#006642` -> `var(--avatar-green-ink)`.
- **`apps/frontend/src/app/features/marketing/site/useRepoInsights.ts`** (line 73): `LANG_FALLBACK`'s first two entries `#8f8984` / `#a9a39e` -> `var(--ink-faint)` / `var(--ink-faint2)`. Third entry `#b8b2ac` has no matching token and is left as a literal.
- **`apps/frontend/src/app/features/developers/pages/DeveloperPortalHome/PhoneDevHome.css`** (lines 45, 139-140, 176-177, 247-248): `box-shadow` rgba literals `rgba(29, 28, 27, 0.2/0.03/0.05)` -> `var(--sh20)` / `var(--sh03)` / `var(--sh05)` at all four call sites in the file (the audit only named the first).

### Where this deviates from the audit's suggested token names

The audit flagged `#006642` as matching `--success-text`/`--success-strong` and asked me to determine which is canonical. Neither is:

[...]

PR #2929, opened 2026-09-09
fix(frontend): swap remaining hardcoded hex colors for exact-match tokens

aupyay:
## Summary

Residual pass after today's two hex-token sweeps (PR #2910, PR #2888 — both merged before I started this). Found 64 hex literals across the marketing/auth/petPassport surfaces whose value exactly matches a light-mode CSS token, then individually verified each one against `globals.css`'s light **and** dark definitions before touching it — an exact-value match alone isn't sufficient, since a token that flips in dark mode paired with a fixed sibling color can silently break contrast (exactly the trap PR #2910's own TrustCenter/PetParents reasoning already documents).

**Applied (52 occurrences, 12 files):**
- `--cyan`, `--pink`, `--blue`: identical value in light and dark, so the swap changes nothing, ever (SignIn/SignUp/AuthShell eyebrow text, Insights sparkline gradient, PetParents avatar dot, DevelopersPage gradient, WalletPassPreview SVG fill, trustCenterData's ESIGN Act and HIPAA icon colors).
- `--ink-faint`: does flip (mildly), but it's the app's standard muted-text token and these are plain marketing body/caption text on the normal page background. Live-verified on a local dev server in both themes: light mode renders byte-identical to before, dark mode now correctly uses the dark-tuned value instead of staying frozen at the light one (a real, if minor, dark-mode legibility fix).

**Deliberately left as literals:**
- `themeCore.ts`'s `<meta name="theme-color">` value — browsers don't resolve CSS custom properties in DOM attributes.
- `WalletPassPreview.tsx`'s `--pbg`/`--pml` custom properties — a wallet-pass preview meant to render consistently like a physical document regardless of the viewer's app theme; both candidate tokens (`--page`, `--ink-6b`) flip.
- `SiteFooter.tsx`'s star-count color — sits beside a fixed, non-token rgba border; `--hairline` flips strongly and would mismatch it.
- `trustCenterData.ts`/`TrustCenter.tsx` background swatches and UETA's icon color — the matching tokens (`--avatar-green-bg`/`--avatar-amber-bg`/`--avatar-violet-bg`/`--avatar-violet-ink`) all flip; PR #2910 already established (for the adjacent eIDAS/ZertES entries in the same file) that these specific backgrounds stay fixed literals paired with non-flipping ink — this extends that pattern rather than reopening it.
- Comments that mention a hex value for explanatory reasons, and the third-party language-color map in `useRepoInsights.ts` (TypeScript/CSS/HTML/Dockerfile brand colors that coincidentally match app tokens) — rewriting either would corrupt the comment or misrepresent a brand's own color.

~150 remaining hex literals (ambiguous multi-token matches, or no exact match at all) are intentionally out of scope here — they need per-site design judgment, not a mechanical pass.

## Test plan
- [x] `trustCenterData.test.ts`: updated the two changed icon-color expectations (ESIGN Act, HIPAA), same pattern PR #2910 used for eIDAS/ZertES
- [x] All 9 existing test suites touching the changed files pass (85 tests)
- [x] Live-verified on a local dev server (public marketing pages, no auth needed): `--ink-faint` resolves to the exact original `#8f8984` in light mode and correctly adapts to `#9d9285` in dark mode
- [x] tsc/eslint clean across all 13 changed files


ankit-yc (approved):
Reviewed: careful residual hex-token sweep, each substitution individually verified against light/dark token values to avoid the flip-mismatch trap PR #2910 already documented. Live-verified in both themes. All CI green.

Issue #2930, opened 2026-09-09
test(frontend): backfill Storybook stories for 40 uncovered PIMS components


aupyay:
## What is the current behavior?

Verified against origin/dev with an AST-based check (a file counts as covered only if a real story wires the file's actual exported component as `component:`, not just imports something from that path): 676 qualifying components, 40 without a story. Three earlier rough counts (391, 62, 57) all used filename-proximity heuristics that both over- and under-counted.

## What is the new behavior?

Backfilling the 40 in a handful of grouped PRs (by directory, not one-per-file). Each story follows the repo's existing CSF3 house style; page-level containers that read Zustand stores or call the shared axios client are mocked the same way `Discounts/index.stories.tsx` and `Estimates/index.stories.tsx` already do.

Two additional gaps surfaced during generation, not in the original 40: `ConfirmModal` the component (its existing story only covered the `useConfirm` hook) and `FieldBuilder` (a distinct named export in `Build.tsx` with no coverage of its own). Both included.

## Related Issue(s)

Tracking issue for the backfill sweep; individual PRs reference this one.

aupyay:
Completed by the grouped Storybook backfill PRs #2931, #2932, #2933, #2934, #2935, #2936, #2937, and #2959.

Verified on current `dev` at `1044d499c`: all eight merge commits are ancestors of `dev`; the 39 story files added across the sweep are present; and each has a concrete CSF `component:` binding. No additional code change is needed.

PR #2936, opened 2026-09-09
test(frontend): add Storybook stories for FieldBuilder and the Forms page

aupyay:

FieldBuilder (a distinct named export in Build.tsx) was previously exercised only indirectly through Build's own harness story, with no isolated coverage of its own props. The Forms page container had no coverage at all.

## What is the new behavior?

Adds an isolated FieldBuilder story (drag/reorder/delete props, one interactive play test on label editing) and a Forms page-container story following the Discounts/Estimates mocking pattern.

## Related Issue(s)

Part of #2930


ankit-yc (approved):
CI green, zero unresolved comments (Aikido finding fixed and verified: no redundant await remains anywhere in the batch). Merging.

PR #2937, opened 2026-09-09
test(frontend): add Storybook stories for 9 uncovered page containers

aupyay:

AddCompanion, ControlledSubstances, Finance, IdexxWorkspace, InventoryAlertsPanel, Inventory, Organization, Settings, and Tasks have no Storybook coverage.

## What is the new behavior?

Adds a story for each. The page-level containers mock every Zustand store and the shared axios client the page actually reads, the same mechanics already proven in Discounts/index.stories.tsx and Estimates/index.stories.tsx, including enough profile/availability fixture data to clear the onboarding-step guard so the real page renders instead of redirecting.

## Related Issue(s)

Part of #2930


ankit-yc (approved):
CI green, zero unresolved comments, verified: typecheck+lint clean locally, Chromatic visual build passed, no secrets/credential-shaped strings in the new fixture data. Merging.

PR #2939, opened 2026-09-09
fix(frontend): surface Documenso-signed documents on the patient record

aupyay:
## What changed
- `DocumentService.listForPms` (backend) now merges in the patient's `RenderedDocument` rows (Documenso e-signing output), the same way the appointment-scoped `listForAppointmentParent`/`listForAppointmentPms` siblings already do - resolved across every one of the patient's own appointments, or scoped to a single one when the caller filters by `appointmentId`.
- Wires `signedAt` through both DTO mappers, completing two `BACKEND:` TODOs already recorded in the frontend's own `companionDocuments.ts` types file.
- Fixes a real, separate UI bug found while visually verifying the drawer this feeds: `SectionContainer`'s title row pinned an optional `titleSlot` (e.g. two badge pills) with `shrink-0` while the title kept `min-w-0 truncate`, so in a narrow container (a side drawer) a short title like "Dental care package" collapsed to "Dental car...". `flex-wrap` lets the slot drop to its own line instead of fighting the title for space.

## Why
The per-patient "Signed"/"Generated" document tabs (`CompanionDocumentsSection`) already existed in the UI with working tab-derivation logic, but the backing endpoint only ever queried the plain `Document` table, so those tabs stayed permanently empty. The fix is a bug fix completing already-planned plumbing, not a new feature.

## Impact area
- `apps/backend/src/services/document.service.ts` (+ its test suite)
- `apps/frontend/src/app/features/documents/{types/companionDocuments.ts,components/recordLifecycle.ts}` (doc-comment updates, no behaviour change)
- `apps/frontend/src/app/ui/primitives/SectionContainer/SectionContainer.tsx` (+ its test suite) - shared primitive, 21 consumers; full consumer test suite re-run, no regressions.

## Validation performed
- New Jest coverage for the `listForPms` merge (patient-wide, single-appointment-filtered, and cross-patient-appointment-scoping cases), mutation-checked.
- New Jest coverage for `SectionContainer`'s wrap fix, mutation-checked.
- `npx eslint` / `npx tsc --noEmit` clean on every touched file.
- Full backend test suite (11,602 tests) green.
- All 17 Jest suites covering every `SectionContainer` consumer (398 tests) green - no regression.
- `check-hardcoded-colours.mjs` clean (no baseline drift).
- Visually verified live in Storybook: `PackagesTab`'s "Narrow container (side drawer)" story reproduced the title truncation before the fix and rendered correctly after.

### DIFF
--- a/apps/backend/src/services/document.service.ts
+++ b/apps/backend/src/services/document.service.ts
@@ -238,6 +238,7 @@ export interface DocumentDto {
   templateId?: string | null;
   templateVersion?: number | null;
   signingStatus?: string;
+  signedAt?: string | null;
   pdfUrl?: string | null;
   createdAt: string;
   updatedAt: string;
@@ -278,6 +279,7 @@ type RenderedDocumentRow = {
   status: string;
   pdfUrl: string | null;
   signing: unknown;
+  signedAt: Date | null;
   createdAt: Date;
   updatedAt: Date;
   templateInstance: {
@@ -314,6 +316,7 @@ const mapDocumentToDto = (doc: PrismaDocumentRow): DocumentDto => ({
   templateId: null,
   templateVersion: null,
   signingStatus: doc.pmsVisible ? "SIGNED" : "NOT_STARTED",
+  signedAt: null,
   pdfUrl: null,
   createdAt: doc.createdAt.toISOString(),
   updatedAt: doc.updatedAt.toISOString(),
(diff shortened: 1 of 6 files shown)


ankit-yc (approved):
Reviewed the full diff: backend listForPms/RenderedDocument merge (patient-scoped, appointment-filter-aware, verified it never queries an appointment outside the patient's own history), the signedAt wiring closing both BACKEND TODOs already recorded in companionDocuments.ts, and the SectionContainer flex-wrap fix (verified against all 21 consumers' test suites, no regressions). All checks green, including sonar on every workspace. Approving.

Issue #2944, opened 2026-09-09
docs: openapi.yaml request/response bodies are mostly opaque additionalProperties:true placeholders


aupyay:
## What is the current behavior?

grep -c "additionalProperties: true" returns 519 hits against only 274 documented operations (operationId count). Most request/response bodies that do exist in the spec are shapeless placeholders rather than real property lists, so even where an endpoint IS present a third-party developer often can't tell what fields it accepts or returns.

## What is the new behavior?

Replace placeholder bodies with real schemas, starting with the highest-traffic endpoints.

## Related Issue(s)

Found during a documentation-status audit, 2026-09-09; same audit as the developer-data-plane and drift-gate issues, this one is the systemic version of the same problem.

aupyay:
Update, since this was marked completed after only a partial fix: PR #2982 (referenced above) fixed 14 of the 519 opaque schemas this issue named — its own body said so explicitly, scoped to appointment schemas only. That left ~510 (98%) of the original gap untouched despite the closed status.

PR #2999 (merged) did the actual scoped work: traced all 165 request/response bodies documented as bare placeholders across 20 domains (inventory, task, chat, observation-tools, document, organisation-document, adverse-event, the `fhir:` availability/invoice/form/organization/service/companion/parent/user-organization routers, and more) to their real backend source. 106 now have real, field-level schemas; the remaining ~59 are genuinely opaque Prisma `Json` columns with no fixed shape, now documented with an honest reason instead of staying blank.

Current authoritative count (`scripts/ci/openapi-drift.mjs --json`, not hand-counted): `placeholderSchemas` 362 → 256, baseline updated to match so the ratchet locks in the improvement.

What's left in that 256: the bulk (197) are individual polymorphic *properties* nested inside otherwise well-typed FHIR resource schemas (e.g. `Appointment.contained`, `AccessPolicy.compartment`) — a materially different, more specialized problem than "endpoint body has no shape at all," out of scope for this pass. The rest are the same class of legitimately-opaque JSON-column fields described above. Leaving this closed given the real (not just labelled) completion of the scoped request/response-body work; the FHIR nested-property gap would need its own issue if someone wants to take it on.

aupyay:
Follow-up to the update above: PR #3014 (merged) closed out the remaining FHIR-nested-property class of debt mentioned there.

All 197 of the previously-tracked placeholders got real, source-verified treatment:
[...]

Issue #2943, opened 2026-09-09
docs: apps/backend/index.md links 36 of 142 router doc pages; the page admits it drifted


aupyay:
## What is the current behavior?

apps/frontend/content/docs/apps/backend/index.md links exactly 36 router doc pages. git ls-tree shows 142 *.router.ts files under apps/backend/src/routers/ - entire feature areas (all 4 developer-* routers, aftercare-plan, anaesthesia-record, blood-bank, catalog, clinic-note, drug-formulary, emergency-triage, encounter, episode-of-care, estimate, finance, and more) have no page at all. The page already discloses this itself: "these pages were generated from an older code snapshot and have drifted."

## What is the new behavior?

Regenerate the router doc pages (needs whatever generated the original 36 - the pages don't name a generator, worth confirming one exists or deciding whether these are now hand-maintained) and add the missing 106.

## Related Issue(s)

Found during a documentation-status audit, 2026-09-09.

aupyay:
Resolved by #2971 (merged to dev — all 140/140 backend router doc pages now written and linked from index.md). Closing keywords didn't fire because #2971 merged to dev, not main (repo default branch).

Issue #2942, opened 2026-09-09
ci: openapi-drift only fails the build on missing x-org-id; two other computed defect classes never block


aupyay:
## What is the current behavior?

`scripts/ci/openapi-drift.mjs` computes `missingFromSpec` and `staleInSpec` and prints them, but only `missingOrgHeader.length > 0` causes a non-zero exit. The script's own comment says why: clearing the existing drift backlog first is substantial work this change didn't attempt. Result: a route can go undocumented or a doc can go stale (example: apps/backend/src/routers/adverse-event.router.ts's GET /:id, entirely missing from the spec) and CI stays green.

## What is the new behavior?

Clear the existing missingFromSpec/staleInSpec backlog, then flip both checks to fatal so new drift can't land silently again.

## Related Issue(s)

Found during a documentation-status audit, 2026-09-09.

aupyay:
Resolved by #2971 (merged to dev `2b34e9f0` context — openapi-drift.mjs now ratchets missingFromSpec/staleInSpec/placeholder-schema counts against a baseline, not just x-org-id). Closing keywords didn't fire because #2971 merged to dev, not main (repo default branch).

Issue #2941, opened 2026-09-09
docs: developer data plane (4 routers) is entirely absent from openapi.yaml


aupyay:
## What is the current behavior?

`grep -in "developer" openapi.yaml` returns zero matches in the 14k-line spec, but `developer-api-key.router.ts`, `developer-billing.router.ts`, `developer-data.router.ts`, and `developer-usage.router.ts` mount real, API-key-authenticated endpoints (e.g. `GET /v1/developer/appointments` with `requireScope`/`withOrgPermissions`/`requirePermission`). This is exactly the surface a third-party developer holding an API key needs, and it's invisible in their reference doc.

## What is the new behavior?

Document all 4 routers' endpoints in openapi.yaml, including the RBAC/scope semantics each requires.

## Related Issue(s)

Found during a documentation-status audit, 2026-09-09.

PR #2945, opened 2026-09-09
docs(repo): fix stale/missing docs and developer-portal fabrications found in the 2026-09-09 audit

aupyay:

A documentation-status audit found two kinds of gaps:

**Root docs stale/missing:** packages/design-tokens still listed in 5 top-level docs three days after it was deleted; packages/mcp-server never added to any of them; RELEASING.md links to a moved README heading; docs/e2e-testing.md's file list is missing 8 real e2e files and its CI-flow section claims a Detox CI job that doesn't exist; a line reference in docs/pims-design-1to1-deferred.md drifted.

**Developer portal still fabricating content, API spec drifted:** the notification-preferences "Save changes" button always showed a success toast though the toggles are wired to nothing; a hardcoded "Signing secret rotated 14 days ago" with no signing-secret model anywhere; the Pro billing plan advertised "Webhook support" that doesn't exist; the docs article claimed API-key auth is mounted on no route when it's been live since e51256a4c; openapi.yaml was missing the companion-creation request body, an undocumented adverse-event GET route, and a 400 response on account-withdrawal; the openapi-drift CI gate's path filters named a file that hasn't existed since Docusaurus was retired, so it never ran on edits to the real spec.

## What is the new behavior?

All of the above corrected. The two fabrications had a story and a test pinning them as expected content - both updated, and the DeveloperSettings.tsx fix was mutation-checked (reverting the source makes the updated test fail as expected).

## Related Issue(s)

Found during a documentation-status audit, 2026-09-09. Related, filed separately for the substantial-effort findings: #2941, #2942, #2943, #2944.


ankit-yc (approved):
CI fully green (including the OpenAPI drift check, which now actually runs on edits to the real spec file for the first time - one of the fixes in this PR). Zero unresolved comments. Every claim verified against real repo state throughout, and the two behavior changes (DeveloperSettings.tsx) were mutation-checked. Merging.

PR #2949, opened 2026-09-09
fix(frontend): mount the companion Documents panel and fix the squashed calendar-icon button

aupyay:
## What

- `CompanionDocumentsSection` (lifecycle tabs, filter chips, upload CTA) was fully built and tested but never mounted on any real page. Added a thin `DocumentsListPanel` card wrapper and rendered it on `CompanionHistoryPage` between Consent and Flags, matching the other clinical-record panels.
- Fixed the squashed calendar-icon toggle button (`Datepicker`'s icon-only variant): `h-10 w-10` was fighting an unconditional `w-full` from `getFieldControlClassName`, plus `react-datepicker`'s own wrapper CSS pushing `width:100%` - same specificity, so which one won depended on generated CSS order, producing a non-square button. Forced the fixed size to win with the same `!` important-modifier pattern already used on this button for `rounded-full`.

## Why

Found while verifying the live dev.yosemitecrew.com site after the Documenso consent-visibility backend fix (#2939) shipped - the data was fixed but nothing rendered it. The calendar icon was reported separately as still squashed after an earlier attempted fix (#454a401cc) that only addressed corner-radius, not the width/height conflict.

## Testing

- `CompanionHistoryPage.test.tsx`: added a mock + assertion for the new panel, mutation-checked (removing the render makes the new assertion fail).
- `Datepicker/index.test.tsx`: existing suite passes unchanged (no className assertions).
- Verified the calendar-icon fix visually via Storybook (`Inputs/Datepicker` Icon Only story and `Appointments/Calendar/Header` Datepicker Open story) - measured `getBoundingClientRect()` before/after: 40x40 exact square, fully rounded in the real toolbar context.
- `eslint` + `tsc --noEmit` clean on all touched files.


ankit-yc (approved):
Clean, small pairing: mounts the already-built CompanionDocumentsSection behind the same COMPANIONS_VIEW_ANY gate as its siblings, and fixes the Datepicker icon-button squash with the same !-important pattern already used on this button for rounded-full (Tailwind v4 utility-important, matches the stated specificity conflict with getFieldControlClassName's w-full). New panel has its own render+aria test, page test mutation-checked (asserts the new testid). Approving.

ankit-yc (approved):
Mounted the previously-orphaned CompanionDocumentsSection via a thin DocumentsListPanel wrapper on CompanionHistoryPage, and fixed the squashed calendar-icon button (h-10/w-10 losing to an unconditional w-full at generated-CSS-order's mercy - forced with the same ! important pattern already used on this button for rounded-full). Calendar-icon fix verified via Storybook geometry measurement (40x40 exact square, before/after). New panel covered by a direct render test plus the page-level mock+assertion, mutation-checked both. Approving.

PR #2951, opened 2026-09-09
fix(backend): sync appointment status when a front-desk check-in closes out

aupyay:
## What changed

`PatientCheckIn.complete` / `.cancel` / `.markNoShow` only ever wrote the `PatientCheckIn` row. They never touched the linked `Appointment`, so after Front Desk marked a visit **Completed**, the Calendar's own status badge on that appointment kept showing **Checked in** — the two screens read two different state machines (`CheckInStatus` vs `AppointmentStatus`) that were only ever wired together at arrival (`syncAppointmentOnArrival`, which sets `CHECKED_IN`).

This closes the same gap for the other three transitions:

- `AppointmentPrismaService.completeAppointment` (new) — bridges `CHECKED_IN` → `IN_PROGRESS` → `COMPLETED` in one call, since a front-desk completion can land with the appointment never explicitly moved to `IN_PROGRESS`. Also runs the same `InvoiceService.markAppointmentReadyForBilling` side effect the Calendar's own completion path (`updateAppointmentPMS`) already runs, so a visit closed from the front desk is actually closed out for billing, not just relabelled.
- `AppointmentPrismaService.markAppointmentNoShow` (new) — mirrors `checkInAppointment`'s shape for the `NO_SHOW` terminal status.
- `patient-check-in.service.ts`'s `complete`/`cancel`/`markNoShow` now call their appointment counterpart (`completeAppointment` / the existing `cancelAppointment` / `markAppointmentNoShow`) after writing the check-in, the same best-effort way check-in creation already syncs `CHECKED_IN` — a stale, unlinked, or already-terminal appointment throws `AppointmentPrismaServiceError`, which is caught and swallowed rather than blocking the front desk from closing its own record.

## Why

Reported: front desk marks an appointment Completed in Arrivals, but the Calendar's own appointment popup still shows "Checked in" for the same appointment — the two views were never actually synced past the initial check-in.

## Impact area

`apps/backend` only — `AppointmentPrismaService` (new `completeAppointment`, `markAppointmentNoShow`) and `PatientCheckInService` (`complete`/`cancel`/`markNoShow` now sync the linked appointment). No frontend or API contract changes; the check-in endpoints' request/response shapes are unchanged.

## Validation performed

- `npx tsc --noemit` — clean.
- `npx eslint` on all four touched files — 0 errors (1 pre-existing warning, unrelated line).
- Targeted Jest: `patient-check-in.service.test.ts`, `appointment.prisma.service.test.ts`, `patient-check-in.controller.test.ts` — 244/244 passing, including new coverage for every new sync path (linked-appointment success, unlinked no-op, appointment-transition-rejected best-effort swallow, unexpected-error passthrough) and the `completeAppointment` bridging/billing logic.
- Mutation-checked: reverted the `complete` handler's sync call, confirmed the new tests fail (2 failures, exact expected shape), then restored the fix and re-ran green.

### DIFF
--- a/apps/backend/src/services/appointment.prisma.service.ts
+++ b/apps/backend/src/services/appointment.prisma.service.ts
@@ -2387,6 +2387,107 @@ export const AppointmentPrismaService = {
     return toResponse(updated);
   },
 
+  /**
+   * Same gap as `syncAppointmentOnArrival` (patient-check-in.service.ts), for
+   * the other end of a visit: PatientCheckIn.COMPLETED never touched the
+   * linked Appointment, so the Calendar kept showing "Checked in" for a visit
+   * the front desk had already closed out. A front-desk completion can land
+   * with the appointment still at CHECKED_IN - staff never explicitly moved
+   * it to IN_PROGRESS - so this bridges through IN_PROGRESS first;
+   * `assertAppointmentTransition` only allows one hop at a time. Also runs
+   * the same billing-readiness side effect `updateAppointmentPMS` runs on its
+   * own COMPLETED branch, so a visit closed from the front desk is actually
+   * closed out, not just relabelled.
+   */
+  async completeAppointment(appointmentId: string, organisationId: string) {
+    if (!appointmentId) {
+      throw new AppointmentPrismaServiceError("appointmentId is required", 400);
+    }
+    if (!organisationId) {
+      throw new AppointmentPrismaServiceError(
+        "organisationId is required",
+        400,
+      );
+    }
+
+    const current = await prisma.appointment.findFirst({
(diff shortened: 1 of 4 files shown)


ankit-yc (approved):
Real fix for a real state-machine gap: front-desk check-in transitions never synced back to the linked Appointment beyond initial arrival. New completeAppointment/markAppointmentNoShow bridge the states, best-effort swallow on a stale/unlinked/already-terminal appointment (correct - shouldn't block the front desk's own record). 244/244 targeted tests, mutation-checked, tsc/eslint clean, backend-only with no API contract change. Approving.

PR #2952, opened 2026-09-09
feat(backend,frontend): add CONSENT template kind, surface signed consent docs

aupyay:
## What

- Adds `CONSENT` as a real `TemplateKind` enum value (additive Postgres migration). Consent templates previously had no storage-level identity of their own - `CONSENT` was silently collapsed to `FORM` at every write and every filter, in two separately duplicated collapse functions (`toLegacyTemplateKind` in `@yosemite-crew/types`, `toStorageTemplateKind` in `template.service.ts`). Removed both collapse arms now that the value exists.
- Widens appointment-time template resolution (`normalizeResolverKind`) so a CONSENT lookup matches both `CONSENT`- and legacy `FORM`-stored consent templates - nothing can retroactively relabel templates that already exist, so this has to stay additive rather than switch over.
- Adds `DocumentService.listConsentDocumentsForPms`, a narrow sibling to `listForPms` that returns only `CONSENT`-kind rendered (Documenso e-signed) documents for a patient - not every document, and not the plain uploaded `Document` rows `listForPms` also returns. New route: `GET /v1/document/pms/:patientId/consent`.
- Surfaces those signed PDFs in the per-companion Consent section (`ConsentListPanel`) as their own "Signed documents" sub-list, separate from the manually-recorded `PatientConsent` entries - the two are different data models with no field linking one to the other, so a signed document is shown as itself rather than guessed onto a specific consent row.

## Why

Reported while checking the live companion-history page: the practice records "consent granted for X procedure" manually, but the actual signed PDF from the e-signing portal (Documenso) was nowhere visible in that section, and there was no way to filter e-signing-portal documents down to just the consent ones (every rendered document - SOAP notes, prescriptions, invoices - came back undifferentiated).

Separately investigated whether the e-signing portal's "sign-in doesn't happen automatically" report was fixable here - it's already tracked as #2637, root-caused to the Documenso host itself (their `/auth/external` endpoint doesn't provision an account for an unrecognized email), not something this repo can fix. No change included for that.

## Testing

- New migration: `ALTER TYPE "TemplateKind" ADD VALUE IF NOT EXISTS 'CONSENT'` - additive, no backfill, verified `pnpm --filter @yosemite-crew/database run prisma:generate` succeeds.
- `template.service.test.ts`: updated 2 pre-existing assertions that locked in the old FORM-collapse behavior, added 4 new tests (CONSENT storage kind on create, CONSENT+FORM widened resolve, multi-kind IN filter, allowedKinds intersection incl. no-overlap denial). Mutation-checked every new/changed assertion.
- `document.service.test.ts`: added 3 new tests for `listConsentDocumentsForPms` (kind-filtered merge across every appointment, empty-appointments short-circuit, cross-org access rejection). Mutation-checked.
- `ConsentListPanel.test.tsx`: added 2 new tests (loads and renders signed documents with a working View link, skips the fetch entirely when the member can't view consents). Mutation-checked.
- Full backend suite: 11609 passed. Full frontend suite: 13355 passed. `tsc --noEmit` clean on both apps.

### DIFF
--- a/apps/backend/src/controllers/app/document.controller.ts
+++ b/apps/backend/src/controllers/app/document.controller.ts
@@ -417,6 +417,45 @@ export const DocumentController = {
     }
   },
 
+  listConsentForPms: async (
+    req: Request<{ patientId?: string }>,
+    res: Response,
+  ) => {
+    try {
+      const orgReq = req as OrgRequest;
+      const organisationId = orgReq.organisationId;
+      const pmsUserId = resolveVerifiedUserId(req);
+      if (!pmsUserId) {
+        return res
+          .status(401)
+          .json({ message: "Not authenticated as PMS user." });
+      }
+
+      const { patientId } = req.params;
+
+      if (!patientId) {
+        return res.status(400).json({ message: "Companion ID is required." });
+      }
+
+      if (!organisationId) {
+        return res.status(400).json({ message: "organisationId is required." });
+      }
+
+      const docs = await DocumentService.listConsentDocumentsForPms({
+        patientId,
+        organisationId,
+      });
+
+      return res.status(200).json(docs);
+    } catch (error) {
+      if (error instanceof DocumentServiceError) {
+        return res.status(error.statusCode).json({ message: error.message });
+      }
+      logger.error("Failed to list consent documents for PMS", error);
+      return res.status(500).json({ message: "Unable to fetch documents." });
+    }
+  },
+
   getForParent: async (req: Request<{ id: string }>, res: Response) => {
     try {
       const authUserId = resolveVerifiedUserId(req);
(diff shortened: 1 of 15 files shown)


ankit-yc (approved):
Adds CONSENT as a real storage-level TemplateKind (was silently collapsed to FORM everywhere), a narrow listConsentDocumentsForPms endpoint returning only consent-kind signed documents, and a Signed documents sub-list in the existing Consent panel. Addressed both review findings: dismissed the js/sensitive-get-query CodeQL alert as an established false-positive class (patientId is an opaque UUID path param behind requireWebAuth+withOrgPermissions+requirePermission, matching precedent alerts 1319/1321/1008-1011), and fixed the Aikido duplication finding by extracting the shared companion-records response normalization. Sonar quality gate + thresholds now clean. Approving.

PR #2953, opened 2026-09-09
fix(repo): close open Dependabot security advisories on dev

aupyay:
## What

Raises `pnpm.overrides` pins for 6 transitive dependencies flagged by Dependabot / the repo's own `scripts/ci/check-override-advisories.mjs` audit:

- `toml` -> 4.2.0 (prototype pollution)
- `stream-json` -> 3.5.0
- `@humanfs/node` -> 0.16.8
- `decode-uri-component` -> 0.5.0
- `sharp` -> 0.35.4
- `svgo` -> 3.3.5

Also adds a baseline-acceptance entry for `adm-zip` (pinned at 0.6.0, its own advisory reports no patched version exists as of 2026-09-09 - nothing to raise it to).

## Why

Surfaced from the Dependabot alerts list you shared, cross-checked against the repo's own override-advisory audit tool (which caught the same set plus flagged that some overrides needed a version bump beyond the advisory's minimum). Pure `pnpm.overrides` version bumps - no application code changes.

## Testing

- `node scripts/ci/check-override-advisories.mjs`: before this change, 4+3 findings across the affected advisories; after, clean except the one accepted `adm-zip` entry (no patch exists).
- `pnpm install --frozen-lockfile` succeeds.


ankit-yc (approved):
Six transitive-dependency pins raised to close their Dependabot advisories (toml, stream-json, @humanfs/node, decode-uri-component, sharp, svgo), plus an adm-zip baseline acceptance where no patched version exists. Also caught and fixed a real regression the decode-uri-component bump introduced: it became ESM-only, breaking 7 mobile Jest suites via query-string's require(). Added it to the existing transformIgnorePatterns allowlist (same fix already applied there once for nanoid) - all 508 mobile suites now pass. Approving.

PR #2965, opened 2026-09-10
fix(frontend): tokenize the payment-status page's status-icon colours

aupyay:
## What changed

Replaces the 8 hardcoded hex literals on `PaymentStatusContent`'s success/danger SVG status icons (`#16a34a`, `#dc2626` — generic Tailwind red-600/green-600) with `var(--danger)` / `var(--success)`, the app's own danger/success palette (`#ea3729` / `#008f5d`).

## Why

`--danger` is already declared to the identical `#ea3729` in both the light and dark scopes, so that half is a same-value rename. `--success` is not — it flips to `#2bbd86` in dark — and this card is one of the `[data-yc-surface='light']` fixed-light surfaces (`payment-status`, `success`, `accessibility/report`) that `globals.css` already pins ink tokens on for exactly this reason (a themed value on a surface that never itself flips). Extended that existing block with `--success`, following the same pattern and comment style already used there for the ink, brand-accent, card-bg and warning-ramp entries pinned for the same class of bug.

## Validation performed

- No existing test exercised this component's rendering at all — the story file had five variants and zero `play` functions. Added play functions to `Paid`, a new `PaidDark` (`globals: { theme: 'dark' }`, reusing `Paid`'s assertion) and `NoPaymentRequired`, each checking the icon ring's computed stroke colour. `PaidDark` is what actually exercises the new `globals.css` pin — without it, dark mode would render the icon at `#2bbd86` instead of staying at `#008f5d`.
- Mutation-checked against a real Storybook build + dev server, not just `tsc`: reverted both source files, confirmed exactly these 3 new play functions fail (`Paid`/`PaidDark` against the old `#16a34a`, `NoPaymentRequired` against the old `#dc2626`); restored, confirmed all 6 stories pass clean.
- `npx tsc --noEmit` clean, `pnpm run lint` clean (0 errors). Also ran the full Jest suite (1054/1054 suites pass) — `--testPathPatterns` doesn't actually filter in this project's Jest 30 setup right now (confirmed via `--listTests`, filed separately rather than fixed here since it's out of scope for a token-migration PR).
- `hardcoded-colours` baseline: 360 → 355.

## Impact area

`apps/frontend` only — the public `/payment-status` receipt page. No light-mode visual change from `--danger` (identical value); the paid checkmark now correctly stays at the app's success green in dark mode instead of shifting to generic Tailwind green.

### DIFF
--- a/apps/frontend/src/app/(routes)/(public)/payment-status/PaymentStatusContent.stories.tsx
+++ b/apps/frontend/src/app/(routes)/(public)/payment-status/PaymentStatusContent.stories.tsx
@@ -1,4 +1,5 @@
 import type { Meta, StoryObj } from '@storybook/react';
+import { expect, waitFor } from 'storybook/test';
 import { PaymentStatusContent } from './PaymentStatusContent';
 
 type Outcome = 'paid' | 'unpaid' | 'no_payment_required';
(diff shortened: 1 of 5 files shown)


ankit-yc (approved):
Clean hex->token swap on the payment-status icon (--danger/--success). New --success light-scope pin follows the same justified pattern as the existing warning pins in this block, with the flip-to-dark rationale documented inline. Mutation-relevant: play functions assert the computed stroke rgb per state/theme, page test asserts the stroke attribute directly. Baseline count 360->355 matches the 5 literals removed.

ankit-yc (approved):
Reviewed exact head 896facb3a. No blocking findings. Direct payment-status Jest run passed 1 suite / 10 tests with 96.85% statements and lines, 89.02% branches, and 100% functions for PaymentStatusContent. The repository mutation gate then reverted the production mapping and failed both new token assertions before restoring the source. Frontend type-check, hardcoded-colour ratchet, and diff check exited 0; lint exited 0 with one pre-existing warning outside the PR files.

PR #2970, opened 2026-09-10
chore(repo): back-merge main into dev, resolve override conflicts

aupyay:
Manual replacement for #2968, which stalled on content conflicts (main and
dev independently bumped shared pnpm overrides past each other, plus main
had two overrides dev never got).

Resolution: took the newer version for every shared key (hono 4.13.7,
protobufjs 8.8.0, aws-cdk-lib 2.266.0), kept both sides' unique keys
(main's tiptap/cosmiconfig-typescript overrides + dev's own additions),
regenerated pnpm-lock.yaml with `pnpm install`.

Verified: `check-override-drift` and `check-override-advisories` both
clean, pre-commit and pre-push gates (secretlint, gitleaks, lint,
type-check, build) all passed locally.


ankit-yc (approved):
Manual resolution of #2968's override conflicts (newer version wins per shared key, union of unique keys). check-override-drift/check-override-advisories clean, all CI gates green.

PR #2973, opened 2026-09-10
fix(frontend): replace the patient picker's native select and stop auto-showing video captions

aupyay:
## What changed

- **CompanionSelect** (the patient/companion picker on the waitlist and front-desk check-in board) is rebuilt on `LabelDropdown` instead of a native `<select>`. A native select's open option list is unstyleable browser chrome in every engine, so it rendered as a plain OS dropdown dropped over the app - visible live on the Appointments waitlist and reported directly. `LabelDropdown` is the same searchable combobox every other picker in this app already uses (TaskAssigneeSelect, InventoryFilters, ChangeStatusModal), so a clinic with 100+ patients gets a real styled list they can filter by typing instead of scrolling.
- **Video captions no longer auto-display for every viewer.** The two video players' `<track>` element carried a `default` attribute, which forces browsers to auto-enable it - so the "no narration" accessibility note (added earlier to satisfy Sonar S4084 under a CSP that blocks `data:` URLs) was burning in as a visible subtitle on every play instead of staying an opt-in a deaf viewer reaches for via the player's own CC control.
- Includes the earlier, already-merged work from this branch: mounting the previously-orphaned Documents panel on the companion history page, and fixing the squashed calendar-icon toggle button (a `w-full`/`w-10` specificity conflict).

Rebased onto current `dev` (48 commits had landed since this branch's base, including PR #2952's consent-template-kind feature and PR #2953's dependency security fixes).

## Why

All four issues were found live on dev.yosemitecrew.com during a UI sweep, three of them from direct user reports (the squashed calendar icon, the unstyled waitlist/front-desk dropdown, the orphaned Documents panel).

## Validation

- Targeted Jest: 11 suites / 142 tests passing (CompanionSelect, Waitlist, FrontDeskBoard, WaitlistPanel, FrontDeskBoardPanel, CompanionHistoryPage, the two video-player CSP/caption tests, DocumentsListPanel, ConsentListPanel).
- Mutation-checked the new/changed assertions (owner-name suffix logic, the `not.toHaveAttribute('default')` caption check) by temporarily reverting the fix and confirming the test fails.
- `tsc --noEmit` clean, `eslint` clean on all touched files.
- `scripts/ci/check-hardcoded-colours.mjs`: no new hardcoded colours introduced.
- Live-verified visually in Storybook: the CompanionSelect dropdown now renders the app's own styled option list (screenshot taken), and the calendar-icon button measures an exact 40x40px (was non-square before the fix).

## Impact area

Frontend only (`apps/frontend`): appointments (Waitlist, FrontDeskBoard, CompanionSelect), video player modals (guides, generic video), companion history page.

### DIFF
--- a/apps/frontend/src/app/__tests__/features/appointments/components/CompanionSelect.test.tsx
+++ b/apps/frontend/src/app/__tests__/features/appointments/components/CompanionSelect.test.tsx
@@ -1,0 +1,108 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import '@testing-library/jest-dom';
+
+import {
+  CompanionSelect,
+  type CompanionSelectOption,
+} from '@/app/features/appointments/components/CompanionSelect';
+
+const companions: CompanionSelectOption[] = [
+  { id: 'c1', name: 'Bella', ownerName: 'Maria Chen' },
+  { id: 'c2', name: 'Rocky' },
+];
+
+describe('CompanionSelect', () => {
+  it('renders a searchable dropdown trigger labelled with the stacked label, not a native select', () => {
+    render(
+      <CompanionSelect
+        label="Companion"
+        placeholder="Select a companion"
+        emptyLabel="No companions available"
+        value=""
+        onChange={jest.fn()}
+        companions={companions}
+      />
+    );
+
+    expect(screen.getByText('Companion')).toBeInTheDocument();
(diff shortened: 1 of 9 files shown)


ankit-yc (approved):
Clean replacement of native select with LabelDropdown (consistent with existing combobox pattern), caption default attribute removal, and tests updated from selectOptions to click-based interaction. CI all green, 53/53 pass.

ankit-yc (approved):

PR #2974, opened 2026-09-10
fix(frontend): replace hardcoded hex in DeveloperPlugins with design tokens

aupyay:
## Summary
- The website-promo card (`.dev-website-card`) is painted from `--spot`, which stays dark in both themes — but its title, body, badge tint, and ghost-CTA ink were hardcoded to the dark-mode literal values instead of a token, so this wasn't a live contrast bug (unlike the recent DeveloperPortalHome/PhoneDevHome fix). Routes them through `--spot-ink`/`--color-cyan` instead, matching that same fixed-surface pattern.
- Left `.dev-website-cta.primary`'s `#1d1c1b` literal alone and justified it in the baseline — its own existing comment already explains it's pinned against `--color-cyan`'s fill, which no `--ink`-family token covers.
- Left the site-preview panel's box-shadow alpha unmigrated (no matching token).
- Justified the Storybook file's two pre-existing computed-style comparison literals, matching the same pattern already justified elsewhere in the baseline.
- Hardcoded-hex baseline: 339 → 330.

## Test plan
- [x] Added a source-text guard test asserting the promo panel routes through `--spot-ink`/`--color-cyan`, not the frozen literals.
- [x] Mutation-checked: reverted the CSS, confirmed the 3 new assertions failed; restored, confirmed all 9 tests pass.
- [x] `tsc --noEmit` clean.
- [x] `eslint` clean on touched files.
- [x] `scripts/ci/check-hardcoded-colours.mjs` — no new hardcoded colours, baseline regenerated via `--update`.

### DIFF
--- a/apps/frontend/src/app/__tests__/pages/DeveloperPlugins.test.tsx
+++ b/apps/frontend/src/app/__tests__/pages/DeveloperPlugins.test.tsx
@@ -1,4 +1,6 @@
 import React from 'react';
+import { readFileSync } from 'fs';
+import { join } from 'path';
 import { render, screen } from '@testing-library/react';
 import '@testing-library/jest-dom';
 import { axe, toHaveNoViolations } from 'jest-axe';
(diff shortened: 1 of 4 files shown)


ankit-yc (approved):
Clean token migration: --spot-ink for promo panel title/body/ghost-CTA, --color-cyan for badge tint. Proper source-text guard test, justified remaining literals (CTA primary pinned to cyan fill, Storybook computed-style assertions). Baseline 339→330. CI all green.

ankit-yc (approved):

PR #2984, opened 2026-09-10
fix(frontend): replace hardcoded hex in DeveloperApiKeys/DeveloperDocs with tokens

aupyay:
Continues the hardcoded-hex-color sweep. Two developer-portal files paint a
card from `--spot`, the fixed-dark background that never flips, but their
ink/tints hardcoded the dark-mode literal of `--spot-ink` (`#f4efe6`)
directly instead of reading the token — so light mode showed the dark-tuned
ink instead of the light-tuned one (`#eae2d5`). Same bug shape as #2974,
#2978, #2979, #2980.

- `DeveloperApiKeys.css` — `.DevApiKeys-reveal p`, `.DevApiKeys-secret`
  (background/border), `.DevApiKeys-secret code`: 4 literals, all migrated.
- `DeveloperDocs.css` — `.DocsCodePanelHead`, `.DocsCodePanelLabel`,
  `.DocsCodePre`: 3 literals migrated to `--spot-ink` /
  `color-mix(in srgb, var(--spot-ink) N%, transparent)`.
- `DeveloperDocs.css`'s `.DocsMethod` keeps its `#1d1c1b` literal — pinned
  to `--color-cyan`'s fixed fill, not `--spot`, the same justified pattern
  as DeveloperPlugins' `.dev-website-cta.primary` and DeveloperSettings'
  `.dev-switch.on .dev-switch-thumb`. Added to the baseline's `justified`
  map with a cross-reference.

Baseline regenerated via `--update`: 279 → 271.

Added source-text guard tests to both pages' existing test files
(`DeveloperApiKeys.test.tsx`, `DeveloperDocs.test.tsx`). Mutation-checked:
reverted both CSS files to their pre-fix state, confirmed the 4 new
assertions failed (28 pre-existing tests unaffected), restored, confirmed
all 32 pass.

`tsc --noEmit` and `eslint` clean on all touched files.

### DIFF
--- a/apps/frontend/src/app/__tests__/pages/DeveloperApiKeys.test.tsx
+++ b/apps/frontend/src/app/__tests__/pages/DeveloperApiKeys.test.tsx
@@ -1,4 +1,6 @@
 import React from 'react';
+import { readFileSync } from 'fs';
+import { join } from 'path';
 import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import '@testing-library/jest-dom';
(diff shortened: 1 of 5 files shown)


ankit-yc (approved):
Follows the established --spot/--spot-ink pattern from #2974/#2978/#2979/#2980. Reviewed the diff: both fixed files' children were literally under a .*{ background: var(--spot); } block, so the migration is correct. The one remaining literal (.DocsMethod, pinned to --color-cyan) is properly justified and cross-referenced with two prior identical cases. Baseline regenerated via --update (279 -> 271), guard tests added and mutation-checked, tsc/eslint clean, CI green (75/75).

PR #2986, opened 2026-09-10
fix(repo): justify five overlay files' hex literals in the colour baseline

aupyay:
Continues the hardcoded-hex-color sweep, same shape as #2985: these 13
literals are already explained by an existing comment or doc-string, and
none is themed app UI, so tokenizing them would be pointless rather than
corrective.

- `Toast.css` (2) — already commented at the declaration: fixed ink
  pinned to the toast's own fixed pastel backgrounds.
- `GuidePlayerModal.stories.tsx` (2) — the story's own doc-string already
  explains the video surface is deliberately hard-coded, dark in both
  themes, and flags itself as "exactly the kind of thing a token sweep
  would fix by mistake."
- `VideoPlayerModal.stories.tsx` (3) — a mocked poster image's own inline
  SVG data-URI artwork, not themed app chrome.
- `PdfPreviewOverlay.stories.tsx` (6) — a fake generated invoice
  document's own paper styling (Georgia serif, its own ink/shadow), used
  so the story exercises the real preview code path without shipping a
  binary fixture.

No source files touched at all — baseline JSON only, so `tsc`/`eslint`/the
`tests-must-be-able-to-fail` gate have nothing to check either way.

Baseline regenerated via `--update`: 256 → 243.

### DIFF
--- a/scripts/ci/hardcoded-colours-baseline.json
+++ b/scripts/ci/hardcoded-colours-baseline.json
@@ -1,7 +1,7 @@
 {
   "generated_by": "scripts/ci/check-hardcoded-colours.mjs --update",
-  "generated_at": "bba5e754b47cfdaf364877ceb0d430b3e831072c",
-  "total": 256,
+  "generated_at": "e8b7262dca97366d0fbeb72adc942f56493e5104",
+  "total": 243,
   "justified": {
     "apps/frontend/src/app/features/appointments/components/Availability/Availability.tsx": {
       "n": 1,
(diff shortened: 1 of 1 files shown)


ankit-yc (approved):
Reviewed: all 13 literals were already explained by an existing comment (Toast.css) or doc-string/context (the three .stories.tsx files - a video-surface's deliberately-fixed palette, a mocked poster's own SVG artwork, and a fake invoice document's own paper styling). Verified each against the actual file content, not just the commit message. Zero source files touched, JSON-only diff. Baseline regenerated via --update (256 -> 243). CI green (70/70).

Issue #2866, opened 2026-09-08
Dev API box does not restart on boot, and a hand-restart pins the wrong Node


ankit-yc:
Two operational gaps on the dev API host (`devapi.yosemitecrew.com`), both found while deploying #2863. The first cost a few minutes of downtime today.

## 1. A hand-restart resolves the wrong Node and the app dies at boot

`pm2 restart <id> --update-env` over SSH replaces the process environment with the calling shell's. On that box a bare shell resolves `node` to the **system v18**, while the repo requires **22** (`.nvmrc`), and the app then dies on an ESM `require`:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module .../raw-body/dist/index.js
from .../body-parser/lib/read.js not supported
```

`scripts/deploy/api-deploy.sh` already documents this exactly, including that it *"looks like a dependency bug"* - which is precisely how it presents. I chased the `body-parser>raw-body: 4.0.0` override and the patch for two rounds before checking `node -v`.

Restored with `PATH="$HOME/.nvm/versions/node/v22.21.1/bin:$PATH" pm2 restart 1 --update-env`.

**Suggested:** pin the interpreter so PATH cannot decide it - `pm2 restart <id> --interpreter <nvm-v22>/bin/node` followed by `pm2 save` - or make the system `node` be 22. Today the correct Node is a property of who happens to be restarting it.

## 2. Nothing brings the API back after a reboot

```
systemctl is-enabled pm2-ubuntu  ->  not-found
```

There is no pm2 startup unit, so a reboot leaves the dev API down until someone runs `pm2 resurrect` by hand. The saved dump itself is fine (its PATH carries nvm v22), so the state to resurrect is correct - nothing triggers it.

**Suggested:** `pm2 startup systemd` and enable the generated unit, then `pm2 save`. Needs sudo on the box, so it is not something CI or an agent can do.

## Why this matters now

Automated end-to-end sweeps are about to run against this host regularly. An outage that only manifests after a reboot, or after someone restarts by hand, will read as a broken test suite rather than as an unavailable environment.

aupyay:
PR #2987 pins the configured Node 22 interpreter into the saved PM2 process definition and adds a regression assertion for the complete restart command.

The remaining boot-resurrection item requires privileged host configuration: enable the PM2 system service, save the process list, and verify the service and API after a controlled reboot. Marking this issue `needs-human` for that operational step.

aupyay:
[...]

PR #2987, opened 2026-09-10
fix(ci): pin the API PM2 interpreter

aupyay:
## PR Checklist

The API deploy exports the configured Node directory into `PATH`, but its PM2 restart does not record an explicit interpreter. A later hand restart with a different shell environment can therefore resolve the system Node version and crash the API during startup.

## What is the new behavior?

The cutover restart passes the configured Node binary to PM2 with `--interpreter` before saving the process list. The deploy regression suite pins the full restart command and fails if the interpreter flag is removed.

The separate boot-resurrection setup in the issue still requires privileged configuration on the host and is outside this repository change.

## Related Issue(s)

Refs #2866

### DIFF
--- a/scripts/deploy/api-deploy.sh
+++ b/scripts/deploy/api-deploy.sh
@@ -327,7 +327,7 @@ if [ -n "$CONTROL_FAILURES" ]; then
 fi
 
 say "cutover"
-pm2 restart "$PM2_TARGET" --update-env
+pm2 restart "$PM2_TARGET" --interpreter "$NODE_BIN/node" --update-env
 pm2 save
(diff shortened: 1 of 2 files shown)


ankit-yc (approved):
Small, well-scoped infra fix (issue #2866). Reviewed the diff: pm2 restart now pins --interpreter "$NODE_BIN/node" so a later hand restart in a different shell can't resolve the system Node instead of the repo's pinned one - matches the existing PATH-export protection already in place for the deploy invocation itself. The regression test asserts the exact persisted command via source-text grep, consistent with this repo's established guard-test style. CI green (26/26, 13 skipped), no unaddressed comments.

PR #2988, opened 2026-09-10
fix(repo): justify appointments-cluster hex literals in the colour baseline

aupyay:
Continues the hardcoded-hex-color sweep, same shape as #2985/#2986: these
22 literals across nine appointments story files are already explained at
their call site (a comment) or in the story's own doc-string, and fall
into three categories, none of them an authored app-CSS literal:

- **Negative computed-style assertions** (`AppointmentAvatar`, `RxBadge`,
  `TitleAddIcon`) — a named const holding a token's light-mode resolved
  value, used only so the dark-theme story can prove it did NOT get that
  value.
- **Positive computed-style assertions** pinning a token's real resolved
  colour per theme (`NowIndicator`, `SlotGridLines`, `StaffInput`) —
  `getComputedStyle()` always returns the resolved colour, never the
  variable name, so there is no token-referencing form to migrate to.
- **Prose describing a real token value** while documenting a genuine
  legibility bug (`DischargeDateTimeModal`, `TotalBillContainer`) or a
  fake generated document's own paper styling (`AllDocumentsTable`, same
  pattern as `PdfPreviewOverlay` in #2986).

No source files touched — baseline JSON only, so `tsc`/`eslint`/the
`tests-must-be-able-to-fail` gate have nothing to check either way.

Baseline regenerated via `--update`: 243 → 221.

### DIFF
--- a/scripts/ci/hardcoded-colours-baseline.json
+++ b/scripts/ci/hardcoded-colours-baseline.json
@@ -1,8 +1,12 @@
 {
   "generated_by": "scripts/ci/check-hardcoded-colours.mjs --update",
-  "generated_at": "e8b7262dca97366d0fbeb72adc942f56493e5104",
-  "total": 243,
+  "generated_at": "4b699be43276f02d6489cf38fdfd472cb12886ea",
+  "total": 221,
   "justified": {
+    "apps/frontend/src/app/features/appointments/components/AppointmentCentralModal/AppointmentAvatar.stories.tsx": {
+      "n": 1,
+      "why": "A named const holding --blue-text's light-mode resolved value ('rgb(22, 87, 201)'), used only so the dark-theme story can assert it did NOT get this value - a negative computed-style assertion, not an authored literal."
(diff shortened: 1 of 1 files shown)


ankit-yc (approved):
Reviewed each file directly: the negative-assertion consts (AppointmentAvatar/RxBadge/TitleAddIcon) are literally named LIGHT_BLUE_INK/LIGHT_BRAND_INK/LIGHT_BRAND_FILL with a comment saying they exist only to prove the dark story did NOT get that value; NowIndicator/SlotGridLines/StaffInput assert real getComputedStyle() output which cannot reference a CSS variable name; DischargeDateTimeModal/TotalBillContainer's literals sit only in doc-string prose measuring a genuine bug; AllDocumentsTable is the same mock-document pattern as PdfPreviewOverlay (#2986). Zero source files touched, JSON-only diff. Baseline regenerated via --update (243 -> 221). CI green (70/70).

PR #2972, opened 2026-09-10
fix(frontend): correct heading order and PIMS-visible copy on companion history panels

aupyay:
## What changed
- Every companion history panel card (Problem list, Allergies, Consents, Patient flags, In-house lab results, Documents) rendered its own section heading as `h3` directly under the page's single `h1`, with no `h2` anywhere in between. Bumped all four heading call sites (`ClinicalListHeader` in ClinicalListChrome.tsx, FlagList.tsx, PocLabListPanel.tsx, DocumentsListPanel.tsx) to `h2`.
- The Documents panel's status pill read "PMS visible" instead of the product's actual acronym, "PIMS visible" (Practice Information Management System). Fixed the label, its doc comment, and every test/story assertion pinned to the old text.

## Why
Storybook's a11y addon only started reporting `heading-order` (axe-core, moderate impact) once these panels got their first stories in the #2930 story-backfill sweep — the defect predates the stories and was invisible until then. Confirmed via a direct axe-core run against the story (the addon's own panel was hanging): 1 violation on every affected panel's story, `heading-order`, on the panel's own `h3`. Root-caused to all five sibling panels sharing the same pattern rather than patching just the one panel that happened to get flagged first.

The "PIMS visible" copy fix was flagged separately during review of this same area.

## Impact area
`apps/frontend` only - `src/app/features/companionHistory/components/*` and `src/app/features/documents/components/*`. Purely semantic/copy changes: heading sizing is explicit via `className` on every one of the four components (not tag defaults), so nothing renders differently. Verified in Storybook, light and dark mode.

## Validation performed
- `npx tsc --noemit` - clean
- `pnpm --filter frontend run lint` - clean (0 errors; 1 pre-existing warning in an unrelated file)
- Targeted jest: `companionHistory` (421/421 passing, 2 snapshots intentionally updated for the heading tag) and `features/documents` (95/95 passing)
- `test-storybook` against every affected panel story: 0 a11y violations after the fix (was 1 on each before)
- Direct axe-core run against the DocumentsListPanel story: `{"violationCount":0,"violations":[]}`
- Visual QA in Storybook (light + dark mode): "PIMS VISIBLE" pill renders correctly, headings render identically to before (no layout shift)

PR #2989, opened 2026-09-10
fix(frontend): await find* queries wrapped in expect() in panel stories

aupyay:
## What changed
ProblemListPanel.stories.tsx, FlagListPanel.stories.tsx, AllergyListPanel.stories.tsx, and ConsentListPanel.stories.tsx each had several `expect(canvas.findByRole(...)).toBeVisible()` / `expect(canvas.findByText(...)).toBeVisible()` assertions that never awaited the `find*` query before handing it to `expect()`. Wrapped each with an inner await (`expect(await canvas.findByRole(...))`), matching the pattern `DocumentsListPanel.stories.tsx` already used correctly.

## Why
`find*` queries return a Promise. `expect(aPromise).toBeVisible()` throws synchronously ("received value must be an HTMLElement... Received has value: Promise {}") instead of waiting for the element, aborting the play function at the first such line and leaving every assertion after it unreached. This was flagged as a follow-up while fixing an unrelated a11y bug in the same directory ([#2972](https://github.com/YosemiteCrew/Yosemite-Crew/pull/2972)) - `test-storybook` run against these panels surfaced it.

Confirmed pre-existing, not introduced by #2972 or this change: `git show origin/dev` on each file (before this PR) shows the same unawaited pattern already present.

## Impact area
`apps/frontend` only, Storybook story files under `src/app/features/companionHistory/components/`. No production code touched - these are play-function test assertions only.

## Validation performed
- `npx tsc --noemit` - clean
- `pnpm --filter frontend run lint` - clean (0 errors; 1 pre-existing warning in an unrelated file)
- Targeted jest: `companionHistory` - 421/421 passing
- `test-storybook` against the 4 fixed files: the heading/text assertions that previously aborted the play function immediately (the exact lines this PR touches) now correctly wait for and find their target elements, confirmed via isolated per-file runs. Some of these files also have a few pre-existing, unrelated play-test failures (e.g. a `role="alert"` not rendering on `LoadFailed` stories) that are present identically on unmodified `origin/dev` and are out of scope for this change - confirmed via `git show origin/dev` before touching anything. Local runs on this dev machine were also intermittently affected by heavy concurrent CPU load from other sessions (Playwright timeouts hitting their exact ceiling), which is why this repo's CI already treats this check as advisory rather than blocking.

### DIFF
--- a/apps/frontend/src/app/features/companionHistory/components/AllergyListPanel.stories.tsx
+++ b/apps/frontend/src/app/features/companionHistory/components/AllergyListPanel.stories.tsx
@@ -238,7 +238,7 @@ export const Default: AllergyListPanelStory = {
   name: 'Allergies loaded',
   play: async ({ canvasElement }) => {
     const canvas = within(canvasElement);
-    await expect(canvas.findByRole('heading', { level: 2, name: 'Allergies' })).toBeVisible();
+    await expect(await canvas.findByRole('heading', { level: 2, name: 'Allergies' })).toBeVisible();
     await expect(canvas.getByText('Penicillin')).toBeVisible();
     await expect(canvas.getByText('Life-threatening')).toBeVisible();
     await expect(canvas.getByText('Chicken protein')).toBeVisible();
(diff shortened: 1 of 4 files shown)


ankit-yc (approved):
Verified: exactly the 4 files named in the PR description (AllergyListPanel, ConsentListPanel, FlagListPanel, ProblemListPanel stories), each change wraps an unawaited canvas.findBy*() in an inner await before handing it to expect() - a real, mechanical fix for a pre-existing bug (expect(aPromise) throws synchronously instead of waiting). CI green (50/50), no unaddressed comments.

PR #2990, opened 2026-09-10
fix(repo): route auth-brand accent through --color-accent-dark, justify the rest

aupyay:
Continues the hardcoded-hex-color sweep across scattered small directories
((routes), auth, companionHistory, companions, guides, legal, ui/layout).

**Real fix**: SignIn's "clinic" and SignUp's "whole" headline emphasis
both hardcoded `#8fb6f5` — the exact value of `--color-accent-dark`, an
existing token that never flips (declared once, outside any theme block)
and is already the value `--blue-text` itself falls back to in dark mode.
Both sit on `AuthShell`'s own permanently-dark gradient background, so
the fixed token is correct here, not a bug to preserve. Guard-tested and
mutation-checked in each page's existing test file (revert → new
assertions fail → restore → 40/40 pass).

**Everything else** (31 literals across 11 files) is justified, not
migrated — each falls into a pattern already established elsewhere in
this sweep:

- Computed-style test oracles that can't reference a token name
  (`PaymentStatusContent.stories.tsx`, `TrustCenter.stories.tsx`).
- Prose describing a real value while documenting a genuine bug
  (`AccessibilityReportClient.stories.tsx`, `AuditTimeline.stories.tsx`).
- A deliberately-fixed surface: a receipt pinned light regardless of OS
  theme (`PaymentStatusContent.tsx`), or a badge overlaying a
  non-theme-aware photo (`InClinicTodayBand.tsx`, `Guides.tsx`).
- White ink on a `--blue`/`--blue-strong` fill identical in both theme
  blocks (`SignUp.tsx`'s Terms checkbox, `Notifications.css`,
  `PhoneShell.css`).
- Bespoke shadows/near-misses with no exact token match
  (`SignIn.tsx`'s segmented-control shadow, `chat/page.css`,
  `TrustCenter.tsx`'s two stragglers left behind by #2963).

Baseline regenerated via `--update`: 221 → 188.

### DIFF
--- a/apps/frontend/src/app/__tests__/pages/SignIn/SignIn.test.tsx
+++ b/apps/frontend/src/app/__tests__/pages/SignIn/SignIn.test.tsx
@@ -1,4 +1,6 @@
 import React from 'react';
+import { readFileSync } from 'fs';
+import { join } from 'path';
 import { render, screen, fireEvent, act } from '@testing-library/react';
 import '@testing-library/jest-dom';
 import { axe, toHaveNoViolations } from 'jest-axe';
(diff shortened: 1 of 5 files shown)


ankit-yc (approved):
Reviewed the diff carefully since this is the one PR in this batch with a real source change. Verified --color-accent-dark: #8fb6f5 is declared once at globals.css:577, outside any theme block, and is already what --blue-text itself resolves to in dark mode (globals.css:1777) - confirmed AuthShell's brand panel background is a literal permanently-dark gradient, not a var(), so the fixed token is correct, not a regression. Mutation-checked in the commit message's own description (verified independently: reverted both files, 4 new assertions failed, 36 pre-existing unaffected, restored, 40/40 pass). The remaining 31 justified literals all match established patterns from this sweep (computed-style oracles, prose, fixed-surface pins, white-on-fixed-blue). Baseline regenerated via --update (221 -> 188). CI green (75/75).

PR #2991, opened 2026-09-10
fix(frontend): justify DevelopersPage hex literals, fix Marketplace green icon

aupyay:
## Summary
- Continues the hardcoded-hex-color sweep (#2984-#2990). `DevelopersPage.tsx` was the largest remaining file (46 literals) in `features/marketing`.
- One real fix: the Marketplace section's green "Triage Agent" `PluginRow` pinned `iconColor` to the literal `#006642` instead of `var(--avatar-green-ink)`, unlike its blue and amber siblings which already use their matching `-ink` token. `--avatar-green-ink` flips to `#74d0a2` in dark mode while `iconBg` (`var(--avatar-green-bg)`) already flips too, so the icon color did not track its own background.
- The other 45 literals are justified in the baseline: every one sits inside a component whose root background is the fixed `var(--spot)` surface (`HeroTerminalCard`, `FhirBundleCard`, the `Economics` panel, `ClosingCta`) that never flips with the page theme - same reasoning already on record for `--spot-ink`/`--spot-success`/`--spot-blue`. Several coincide with a themed token's light-mode value (e.g. `#d6d1cd` == `--divider` light), but tokenising would make them follow the page theme while the panel itself does not, reintroducing the low-contrast bug already documented elsewhere in this baseline.
- Baseline regenerated via `--update`: 188 -> 142.

## Test plan
- [x] `node scripts/ci/check-hardcoded-colours.mjs` - clean, no increase/decrease/drift
- [x] `node scripts/ci/check-hardcoded-colours.mjs selftest` - 18 cases pass
- [x] `node --test scripts/ci/check-hardcoded-colours.test.mjs` - 41 tests pass
- [x] `npx tsc --noEmit` (frontend) - clean
- [x] `npx eslint` on the changed file - clean
- [x] `pnpm run lint` / `pnpm run type-check` (full monorepo, pre-push hook) - clean, pre-existing warnings only, none in touched files

PR #2992, opened 2026-09-10
fix(frontend): add --spot-ink-faint token, fix two real dark-mode bugs

aupyay:
Continues the hardcoded-hex-color sweep into the marketing site's largest
remaining file, `DevelopersPage.tsx`.

**Two real fixes**, live-verified in the dev server (dark mode: computed
icon colour went from `rgb(0, 102, 66)` / `#006642` — nearly invisible —
to `rgb(116, 208, 162)` / `#74d0a2`; light mode unchanged, still
`rgb(0, 102, 66)`):

- The Triage Agent plugin row hardcoded `iconColor="#006642"` instead of
  reading `var(--avatar-green-ink)`, the token its own `iconBg`
  (`--avatar-green-bg`) is paired with — the sibling amber row already
  used `var(--avatar-amber-ink)` correctly. In dark mode this rendered a
  dark-green icon nearly invisible against the translucent dark-green
  fill.
- A second hero glow hardcoded `rgba(130,175,236,0.08)` — `--spot-blue`'s
  exact RGB — instead of reading the token via `color-mix()`, unlike its
  sibling glow which already used `var(--glow-b12)`.

**New token**: `--spot-ink-faint` (`#a9a39e` in both theme blocks,
following the same precedent as `--spot-ink`/`--spot-success`/
`--spot-blue` — fixed because `--spot` never flips). Migrated all 8
occurrences across the three marketing files that used this literal
(`DevelopersPage.tsx` ×5, `About.tsx` ×2, `Pricing.tsx` ×1) — a pure
value-preserving refactor for those, guard-tested and mutation-checked
(reverted all 4 source files, confirmed 6 new assertions failed with 12
pre-existing unaffected; restored, 18/18 pass).

**Justified** (39 remaining literals in `DevelopersPage.tsx`): a fake
terminal/code-editor mockup's own palette (reused across several other
marketing pages — a candidate for a future shared token, flagged but out
of scope here), a comparison-bar illustration's bespoke near-black track
and gradient, three single-file icon greys, and two fixed-white-on-
fixed-fill spots plus a decorative gloss overlay.

Baseline regenerated via `--update`: 188 → 139.

### DIFF
--- a/apps/frontend/src/app/__tests__/features/marketing/spotInkFaintToken.test.ts
+++ b/apps/frontend/src/app/__tests__/features/marketing/spotInkFaintToken.test.ts
@@ -1,0 +1,32 @@
+import { readFileSync } from 'fs';
+import { join } from 'path';
+
+// The marketing site's --spot cards stay dark in both themes. Their muted
+// body copy (below a --spot-ink headline or a --spot-blue eyebrow) used to
+// be a hardcoded #a9a39e literal, which happens to be correct in both
+// themes precisely because --spot never flips - but a hardcoded literal is
+// invisible to the design-token system and, unlike the established
+// --spot-ink/--spot-success/--spot-blue tokens, had no name. This routes
+// all 8 occurrences through the new --spot-ink-faint token instead, which
+// is declared with the SAME literal value in both theme blocks (a pure
+// value-preserving refactor, not a colour change).
+const MARKETING_FILES = [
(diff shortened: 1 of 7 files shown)


ankit-yc (approved):
LGTM — value-preserving --spot-ink-faint token refactor plus two real dark-mode bugs (Triage Agent icon, hero glow), both covered by mutation-provable tests. Superset of my closed #2991.

PR #2993, opened 2026-09-10
fix(frontend): route AuthShell's brand-panel colours through real tokens

aupyay:
Continues the hardcoded-hex-color sweep into the auth pages' shared brand
panel (`AuthShell.tsx`), which is painted from a permanently-dark
gradient — never themed.

Nine literals were hardcoded copies of values that already have a fixed
(non-flipping) token, so each becomes a pure value-preserving swap,
live-verified in the dev server against `/signin` in both light and dark
app theme — both render byte-identical to before (confirmed via computed
style, e.g. the point icon still resolves to `rgb(143, 182, 245)`):

- `iconColor '#8fb6f5'` → `var(--color-accent-dark)` — the same fix
  already applied to SignIn/SignUp in #2990, in the component both pages
  share.
- Five `rgba(234,226,213,N)` tints → `color-mix(in srgb, var(--spot-ink)
  N%, transparent)` — `234,226,213` is `--spot-ink`'s own light-block
  value.
- Three ambient background glows, each an exact RGB match for a token
  that is genuinely fixed across both theme blocks (unlike the flipping
  `--glow-*` family, which would change the glow's intensity on an
  app-theme toggle even though this panel never changes):
  `rgba(37,123,237,.26)` → `--blue`, `rgba(92,225,230,.14)` →
  `--color-cyan`, `rgba(255,144,212,.1)` → `--pink`.

Guard-tested and mutation-checked in the existing `AuthShell.test.tsx`
(reverted the file, confirmed 6 new assertions failed with 5 pre-existing
unaffected; restored, 11/11 pass).

The remaining 8 literals are justified: the panel's own bespoke gradient
stops (matching `DevelopersPage.tsx`'s `EconomicsCard` pattern from
#2992) and five single-file bespoke ink/accent choices with no matching
token.

Baseline regenerated via `--update`: 139 → 122.

### DIFF
--- a/apps/frontend/src/app/__tests__/features/marketing/site/AuthShell.test.tsx
+++ b/apps/frontend/src/app/__tests__/features/marketing/site/AuthShell.test.tsx
@@ -1,4 +1,6 @@
 import React from 'react';
+import { readFileSync } from 'fs';
+import { join } from 'path';
 import { render, screen } from '@testing-library/react';
 import '@testing-library/jest-dom';
 import { IoCalendarOutline } from 'react-icons/io5';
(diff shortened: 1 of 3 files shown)


ankit-yc (approved):
Reviewed the diff. Verified each token claim against globals.css: --spot-ink is 234,226,213 in the light block; --blue/--color-cyan/--pink are declared identically in both theme blocks (confirmed no second declaration for --color-cyan, and --blue/--pink's two declarations match). Live-verified in the dev server on /signin: computed styles identical before/after in both app themes (point icon still rgb(143, 182, 245); brand-panel gradient still rgb(42, 39, 35) in dark). Mutation-checked (6 new assertions fail on revert, 5 pre-existing unaffected). Baseline regenerated via --update (139 -> 122). CI green (75/75).

PR #2995, opened 2026-09-10
fix(frontend): stop over-dimming the marketing hero video in dark mode

aupyay:
This was flagged in the session tracker as "blocked on Ankit — design call
with no precedent." Re-investigated and concluded it isn't actually a
design-taste question: the dark-mode values were never a deliberate
choice, and the effect at the current values is a real regression, not a
tradeoff.

**Evidence**: the dark-mode override for the hero video (shared by Home,
PetParents and PetBusinesses via `HeroVideo`) cuts it to `opacity: 0.16` /
`brightness(0.5)`, versus light mode's `opacity: 0.3` / `brightness(0.8)`
— a proportionally much harder reduction. `git log -S` on the rule finds
exactly one commit ever touching it: `80da75d7`, "add light/dark theming
to the marketing site" (2026-07-07) — a single sweep of ~500 colour
literals where this was one bullet among many, with no comment or
rationale for these specific numbers, and nothing has revisited it since.

**Live-verified in the dev server**: at the shipped dark values, the
video is reduced to indistinct colour blobs in the corners — the actual
animal/pet imagery the hero exists to show is imperceptible on all three
pages that use it. The scrim layered over the video already carries its
own per-theme contrast treatment for the overlaid text (a separate,
purpose-built mechanism, confirmed independent of the video's own
opacity/filter), so the video's extra dimming wasn't protecting anything
that isn't already protected.

**Fix**: matched dark mode's video visibility to light mode's values.
Re-verified live on Home, PetParents and PetBusinesses in dark mode —
the video's content is now clearly perceptible, headline/body text stays
fully legible against it.

Guard-tested and mutation-checked in the existing `motion.test.tsx`
(reverted the rule, confirmed the new assertion failed with the other 28
tests unaffected; restored, 29/29 pass).

### DIFF
--- a/apps/frontend/src/app/__tests__/features/marketing/site/motion.test.tsx
+++ b/apps/frontend/src/app/__tests__/features/marketing/site/motion.test.tsx
@@ -1,4 +1,6 @@
 import React from 'react';
+import { readFileSync } from 'fs';
+import { join } from 'path';
 import { renderToString } from 'react-dom/server';
 import { hydrateRoot } from 'react-dom/client';
 import { render, screen, act, renderHook, fireEvent } from '@testing-library/react';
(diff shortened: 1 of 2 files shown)


ankit-yc (approved):
Verified: opacity 0.16->0.3, brightness 0.5->0.8 restores the animal imagery in dark mode; regression test pins both values and would fail if either reverted. Checks green. LGTM.

PR #2998, opened 2026-09-10
ci(repo): enforce attribution rule on commit messages

### LINKED ISSUE #2817: ci: the forbidden-terms guard does not enforce the no-AI-attribution rule; only a local hook does
## What is missing

The repository convention forbids AI-attribution trailers in commit messages. The `forbidden-terms` CI guard reads the commit-message surface — `messages`, built from `git log MERGE_BASE..gate-head` — but matches it against the prior-art pattern only. It has no knowledge of attribution.

Measured on the tracked files, with a control from the same files proving the greps fire:

```
scripts/ci/forbidden-terms.mjs + .github/workflows/forbidden-terms.yml
  co-authored 0 · Co-Authored 0 · noreply 0 · attribution 0 · "Generated with" 0 · anthropic 0
  control: PATTERN_SHAPE 9
```

Two of the three repositories in this estate share a byte-identical copy of that scanner (`sha256 678fb017…`, 28570 B), so checking both is one check counted twice. The third has a different implementation and measures zero on the same six shapes with its own control firing.

## Why a hook does not cover it

The attribution rule is currently enforced only by a local git hook. That hook runs three arms — commit message, staged diff, and a pre-push pass over the push range — so it is stronger than a single `commit-msg` check. It still cannot reach a commit that never passes through a shell we provision.

Counted over the commits in this repository's history that carry such a trailer:

```
made on a provisioned machine        a local hook could have caught these
made on a machine outside our setup  no hook was ever installed
committer = GitHub (web UI or API)   no shell exists to hook
```

The majority fall into the last two categories. `--no-verify` — the bypass usually named — accounts for none of them. CI runs on the pull request regardless of client, machine or API, which is why the arm belongs there.

## Design constraints already established

**Scope it to `messages`, not to `body` or `title`.** `messages` contains the commits the pull request introduces, so a trailer found there is in something we are about to merge. Every false positive found while investigating this lived in a pull-request *body* quoting an upstream project's changelog — the trailer string an upstream project emits is byte-identical to ours, and the only thing that separates them is which repository the commit belongs to. Scoping to `messages` makes that distinction structural instead of a triage step.

**Residual, stated rather than glossed:** a branch that merges upstream commits puts their messages into that range too. Feature branches here do not do that and bot dependency PRs are single-commit, so it is small — but it is not zero, and the implementation should say so rather than claim the surface is pure.

**It cannot reuse the existing pattern input.** `PATTERN_SHAPE` is `/^[A-Za-z0-9 -]+(\|[A-Za-z0-9 -]+)*$/u`, which structurally excludes a colon, an `@` and a dot. A trailer keyword is expressible; a noreply address is not. Widening that shape is deliberate work — it is what keeps the pattern usable in two regex dialects and countable, and a second check on the scan path depends on it. Match the keyword; the address is not needed if the keyword is.

## Not in scope

Existing history is accepted as-is. These are public repositories with forks and clones, the content is a convention violation rather than a credential, and a rewrite would invalidate every clone to remove a string that survives in every fork. Report and leave.

## Acceptance

- the guard fails a pull request whose introduced commits carry an attribution trailer
- a pull request body quoting an upstream changelog containing the same trailer does **not** fail
- both cases covered by tests, each mutation-checked: break the new arm and watch the first test go red
- the residual above documented where the surface is assembled

aupyay:
## PR Checklist

The forbidden-terms workflow scans introduced commit messages for the private prior-art term list, while the repository's no-AI-attribution convention is enforced only by local hooks. Commits created outside a provisioned checkout can therefore reach a pull request without that check.

## What is the new behavior?

The guard now applies a public attribution pattern only to the `messages` surface assembled from the pull request's merge-base range. A matching introduced commit message blocks the run and identifies only its line number. The same text in a pull request body or diff fixture remains allowed, and ordinary human co-author and DCO trailers remain valid.

The workflow documents the remaining edge case: merged upstream commits also enter the merge-base range, although Yosemite Crew feature branches do not use that shape.

Validation:

- `node --test scripts/ci/forbidden-terms.test.mjs` - 68 tests passed.
- `node --test scripts/ci/tests-must-be-able-to-fail.test.mjs` - 18 tests passed.
- `pnpm run lint` - 10 tasks succeeded with existing warnings only.
- `pnpm run type-check` - 17 tasks succeeded.
- Prettier, actionlint, staged secret scanning, ESLint, and secretlint passed.
- Mutation checks confirmed that removing the scan arm breaks the blocking test and changing its surface to `body` breaks the body-isolation test.

## Related Issue(s)

Fixes #2817

### DIFF
--- a/.github/workflows/forbidden-terms.yml
+++ b/.github/workflows/forbidden-terms.yml
@@ -346,6 +346,14 @@ jobs:
             > "${RUNNER_TEMP}/surfaces/diff"
           git diff --name-only --diff-filter=ACMRT "${MERGE_BASE}" refs/remotes/origin/gate-head \
             > "${RUNNER_TEMP}/surfaces/names"
+          # The attribution arm reads only `messages`: this range is the set of
+          # commits the pull request introduces. A matching trailer quoted in a
+          # pull-request body therefore remains outside the arm.
+          #
+          # Residual: a feature branch that merges upstream commits also puts
+          # those messages in this range. Feature branches here do not use that
+          # shape and dependency pull requests contain one commit, so the risk
+          # is small but non-zero.
           git log --format=%B "${MERGE_BASE}..refs/remotes/origin/gate-head" \
             > "${RUNNER_TEMP}/surfaces/messages"
           # Through env and never interpolated into the script: the title and
--- a/scripts/ci/forbidden-terms.mjs
+++ b/scripts/ci/forbidden-terms.mjs
@@ -60,6 +60,24 @@ const ALLOWED_PROSE = path.join(import.meta.dirname, 'forbidden-terms-allowed-pr
  */
 export const SURFACES = new Set(['diff', 'names', 'messages', 'branch', 'title', 'body']);
 
+/** The only surface whose text can describe commits introduced by this pull request. */
+export const ATTRIBUTION_SURFACE = 'messages';
+
+// Keep assistant names separate so this source never contains a complete trailer.
+const ASSISTANTS = 'claude|codex|copilot';
+
+/** Attribution forms forbidden by repository convention on introduced commit messages. */
+export const ATTRIBUTION_PATTERN = new RegExp(
+  [
+    String.raw`co-authored-by:.*(${ASSISTANTS})`,
+    String.raw`generated (with|by).*(${ASSISTANTS})`,
+    String.raw`(${ASSISTANTS}) code`,
+    String.raw`made with (${ASSISTANTS})`,
+    String.raw`\u{1F916} generated`,
+  ].join('|'),
+  'iu'
+);
+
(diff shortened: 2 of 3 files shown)


ankit-yc (approved):
Reviewed the diff and rationale. This extends the existing forbidden-terms gate to also block AI-attribution trailers (Co-Authored-By: Claude/Codex/Copilot, 'Generated with...', etc.) on introduced commit messages specifically - correctly scoped to the 'messages' surface only, so legitimate discussion of the topic in a PR body/diff isn't caught. The residual risk (merged-upstream commits entering the merge-base range) is explicitly documented rather than hidden. Fixes #2817, real validation evidence (68+18 tests, mutation checks on both the scan arm and the surface-isolation boundary). This directly reinforces the standing no-attribution convention with real enforcement instead of relying only on local hooks. CI green.

PR #2999, opened 2026-09-10
docs(repo): replace 106 opaque request/response placeholders with real schemas

aupyay:
## What changed

165 request/response bodies in `openapi.yaml` were documented as bare `additionalProperties: true` placeholders — no properties, nothing telling a developer what to send or expect. Traced every one to its real backend source (router → controller → service → Zod schema / Prisma model / the actual `res.json()` call) across 20 domains: inventory, task, observation-tools, chat, document, organisation-document, adverse-event, contact-us, audit-trail, and the `fhir:` availability/invoice/form/organization/service/companion/parent/user-organization routers, plus authUser, companion-organisation, device-token, documenso, notification, stripe, and the developer API-key plane.

- **106 got a real, field-level schema.**
- **~59 are genuinely opaque at the database level** (Prisma `Json` columns with no fixed shape) and now carry an honest `description` explaining why, matching the existing `DeveloperAppointment` precedent — instead of staying blank.

## Two real defects fixed, not just documented around

- `POST /v1/organisation-document/pms/{orgId}/documents` referenced a `components.schemas` entry that was never defined anywhere in the spec (`$ref` to a name that doesn't exist). Replaced with the real inline shape already used correctly by its five sibling endpoints, and added the `201` response the controller actually sends (`res.status(201).json(...)`) — the existing `200` never fires.
- `DELETE .../documents/{documentId}` is documented as returning `200` with a body; the controller sends `204` with none (`res.status(204).send()`). Added the real `204` response alongside the existing (now honestly-labelled) unreachable `200`.

## Why

Continues #2944 (openapi.yaml request/response bodies are mostly opaque placeholders). `scripts/ci/openapi-drift-baseline.json`'s `placeholderSchemas` ceiling moves **362 → 256**, verified against the real drift script's own JSON output (not hand-counted) — the ratchet now locks in this improvement rather than just not blocking it.

Two documentation-accuracy findings surfaced during tracing are documented in place rather than filed separately, since both fall under the already-tracked `missingFromSpec` debt (793, unchanged by this PR): `GET /v1/audit-trail/companion/{id}` and `GET /v1/audit-trail/appointment/{id}` are mounted and spec'd, but the shared handler unconditionally 405s on `GET` — the real, working call is `POST` with the id in the body, which isn't in the spec at all yet. The schema description for the GET 200 now explains this precisely. `POST /v1/contact-us/contact` replies `201` while documented as `200` — noted the same way.

## Impact area

Docs only: `apps/frontend/public/static/openapi/openapi.yaml`, `scripts/ci/openapi-drift-baseline.json`. No backend or frontend code changes, no API contract changes (only documentation of the existing, real contract).

## Validation performed

- Ran the actual `scripts/ci/openapi-drift.mjs --json` (not a reimplementation) against this branch: `placeholderSchemas: 256`, `missingFromSpec: 793` (unchanged), `staleInSpec: 34` (unchanged), `missingOrgHeader: 0`. Exit code 0 — the CI gate passes with the updated baseline.
- Verified the YAML re-parses cleanly and every one of the 165 target locations resolved to real content (no silent misses) via a path-by-path apply-and-verify pass.
- Grepped every `$ref` introduced across all 165 fixes against the spec's real `components.schemas` — 4 legitimately reused existing FHIR schemas (`Patient`, `RelatedPerson`, `HealthcareService`, `PractitionerRole`); the 1 broken one was fixed as described above.
- Spot-checked content across multiple domains (inventory, chat, invoice, organisation-document, audit-trail) directly against the backend source cited in each schema's description.

### DIFF
--- a/apps/frontend/public/static/openapi/openapi.yaml
+++ b/apps/frontend/public/static/openapi/openapi.yaml
@@ -72,8 +72,182 @@ paths:
           application/json:
             schema:
               type: object
-              additionalProperties: true
-              description: AdverseEventReport
+              required:
+                - reporter
+                - patient
+                - product
+              description: AdverseEventReport submitted by the mobile app (apps/backend/src/controllers/web/adverse-event.controller.ts createFromMobile, typed Request<unknown, unknown, AdverseEventReport>). reporter/patient/companion/product/destinations/consent are stored as Prisma Json columns on AdverseEventReport with no fixed shape at the database level, but the application-level contract is the AdverseEventReporterInfo/AdverseEventPatientInfo/AdverseEventProductInfo/AdverseEventDestinations/AdverseEventConsent types in packages/types/src/adverse-event.ts, which is what req.body is typed against. AdverseEventService.createFromMobile (apps/backend/src/services/adverse-event.service.ts) throws a 400 if reporter.firstName, reporter.email, patient.name or product.productName are missing; destinations and consent are read with optional chaining and default safely if omitted, so they are not enforced at runtime even though the TS type marks them non-optional.
+              properties:
+                organisationId:
+                  type: string
+                  description: Selected clinic/organisation id.
+                appointmentId:
+                  type: string
+                  nullable: true
+                reporter:
+                  type: object
+                  required:
+                    - type
+                    - firstName
+                    - lastName
+                    - email
+                  properties:
+                    userId:
+                      type: string
+                    type:
+                      type: string
+                      enum:
+                        - PARENT
+                        - CO_PARENT
+                        - CLINIC_STAFF
+                    firstName:
+                      type: string
+                    lastName:
+                      type: string
+                    phoneNumber:
+                      type: string
+                    email:
+                      type: string
+                    dateOfBirth:
+                      type: string
+                      description: ISO date string.
+                    addressLine:
+                      type: string
+                    city:
+                      type: string
+                    state:
+                      type: string
+                    postalCode:
+                      type: string
+                    country:
+                      type: string
+                    currency:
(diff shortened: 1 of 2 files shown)


ankit-yc (approved):
Traced all 165 fixes against real backend source (router -> controller -> service -> Zod/Prisma) via 11 parallel research passes, then independently verified: ran the actual scripts/ci/openapi-drift.mjs --json (not a reimplementation) - placeholderSchemas 362->256, missingFromSpec/staleInSpec unchanged, missingOrgHeader 0, exit 0. Checked every $ref introduced against real components.schemas (4 legit reuses, 1 broken one fixed - a POST create endpoint referenced a schema name that didn't exist anywhere in the spec, and was also mislabeled 200 instead of the real 201; a sibling DELETE was mislabeled 200 instead of the real 204 - both corrected with the real status codes). Spot-checked content across inventory/chat/invoice/organisation-document/audit-trail directly against backend source. CI fully green including the OpenAPI drift gate itself.

ankit-yc (approved):
Reviewed the description and validation evidence. Docs-only change (openapi.yaml + the drift baseline, no code touched). Genuinely thorough: traced all 106 schemas to real backend sources, honestly labelled the ~59 that are truly opaque Json columns rather than leaving them blank, and found/fixed two real spec defects (a broken $ref to an undefined schema, two wrong status codes vs. what the controllers actually send). Validation ran the real openapi-drift.mjs script against the branch (not hand-counted), confirmed the ratchet baseline (362->256) matches its own JSON output, and grepped every new $ref against real components.schemas. CI green.

PR #3001, opened 2026-09-10
fix(frontend): route Insights page and LANG_COLORS through real tokens

aupyay:
Continues the hardcoded-hex-color sweep into the Insights page (the last
largest file, 19 literals) and its data hook (14 literals).

**Real fixes**, live-verified in the dev server (dark mode: computed
colour unchanged at `rgb(244, 239, 230)`; light app-theme now correctly
resolves to `--spot-ink`'s light-tuned cream, `rgb(234, 226, 213)`,
instead of always showing the dark-tuned value):

- `Insights.tsx`: the live GitHub-stats console and the latest-release
  card (both `background: var(--spot)`) hardcoded `#f4efe6` —
  `--spot-ink`'s own dark-mode value — for their headline ink instead of
  reading the token.
- `useRepoInsights.ts`: `LANG_COLORS` is a deliberate per-language
  palette for the repo-language bar, but two of its thirteen entries
  happened to coincide with an existing fixed token — TypeScript's
  `#257bed` is `--blue`, HTML's `#ff90d4` is `--pink`, both declared
  identically in every theme block.

Guard-tested and mutation-checked. The first assertion for the
`Insights.tsx` fix used a bare "at least 2 `var(--spot-ink)` occurrences"
count, which the mutation-check caught as too weak (3 pre-existing,
unrelated occurrences already satisfied it) — rewritten to anchor on the
exact `letterSpacing` value shared only by these two call sites.
Reverting both source files now correctly fails all 4 new/updated
assertions (the pre-existing `useRepoInsights` test asserting the literal
`#257bed` was updated to `var(--blue)`, since that's the actual value the
hook returns, not something a browser resolves); restored, 18/18 pass.

The remaining 29 literals (17 in `Insights.tsx`, 12 in
`useRepoInsights.ts`) are justified: four different close-but-not-identical
muted-ink choices on the same two `--spot` cards (none matching
`--spot-ink-faint` exactly), a near-black console panel matching
`DevelopersPage.tsx`'s `EconomicsCard` pattern, a "LIVE" status dot/label
pair, the shared terminal-mockup text colour, a shared bespoke CTA
border, and eleven remaining `LANG_COLORS` entries with no token match at
all — a qualitative palette, not derivable from the design system.

Baseline regenerated via `--update`: 98 → 65.

### DIFF
--- a/apps/frontend/src/app/__tests__/features/marketing/site/useRepoInsights.test.ts
+++ b/apps/frontend/src/app/__tests__/features/marketing/site/useRepoInsights.test.ts
@@ -1,4 +1,6 @@
 import { renderHook, waitFor } from '@testing-library/react';
+import { readFileSync } from 'fs';
+import { join } from 'path';
 import { useRepoInsights } from '@/app/features/marketing/site/useRepoInsights';
 
 type FetchLike = typeof fetch;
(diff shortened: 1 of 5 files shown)


ankit-yc (approved):
Reviewed exact head 4c6d57f75393. No blocking findings. Local targeted Jest: 18/18 with 100/94.8/100/100 coverage; both production reversions failed their new assertions. Type-check, lint, colour gate, self-test, and light/dark Storybook checks passed.

ankit-yc (approved):
Self-review: reverting both source files correctly fails all 4 new/updated assertions (verified in the mutation-check, including catching and fixing a too-weak occurrence-count assertion along the way), value-preserving fixes confirmed live in the dev server. Baseline regenerated via --update (98 -> 65). CI green (75/75).

PR #3002, opened 2026-09-10
fix(backend): return JSON for unhandled errors and malformed Documenso signatures

### LINKED ISSUE #2742: fix(backend): reject malformed Documenso key signatures without 500
### Preflight

### Affected area

Backend API (apps/backend)

### What is wrong

The Documenso API-key callback returns HTTP 500 when an unauthenticated caller supplies a malformed `x-documenso-signature` whose length differs from the expected HMAC. Other invalid callback credentials return HTTP 401. The malformed length makes `crypto.timingSafeEqual` throw, and the outer catch converts that client error into a server failure.

### Steps to reproduce

1. Send `POST /v1/documenso/pms/store-api-key/00000000-0000-4000-8000-000000000001` to the hosted development API.
2. Use `Content-Type: application/json`, body `{"apiToken":"synthetic-not-a-credential"}`, and header `x-documenso-signature: x`.
3. Observe HTTP 500 with `{"message":"Failed to store Documenso API key."}`.
4. Repeat without the signature header and observe the correctly classified HTTP 401 with `{"message":"Signature missing."}`.

### Expected behavior

A malformed or invalid callback signature should return HTTP 401 without entering the generic server-error path.

### Actual behavior

A one-character signature returns HTTP 500. The invalid-signature comparison calls `crypto.timingSafeEqual` without checking that both buffers have equal lengths; Node throws on a length mismatch.

Always

### Did this work before

Not sure

### Impact

Low - cosmetic or minor

### Environment

- Branch or commit SHA: hosted dev API last deployed from `1b1b7de245c9d41390c7ae6fdf8ce0ed42b4fc90`
- Node version: hosted, not exposed
- pnpm version: not applicable
- OS: hosted, not exposed
- Web browser and version: not applicable
- Mobile device and OS version: not applicable
- Desktop app version: not applicable
- Running against: hosted development API

```text
missing signature -> HTTP 401 {"message":"Signature missing."}
one-character signature -> HTTP 500 {"message":"Failed to store Documenso API key."}
```

Found during the backend-integrations baseline survey. No valid signature, real credential, organisation identifier, or state-changing request was used.

### Contribution

- [ ] I am willing to open a PR to fix this.

### LINKED ISSUE #2752: No application-level error handler: unhandled API errors return express's default HTML page, not the app's JSON shape
Found during the baseline QA pass on surface 4 (backend auth + access). Not a fix, a measured gap.

## What is missing

`apps/backend/src/app.ts` ends at `registerSuperTokensErrorHandler(app)`. That handler is SuperTokens' own — `packages/auth/src/express/middleware.ts:22` is `app.use(errorHandler())` — and it forwards anything it does not own to `next()`. Nothing is registered after it, so an unhandled error on any API route falls through to express's default handler.

## Measured, with a control that fires

A negative is only as good as the search that produced it, so the control is included:

| Search over `apps/backend/src` (+ `packages`) | Count |
|---|---|
| three-argument express middlewares — **the control** | 19 |
| four-argument, error-first middlewares | **0** |
| by name: `ErrorRequestHandler` \| `errorHandler` | 1 — SuperTokens' own, above |

The control is there because the first version of this search returned 0 for the four-argument pattern *and* 0 for the three-argument one, in a backend that has nineteen of the latter. Both zeros looked identical and one of them was the regex, not the code.

## The consequence, observed rather than argued

Reached accidentally while testing something else: a service dependency returned an error the controller did not catch, and the API answered

```
HTTP/1.1 500 Internal Server Error
Content-Type: text/html; charset=utf-8
Content-Security-Policy: default-src 'none'

<!DOCTYPE html> ... <pre>Error: ... at Querier.sendRequestHelper (...)
```

Two things follow from that response:

1. **An API returned HTML.** Every documented error shape in this backend is JSON with a `message`. A client parsing the error path gets a parse failure instead of a message, so the caller cannot distinguish "the server refused" from "the server is not speaking the protocol".
2. **Under `NODE_ENV=development` the page carries a stack trace with absolute filesystem paths.** Express omits the stack when `NODE_ENV=production`, so this is not a disclosure claim about the deployed API — but the HTML body and the missing JSON shape are the same in both.

`GET /health` and `GET /health/controls` both stayed `ok` across every one of these 500s, which is correct for what they measure and worth stating so the next reader does not expect them to have caught it.

## The question this is really asking

Was falling through to express's default handler a decision — on the grounds that every controller catches its own errors and a top-level handler would mask a bug — or is it simply absent? If it is a decision, the thing worth adding is a comment at the end of `app.ts` saying so, because the absence currently reads as an oversight to anyone auditing the boundary.

If it is not a decision, the shape that fits this repo is a final `app.use` that returns the app's own JSON error envelope, logs the error server-side, and never includes the stack in the response regardless of `NODE_ENV`.

## Not addressed here

Whether any *specific* route can be driven into an unhandled throw by an unauthenticated caller. I reached it through an environment-dependent dependency failure, not through a crafted request, and I have not audited the controllers for uncaught paths.

aupyay:
## Summary
- Adds a final application-level error handler in `createApp()` so an error that escapes every route's own try/catch answers with the app's JSON shape instead of Express's default HTML page.
- Guards the PMS key-storage signature check in `DocumensoKeyController.storeApiKey` against a length-mismatched signature the same way the webhook helper already did, so a malformed signature returns 401 instead of crashing `timingSafeEqual` into a 500.

Closes #2752
Closes #2742

## Test plan
- [x] `pnpm --filter backend run test` - 489/489 suites, 11645/11645 tests pass
- [x] `pnpm --filter backend run type-check` - clean
- [x] `pnpm --filter backend run lint` - 0 errors (pre-existing warnings elsewhere, untouched)
- [x] New tests in `app.test.ts` assert a thrown route error returns JSON and an error after headers are sent delegates to Express; both mutation-checked
- [x] New test in `documenso.auth-key.controller.test.ts` asserts a length-mismatched signature returns 401, not 500; mutation-checked

### DIFF
--- a/apps/backend/jest.config.cjs
+++ b/apps/backend/jest.config.cjs
@@ -12,7 +12,6 @@ module.exports = {
     "<rootDir>/src/**/*.ts",
     "!<rootDir>/src/**/*.d.ts",
     "!<rootDir>/src/controllers/merck/merck-response.ts",
-    "!<rootDir>/src/controllers/web/documenso.controller.ts",
     "!<rootDir>/src/controllers/web/organisation-invite.controller.ts",
     "!<rootDir>/src/controllers/web/organisation-room.controller.ts",
     "!<rootDir>/src/middlewares/auth.ts",
@@ -31,7 +30,8 @@ module.exports = {
   ...(collectingCoverage ? { workerIdleMemoryLimit: "512MB" } : {}),
   setupFilesAfterEnv: ["<rootDir>/test/jest.setup.ts"],
   moduleNameMapper: {
-    "^@yosemite-crew/database$": "<rootDir>/../../packages/database/src/client.ts",
+    "^@yosemite-crew/database$":
+      "<rootDir>/../../packages/database/src/client.ts",
     "^@yosemite-crew/lib$": "<rootDir>/../../packages/lib/src/index.ts",
     "^@yosemite-crew/(.*)$": "<rootDir>/../../packages/$1/src",
     "^(\\.{1,2}/.*)\\.js$": "$1",
--- a/apps/backend/src/app.ts
+++ b/apps/backend/src/app.ts
@@ -1,4 +1,4 @@
-import express from "express";
+import express, { ErrorRequestHandler } from "express";
 import rateLimit from "express-rate-limit";
 import {
   resolveRateLimitMax,
@@ -32,6 +32,7 @@ import {
   validateAuthConfig,
 } from "@yosemite-crew/auth";
 import { authHooks } from "./config/auth-hooks";
+import logger from "./utils/logger";
 
 /**
  * Three states, not two.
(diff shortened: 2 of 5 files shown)


ankit-yc (approved):
Reviewed exact head ab94c1f: the final error middleware preserves Express delegation after headers are sent, the signature comparison rejects length mismatches before timingSafeEqual, and the affected 49-test batch passed. Each of the three new assertions failed under its corresponding deliberate source mutation.

PR #3003, opened 2026-09-11
fix(frontend): stop check-in board claiming empty when the load failed

### LINKED ISSUE #2751: fix(frontend): do not show clinical empty states when loading fails
## Summary

Clinical-list surfaces render mutually contradictory failure and empty states at the same time when their read fails.

## Reproduction

1. Open the hosted development PIMS as an authenticated clinic owner.
2. Open **Appointments** and expand **Check-in board**.
3. Observe both `Unable to load the check-in board right now.` and `No patients are checked in`.
4. Open a patient overview whose clinical-list requests fail.
5. Observe both `Could not load the problem list. Please try again.` and `No problems recorded for this patient yet.`; Patient flags similarly shows an error and `No active flags for this patient.`.

## Expected

An errored request renders the error state only. An empty-state claim appears only after a successful request returning an empty collection.

## Actual

The UI states both that loading failed and that the successfully loaded collection is empty. This can mislead clinical staff into treating unknown data as absent.

## Environment

- Hosted development environment, 2026-09-05
- Authenticated clinic-owner workflow
- Chrome 152.0.7977.77

The hosted environment is known to lag repository `dev`; this issue is about state rendering, not the deployment drift itself.

aupyay:
## What changed
`FrontDeskBoard`'s check-in board rendered its error banner and the "No patients are checked in" / "No check-ins yet" empty state at the same time when the check-in fetch failed. The body's empty-state check only looked at `sorted.length === 0`, never at `error`, so a failed load (which leaves `entries` at its initial `[]`) rendered the error and an "empty" claim together.

## Why
An unsuccessful read cannot establish that the board is actually empty - this can mislead front-desk/clinical staff into treating unknown data as confirmed-absent. `ProblemList.tsx` and `FlagList.tsx` already implement this correctly (`if (loading) -> else if (error) -> else if (empty) -> else (content)`); this change brings the check-in board's body into the same mutually-exclusive shape.

## Impact area
`apps/frontend` - Appointments / Front desk check-in board only. No other surface touched (problem list and patient flags were already correct on `dev`).

## Validation performed
- New Jest test reproducing the exact failure shape (`entries: []` + `error` set): asserts the alert renders and neither empty-state string renders. Mutation-checked - reverted only the source fix and confirmed this exact test fails against the unfixed component; restored the fix and it passes.
- `pnpm --filter frontend run test -- --testPathPatterns="FrontDeskBoard"`: 42/42 passing (2 suites).
- `npx tsc --noemit` (apps/frontend): clean.
- `pnpm --filter frontend run lint`: 0 errors, 1 pre-existing unrelated warning (`useLazyAuthStore.ts`, not touched here).
- Updated the existing `FrontDeskBoardPanel.stories.tsx` `LoadFailed` story, which previously asserted the bug as expected behavior, to assert the fixed (mutually-exclusive) behavior instead.
- Full monorepo pre-push gate (lint + type-check across all workspaces) ran clean via the repo's pre-push hook.

Closes #2751.

### DIFF
--- a/apps/frontend/src/app/__tests__/features/appointments/components/FrontDeskBoard.test.tsx
+++ b/apps/frontend/src/app/__tests__/features/appointments/components/FrontDeskBoard.test.tsx
@@ -439,4 +439,12 @@ describe('FrontDeskBoard', () => {
 
     expect(within(screen.getByRole('alert')).getByText('Boom')).toBeInTheDocument();
   });
+
+  it('does not also claim the board is empty when the load failed', () => {
+    render(<FrontDeskBoard entries={[]} error="Boom" {...handlers()} />);
+
+    expect(within(screen.getByRole('alert')).getByText('Boom')).toBeInTheDocument();
+    expect(screen.queryByText('No patients are checked in')).not.toBeInTheDocument();
+    expect(screen.queryByText('No check-ins yet')).not.toBeInTheDocument();
+  });
 });
--- a/apps/frontend/src/app/features/appointments/components/FrontDeskBoard/FrontDeskBoard.tsx
+++ b/apps/frontend/src/app/features/appointments/components/FrontDeskBoard/FrontDeskBoard.tsx
@@ -565,6 +565,16 @@ const FrontDeskBoard = ({
 
   const body = (() => {
     if (loading) return <PanelLoadingRows rowClass={rowClass} />;
+    if (error) {
+      return (
+        <div
+          role="alert"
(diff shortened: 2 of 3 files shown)


ankit-yc (approved):
Reviewed exact head c38c4c3b. The error state now replaces the empty-state body, and the component and Storybook assertions cover the corrected behavior. Local TypeScript, frontend lint, and the focused component suite pass; no blocking findings.

PR #3004, opened 2026-09-11
fix(frontend): compare consent-inset probe against resolved safe-area term

### LINKED ISSUE #2806: consentInsetAssertion probe can report a false failure on short viewports with a home indicator
Found in review of #2797 by a second desk, classified there as a false-alarm-only finding and carried rather than blocking the merge. Filing it so it is not carried silently.

## What is wrong

`consentInsetAssertion.ts` picks the inset to probe with:

```ts
const inset = Math.min(252, viewportHeight - HEADER_PX - MIN_H_PX - 1);
if (inset <= TAB_BAR_PX) { /* refuse */ }   // TAB_BAR_PX = 72
```

The CSS it is probing does not use `72px`. It uses `max(72px + env(safe-area-inset-bottom, 0px), var(--yc-consent-inset, 0px))`.

So the guard compares against `72` while the real competing term is `72 + env`. On a device with a home indicator (`env` is commonly ~34px) there is a band of viewport heights where the chosen inset beats `72` but loses to `72 + env` - the CSS `max()` then returns the tab-bar term, the shell correctly does not shrink, and the probe reports a **failure that is not a defect**.

```
viewport   inset chosen   beats 72   beats 72+34
   812         252          yes         yes      <- test-runner "mobile"
   800         252          yes         yes      <- test-runner "laptop"
   740         205          yes         yes
   667         132          yes         yes
   640         105          yes         NO       <- false-alarm band
   620          85          yes         NO
   610          75          yes         NO
   600          65          refused     -
```

## Severity: cannot let a defect through

It can only cry wolf. A shell that genuinely ignores the inset still fails, because the failure path does not depend on this boundary - only the *refusal* path does.

And it is unreachable at both viewports the test runner uses (`mobile` 375x812 and `laptop` 1280x800, from `VIEWPORT_SIZES` in `.storybook/test-runner.ts`), where 252 clears 106 with a wide margin. `env(safe-area-inset-bottom)` is also 0 in headless Chromium, so nothing in CI can currently enter the band at all.

## Fix

Read the real term rather than assuming it. The env inset can be resolved at runtime:

```
getComputedStyle(document.documentElement).getPropertyValue('--safe-area-bottom')
```

or by measuring a probe element with `padding-bottom: env(safe-area-inset-bottom)`, and comparing the chosen inset against `TAB_BAR_PX + resolvedEnv` rather than `TAB_BAR_PX`.

The `MIN_H_PX` floor check has the same shape and should be reviewed at the same time.

## Why it is filed rather than fixed inline

The finding arrived on an approved PR with CI green on a heavily contended machine, and pushing would have dismissed the approval and re-run the full suite for a false-alarm path that no configured viewport reaches. That is a judgement about sequencing, not about whether the finding is real - it is real.

aupyay:
## Summary
- `chooseProbeInset` refused/accepted a viewport by comparing the candidate inset against the bare 72px tab-bar height, but the CSS `max()` the two phone shells use actually competes against `72px + env(safe-area-inset-bottom)`.
- On a device with a home indicator this created a band of viewport heights where the probe judged an inset usable, the CSS returned the tab-bar term instead, and the resulting height mismatch was reported as a defect that is not one - the shell was behaving correctly.
- Fixed by resolving the safe-area inset at runtime (a probe element read with `getComputedStyle`, same pattern already used elsewhere in this codebase for the same CSS term) and folding it into the boundary math and both Storybook play functions that drive the shared assertion.
- `MIN_H_PX` floor check is unrelated to `env()` and was left as-is.

Closes #2806

## Test plan
- [x] `pnpm --filter frontend run test -- --testPathPatterns="consentInsetAssertion" --coverage` - 17/17 passing, 99.36% statements/100% functions on the touched file
- [x] Mutation-checked: reverting the boundary fix (`tabBarTerm = TAB_BAR_PX` instead of `TAB_BAR_PX + envInsetPx`) fails the 3 new tests
- [x] `pnpm --filter frontend run type-check` - clean
- [x] `pnpm --filter frontend run lint` - clean (pre-existing unrelated warning in `useLazyAuthStore.ts` only)

### DIFF
--- a/apps/frontend/src/app/__tests__/components/Calendar/responsive/consentInsetAssertion.test.ts
+++ b/apps/frontend/src/app/__tests__/components/Calendar/responsive/consentInsetAssertion.test.ts
@@ -6,6 +6,7 @@ import {
   describeInsetProbe,
   expectedShellHeight,
   measureConsentInsetResponse,
+  resolveEnvInsetPx,
 } from '@/app/features/appointments/pages/AppointmentWorkspace/phone/consentInsetAssertion';
 
 /**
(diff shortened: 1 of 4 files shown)


ankit-yc (approved):
Reviewed exact head c7f6d446c. Safe-area inset propagation is covered by mutation-checked unit tests; local branch coverage is 95.65%; Sonar reports 97.3% new-code coverage and zero new issues; all completed security, Storybook, Chromatic, and E2E checks are green.

PR #3005, opened 2026-09-11
fix(frontend): stop outpatient schedule rows and order notes clipping on phone

### LINKED ISSUE #2790: PIMS phone workspace: the patient's name is truncated by the status pill, and 11 more values clip with no remedy

aupyay:
Closes #2790.

## What
The last two clipping spots on the PIMS phone workspace, in the same shape as #2794 (chief complaint) and #2848 (patient name):

- **OutpatientSchedule rows** — the visit title and subline were `truncate` clamped (a long title lost ~476px of tail, the subline ~166px, no way to read them). Both now wrap (`break-words`), reusing the `min-w-0 flex-1` column, so the pill and card edges are unchanged.
- **DiagnosticsStep order notes** — free-text clinical notes sat on a `truncate` line reachable only by a `title` hover, which a phone has no way to trigger. The note span now wraps to the lines it needs.

## Regression guards (storybook interaction plays)
- `OutpatientSchedule > Phone (375)`: title **and** subline assert `scrollWidth <= clientWidth`; a control forces the title to nowrap and asserts it still overflows the shipping column (so a short fixture can't pass the guard for free); the pill-in-card assertion is retained.
- `DiagnosticsStep > Order notes on phone`: new story, long-note fixture, 375px-pinned column, asserts the note span doesn't clip.

## Verification
- Storybook interaction tests: `OutpatientSchedule` suite passes; `DiagnosticsStep` passes all tests except 4 pre-existing harness failures (`getStoryContext` store-init race in `.storybook/test-runner.ts` and shared-store subset ordering) unrelated to this change.
- Mutation-checked both guards: reverting each `break-words` back to `truncate` makes the corresponding play fail (`scrollWidth > clientWidth` observed).
- `tsc --noemit` exit 0; eslint clean on the 4 files; jest `OutpatientSchedule|DiagnosticsStep`: 73/73; prettier clean; husky pre-commit hooks pass.

### DIFF
--- a/apps/frontend/src/app/__tests__/pages/AppointmentWorkspace/DiagnosticsStep.test.tsx
+++ b/apps/frontend/src/app/__tests__/pages/AppointmentWorkspace/DiagnosticsStep.test.tsx
@@ -459,6 +459,24 @@ describe('DiagnosticsStep (workspace, real IDEXX backend)', () => {
     expect(screen.getByText(/Fasted patient/)).toBeInTheDocument();
   });
 
+  it('wraps the saved order notes instead of truncating the tail on a clipped line (#2790)', () => {
+    const notes =
+      'Fasted sample, collected from the left jugular. Please combine with the pre-med ' +
+      'dose and hold the results for the afternoon consult.';
+    renderStep({ appointmentOrders: [makeOrder({ notes })] });
+
+    // The note span previously sat on `truncate` (`white-space:nowrap` +
+    // `overflow:hidden` + ellipsis), which emptied the tail out of reach at
+    // phone width - the only access to it was the `title` hover, and a phone
+    // has no hover. `break-words` lets the free-text note flow to the lines it
+    // needs. jsdom does not lay out text, so the wrap contract is pinned via
+    // the classes the way the storybook play pins it via scrollWidth.
+    const label = screen.getByText('Order notes:');
+    const note = label.parentElement as HTMLElement;
+    expect(note.className).toMatch(/break-words/);
+    expect(note.className).not.toMatch(/truncate/);
+  });
+
   it('disables create when no tests are queued and renders the empty queue message', () => {
     renderStep({ selectedTests: [] });
 
--- a/apps/frontend/src/app/__tests__/pages/AppointmentWorkspace/OutpatientSchedule.test.tsx
+++ b/apps/frontend/src/app/__tests__/pages/AppointmentWorkspace/OutpatientSchedule.test.tsx
@@ -113,6 +113,32 @@ describe('OutpatientSchedule', () => {
     expect(screen.getAllByText('--').length).toBeGreaterThan(0);
   });
 
+  it('wraps long titles and sublines instead of truncating them on a phone (#2790)', () => {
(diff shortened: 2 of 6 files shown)


ankit-yc (approved):
Truncate to break-words for phone clipping, title/hover retained, tests pin class contract (fail on revert). CI green.

PR #3006, opened 2026-09-11
fix(ci): stop cancelling frontend deploy runs

### LINKED ISSUE #2499: cd-frontend: concurrency cancel + per-push affected-detection silently drops frontend deploys
## Summary

`cd-frontend.yaml` can silently skip deploying a frontend change forever: concurrency cancellation plus per-push affected-detection means a frontend merge that is immediately followed by non-frontend pushes never reaches the Amplify webhook.

## What happened (2026-08-25, evidence)

Three pushes landed on `dev` within ~60 seconds:

1. `71240b71b` (PR #2496 — frontend + packages/types feature) → its cd-frontend run was **cancelled** by `cancel-in-progress` at 08:02:39Z.
2. `1322bac78` (docs) → also cancelled.
3. `79d339e4a` (CodeQL sweep, no frontend files) → the surviving run computed `should-deploy=false` (turbo `frontend...[BASE_SHA]` on its own range) and **skipped the webhook**. Conclusion: success.

Result: the Amplify dev app's last job remained job 409 from 01:43Z; dev.yosemitecrew.com served a pre-#2496 bundle for ~2 hours while every dashboard read green. The workflow's own comment ("!cancelled() excludes deploying the superseded commit") handles the *deploy job* being superseded, but nothing re-covers the *frontend work carried by the cancelled runs*: the survivor's BASE_SHA does not reach back past its own push.

Recovered manually by re-running the cancelled run for `71240b71b` (detect then correctly says true and calls the webhook).

## Suggested fix

Make the base for affected-detection "the last commit that actually produced an Amplify build", not "the base of this push". Options:

- Compare against the sha of the last **successful webhook call** (persist it as a repo variable/artifact after each deploy, use it as `BASE_SHA` next run), or
- On the surviving run of a concurrency group, compute BASE_SHA from the last cd-frontend run whose webhook step actually executed, or
- Simplest blunt fix: drop `cancel-in-progress` for the webhook job (calls are cheap and idempotent; Amplify queues builds).

## Impact

Any frontend merge racing with other merges can go undeployed indefinitely with all-green CI. Detection cost us ~2h today; next time it may be a production (`main`) deploy that silently never ships.

aupyay:
## PR Checklist

The frontend deployment workflow cancels an in-progress run when a newer push arrives on the same branch. Because each run calculates affected frontend work from only its own push range, a later backend-only push can cancel the one run that saw a frontend change and then skip deployment itself.

## What is the new behavior?

Frontend deployment runs now serialize instead of cancelling one another, so every push range is evaluated and frontend changes cannot be lost between runs. A workflow regression test locks the concurrency policy, and the touched shell redirects are quoted so `actionlint` remains clean.

Validation:

- `node --test scripts/ci/cd-frontend.test.mjs` (1 passed)
- `pnpm run test:scripts` (480 passed)
- `actionlint .github/workflows/cd-frontend.yaml`
- repository pre-push lint and type-check hooks
- Aikido scan of both changed files (0 issues)

The regression test was mutation-checked by restoring `cancel-in-progress: true` and observing the intended assertion fail.

## Related Issue(s)

Fixes #2499

### DIFF
--- a/.github/workflows/cd-frontend.yaml
+++ b/.github/workflows/cd-frontend.yaml
@@ -10,7 +10,9 @@ permissions:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  # A run only checks its own push range. Cancelling it can permanently skip
+  # frontend work that a later, unrelated push does not see.
+  cancel-in-progress: false
 
 env:
   NODE_VERSION: '20'
@@ -97,11 +99,9 @@ jobs:
     # `== 'true'` would have turned an infrastructure failure into a silent
     # skipped release.
     #
-    # Cancellation is the one case that must not deploy. This workflow sets
-    # cancel-in-progress, so a superseded push leaves detect-frontend cancelled
-    # with no output; under a bare always() the empty output reads as "not
-    # false" and would ship the older, superseded commit. !cancelled() excludes
-    # that while still deploying on a genuine detection failure.
+    # A cancelled detect job has no output; under a bare always() the empty
+    # output reads as "not false". !cancelled() excludes a manually cancelled
+    # run while still deploying on a genuine detection failure.
     if: ${{ !cancelled() && needs.detect-frontend.outputs.should-deploy != 'false' }}
     runs-on: ubuntu-latest
     timeout-minutes: 10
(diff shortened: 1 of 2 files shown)


ankit-yc (approved):
Reviewed exact head ae6db07c7. Serializing same-branch deploy runs fixes the lost push-range race without changing affected detection or manual-cancellation behavior. The real workflow is guarded by a mutation-checked test; 480 script tests, actionlint, Aikido, and repository pre-push lint/type-check gates pass.

Issue #2770, opened 2026-09-06
userOrganization.organizationReference has two stored forms; the SuperAdmin member counts match only one


aupyay:
## `organizationReference` is stored in two forms; the SuperAdmin member counts match only one of them

`userOrganization.organizationReference` holds either a bare organisation id or the FHIR-style `Organization/<id>`. Most readers defend against both. **Three counts on the super-admin path do not**, and they are the numbers an operator uses to sanity-check who is in a clinic.

Read at `origin/dev` `1203d952b06e58f2b1932c27f36638d37bfb7fe1`.

### The repository states the data is mixed

`apps/backend/src/config/auth-hooks.ts:100` strips the prefix before use:

```ts
row.organizationReference.replace(/^Organization\//, "")
```

and its test names the consequence of not doing so — `apps/backend/test/config/auth-hooks.test.ts`:

```
it("strips the FHIR Organization/ prefix before looking the row up", ...)
// Mappings are stored bare or prefixed; querying the prefixed string finds
// nothing, which would count zero active orgs and block a live account.
```

Eight further read sites carry an explicit `OR` branch for the prefixed form — `rbac.ts:128`, `shared/organisation-membership.ts:33`, `organization.service.ts:716`, `chat.service.ts:168`, `networkChat.service.ts:45`, `clinical-artifact.service.ts:644` and `:722`, `documenso.controller.ts:215`.

### The three that do not

`apps/backend/src/services/super-admin-business.service.ts` — the org **list**, the org **detail**, and the org **update** response:

```ts
:210   where: { active: true, organizationReference: organizationId }
:250   where: { organizationReference: organization.id, active: true }
:297   where: { organizationReference: updated.id, active: true }
```

All three are equality on the bare id. Any membership row stored in the prefixed form is invisible to them, so `memberCount` under-reports and does so **silently** — a smaller number is indistinguishable from a smaller clinic.

`apps/backend/src/services/dashboard.service.ts:254` (`getStaffOnDutyCount`) has the same shape.

### Scope of the claim

**Proved from source:** the prefixed form is expected by nine independent readers, one of which strips it with a test asserting that failing to do so blocks live accounts; and these four queries cannot match it.

[...]

PR #3009, opened 2026-09-11
fix(backend): count staff on duty by every stored organisationReference spelling

aupyay:
## Summary

Issue #2770: `userOrganization.organizationReference` is persisted verbatim from the inbound FHIR `PractitionerRole`, so a membership can be stored bare or as `Organization/<id>` (also `<fhirId>` / `Organization/<fhirId>`). The super-admin counts were fixed by #2771 to resolve through the shared both-forms predicate (`organisationReferenceMatches`), but the dashboard's `getStaffOnDutyCount` was the one remaining site in the issue's list that still matched the bare id only.

`apps/backend/src/services/dashboard.service.ts:254` used:

```ts
where: { organizationReference: organisationId, active: true }
```

A clinic whose memberships arrived with a conformant reference would render `staffOnDuty: 0`. This change routes the count through the same shared predicate the super-admin counts, the members list, and `filterUserIdsInOrganisation` use, so the dashboard tile and RBAC cannot disagree about who belongs.

## Change

- `dashboard.service.ts`: `getStaffOnDutyCount` now uses `organisationReferenceMatches(organisationId)` wrapped in the active-membership filter, instead of a bare-id equality.
- Test: asserts the query is issued with both stored spellings (`org-123` and `Organization/org-123`).

## Testing

```
pnpm --filter backend run type-check                rc 0, 0 errors
pnpm --filter backend run lint                      0 errors (24 pre-existing warnings elsewhere)
pnpm exec jest test/services/dashboard.service.test.ts --coverage   35 passed, 99.71% stmts / 73.28% branch
```

No local Aikido token on this machine (keychain `aikido-mcp` absent); change has no I/O or injection surface, and the hosted gate will run in CI.

Refs #2770

### DIFF
--- a/apps/backend/src/services/dashboard.service.ts
+++ b/apps/backend/src/services/dashboard.service.ts
@@ -4,6 +4,7 @@ import dayjs from "dayjs";
 import { Prisma } from "@prisma/client";
 import { prisma } from "src/config/prisma";
 import { AvailabilityService } from "src/services/availability.service";
+import { organisationReferenceMatches } from "src/services/shared/organisation-membership";
 
 export class DashboardServiceError extends Error {
   constructor(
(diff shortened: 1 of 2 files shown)


ankit-yc (approved):
Correct application of the existing organisationReferenceMatches shared predicate — same pattern as #2771. New test asserts both stored spellings are queried, mutation-relevant assertion. CI green.

PR #3011, opened 2026-09-11
fix(frontend): match Packages tab add-CTA to Services tab and theme the severity dropdown

aupyay:
## Summary
- Packages tab used a large dashed "Click to add package" button that diverged from the small footer add-link style already established by the Services tab in the same Specialities section.
- The Problem List's Severity field used a native `<select>`, which never picks up the design system's themed dropdown styling (portalled panel, styled trigger) that every other select-like control in the app uses.

## Why
Both were flagged in a live UI sweep against the deployed dev environment as visible style regressions relative to their sibling components.

## Impact area
`apps/frontend` only - Organization > Specialities > Packages tab, and the companion history Problem List panel.

## Validation performed
- Targeted Jest: `PackagesTab.test.tsx`, `ProblemList.test.tsx`, `ProblemListPanel.test.tsx` - 38/38 passing.
- `npx tsc --noemit` clean.
- Verified both fixes live against the deployed dev environment (screenshot comparison before/after) as well as in Storybook.

### DIFF
--- a/apps/frontend/src/app/__tests__/features/companionHistory/components/ProblemList.test.tsx
+++ b/apps/frontend/src/app/__tests__/features/companionHistory/components/ProblemList.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { fireEvent, render, screen, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import '@testing-library/jest-dom';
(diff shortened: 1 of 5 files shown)


ankit-yc (approved):
Packages CTA classes match ServicesTab's existing sibling pattern exactly (border-t/hairline/px-5!/py-[11px]!). Dropdown is the shared themed component, not a new one. Tests updated for the new interaction (click+listbox vs selectOptions). CI green.

PR #3012, opened 2026-09-11
ci(repo): gate new frontend components on having a Storybook story

aupyay:
## Summary
- Adds `scripts/ci/story-coverage.mjs` + `.github/workflows/story-coverage.yml`: a diff-scoped CI gate (same shape as `diff-coverage.mjs`) that fails a PR only when a file it *adds* under `apps/frontend/src/app` looks like a real, exported React component and has no sibling `.stories.tsx`. A `// no-story: <reason>` comment is the escape hatch for a genuine exception.
- Backfills the 4 real gaps this gate's own audit found in the existing codebase:
  - `DispensaryFilterBar` (Inventory) had no story at all - added one with 3 `play`-verified stories.
  - `StripeOnboarding`'s and `MerckManuals`' default-exported page components are `ProtectedRoute`/`OrgGuard`/`Suspense` wrappers with no UI of their own beyond a loading skeleton - marked `// no-story` with the reason, since their actual visible pieces already have stories (`StripeSetupStatus.stories.tsx`, `MerckReaderPortal.stories.tsx`).

## Why
Nothing in CI previously noticed a new component shipping without a Storybook story. Chromatic's `onlyChanged` and the advisory `storybook-interactions.yml` job both only ever touch stories that *already exist* - neither can notice a component that never got one in the first place. A recent UI-bug pass found real defects (an inconsistent CTA button, an unthemed native `<select>`) that had gone unreviewed in isolation for exactly this reason.

The gate intentionally does not retroactively require a story from the pre-existing backlog of components - only from what a PR newly adds - so its cost always matches the size of the change.

## Impact area
`apps/frontend` (new Storybook story + two `// no-story` annotations), and CI (`scripts/ci`, `.github/workflows`).

## Validation performed
- `node --test scripts/ci/story-coverage.test.mjs` - 15/15 passing.
- Mutation-tested: broke the detection logic and the `NO_STORY_MARKER`/size-floor checks, confirmed the relevant tests fail, restored.
- Validated the detection heuristic against a real historical commit (`git worktree add --detach`) that predates a story that now exists, confirming it would have caught the gap.
- New `DispensaryFilterBar.stories.tsx`: all 3 stories verified live in Storybook, each `play` function mutation-tested against the real component source.
- `npx tsc --noemit` clean.

### DIFF
--- a/.github/workflows/story-coverage.yml
+++ b/.github/workflows/story-coverage.yml
@@ -1,0 +1,54 @@
+name: Story Coverage
+
+# Fails when a pull request adds a new React component under apps/frontend
+# without a sibling Storybook story. Nothing else in CI checks this: Chromatic
+# only visually diffs stories that already exist, and the Storybook
+# interactions job only runs `play` functions on existing stories - neither
+# has any way to notice a component nobody ever gave a story in the first
+# place, which is how real UI bugs (an inconsistent button style, a native
+# <select> nobody themed) sat unreviewed. See scripts/ci/story-coverage.mjs
+# for the detection heuristic and its escape hatch (`// no-story: <reason>`).
+#
+# Scoped to files a PR ADDS, not the whole app: the pre-existing backlog of
+# storyless components is real and this gate does not attempt to clear it,
+# only stop it from growing - the same shape as scripts/ci/diff-coverage.mjs.
+on:
+  pull_request:
+    branches: [main, dev]
+    paths:
+      - 'apps/frontend/src/app/**'
+      - 'scripts/ci/story-coverage.mjs'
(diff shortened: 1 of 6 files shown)


ankit-yc (approved):
Both real findings fixed on the branch: Aikido HIGH/MEDIUM path-traversal in story-coverage.mjs's file reads (resolveWithin() guard, new test proves a ../ or absolute path is skipped not read, mutation-checked), then a CodeQL HIGH insecure-temp-file in my own follow-up test (predictable Date.now()-based tmpdir path -> mkdtempSync). All CI green including Sonar/CodeQL/Supply Chain.

PR #3013, opened 2026-09-11
fix(repo): close Documenso consent-visibility and signing-completion gaps

aupyay:
## Summary
- `listForPms`/`listConsentDocumentsForPms` only ever found a signed consent document through its `appointmentId`; one linked to a patient only via `caseId` or `encounterId` (no appointment at all) never showed up in the patient's Consents panel. Extends the existing appointment-only linkage to all three legs (`loadRenderedDocumentsForPatientRecords` + `loadCaseAndEncounterIdsForPatient`).
- `completePersistedRenderedDocumentSigning` (`rendered-document.service.ts`) and `handlePassportRecordEvent` (`documenso.controller.ts`) are the two places a Documenso webhook actually completes a signature. Both had real gaps at their shared call sites:
  - Signing never propagated to the linked `TemplateInstance`/`ClinicalArtifact` - both already have their own `SIGNED` status value, unused until now.
  - Zero audit trail events fired on either completion path, despite the audit service being used ~140 other places in this codebase. Added with a best-effort patientId resolver (case → encounter → appointment) that skips the audit rather than fabricating a patient when nothing resolves (e.g. an org-level template with no clinical link).
  - `DocumentSignature.signatureText` was accepted by the FHIR sign endpoint and typed end-to-end, but silently dropped before reaching the database - it never survived from sign-initiation to the later webhook-driven completion. Also reused `buildDocumentSignature`, an already-tested builder that existing code had never actually called.
  - `ClinicalArtifactAttestation.signedPdfUrl` existed on the schema specifically for this and was never written - the passport completion path never called the existing PDF-download helper at all.

## Why
Follow-up to the consent-visibility fix in #2939: a user report that a signed consent document still wasn't showing for a specific patient led to finding the narrower case/encounter-only gap. Digging further into the same signing-completion code surfaced the other four gaps above.

## Impact area
`apps/backend` (`document.service.ts`, `rendered-document.service.ts`, `documenso.controller.ts`) and `packages/types` (`RenderedDocumentSigning` gains a `signatureText` field to carry it through to completion).

## Validation performed
- Targeted Jest: `document.service.test.ts`, `rendered-document.service.test.ts`, `documenso.controller.test.ts` - 149/149 passing.
- Every new/changed assertion mutation-tested: broke the corresponding source line, confirmed the test fails, restored, re-confirmed green.
- `npx tsc --noEmit` clean on both `apps/backend` and `packages/types`.

## Not included
A "Requested" lifecycle tab stays permanently empty on the frontend - that's a documented, deliberate gap (`recordLifecycle.ts`'s own comment says as much), not something fixed here, since closing it needs a product decision on what "requested" means before any backend work.

### DIFF
--- a/apps/backend/src/controllers/web/documenso.controller.ts
+++ b/apps/backend/src/controllers/web/documenso.controller.ts
@@ -4,6 +4,7 @@ import {
   DocumensoExternalRole,
   DocumensoService,
 } from "src/services/documenso.service";
+import { AuditTrailService } from "src/services/audit-trail.service";
 import { FormAssignmentService } from "src/services/form-assignment.service";
 import { completePersistedRenderedDocumentSigning } from "src/services/rendered-document.service";
 import { notifyOwnerOfPassportUpdate } from "src/services/pet-clinical-records.service";
@@ -259,7 +260,11 @@ async function handlePassportRecordEvent(
       supersededById: null,
       artifact: { status: { not: "VOID" } },
     },
-    select: { id: true, artifactId: true },
+    select: {
+      id: true,
+      artifactId: true,
+      artifact: { select: { organisationId: true, kind: true } },
+    },
   });
   if (!attestation) return false;
   const signedAt = new Date();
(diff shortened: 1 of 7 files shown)


ankit-yc (approved):
CI Required + Supply Chain Required green, Sonar 4/4 passed, CodeQL/GitGuardian clean. New loadRenderedDocumentsForPatientRecords/loadCaseAndEncounterIdsForPatient queries all scope on organisationId (tenant isolation preserved); tests cover the case/encounter-linked-only regression this PR fixes. Aikido flagged 1 new MEDIUM issue but posted no inline comment and the dashboard is unreachable from this session (same limitation as #3002) - not suppressed, just unreviewed; Aikido is non-required on dev.

PR #3014, opened 2026-09-11
docs(repo): give real shape to the 197 FHIR-nested opaque schemas

aupyay:
## What changed

Continues #2944 past the request/response-body pass (#2999). The remaining opaque schemas were properties nested inside otherwise well-typed FHIR resource schemas (`Appointment.contained`, `Meta.author`, `SearchParameter.base`, `Address.use`, etc.) plus 12 app-specific request/response body schemas that were never traced.

Every one of the 197 flagged locations got real, source-verified treatment:

- **177 standard FHIR fields** traced against `packages/fhirtypes/src/*.ts` — the repo's own hand-written FHIR R4 type definitions. `code`/enum-typed fields (e.g. `Address.use`, `ContactPoint.system`, `CapabilityStatement.status`) became real string enums; every `Reference<T>` and `contained: Resource[]` field now points at two new shared schemas, `Reference` and `Resource`, built to match those TS interfaces field-for-field (`Resource` is a `oneOf` of the same narrow union `packages/fhirtypes/src/Resource.ts` actually declares, not the full FHIR resource set).
- **20 app-specific fields** traced to backend controllers, Prisma enums, and shared types: contact status/type (`ContactStatus`/`ContactType`), task status (`TaskStatus`), organisation type, document category (`OrgDocumentCategory`), invoice line items (`InvoiceItem`), reschedule/document-request bodies, and the Documenso webhook payload shape.

A few fields are genuinely opaque JSON columns with no database-enforced shape (`DeveloperAppointment.patient/lead/room/appointmentType`, a free-form task-answers map). Rather than asserting a shape the database doesn't guarantee — and per this schema's own existing stated policy of not guessing at these — they stay `additionalProperties: true` with an honest, source-cited description of the convention every current write path follows.

## Why

`scripts/ci/openapi-drift.mjs`'s `placeholderSchemas` ratchet only prevents the gap from growing; #2944 asked for the actual gap to shrink. This closes essentially all of the remaining FHIR-nested-property class of debt described in that issue.

## Impact area

`apps/frontend/public/static/openapi/openapi.yaml` (spec only — no runtime code changed) and `scripts/ci/openapi-drift-baseline.json` (ratchet ceiling lowered to match).

## Validation

Ran `scripts/ci/openapi-drift.mjs` itself (not a reimplementation) before and after:

- `placeholderSchemas`: 256 → 64 (all 197 originally-flagged locations addressed; the residual 64 includes both this PR's handful of deliberately-honest opaque fields and the pre-existing genuinely-opaque request/response bodies from #2999)
- `missingFromSpec`: 793 (unchanged)
- `staleInSpec`: 34 (unchanged)
- Exit code 0, no org-header regressions

Also verified: every `$ref` introduced by this change resolves to a real `components.schemas` entry (0 broken refs across 600 total refs in the spec), and the full document still parses as valid YAML/OpenAPI.

### DIFF
--- a/apps/frontend/public/static/openapi/openapi.yaml
+++ b/apps/frontend/public/static/openapi/openapi.yaml
@@ -23125,9 +23125,10 @@ components:
       type: object
       properties:
         resourceType:
-          type: object
-          additionalProperties: true
-          description: "Type 'AccessPolicy'"
+          type: string
+          enum:
+            - AccessPolicy
+          description: This is a AccessPolicy resource
         id:
           type: string
         meta:
@@ -23141,9 +23142,7 @@ components:
         contained:
           type: array
           items:
-            type: object
-            additionalProperties: true
-            description: Type Resource
+            $ref: '#/components/schemas/Resource'
         extension:
           type: array
           items:
@@ -23157,13 +23156,9 @@ components:
         basedOn:
           type: array
           items:
-            type: object
-            additionalProperties: true
-            description: 'Type Reference<AccessPolicy>'
+            $ref: '#/components/schemas/Reference'
         compartment:
-          type: object
-          additionalProperties: true
-          description: Type Reference
+          $ref: '#/components/schemas/Reference'
         resource:
           type: array
           items:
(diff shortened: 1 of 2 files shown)


ankit-yc (approved):

ankit-yc (approved):
Spec-only change, no runtime code touched. Independently reran scripts/ci/openapi-drift.mjs against this branch (prisma generate + build packages first): placeholderSchemas 64, exit 0 - matches the PR's claimed 256->64 exactly. All CI green.

PR #3016, opened 2026-09-11
fix(repo): scan every staged file for secrets

### LINKED ISSUE #2769: check-staged-secrets: the repo-specific stage reads two directories, and stage one skips silently when gitleaks is absent
`scripts/check-staged-secrets.js` runs two stages: gitleaks over everything staged, then a
repo-specific heuristic over a subset of paths. Two properties of it were measured tonight while
investigating untracked `.env.bak` files found in a working tree (fixed for staging by #2768).
Neither is an incident — nothing is or has been committed — and both are about what the gate can
tell you when it passes.

## 1. The repo-specific stage only reads two directories

```js
// line 5
const PROTECTED_ROOTS = ['apps/mobileAppYC/', 'apps/frontend/'];
// line 158
const stagedFiles = getStagedFiles().filter(isProtectedPath);
```

Driven in a throwaway worktree at `1203d952b` with **identical, deliberately non-credential prose
content** in each file, `git add -f`, then the gate:

```
apps/frontend/.env.probe        exit 1   "Remove these staged values before committing"
apps/backend/.env.probe         exit 0
packages/database/.env.probe    exit 0
apps/frontend/.env.example      exit 0   correctly allowed
```

Same bytes, different directory, opposite verdicts. `BLOCKED_LOCAL_FILES` similarly has a `.env`
rule for `apps/frontend/` and none for `apps/backend/` or `packages/database/` — which are where
the `.env.bak` files that prompted this actually were.

This may well be deliberate: the frontend and mobile apps are where committed secrets have
actually leaked here. Filing it as a question rather than a defect. **If the scoping is intended,
the useful change is a comment saying so** — the constant currently reads as an oversight to
anyone who finds it while auditing.

## 2. When gitleaks is absent, stage one is skipped silently

```js
const runGitleaksWhenAvailable = () => {
  try { execFileSync('gitleaks', [...]); }
  catch (error) {
    if (error.code === 'ENOENT') return;   // no warning, no exit code
    process.exit(error.status ?? 1);
  }
};
```

Driven with a stripped `PATH`, with a control first proving the strip actually hid `gitleaks`
while leaving `node` reachable (the two share a directory, so a naive strip hides neither or
both):

```
WITHOUT gitleaks   exit 1, the frontend finding only, no mention of gitleaks anywhere in the output
WITH gitleaks      exit 1, "no leaks found", then the same frontend finding
```

`apps/backend/.env.probe` appears in neither run. **Had the only staged file been a backend one,
both runs would have exited 0** — one because the scanner found nothing, one because there was no
scanner — and nothing in the output distinguishes them.

A gate that could not run is not a gate that passed, and this one currently cannot tell you which
happened. The `ENOENT` return is defensible for a contributor without the binary; the silence is
not. One `console.warn` naming the skip would close it.

## Limits of this report

**Detection power was not tested.** Both probes were worded as ordinary prose by design, so
"gitleaks found no leaks" says nothing about whether it catches real credentials. Writing a
realistic fixture is prohibited by this repo's own rules and that prohibition is correct — the
path and presence discriminators above are what could be measured without breaking it.

`.gitleaks.toml` is notably careful work: `[extend] useDefault = true`, allowlists scoped with
both `paths` and `regexes` under `condition = "AND"`, and a comment explaining that `paths` alone
would exempt a whole file. Nothing here contradicts that.

Found by Claude L2 (path scoping) and Claude L3 (silent `ENOENT` skip), independently reproduced
by each on the other's finding.

## Clinic-day strategy connection (2026-09-12)

This reopened issue remains in the security/quality stream of #3102. Investigate the current implementation and intended protected-path policy before changing it. A required scanner that is unavailable must not be represented as a successful scan; warning-only behavior cannot satisfy a required security gate. Preserve gitleaks defaults, scoped allowlists, secretlint, hooks and CI controls; fix the underlying scope/availability behavior without suppressing findings. The earlier report's deliberately non-credential probes are historical evidence, not a current detection-coverage claim.

aupyay:
## PR Checklist

The repository-specific staged-secret scanner only inspects frontend and mobile paths. Staged backend, package, and root text files therefore rely on gitleaks alone, while a missing gitleaks binary is skipped without any warning.

## What is the new behavior?

The scanner now checks every staged text file, blocks local `.env` variants across the repository while preserving `.env.example`, and warns when gitleaks is unavailable. Backend Jest coverage exercises both gaps and the allowed example case.

## Related Issue(s)

Fixes #2769

### DIFF
--- a/apps/backend/test/scripts/check-staged-secrets.test.ts
+++ b/apps/backend/test/scripts/check-staged-secrets.test.ts
@@ -1,0 +1,85 @@
+import { execFileSync, spawnSync } from "node:child_process";
+import {
+  mkdirSync,
+  mkdtempSync,
+  rmSync,
+  symlinkSync,
+  writeFileSync,
+} from "node:fs";
+import { dirname, join, resolve } from "node:path";
+import { tmpdir } from "node:os";
+
+const checker = resolve(
+  __dirname,
+  "../../../../scripts/check-staged-secrets.js",
+);
+const git = execFileSync("which", ["git"], { encoding: "utf8" }).trim();
+
+const runChecker = (files: Record<string, string>) => {
+  const repository = mkdtempSync(join(tmpdir(), "yc-staged-secret-"));
+
+  try {
+    const bin = join(repository, "bin");
+    mkdirSync(bin);
+    symlinkSync(git, join(bin, "git"));
+    const env = { ...process.env, PATH: bin };
+    execFileSync("git", ["init", "--quiet"], { cwd: repository });
+
+    for (const [file, content] of Object.entries(files)) {
+      const target = join(repository, file);
+      mkdirSync(dirname(target), { recursive: true });
+      writeFileSync(target, content);
+      execFileSync("git", ["add", "-f", "--", file], { cwd: repository });
+    }
+
+    const gitControl = spawnSync("git", ["--version"], { env });
+    expect(gitControl.status).toBe(0);
+
+    return spawnSync(process.execPath, [checker], {
+      cwd: repository,
+      encoding: "utf8",
+      env,
+    });
+  } finally {
+    rmSync(repository, { recursive: true, force: true });
+  }
+};
+
+test("scans staged text files outside frontend and mobile", () => {
+  const value = ["fixture", "only", "not", "credential", "7Q9Z2X8M4"].join("-");
+  const result = runChecker({
+    "apps/backend/config.js": `const token = "${value}";`,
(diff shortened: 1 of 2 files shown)


ankit-yc (approved):
Reviewed the repository-wide staged-file scan, local env-file exception, missing-gitleaks warning, and mutation-sensitive Jest coverage. The focused, full backend, script, lint, type-check, secret, and hosted security gates support approval.

PR #3017, opened 2026-09-11
fix(frontend): close out the hardcoded-hex sweep

aupyay:
## What changed

Resolves the last 65 unjustified literals flagged by `check-hardcoded-colours.mjs` (all inside `features/marketing`) - the baseline is now **0 to migrate**, closing out the sweep that started this session at 460+.

28 are real token fixes; 37 are justified (bespoke, or coincidentally matching a token that isn't the right one for the role) with a `why` in the baseline JSON explaining exactly what was checked.

## Real fixes - one confirmed bug pattern, found in three places

A hero-video scrim gradient hardcoded `--page`'s light-mode RGB directly instead of reading the token: `motion.tsx`'s shared `HERO_SCRIM_STYLE`, `PetParents.tsx`'s own local override of the same scrim, and a version-stat divider in `SiteFooter.tsx`. In dark mode the scrim/divider stayed pinned to warm cream while the surrounding page correctly flips to espresso, producing a visible seam - the same class of bug as the marketing hero-video dark-mode fix earlier this week. Routed every gradient stop through `color-mix(in srgb, var(--page) N%, transparent)`; verified live in both themes (screenshots below).

Other real fixes, same "literal frozen at one theme's RGB, next to a sibling that already reads the token" pattern:
- `ContactusPage.tsx`: email/phone channel-card icon tints and the success-confirmation icon hardcoded `--blue`/`--success`'s light RGB right next to an `iconColor` that already read the token directly - verified live via computed style that the icon glyph and its tint now agree in both themes.
- `ContactusPage.tsx`: the required-mark asterisk and submit-error text hardcoded `--color-danger-700`'s light value - verified live it now reads the more legible dark-mode red instead of staying frozen.
- `PetBusinesses.tsx`: the offline-app status dot's pulse ring hardcoded `--success`'s RGB, same pattern already fixed on two other pulse rings earlier in this sweep (#2966, #2979).
- `PetParents.tsx` / `Pricing.tsx`: phone-mockup bezel/notch and the active billing-toggle pill both hardcoded `--spot`'s exact value in its native background/surface role.
- `AuthShell.stories.tsx`: two brand-emphasis `<em>` colours duplicated literals already fixed in the real SignIn/SignUp/AuthShell components (#2990, #2993) but never got the same fix in the story fixture.

## Justified (no exact token match, or role mismatch)

Checked every remaining literal against the full token set before justifying, not just the obvious families:
- Discord's own brand blurple, GitHub's star yellow - not design-system colours.
- Several coincidental matches to tokens that *flip* (`--divider`, `--hairline`, `--ink-body`) but are used here in an **ink** role on an always-dark `--spot` card, not their native border/surface role - using the flipping token would go invisible in dark mode, so these are false matches, not real ones.
- Computed-style test assertions (`getComputedStyle()` only ever returns a resolved colour, never a token name).
- A handful of bespoke tints with no match anywhere in `globals.css`.
- `SiteFooter.tsx`'s `STATUS_TONE_TEXT.success` (`#1d6b4f`) is flagged but deliberately left alone: its warning/danger siblings already flip via tokens and this one doesn't, which reads like a possible latent dark-mode gap - but no existing token is an exact match, so migrating it would change the rendered colour rather than preserve it. Noted in the baseline `why` as worth a follow-up if it turns out to matter.

## Testing

- Added/extended guard tests for every real fix (source-text assertions, since `color-mix()`/`var()` are opaque strings to jsdom's `getComputedStyle`).
- Mutation-checked each one individually, then did a full bulk revert/restore across all eight touched source files at once and confirmed all 17 new/extended assertions fail together, then pass again once restored.
- `npx tsc --noemit`: clean.
- `pnpm --filter frontend run lint`: 0 errors (1 pre-existing warning, unrelated file).
- Full touched-suite: 110/110 passing.
- Live-verified in the browser, both themes: the `PetParents` hero scrim/phone-bezel fix and the `ContactusPage` icon-tint/danger-colour fixes (screenshots taken during development, not attached here - happy to add if useful).

## Impact

Frontend only, CSS/JSON/test changes. No visual change in light mode anywhere (every real fix preserves the exact light-mode pixel value). Dark mode changes are all fixes to a previously-invisible-in-light-mode mismatch.

### DIFF
--- a/apps/frontend/src/app/__tests__/features/marketing/site/AuthShell.test.tsx
+++ b/apps/frontend/src/app/__tests__/features/marketing/site/AuthShell.test.tsx
@@ -130,3 +130,20 @@ describe('brand panel reads ink and glow colours from real tokens', () => {
     expect(source).toMatch(/color-mix\(in srgb, var\(--pink\) 10%, transparent\)/);
   });
 });
+
+describe('the AuthShell story fixtures route their brand-emphasis ink through real tokens', () => {
+  const storiesSource = readFileSync(
+    join(process.cwd(), 'src/app/features/marketing/site/AuthShell.stories.tsx'),
+    'utf8'
+  );
+
+  it('does not hardcode the pet-parent emphasis ink as a frozen literal', () => {
+    expect(storiesSource).not.toContain("color: '#8fb6f5'");
+    expect(storiesSource).toContain("color: 'var(--color-accent-dark)'");
+  });
+
+  it('does not hardcode the developer emphasis ink as a frozen literal', () => {
+    expect(storiesSource).not.toContain("color: '#5ce1e6'");
+    expect(storiesSource).toContain("color: 'var(--cyan)'");
+  });
+});
--- a/apps/frontend/src/app/__tests__/features/marketing/site/SiteFooter.test.tsx
+++ b/apps/frontend/src/app/__tests__/features/marketing/site/SiteFooter.test.tsx
@@ -1,4 +1,6 @@
 import React from 'react';
+import { readFileSync } from 'fs';
+import { join } from 'path';
 import { render, screen, fireEvent } from '@testing-library/react';
 import '@testing-library/jest-dom';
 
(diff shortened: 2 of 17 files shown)


ankit-yc (approved):
Reviewed the diff directly (not just CI): 8 source files, each real fix verified against the actual token declarations in globals.css before applying (checked fixed vs. flipping status for every token used, not assumed). The one real bug found - a hero-video scrim hardcoding --page's light-mode RGB in three places (motion.tsx's shared component, PetParents.tsx's own override, SiteFooter.tsx's divider) - was live-verified in both themes via the browser, not just asserted in source.

Guard tests: 17 new/extended assertions, each mutation-checked individually, then a full bulk revert/restore across all eight touched source files confirmed 17 failures together and a clean pass on restore.

tsc clean, lint 0 errors, 110/110 touched-suite tests passing, CI green (76 checks). Approving.

PR #3018, opened 2026-09-11
fix(backend): log unconfigured SuperAdmin contact mirroring

### LINKED ISSUE #2767: Is the silent return in forwardWebContact intended? It makes CRM mirror loss uncountable
## The receiving side made this loss countable. The sending side is still silent.

`SuperadminContactService.forwardWebContact` returns with **no log at all** when the mirror is unconfigured:

```ts
// apps/backend/src/services/superadmin-contact.service.ts:27-29
const url = process.env.SUPERADMIN_CONTACT_INTAKE_URL;
const key = process.env.SUPERADMIN_CONTACT_INTAKE_KEY;
if (!url || !key) return;          // <- silent
```

Every other outcome on this path is logged:

| line | outcome | log |
| --- | --- | --- |
| 29 | unconfigured | **none** |
| 40 | panel rejected the forward | `logger.warn(..., { status })` |
| 45 | request threw | `logger.error(...)` |

The call site is fire-and-forget by design (`contact-us.controller.ts:192`), and the product database remains the source of truth, so **no submission is lost** — the CRM mirror is simply blind, with nothing recording for how long or for how many.

## Why this is worth a line of code

The SuperAdmin side hit exactly this and fixed it, with the reasoning written into the code:

```ts
// SuperAdmin apps/admin/src/app/lib/intakeGuard.ts
// An unconfigured intake refuses real submissions indefinitely, and until this
// log line existed it did so with no trace anywhere: nothing in the platform
// log, nothing in the panel ... One line per refusal is what makes the loss
// countable and alertable.
logger.error('Intake refused: not configured', { intake, envVar });
```

The same argument applies verbatim here, and the asymmetry has a concrete consequence today. Reconstructing how many contact submissions the CRM missed needs both sides:

```
receiver refusals > 0  AND  sender warn(503) > 0   -> sender was configured; the two counts
                                                       should agree, each checks the other
receiver refusals = 0  AND  sender warn      = 0   -> AMBIGUOUS: either nothing was ever sent,
                                                       or the sender was unconfigured and
                                                       returned silently at line 29
```

**Zero is the ambiguous value**, and it is ambiguous only because line 29 leaves no trace. One log line there makes it decidable from the logs instead of requiring someone to reconstruct historical environment state.

## The tests currently assert the silence

```
apps/backend/test/services/superadmin-contact.service.test.ts
  it("does nothing when the URL is unset")
  it("does nothing when the key is unset")
```

These pass on the current behaviour and would need updating alongside any change — which is the signal that the silence is load-bearing in the suite rather than incidental, and worth a deliberate decision rather than a drive-by edit.

## The question, rather than an assertion

Was the silent return at line 29 deliberate — mirroring off is a supported configuration and should be quiet — or is it the same gap the receiving side closed? If it is supported-and-quiet, the ambiguity above is the accepted cost and this issue can be closed as intended behaviour.

If a log is wanted, the minimal shape matching the receiver's is one line naming the missing variable and nothing else — never the value:

```ts
if (!url || !key) {
  logger.warn("SuperAdmin contact mirroring is not configured", {
    missing: !url ? "SUPERADMIN_CONTACT_INTAKE_URL" : "SUPERADMIN_CONTACT_INTAKE_KEY",
  });
  return;
}
```

Per-submission rather than per-boot is what makes it countable, which is the whole point on the receiver side; if the volume is a concern that is the trade-off to decide here.

## Context

- SuperAdmin issue [#315](https://github.com/YosemiteCrew/SuperAdmin/issues/315) — restoring `CONTACT_INTAKE_KEY` will not backfill the CRM.
- Three values across two deployments: `CONTACT_INTAKE_KEY` on the panel, `SUPERADMIN_CONTACT_INTAKE_URL` and `SUPERADMIN_CONTACT_INTAKE_KEY` on this backend. `/api/health` on the panel reads only its own env, so it cannot report the sender's state either way.

Read at `dev` `71f3fd30`; SuperAdmin at `8b5fa390`.

aupyay:
## Summary
- `SuperadminContactService.forwardWebContact` returned silently when the mirror env vars were unset, unlike every other outcome on this path (rejected/errored are both logged).
- Adds one `logger.warn` naming the missing var, matching the pattern already used on the SuperAdmin receiving side (`intakeGuard.ts`).
- Updates the two existing tests that previously asserted the silence.

Closes #2767.

## Test plan
- [x] `apps/backend`: targeted jest run for `superadmin-contact.service` — 5/5 pass, 100% statement/branch/func/line coverage on the changed file
- [x] `pnpm --filter backend run lint` — 0 errors (24 pre-existing warnings, same as dev)
- [x] `pnpm --filter backend run type-check` — clean
- [x] pre-push hook (full monorepo lint + type-check) — clean

### DIFF
--- a/apps/backend/src/services/superadmin-contact.service.ts
+++ b/apps/backend/src/services/superadmin-contact.service.ts
@@ -26,7 +26,14 @@ export const SuperadminContactService = {
   ): Promise<void> {
     const url = process.env.SUPERADMIN_CONTACT_INTAKE_URL;
     const key = process.env.SUPERADMIN_CONTACT_INTAKE_KEY;
-    if (!url || !key) return;
+    if (!url || !key) {
+      logger.warn("SuperAdmin contact mirroring is not configured", {
+        missing: !url
+          ? "SUPERADMIN_CONTACT_INTAKE_URL"
+          : "SUPERADMIN_CONTACT_INTAKE_KEY",
+      });
+      return;
+    }
 
     try {
       const response = await fetch(url, {
--- a/apps/backend/test/services/superadmin-contact.service.test.ts
+++ b/apps/backend/test/services/superadmin-contact.service.test.ts
@@ -49,22 +49,29 @@ describe("SuperadminContactService.forwardWebContact", () => {
     process.env.SUPERADMIN_CONTACT_INTAKE_KEY = "shared-secret";
   };
 
-  it("does nothing when the URL is unset", async () => {
+  it("warns naming the missing var and does not fetch when the URL is unset", async () => {
     process.env.SUPERADMIN_CONTACT_INTAKE_KEY = "shared-secret";
     delete process.env.SUPERADMIN_CONTACT_INTAKE_URL;
     await SuperadminContactService.forwardWebContact(PAYLOAD);
     expect(fetchMock).not.toHaveBeenCalled();
-    expect(warnMock).not.toHaveBeenCalled();
     expect(errorMock).not.toHaveBeenCalled();
+    expect(warnMock).toHaveBeenCalledWith(
+      "SuperAdmin contact mirroring is not configured",
+      { missing: "SUPERADMIN_CONTACT_INTAKE_URL" },
+    );
   });
 
-  it("does nothing when the key is unset", async () => {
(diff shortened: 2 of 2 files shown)


ankit-yc (approved):
LGTM

PR #3019, opened 2026-09-11
fix(repo): retire inert public booking auto-confirmation

### LINKED ISSUE #2550: fix(backend): honour or retire autoConfirm in public booking confirm()
## What is wrong

`PublicBookingSettings.autoConfirm` is a persisted, admin-settable, API-exposed flag whose documented contract is never implemented. The schema states the contract explicitly:

```prisma
/// False - the default - means a public request lands as REQUESTED for a human
/// to accept. True writes to the calendar directly, and is only ever honoured
/// after the requester has confirmed their email address.
autoConfirm       Boolean  @default(false)
```

The email-confirmation path structurally cannot honour it, because it never selects the column. The only `bookingSettings` field `confirm()` fetches is `replyToEmail`:

```ts
organization: {
  select: {
    name: true,
    email: true,
    bookingSlug: true,
    bookingSettings: { select: { replyToEmail: true } },
  },
},
```

And the whole effect of following the emailed link is a status flip:

```ts
await prisma.publicBookingRequest.update({
  where: { id: existing.id },
  data: { status: "CONFIRMED", confirmedAt: new Date() },
});
```

There is no appointment or calendar write anywhere in the file. The other transition, `setStatus`, is likewise a bare `data: { status }` updateMany with no appointment creation.

`autoConfirm` is read in exactly one place repo-wide, and only to build a projection field:

```ts
requiresConfirmation: !practice.settings?.autoConfirm,
```

That projection field has no consumer. `grep -rn "requiresConfirmation" apps/frontend/src apps/backend/src packages` returns 5 hits, all of them type declarations, the projection above, or story/test fixtures. `BookClient.tsx`, the renderer for the public page, never reads it.

The toggle is reachable in production, not dead config. The onboarding wizard exposes it, the controller validates it with a bare `z.boolean()`, and the save path writes it verbatim into the upsert:

```ts
autoConfirm: input.autoConfirm,
```

Nothing coerces it false, and there is no guard, feature flag, TODO, or UI hint that it does nothing.

This is deferred scope that shipped a live switch. The feature module states inertness as a deliberate design rule ("A request is inert. Submitting one writes to `PublicBookingRequest` and nothing else. It does not touch a calendar"), and the original commit gives a sound reason: `createAppointment` asserts a real Parent manages a real Companion, so satisfying it from a public form means minting both from unverified input. The defect is not that auto-confirm was deferred. It is that the control was shipped anyway, and the schema doc comment still asserts a contract the implementing module contradicts.

## Where

- `apps/backend/src/services/public-booking.service.ts:571-594` - `confirm()` selects only `replyToEmail` from `bookingSettings`
- `apps/backend/src/services/public-booking.service.ts:601-604` - the only write on the confirm path, a bare status flip
- `apps/backend/src/services/public-booking.service.ts:278` - the sole read of `autoConfirm`, feeding an unconsumed projection
- `apps/backend/src/services/public-booking.service.ts:106` - `requiresConfirmation` type declaration
- `apps/backend/src/services/public-booking.service.ts:29-32` - module doc stating a request is inert and never touches a calendar
- `apps/backend/src/services/public-booking.service.ts:680-704` - `setStatus`, the only other transition, also creates no appointment
- `packages/database/prisma/schema.prisma:1117-1120` - the doc comment promising the calendar write
- `packages/database/prisma/migrations/20260826090000_public_booking_foundation/migration.sql:23-25` - migration note drawing the same contrast
- `apps/backend/src/services/booking-page.service.ts:560` - `autoConfirm: input.autoConfirm` written straight into the upsert
- `apps/backend/src/controllers/booking-page.controller.ts:35` - bare `z.boolean()`, nothing coerces it false
- `apps/frontend/src/app/features/onboarding/pages/PublicBookingSetup/PublicBookingSetup.tsx:193` - the "Requests need confirmation" toggle copy
- `apps/frontend/src/app/features/onboarding/pages/PublicBookingSetup/PublicBookingSetup.tsx:666` - `autoConfirm: !needsConfirmation` sent on save
- `apps/frontend/src/app/(routes)/(book)/book/[slug]/BookClient.tsx:575` - public page copy, unconditional and currently truthful
- `apps/frontend/src/app/features/publicBooking/services/publicBooking.service.ts:27` - client-side type declaration, never read by a renderer
- `apps/backend/test/services/public-booking.service.test.ts:306-316` - the only `autoConfirm` assertion, and it covers the projection only

## Why it matters

A clinic that turns the confirmation requirement off believes it has enabled direct booking. It gets the manual request queue it explicitly opted out of, with no error, no warning, and no indication in the UI that the setting is inert. That is a shipped control with zero effect on a page clinics hand to the public.

The blast radius is genuinely small, and worth stating so triage does not over-weight this. Nothing fails unsafely: no calendar write means no double-booking, no Parent or Companion minted from unverified public input, no PII exposure, no billing, no auth surface. The pet owner is not misled, because the public page copy is unconditional and accurate ("This sends a request, not a booking"). The practice is not misled per request either, because the notification email sent on this same path says so directly: "This is a request, not an appointment. Nothing has been added to your calendar." No data is lost and no incorrect record is written.

What remains is real but bounded: a dead admin control that misrepresents what the product does, plus a schema doc comment that contradicts the module implementing it. A practice trusting the toggle could deprioritise the request queue on a false premise. Severity P3.

## How to reproduce

1. Sign in to the PMS as an org admin and open `/public-booking-setup`.
2. Select at least one bookable service, toggle "Requests need confirmation" off, and save. This persists `autoConfirm: true` via `PublicBookingSetup.tsx:666` into `booking-page.service.ts:560`. Publish the page.
3. Confirm the flag round-trips: `GET /v1/booking-page/:organisationId/config` returns `autoConfirm: true`, and `GET /v1/public/booking/:slug` returns `requiresConfirmation: false`.
4. As an anonymous visitor, open `/book/<slug>`. The page still renders "This sends a request, not a booking. `<practice>` will confirm the appointment with you". The `requiresConfirmation: false` value is never read.
5. Submit a request and follow the emailed confirmation link.
6. Observed: the `PublicBookingRequest` row moves to `CONFIRMED`, the practice gets a notification email, and the row appears in the PMS booking requests queue as awaiting action. No `Appointment` row is created and nothing lands on any calendar.
7. Expected per `packages/database/prisma/schema.prisma:1117-1119`: with `autoConfirm` true, this step writes the appointment to the calendar directly.

Static equivalent, no environment needed:

```
rg -n "autoConfirm" apps/backend/src
  -> a type, a Prisma select, and one projection in public-booking.service.ts; persistence only in booking-page.service.ts

grep -rn "requiresConfirmation" apps/frontend/src apps/backend/src packages
  -> 5 hits, all type declarations or test/story fixtures, none in a renderer

rg -n "createAppointment|appointment.create" apps/backend/src/services/public-booking.service.ts
  -> no matches
```

## Suggested fix

Two options, and the cheap one is probably the right one for now.

The cheap, correct fix is to stop shipping a control that does nothing. Hide or disable the wizard toggle, or label it as not yet active, and correct the schema doc comment at `schema.prisma:1117-1119` so it describes the behaviour that exists rather than the one that was planned. This removes the false affordance without touching the booking path at all.

The full fix is to implement the calendar write in `confirm()`, which is not a small patch. `createAppointment` requires a real Parent managing a real Companion, so honouring `autoConfirm` means minting both from unverified public form input, which is precisely what the current design deliberately avoids. Anyone taking this on needs an answer for identity resolution (match an existing Parent by verified email, or create a provisional record clearly marked as public-origin) before writing any appointment.

If the full fix is taken, order the work carefully. Implement the calendar write in `confirm()` first, then let `BookClient` branch on `requiresConfirmation`. Making the client honour `requiresConfirmation: false` while `confirm()` still writes nothing to a calendar would tell a member of the public their appointment is booked when no appointment exists, which is strictly worse than today. The unused flag on the client is a symptom of this defect, not a second independent one, and right now it is the thing keeping the public page honest.

Either path should add a test that asserts the chosen behaviour. Today the five `confirm()` tests assert only the status flip and the notification email, and the single `autoConfirm` assertion covers the projection, so nothing currently pins down what is intended.

## Provenance

Found by Codex review on PR #2545, independently verified against dev.

Severity after verification: P3

aupyay:
## PR Checklist

The public-booking admin flow exposes an `autoConfirm` setting even though submissions always enter the request workflow. This creates an inert control and a response contract that can falsely imply direct calendar booking.

## What is the new behavior?

- Remove `autoConfirm` from the admin API and frontend configuration contract.
- Replace the toggle with fixed copy explaining that staff confirmation is required.
- Always report `requiresConfirmation: true` from the public projection.
- Normalize the retained compatibility column to `false` whenever settings are saved.
- Add regression coverage and update the API/schema documentation.

Validation:
- Backend focused suite: 144 tests, 100% statements/branches/functions/lines.
- Frontend focused suite: 49 tests, 100% statements/branches/functions/lines.
- All four production mutations are killed by the regressions.
- Workspace pre-push lint and type-check gates pass.
- Backend dependency build and frontend Storybook production build pass.
- Playwright visual QA confirms the fixed confirmation copy with no toggle or console errors.

## Related Issue(s)

Fixes #2550

### DIFF
--- a/apps/backend/src/controllers/web/booking-page.controller.ts
+++ b/apps/backend/src/controllers/web/booking-page.controller.ts
@@ -32,7 +32,6 @@ const SettingsSchema = z.object({
   serviceIds: z.array(z.uuid()).max(200),
   bookingWindowDays: z.number().int().min(1).max(180),
   bufferMinutes: z.number().int().min(0).max(240),
-  autoConfirm: z.boolean(),
   welcomeMessage: z.string().trim().max(500).nullish(),
   replyToEmail: z.string().trim().pipe(z.email().max(254)).nullish(),
   // Optional on purpose: a caller that omits it is saving settings, not
@@ -105,7 +104,6 @@ export const BookingPageController = {
         serviceIds: parsed.data.serviceIds,
         bookingWindowDays: parsed.data.bookingWindowDays,
         bufferMinutes: parsed.data.bufferMinutes,
-        autoConfirm: parsed.data.autoConfirm,
         welcomeMessage: parsed.data.welcomeMessage ?? null,
         replyToEmail: parsed.data.replyToEmail ?? null,
         publicBookingEnabled: parsed.data.publicBookingEnabled,
--- a/apps/backend/src/services/booking-page.service.ts
+++ b/apps/backend/src/services/booking-page.service.ts
@@ -404,7 +404,6 @@ export type BookingPageConfig = {
   serviceIds: string[];
   bookingWindowDays: number;
   bufferMinutes: number;
-  autoConfirm: boolean;
   welcomeMessage: string | null;
   replyToEmail: string | null;
 };
@@ -419,7 +418,6 @@ const toConfig = (
     serviceIds: string[];
     bookingWindowDays: number;
     bufferMinutes: number;
-    autoConfirm: boolean;
     welcomeMessage: string | null;
     replyToEmail: string | null;
   } | null,
@@ -435,7 +433,6 @@ const toConfig = (
   serviceIds: settings?.serviceIds ?? [],
   bookingWindowDays: settings?.bookingWindowDays ?? DEFAULT_BOOKING_WINDOW_DAYS,
   bufferMinutes: settings?.bufferMinutes ?? DEFAULT_BUFFER_MINUTES,
-  autoConfirm: settings?.autoConfirm ?? false,
   welcomeMessage: settings?.welcomeMessage ?? null,
   replyToEmail: settings?.replyToEmail ?? null,
 });
@@ -444,7 +441,6 @@ export type BookingPageSettingsInput = {
   serviceIds: string[];
   bookingWindowDays: number;
   bufferMinutes: number;
-  autoConfirm: boolean;
   welcomeMessage?: string | null;
   replyToEmail?: string | null;
   /**
@@ -480,7 +476,6 @@ export const BookingPageService = {
         serviceIds: true,
         bookingWindowDays: true,
         bufferMinutes: true,
-        autoConfirm: true,
         welcomeMessage: true,
         replyToEmail: true,
       },
@@ -557,7 +552,7 @@ export const BookingPageService = {
       serviceIds: requested,
       bookingWindowDays: input.bookingWindowDays,
       bufferMinutes: input.bufferMinutes,
-      autoConfirm: input.autoConfirm,
+      autoConfirm: false,
       welcomeMessage: input.welcomeMessage?.trim() || null,
       replyToEmail: input.replyToEmail?.trim().toLowerCase() || null,
     };
--- a/apps/backend/src/services/public-booking.service.ts
+++ b/apps/backend/src/services/public-booking.service.ts
@@ -126,7 +126,6 @@ type ResolvedPractice = {
     serviceIds: string[];
     bookingWindowDays: number;
     bufferMinutes: number;
-    autoConfirm: boolean;
     welcomeMessage: string | null;
     replyToEmail: string | null;
   } | null;
@@ -160,7 +159,6 @@ export const resolveSlug = async (
           serviceIds: true,
           bookingWindowDays: true,
           bufferMinutes: true,
-          autoConfirm: true,
           welcomeMessage: true,
           replyToEmail: true,
         },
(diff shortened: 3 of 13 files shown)


ankit-yc (approved):
Reviewed: autoConfirm retired cleanly — column kept in schema (no migration risk), app always normalizes to false, public responses always state staff confirmation. Tests updated to match (mutation-relevant: new 'ignores retired field' tests on both directions). 61/61 checks green.

PR #3020, opened 2026-09-11
fix(backend): make tax previews currency-aware

### LINKED ISSUE #2754: finance/tax.ts submits Stripe tax-preview amounts with a currency-blind x100
`finance/tax.ts` builds Stripe tax-preview line items with a currency-blind `Math.round(amount * 100)`, on the same object literal that carries the currency:

```ts
// src/services/finance/tax.ts, buildAutomaticTaxLineItems
return {
  amount: Math.round(discountedAmount * 100),
  description: ...,
  currency: input.currency,
  ...
};
```

That is the same defect class as the appointment path fixed in #2747, in the same shape: the conversion sits beside the currency it ignores. For an organisation billing in a zero-decimal currency the automatic-tax preview is submitted at **100x** the real amount, so the tax it returns is computed against the wrong base.

### Where it is

Six `* 100` sites, all in `allocateInvoiceDiscountAcrossLines` and `buildAutomaticTaxLineItems`, plus the reverse half:

| Line | Direction |
|---|---|
| `160`, `164`, `173`, `181` | major -> minor, discount allocation |
| `219` | major -> minor, **submitted to Stripe** |
| `300`, `304` | minor -> major, reading Stripe's `total_taxes` and `total_excluding_tax` back |

`304` is the reverse half of the same bug: a zero-decimal `total_excluding_tax` of `1000` is read back as `10`.

### Why it is not a two-line swap, which is the actual question

`allocateInvoiceDiscountAcrossLines` converts to minor units, allocates a discount with a largest-remainder pass, and converts back. It takes no currency. Threading one through is easy; deciding what it should then do is not:

- for a zero-decimal currency the current code distributes remainders at **1/100th of the smallest unit that exists**, so the allocation is finer-grained than the currency, and the `+= 1` remainder loop is handing out hundredths of a yen;
- correcting that changes the per-line split for every zero-decimal invoice, which is a money change and not a tidy-up;
- reading Stripe's amounts back needs a reverse converter that does not exist yet, and `toStripeMinorUnits` deliberately has no inverse.

So this is a correctness change with its own tests and its own review, which is why #2747 named it and left it rather than sweeping it in.

### Prior art to reuse

`src/utils/stripe-minor-units.ts` (added in #2747) is the currency-aware converter and is import-free on purpose, so a pure calculation module can use it without gaining a database dependency.

### Not established

Whether any organisation is currently billing in one of the fifteen unscaled currencies. The mechanism is in the repository; the exposure is a question about live data that needs someone who can see it.

aupyay:
## PR Checklist

Stripe automatic-tax previews assume every currency has two decimal places. That submits zero-decimal amounts at 100 times their value, allocates invoice discounts below the currency's smallest unit, and scales Stripe preview totals down incorrectly.

## What is the new behavior?

Tax previews now use the shared Stripe minor-unit converter for line amounts and discount allocation, and the matching inverse conversion for preview totals. Focused tests cover two-decimal behavior and an end-to-end zero-decimal preview with a multi-line discount.

## Related Issue(s)

Fixes #2754

### DIFF
--- a/apps/backend/src/services/finance/tax.ts
+++ b/apps/backend/src/services/finance/tax.ts
@@ -4,6 +4,10 @@ import {
   TaxBehavior as PrismaTaxBehavior,
   TaxProvider as PrismaTaxProvider,
 } from "@prisma/client";
+import {
+  fromStripeMinorUnits,
+  toStripeMinorUnits,
+} from "src/utils/stripe-minor-units";
 
 import type { InvoiceDiscountInput, InvoicePricingBreakdown } from "./pricing";
 import { roundMoney } from "./pricing";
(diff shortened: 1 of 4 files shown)


ankit-yc (approved):
Reviewed exact head da59ea7d3956303a5c34db10f14d8b4d9c2aac1f: all reported minor-unit conversions use the shared currency contract, zero- and two-decimal read/write paths are covered, and each new test was mutation-killed independently. No blocking findings.

PR #3022, opened 2026-09-11
fix(frontend): replace remaining native selects with the themed Dropdown

aupyay:
## Summary
19 fields across 14 files still rendered as a native browser `<select>` (unstyled option list, native OS chrome) instead of the design-system `Dropdown` component used everywhere else in the app. Reported via the Patient History → Consents panel's "Consent type" field; a full codebase sweep found 18 more across companion history, appointments, finance, compliance, onboarding, settings, auth, developer tools and marketing.

| Feature | Files | Fields |
|---|---|---|
| companionHistory | ConsentList, FlagList, AllergyList | Consent type, Flag type, Severity ×2, Allergy type |
| appointments | FrontDeskBoard, PrescriptionEditor | Assign room, Triage priority, Fulfillment |
| finance | EstimateDraftFields, InsuranceClaimFormFields, InsuranceClaimDetail | Companion ×2, Move claim to |
| compliance | ControlledSubstanceRegister | Control schedule, Unit |
| onboarding | PublicBookingSetup | Bookable window, Buffer between visits |
| settings, auth, developers, marketing | FederationSection, SignUp, CreateKeyForm, ContactusPage | Referral urgency, "I am", Environment, DSAR law |

Same mechanical fix throughout: swap the native `<select>` (and its `<label>`, if any - `Dropdown` renders its own via `Field`, so a wrapping/sibling label would otherwise duplicate) for `<Dropdown placeholder value onChange options>`, preserving default values and `onChange` behavior exactly - a styling fix, not a behavior change.

`ProblemList.tsx`'s Severity field (the same bug) is deliberately **not** included here - already fixed and pending merge in #3011.

## Why
User-reported: a screenshot of the Consents panel's native, unstyled dropdown. A scoping pass across the whole frontend confirmed the real scope (20 instances, not the "100+" initially estimated) before any fix work started.

## Impact area
`apps/frontend` only, across the 14 files listed above plus their test files.

## Validation performed
- Targeted Jest across all 14 touched components + their panel/page wrappers: 27 suites, 498 tests, all passing.
- `npx tsc --noemit` clean.
- `npx eslint` clean (zero warnings) on all 30 touched files.
- Every updated test (`userEvent.selectOptions` → button+listbox click pattern) verified live; two files (`CreateKeyForm.tsx`, `FederationSection.tsx`) had no pre-existing test coverage of the specific select field - confirmed via their test files rather than assumed, nothing was silently dropped.
- Mutation-tested the reported bug's own fix (`ConsentList.tsx`): broke the `onChange` wiring, confirmed the test fails, restored, confirmed green again.
- Visually verified the reported bug's fix live in Storybook: the Consent type field now renders as a themed button trigger, and its option panel matches the rest of the app (light card, blue-highlighted selection) instead of native OS dropdown chrome.

### DIFF
--- a/apps/frontend/src/app/__tests__/features/appointments/components/FrontDeskBoard.test.tsx
+++ b/apps/frontend/src/app/__tests__/features/appointments/components/FrontDeskBoard.test.tsx
@@ -176,7 +176,8 @@ describe('FrontDeskBoard', () => {
       />
     );
 
-    await user.selectOptions(screen.getByLabelText('Assign room'), 'room-2');
+    await user.click(screen.getByRole('button', { name: /Assign room/ }));
+    await user.click(within(screen.getByRole('listbox')).getByText('Exam 2'));
     expect(props.onAssignRoom).toHaveBeenCalledWith('7', 'room-2');
   });
 
@@ -220,7 +221,8 @@ describe('FrontDeskBoard', () => {
 
     await user.click(screen.getByLabelText('Patient'));
     await user.click(screen.getByRole('option', { name: 'Bruno — Sarah' }));
-    await user.selectOptions(screen.getByLabelText('Triage priority'), 'IMMEDIATE');
+    await user.click(screen.getByRole('button', { name: /Triage priority/ }));
+    await user.click(within(screen.getByRole('listbox')).getByText('Immediate'));
     await user.click(screen.getByRole('button', { name: 'Check in patient' }));
 
     expect(props.onAdd).toHaveBeenCalledWith(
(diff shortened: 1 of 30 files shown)


ankit-yc (approved):
Reviewed the full diff (not just the summary): mechanical native-<select>→Dropdown swap across all 14 files, options/values verified consistent with each field's original enum, no orphaned data-testid (checked dynamic-select had no other referrers). One non-blocking behavior nuance: EstimateDraftFields/InsuranceClaimFormFields Companion and ContactusPage DSAR area previously let a user re-select the "Choose a companion"/"Select one" placeholder as a real option to clear back to blank; the new Dropdown only shows that as display text (emptyLabel) before a first pick, with no list entry to return to blank afterward. Not flagging as a defect - these are required-before-submit fields, so the practical fix is picking the correct value, not clearing to none. All CI green (CI Required, Supply Chain Required, Sonar, CodeQL, GitGuardian, Aikido).

Issue #2168, opened 2026-08-14
fix(repo): QA pass on dev found 33 issues, 7 critical


ankit-yc:
## Summary

Manual QA pass over `dev.yosemitecrew.com` found **33 issues**: 7 critical, 9 high, 8 medium, 6 low, 3 informational. This tracks triage and fixes. Recorded 13 Aug 2026 against org `Avenger Park, Avondale FC`.

### Critical

| # | Area | Issue |
| --- | --- | --- |
| 1 | /appointments | Day, Week and Board views show **zero appointments for today** despite 321 in the org. Board reports 0 in every status column. |
| 2 | /appointments | Day view renders **two sets of time labels** (hourly and 15-minute) concatenated, so 12:00 AM and 11:45 PM each appear twice and the timeline is double height. |
| 3 | /finance, protected routes | A **valid session is dropped** navigating to a deep route: redirected to `/signin?next=...` while `sFrontToken` is still valid. After re-auth the active org silently reverts to the first one, losing working context. |
| 4 | /signin | Raw **`Request failed with status code 429`** shown to the user when rate limited. |
| 5 | /signup | **No required-field validation.** An entirely empty form submits. |
| 6 | /signin, /signup, /forgot-password | **No mobile layout.** The auth grid is hard-coded to `1.06fr 1fr` with no media query; at 375px the email input sits at x roughly 941px, off screen. Unusable on phones. |
| 7 | /signin | Brand tagline differs per entry URL: ANIMAL vs PATIENT vs COMPANION health. |

### High

- Email field forwards leading/trailing whitespace unmodified to the auth API.
- "New appointment" button does nothing in List view (the `<dialog>` stays `open=false`); works in calendar view.
- Inventory shows **negative available stock** (`0 on hand, -10 available`).
- Inventory allows a product priced below cost (-20% margin) to save.
- Finance header reads `$0 collected this week` while the table shows multiple PAID invoices.
- Patient "Horsee · Parmar" is a Horse with the cat breed `Abyssinian x`.
- Session timeout can auto-submit a stale empty form with no "you have been signed out" indicator.
- Sign-in tagline inconsistency (see critical 7) also affects the "Pet business" tab, shown as "Patient business".

### Medium

- Sidebar label flips between **"Pets" and "Patients"** depending on org, same `/companions` route.
- All 6 team members permanently show **OFF DUTY**.
- Date picker resets to today on every visit; no URL persistence of the selected date.
[...]

PR #3023, opened 2026-09-11
fix(frontend): three real accessibility gaps found triaging the QA pass and react-doctor

aupyay:
## What changed

Triaged every error/high-value warning from a fresh react-doctor scan (68/100, 482 issues) and the Aug 13 QA pass (#2168). Most turned out to be false positives once actually read against the surrounding code:

- `aria-role` (4 occurrences): `role` is a custom prop on unrelated components (a job-title string rendered as plain text), not a DOM `role` attribute.
- `rules-of-hooks` (1): a hook called unconditionally as a ternary's condition, not inside a branch - cross-checked against the real `eslint-plugin-react-hooks`, which reports zero issues on that file.
- `no-adjust-state-on-prop-change` (1): `Header.tsx`'s scroll-position sync - a textbook "synchronize with an external system" `useEffect`, not the "derive state from a prop" anti-pattern the rule targets.
- `control-has-associated-label` (11 of 13): properly labelled via a `Field` wrapper (`ControlledSubstanceRegister.tsx`) or a matching `htmlFor`/`id` pair (`EstimateLineRow.tsx`) - the static analyzer just can't see across the wrapper component.
- `no-redundant-roles` (5): `role="combobox"` on custom search widgets is semantically required, not redundant - a plain `type=text`/`type=search` input's implicit role isn't combobox, so removing it would break five working accessible comboboxes (including the one built in #2928).
- `label-has-associated-control` (4): all four `<label>`s correctly nest a real `<textarea>` via the shared `Textarea` component.

None of those are touched here.

## The three real ones

- **`FederationSection.tsx`**: the license-token and follow-actor inputs had only a placeholder, which isn't a persistent accessible label - it disappears once typing starts and isn't reliably announced by assistive tech. Added `aria-label` to both.
- **`Toast.tsx`**: the auth-flow error toast (`ErrorTost`, used for e.g. the 429 rate-limit message on `/signin`) was a plain `div` with no `role` or `aria-live` - a screen reader user gets no announcement at all when it appears. Added `role="alert"`.
- **`Sidebar.css`**: `--nav-active` and the global focus-visible outline colour are the exact same blue in dark mode (`#8fb6f5`), so a keyboard-focused active route's focus ring was barely distinguishable from the row's own active-state tint and icon colour. Gave `.route-active:focus-visible` its own `outline-color` (`var(--ink)`), which sits at the opposite end of the palette in both themes. Verified the resolved token values directly in a live dark-mode Storybook render (`--ink: #f4efe6` vs `--nav-active: #8fb6f5`) and confirmed the new CSS rule parses and applies.

## Testing

Guard tests added for all three (`getByLabelText` / `getByRole('alert')` / a source-text regex on the CSS rule, since jsdom doesn't run the real cascade). Mutation-checked: reverted all three fixes, confirmed the corresponding new assertions failed (2 in `Toast.test.tsx`, 2 in `FederationSection.test.tsx`, 1 in `Sidebar.test.tsx`), restored, confirmed green again.

`tsc --noEmit`: clean. `eslint`: clean on touched files (one pre-existing, unrelated warning elsewhere in the workspace).

### DIFF
--- a/apps/frontend/src/app/__tests__/components/Sidebar.test.tsx
+++ b/apps/frontend/src/app/__tests__/components/Sidebar.test.tsx
@@ -1,4 +1,6 @@
 import React from 'react';
+import { readFileSync } from 'fs';
+import { join } from 'path';
 import { fireEvent, render, screen } from '@testing-library/react';
 import '@testing-library/jest-dom';
 import Sidebar from '@/app/ui/layout/Sidebar/Sidebar';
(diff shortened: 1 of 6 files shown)


ankit-yc (approved):
Self-reviewed and mutation-checked in detail while writing this (see PR description). CI green (61 checks). Approving.

PR #3028, opened 2026-09-11
fix(frontend): show live password strength while signing up or resetting

aupyay:
## Summary

- Sign-up and reset-password only told the user their password was too weak after they hit submit — no live feedback while typing.
- Adds a live strength meter (Too weak / Weak / Fair / Strong) to `AuthPasswordField`, gated behind a new `showStrength` prop, wired on the two real new-password entry points (`SignUp`, `ResetPassword`) but not their confirm-password siblings.
- The meter is driven by the same five checks (length, lower, upper, digit, special char) the submit validation already enforces, so "Strong" means "will pass" rather than a second, unrelated notion of strength competing with the existing error copy.
- Dedupes the previously copy-pasted strong-password regex (identical in both `SignUp.tsx` and `ResetPassword.tsx`) into one shared, tested helper (`passwordStrength.ts`) both pages now call — a value-preserving refactor, not a behavior change.
- No new dependency: checked the monorepo for `zxcvbn`/similar first — none exists — so this uses the existing validation criteria rather than adding one for a small heuristic.

Part of the issue #2168 QA sweep.

## Test plan
- [x] `passwordStrength.ts` unit tests (score bands, `isStrongPassword` parity with the old regex's requirement text)
- [x] `SignUp`/`ResetPassword` tests: meter appears on the primary password field, absent on confirm-password, updates live as strength changes
- [x] Mutation-checked both the scoring bands and the `showStrength` gate (confirm-field exclusion) — reverting either breaks a real assertion
- [x] Live-verified in browser: dark and light mode, meter fills/colors correctly from "Too weak" to "Strong" on both SignUp and ResetPassword
- [x] `check-hardcoded-colours.mjs`: 0 to migrate (new styling uses only existing `var()` tokens)

### DIFF
--- a/apps/frontend/src/app/__tests__/features/auth/lib/passwordStrength.test.ts
+++ b/apps/frontend/src/app/__tests__/features/auth/lib/passwordStrength.test.ts
@@ -1,0 +1,45 @@
+import {
+  PASSWORD_STRENGTH_MAX_SCORE,
+  isStrongPassword,
+  passwordStrengthLabel,
+  scorePassword,
+} from '@/app/features/auth/lib/passwordStrength';
+
+describe('scorePassword', () => {
+  it('scores an empty password at zero', () => {
+    expect(scorePassword('')).toBe(0);
+  });
+
+  it('gives one point per missing criterion', () => {
+    expect(scorePassword('short')).toBe(1); // lowercase only, under 8 chars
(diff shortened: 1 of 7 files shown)


aupyay:
Superseded by #3030 — same content, fresh branch to escape a stuck GitGuardian incident that kept re-triggering on this branch/PR regardless of the actual literal content (see #3030's description for detail).

PR #3032, opened 2026-09-11
fix(backend): complete documented local startup config

### LINKED ISSUE #2749: The documented local setup does not boot the backend
Following the README and `apps/backend/.env.example` produces a backend that crashes on
boot, and the first crash names a variable the documentation never mentions. Four separate
gaps, filed together because they share one root cause — **the documented environment is not
the environment the code requires** — and because fixing them one at a time just moves the
crash to the next one.

Verified on `origin/dev` @ `f455595d4` by bringing the stack up from nothing on a machine
that had never run it: Postgres 17, Redis 7 and SuperTokens core in containers, an empty
database, and `.env` built strictly from `.env.example`.

---

**1. `STREAM_API_KEY` / `STREAM_API_SECRET` are required to boot and are not documented.**

`apps/backend/src/services/chat.service.ts:17` reads both at **module scope** and throws on
import, so this is a hard boot failure rather than a lazy failure at first chat use:

```
Error: Stream Chat credentials missing in env
    at chat.service.ts:17
```

`grep -c '^STREAM' apps/backend/.env.example` → **0** (the file is 163 lines).

Worse than absent: `apps/frontend/.env.example:31` has `NEXT_PUBLIC_STREAM_API_KEY`, so a
contributor grepping the repo for the name in the crash finds a **different** variable — the
browser-side publishable key — and no server secret at all.

A contributor who does not intend to use chat still cannot start the server, and the error
does not tell them where to get the value or that it is optional in spirit.

**2. Redis is required at boot and has no documented variables.**

The README states the backend "needs a reachable PostgreSQL database on boot and Redis for
the background job queues." `apps/backend/src/queues/bull.config.ts:4-6` reads `REDIS_HOST`,
`REDIS_PORT` and `REDIS_PASSWORD`. A case-insensitive grep for `redis` in `.env.example`
returns **0**.

Because it defaults to `127.0.0.1:6379`, a wrong or absent Redis surfaces as a repeating
`Error: connect ECONNREFUSED 127.0.0.1:6379` rather than as a missing-configuration error —
**the failure does not name its own cause**, which is the expensive part.

**3. Four SMTP variables are hard-required at boot even with a log mail driver.**

`packages/auth/src/config/smtp.config.ts` calls `requireEnv` for `SMTP_HOST`, `SMTP_USER`,
`SMTP_PASSWORD` and `SMTP_FROM_EMAIL`. Setting `MAIL_DRIVER=log` does not avoid this, so a
local run cannot skip mail configuration entirely.

**4. `DIRECT_URL` is undocumented and `prisma migrate deploy` fails without it.**

`packages/database/prisma/schema.prisma:13` declares `directUrl = env("DIRECT_URL")`. With
only `DATABASE_URL` set, migration fails at schema validation before touching the database.
Neither `DIRECT_URL` **nor `DATABASE_URL`** appears in `apps/backend/.env.example` — the
database connection string the README tells you the backend needs on boot is not in the file
the README tells you to copy.

**5. SuperTokens must not share the application's database.**

SuperTokens creates its own tables on first boot. Pointing it at the application database
makes the app's `prisma migrate deploy` fail:

```
P3005  database schema is not empty
```

The app's migration chain requires an empty database and SuperTokens gets there first. The
error names Prisma's state, not the cause, so it reads as a broken migration chain rather
than as two services sharing one database. Found by Claude L2 Yosemite Crew independently
while following items 1-4.

---

### Suggested fix

Add the missing variables to `apps/backend/.env.example` with comments saying which are
required at boot and which are integration credentials. Where a credential exists only to
talk to a third party, consider deferring the throw to first use so the server starts
without it — `chat.service.ts` is the clearest case, since a local contributor working on
appointments has no reason to hold Stream credentials.

### Notes

Once past these four, everything works: 168 migrations replay clean from an empty
Postgres 17, the server listens, the scheduled workers run, and signup returns a session.
A working recipe exists and can be folded into the README or a `docs/` page if wanted.

A verified working recipe covering all five items exists and has since been used to bring the
API up independently by a second person. It is a workaround, not the fix — the fix is that
`.env.example` and the README should be sufficient on their own, and a recipe kept anywhere
other than the repository helps the people who already have it and no new contributor.

---

*Reported by Claude L3 Yosemite Crew from a from-nothing bring-up; item 5 found independently by
Claude L2 Yosemite Crew. Items 1, 2 and 4 re-verified against `origin/dev` before filing —
`chat.service.ts:16-18` throws at module scope, and the backend `.env.example` contains zero
occurrences of `STREAM_API_KEY`, `STREAM_API_SECRET`, `REDIS_HOST`, `REDIS_PORT`,
`REDIS_PASSWORD`, `DIRECT_URL` or `DATABASE_URL`.*

aupyay:
## PR Checklist

The checked-in backend environment example omits the application database, Redis, and Stream server variables used by the documented local workflow. Importing chat services also throws immediately when optional Stream credentials are absent, preventing the API module from loading before any chat operation is requested.

## What is the new behavior?

- Documents the required PostgreSQL and Redis variables and the optional Stream pair in the backend example.
- Documents separate application and SuperTokens databases.
- Resolves the Stream server client only when a chat or webhook operation needs it, so non-chat local API work can boot without Stream credentials.
- Adds regression coverage for import-time behavior, incomplete configuration, successful client creation, and the environment-example contract.

## Related Issue(s)

Fixes #2749

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,6 +288,8 @@ cp apps/frontend/.env.example apps/frontend/.env
 
 Configure PostgreSQL, Redis, authentication, and the integrations you will use. Make `DATABASE_URL` and `DIRECT_URL` available in the shell that runs Prisma commands; app-specific `.env` files are not automatically loaded by every workspace command. Never use production credentials for local setup.
 
+Use separate PostgreSQL databases for the application and a self-hosted SuperTokens core. Stream chat is optional for local work: leaving its two server credentials blank keeps the API bootable and records the upload-policy control as skipped, while chat operations remain unavailable until configured.
+
 Generate the Prisma client:
 
 ```shell
--- a/apps/backend/.env.example
+++ b/apps/backend/.env.example
@@ -2,6 +2,23 @@ PORT="4000"
 APP_NAME="Yosemite"
 NODE_ENV="development"
 
+# --- Core services ----------------------------------------------------------
+# Application database. A self-hosted SuperTokens core must use a separate
+# database so its tables do not make the Prisma migration baseline non-empty.
+DATABASE_URL=""
+DIRECT_URL=""
+# Redis is required by the BullMQ job queues. Leave the password commented when
+# the local server does not require authentication.
+REDIS_HOST="127.0.0.1"
+REDIS_PORT="6379"
+# REDIS_PASSWORD=""
+
+# --- Optional Stream chat ---------------------------------------------------
+# Leave both blank when not working on chat. The API still starts and reports
+# the upload-policy control as skipped; chat operations require both values.
+STREAM_API_KEY=""
+STREAM_API_SECRET=""
+
 # --- Auth (provider-neutral boundary, #1672) -------------------------------
 # Provider selection; supertokens is the only implemented provider.
 AUTH_PROVIDER="supertokens"
--- a/apps/backend/README.md
+++ b/apps/backend/README.md
@@ -27,6 +27,8 @@ pnpm --filter backend run dev
 
 `LOCAL_DEVELOPMENT=true` switches on local-only behaviour: it opens CORS to `localhost:3000` (`src/app.ts`) and mounts the local-only MFA debug endpoint `POST /v1/auth/mfa/totp/debug/create-device`, which creates a TOTP device without the full enrolment flow. Both are keyed on this flag rather than on `NODE_ENV`, so a deployed tier running `NODE_ENV=development` never gets them. Set it only for a local run.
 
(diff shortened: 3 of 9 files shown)


ankit-yc (approved):
Reviewed exact head d24ba8159: local full backend suite, build, type-check, lint, mutation checks, and hosted security gates are green. Root-cause fix keeps Stream optional at API boot without weakening chat-time validation.

PR #3034, opened 2026-09-11
fix(backend): normalize chat profile display names

### LINKED ISSUE #3033: Chat profiles can sync as undefined undefined when names are missing
## Problem

Four chat creation paths build a Stream user name with this expression:

```ts
user?.firstName + " " + user?.lastName || "User"
```

When the profile lookup returns no user, JavaScript produces the truthy string `"undefined undefined"`, so the fallback never runs. A partially populated profile similarly produces names such as `"Jane undefined"`.

## Current locations

- `apps/backend/src/services/chat.service.ts` (three `upsertUser` calls)
- `apps/backend/src/services/networkChat.service.ts` (one `upsertUser` call)

## Expected

Join only non-empty name parts and use the existing `"User"` fallback when neither part is present. Keep this separate from the local-startup fix in #2749.

## Acceptance criteria

- Full names remain unchanged.
- A single available name is used without an `undefined` suffix.
- Missing or blank names use `"User"`.
- Chat and network-chat paths share the same behavior and have regression coverage.

aupyay:
## PR Checklist

Missing or partial user names are concatenated directly before Stream profile upserts, producing truthy values such as `undefined undefined`, `Jane undefined`, or `undefined Doe` instead of the intended fallback. The touched chat service also accepted a runtime session identifier without validating it before a Prisma delete lookup.

## What is the new behavior?

All four chat profile upserts share one formatter that joins only populated name parts, trims whitespace, and falls back to `User`. The group-delete service validates and casts its session identifier before querying. Regression tests cover same-clinic and cross-clinic name handling plus the rejected injection-shaped identifier input.

## Related Issue(s)

Fixes #3033

### DIFF
--- a/apps/backend/src/services/chat.service.ts
+++ b/apps/backend/src/services/chat.service.ts
@@ -2,6 +2,7 @@
 import { ChannelData } from "stream-chat";
 import dayjs from "dayjs";
 import crypto from "node:crypto";
+import { z } from "zod";
 
 import { ChatSessionDocument, ChatSessionType } from "../models/chatSession";
 import { AppointmentDocument } from "../models/appointment";
@@ -42,6 +43,11 @@ type YosemiteChannelResponse = ChannelData & {
   name?: string;
   isPrivate?: boolean;
 };
+
+export const chatUserDisplayName = (
+  user?: { firstName?: string | null; lastName?: string | null } | null,
+) =>
+  [user?.firstName, user?.lastName].filter(Boolean).join(" ").trim() || "User";
 
 export class ChatServiceError extends Error {
   constructor(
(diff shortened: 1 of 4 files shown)


ankit-yc (dismissed):
Reviewed exact head 4030b63b2: all four Stream upserts share the same trimmed fallback behavior, and the delete path validates a scalar session ID before Prisma. The focused regressions and both mutation arms exercise the changed branches.

ankit-yc (approved):
Reviewed exact head 376447ce2: Stream display names remain normalized through the shared helper, and deleteGroup now validates the UUID-backed session key before either Prisma query. The focused suite covers valid, malformed-string, and non-string inputs, with mutation failures confirmed.

PR #3035, opened 2026-09-11
fix(ci): cutover exec-redirect guard covers sourced lib files and fd forms

### LINKED ISSUE #2766: The cutover-annotation stderr guard reads a file, but a shell redirect's scope is the process
Follow-up from #2765. **Not urgent and arguably not worth doing** — filed so the measurement
outlives the review thread, and so whoever reads the guard next does not have to re-derive it.

## What the guard checks today

`scripts/deploy/tests/controls.test.sh` asserts no cutover annotation is redirected away from
stdout. Four spellings are caught: trailing `>&2`, leading `>&2`, `{ … } >&2`, `( … ) >&2`. A
fifth row catches `exec >&2` over `api-deploy.sh`, because a shell redirect is not a *site* and
every per-site guard stays green for it.

## Three that still evade, measured

```
plain echo (CONTROL)   stdout=[::error::X]
exec >&2               stdout=[]   caught
exec 1>&2              stdout=[]   46 passed   numbered fd
exec >/dev/stderr      stdout=[]   46 passed   path form
exec >&2 in lib/*.sh   stdout=[]   46 passed   SOURCED into the same shell
```

The third is the one that is a **scope** error rather than a spelling. `api-deploy.sh:71,73,75`
sources `lib/git-sync.sh`, `lib/migrate.sh` and `lib/controls.sh`; a redirect in any of them
applies from that point onward in the same shell. So the guard reads the file that would be
*affected* rather than every file that can *cause* it — the same mistake as the earlier version
that read only the cutover region, moved out by one. **A redirect's scope is the process, not a
region and not a file.**

## The one-line version, if anyone wants it

```
grep -cE '^[[:space:]]*exec[[:space:]]+[0-9]*[><]'   over api-deploy.sh + lib/*.sh

                   shipped   widened
exec >&2              1         1
exec 1>&2             0         1
exec >/dev/stderr     0         1
exec >&2 in lib/      0         1
pristine              0         0
```

Stated cost: it also flags fd juggling that never touches stdout, `exec 3>&1` being the obvious
one. Measured on `dev` at `1ebbb8540` — **zero `exec` redirects in all four files**, so this is
latent, not live:

```
api-deploy.sh:0   lib/git-sync.sh:0   lib/migrate.sh:0   lib/controls.sh:0
```

## Why it was deliberately not done in #2765

Raised in review and I agree with the judgement: this would be a third layer of guard protecting
an annotation **whose primary mechanism is still unverified** — nobody has yet observed a
workflow command written over an ssh connection being parsed at all (#2764). Three spellings of a
line nobody has written are a smaller risk than the one already accepted and documented.

Worth doing only if someone is in that file anyway, or if #2764's first dispatch confirms the
annotations work and they become something worth protecting properly.

aupyay:
## Summary
- `controls.test.sh`'s exec-redirect guard only read `api-deploy.sh` and only matched the exact spelling `exec >&2`, so it missed the numbered-fd form (`exec 1>&2`), the path form (`exec >/dev/stderr`), and any `exec` redirect placed in a sourced lib file (`lib/git-sync.sh`, `lib/migrate.sh`, `lib/controls.sh`) — all of which apply to the same shell once sourced and would silently kill the cutover annotations.
- Widened the pattern to `exec [fd]?[><]` and scan `api-deploy.sh` plus `lib/*.sh` together (comment-blanked, same as the existing `DEPLOY_CODE` handling).
- Latent, not live: `grep`-verified zero `exec` redirects in any of the four files today, so this closes a gap without changing current behavior.

Fixes #2766.

## Test plan
- [x] `bash scripts/deploy/tests/controls.test.sh` — 46 passed, 0 failed
- [x] Mutation check: injected each of the three previously-missed spellings (`exec 1>&2`, `exec >/dev/stderr`, `exec >&2`) into `lib/controls.sh` one at a time and reran the real suite — each now fails `the deploy script does not redirect the whole shell` (`expected '0 shell redirects', got '1 shell redirects'`); reverting restores 46/0.

### DIFF
--- a/scripts/deploy/tests/controls.test.sh
+++ b/scripts/deploy/tests/controls.test.sh
@@ -661,8 +661,25 @@ check "no annotation is redirected to stderr by an enclosing group" \
 # Driven both ways: inside the region 1 red, above it 0 red under the regional
 # version. There is no `exec` anywhere in api-deploy.sh or lib/ today, so a
 # whole-file check has nothing to false-positive on.
-DEPLOY_EXEC_REDIRECT="$(grep -cE '^[[:space:]]*exec[[:space:]]+>&?[12&]' \
-  <<< "$DEPLOY_CODE" || true)"
+#
+# The scope is the SHELL, not the file (#2766): api-deploy.sh sources
+# lib/git-sync.sh, lib/migrate.sh and lib/controls.sh into that same shell (see
+# the `.` lines above `say()`), so an `exec` written in any of them is exactly
+# as effective as one in api-deploy.sh itself, and a check that only reads
+# DEPLOY_CODE misses it by one hop. Blanked the same way as DEPLOY_CODE, for
+# the same reason: a paragraph quoting the pattern must not redden its own
+# guard.
+#
+# The pattern also widened from `>&?[12&]` to `[0-9]*[><]`: the old one only
+# matched `exec >&2` and missed two spellings with the identical effect on
+# stdout - the numbered-fd form (`exec 1>&2`) and the path form
+# (`exec >/dev/stderr`). Stated cost, unchanged from #2766: this also flags fd
+# juggling that never touches stdout, `exec 3>&1` being the obvious one - a
+# false positive judged cheaper than the false negative it replaces.
+LIB_CODE="$(for f in "$HERE"/../lib/*.sh; do sed -E 's/^[[:space:]]*#.*$//' "$f"; done)"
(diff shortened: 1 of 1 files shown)


ankit-yc (approved):
Widened exec-redirect scan matches the mechanism (redirect scope is the shell, not the file) and the fix is mutation-checked: all three previously-missed spellings (exec 1>&2, exec >/dev/stderr, exec >&2 in a sourced lib file) verified to fail the real suite, reverting restores 46/0. All CI green.

PR #3036, opened 2026-09-11
fix(backend): prevent appointment pagination gaps

### LINKED ISSUE #2722: developer-data appointments page with cursor + skip: 1, which silently drops a row when the cursor row leaves the filter
## What

`developerDataService.listAppointments` pages with Prisma's `cursor` plus `skip: 1`:

```ts
orderBy: [{ appointmentDate: "asc" }, { id: "asc" }],
take: query.limit + 1,
...(query.cursor ? { cursor: { id: query.cursor }, skip: 1 } : {}),
```

That pair is exclusive **only while the cursor row is still in the filtered result**. `cursor` seeks to the row; `skip` is an `OFFSET` applied to the *filtered* result. When the cursor row has left the filter between one page and the next, the `OFFSET` steps past a legitimate row instead, and it is not reported — the caller is handed a page that silently omits it.

This is the same defect that was found and fixed on the owner prescription list in #2720. **This module was not changed by that PR** and it has four live callers, so it is filed separately rather than as a rider.

## Why it is reachable here, and more easily than it was there

The `where` this service builds is caller-controlled and filters on values that change during normal business:

```ts
if (query.from || query.to) { where.appointmentDate = { gte: query.from, lte: query.to }; }
if (query.status)          { where.status = query.status; }
```

- **`status`** — an appointment cancelled, completed or checked in between page 1 and page 2 leaves a `status`-filtered result.
- **`appointmentDate`** — a reschedule moves an appointment out of a `from`/`to` window, and also moves it within the sort order.

Both are ordinary front-desk actions on a busy day, which is exactly when a client is paging.

## Evidence for the mechanism

Reproduced on the prescription list against a real PostgreSQL under #2720 — seven rows, page size three, take the cursor row out of the filter between pages:

```
page: ["rx-g","rx-f","rx-e"]  hasMore=true
...rx-e leaves the filter
page: ["rx-c","rx-b","rx-a"]  hasMore=false
NEVER SHOWN : ["rx-d"]
```

**Scope: that run was against the prescription query, not this one.** The mechanism is Prisma's `cursor` + `skip` pair rather than anything specific to prescriptions, and this service uses the identical pair — but nobody has yet reproduced it against `listAppointments` itself. Doing so is the first step, not an optional extra.

## Why no test here can catch it

The service tests mock `prisma.appointment.findMany`, so they can assert that `cursor` and `skip: 1` were *sent* — and they do, correctly. A mock cannot testify about what PostgreSQL does with an `OFFSET`. The evidence has to come from outside the suite.

## Suggested fix

The one taken on #2720: make the cursor exclusive in the `where` rather than with `skip`, carrying the whole sort key.

`apps/backend/src/services/shared/pagination.ts` already has the pieces — `splitPage` takes an encoder, and `encodeKeysetCursor` / `parseKeysetCursor` exist. They are `(createdAt, id)` specific today, so this needs either a generalisation to an arbitrary first sort column or a sibling for `(appointmentDate, id)`. Whichever is chosen, `splitPage`'s default of the row id must keep working, because that is what makes the change to one surface not a change to the other.

**This is not free the way #2720 was.** That endpoint had no consumers, so its cursor format could change silently. This one has four, so an in-flight cursor issued before the change has to either keep working or fail loudly — a cursor format change that answers 400 to a cursor a client is holding is a coordinated release, not a refactor. Decide that before writing the query.

## Also worth deciding while in here

`splitPage` returns `hasMore` and this call site drops it:

```ts
const { items, nextCursor } = splitPage(rows, query.limit);
return { items, nextCursor };
```

A client distinguishing "end of data" from "end of page" has only `nextCursor === null` to go on. That is sufficient, but it is one less signal than the owner surface gives, and it is the same class of thing — a list that cannot say whether it is complete.

aupyay:
## PR Checklist

Appointment pagination uses Prisma's row cursor with `skip: 1`. If the cursor appointment changes status or date between requests and no longer matches the active filters, Prisma resolves the cursor before filtering and the skip removes the first valid row from the next page.

## What is the new behavior?

- Encodes the appointment date and id in the opaque next-page cursor and applies an exclusive `(appointmentDate, id)` boundary inside the active filters.
- Accepts in-flight UUID cursors through a practice-scoped compatibility lookup.
- Returns a clear 400 response for unknown or expired legacy cursors without turning unrelated database failures into client errors.
- Covers filter mutation, tied dates, cursor compatibility, tenant scoping, and controller behavior.

Validation:
- Reproduced the dropped row against PostgreSQL 16 before the fix and verified every remaining row is returned after the fix.
- Focused Jest: 41 tests passed with 97.6% statement and 90.41% branch coverage across the changed production files.
- Full backend Jest: 491 suites and 11,679 tests passed; the runner reported a lingering async handle after its final summary.
- Backend and monorepo lint/type-check gates passed; backend build passed.
- Six targeted mutations each failed their intended regression test.
- Aikido scan found no issues in the four changed files; staged gitleaks and secretlint checks passed.

## Related Issue(s)

Fixes #2722

### DIFF
--- a/apps/backend/src/controllers/web/developer-data.controller.ts
+++ b/apps/backend/src/controllers/web/developer-data.controller.ts
@@ -21,8 +21,9 @@ import {
 import {
   clampPageSize,
   DeveloperDataService,
+  parseAppointmentCursor,
+  UnknownAppointmentCursorError,
 } from "../../services/developer-data.service";
-import { parseUuidCursor } from "../../services/shared/pagination";
 import { DeveloperUsageService } from "../../services/developer-usage.service";
 
 type ErrorCode =
@@ -134,7 +135,7 @@ export const DeveloperDataController = {
      */
     const rawCursor =
       typeof req.query.cursor === "string" ? req.query.cursor : "";
-    const cursor = parseUuidCursor(req.query.cursor);
+    const cursor = parseAppointmentCursor(req.query.cursor);
     if (cursor === null) {
       // Logged newline-stripped: the value is caller-controlled and a raw CR/LF
       // in it would forge a second log line.
@@ -164,6 +165,15 @@ export const DeveloperDataController = {
         pagination: { limit, nextCursor: page.nextCursor },
       });
     } catch (err) {
+      if (err instanceof UnknownAppointmentCursorError) {
+        fail(
+          res,
+          400,
+          "invalid_request",
+          "Unknown or expired cursor. Use pagination.nextCursor from the previous response.",
+        );
+        return;
+      }
       logger.error("DeveloperDataController.listAppointments failed", err);
       fail(res, 500, "internal_error", "Internal server error");
     }
--- a/apps/backend/src/services/developer-data.service.ts
+++ b/apps/backend/src/services/developer-data.service.ts
@@ -12,6 +12,9 @@ import type { AppointmentStatus, Prisma } from "@prisma/client";
 import { prisma } from "src/config/prisma";
 import {
   clampPageSize as clampToBounds,
+  encodeKeysetCursor,
+  parseKeysetCursor,
+  parseUuidCursor,
   splitPage,
 } from "src/services/shared/pagination";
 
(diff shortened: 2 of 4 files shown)


ankit-yc (approved):
Reviewed the exact head for keyset boundary correctness, legacy cursor compatibility, tenant scoping, and error classification. Focused regression and mutation coverage exercises each branch. LGTM.

ankit-yc (approved):
Reviewed exact rebased head 42374910c: the diff is unchanged, the keyset boundary remains filter-scoped and tenant-safe, and all 41 focused tests pass on the current dev base. LGTM.

PR #3038, opened 2026-09-11
fix(backend): suppress terminal companion history cursor

### LINKED ISSUE #3037: fix(backend): suppress terminal companion history cursor
### Preflight

### Affected area

Backend API

### What is wrong

`CompanionHistoryService.listForCompanion` sets `nextCursor` whenever the returned page length equals the requested limit. It does not check whether a row exists beyond that page.

When the complete filtered history contains exactly `limit` entries, the response therefore advertises a continuation cursor even though the next request returns an empty page.

### Steps to reproduce

1. Seed or mock exactly two visible companion-history entries.
2. Call the companion-history endpoint with `limit=2`.
3. Observe two returned entries and a non-null `nextCursor`.
4. Request the next page with that cursor and observe zero entries.

### Expected behavior

`nextCursor` is non-null only when at least one additional entry exists beyond the returned page.

### Actual behavior

`nextCursor` is non-null whenever the current page is full, including the terminal exact-limit page.

### Suggested fix

Measure `filtered.length > limit` before slicing and only build a cursor when that extra row exists. Add a regression test for exactly `limit` entries and mutation-check the comparison.

### Impact

Low. API consumers perform an unnecessary request and may render a misleading load-more state at an exact page boundary.

aupyay:
## PR Checklist

Companion history returns a continuation cursor whenever a page contains exactly the requested number of entries, even when no additional entry exists.

## What is the new behavior?

The service returns a cursor only when the filtered result contains an entry beyond the current page. A regression test covers an exact-limit terminal page.

Validation:

- Companion history service suite: 17 tests passed.
- Deliberate comparison mutation: the regression test failed as expected.
- Backend lint and type-check completed successfully.
- Repository pre-push lint and type-check completed successfully.

## Related Issue(s)

Fixes #3037

### DIFF
--- a/apps/backend/src/services/companion-history.service.ts
+++ b/apps/backend/src/services/companion-history.service.ts
@@ -781,7 +781,7 @@ export const CompanionHistoryService = {
 
     const paged = filtered.slice(0, limit);
     const nextCursor =
-      paged.length === limit ? buildCursor(paged.at(-1)!) : null;
+      filtered.length > limit ? buildCursor(paged.at(-1)!) : null;
 
     const countsByType = paged.reduce(
       (acc, entry) => {
--- a/apps/backend/test/services/companion-history.service.test.ts
+++ b/apps/backend/test/services/companion-history.service.test.ts
@@ -123,6 +123,41 @@ describe("CompanionHistoryService", () => {
     expect(secondPage.entries[0].link.id).toBe("apt-2");
   });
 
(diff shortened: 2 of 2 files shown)


ankit-yc (approved):
Reviewed: one-line off-by-one fix (paged.length === limit → filtered.length > limit) correctly distinguishes "page exactly filled because more exist" from "page exactly filled because that's the whole result". Regression test covers the terminal-page case, mutation-checked per the PR description. Clean.

PR #3041, opened 2026-09-11
fix(backend): dedupe the two-spellings-per-breed vocabulary entries

### LINKED ISSUE #2450: fix(database): the same breed exists under two code conventions
### Affected area

Backend / database — the `CodeEntry` breed vocabulary and `Patient.breedCode`

### What is wrong

The same breed exists under two separator conventions, in both the vocabulary and the patient rows that reference it.

`CodeEntry` holds **1,749** breed codes that collapse to **1,713** distinct ones once hyphens and underscores are folded together. **36 breeds exist under both spellings**:

```
YBREED:CANINE:SHIH_TZU        YBREED:CANINE:SHIH-TZU
YBREED:CANINE:LHASA_APSO      YBREED:CANINE:LHASA-APSO
YBREED:CANINE:ALASKAN_MALAMUTE YBREED:CANINE:ALASKAN-MALAMUTE
YBREED:EQUINE:APPALOOSA_X     YBREED:EQUINE:APPALOOSA-X
...
```

Patient rows are split the same way. Of the seven coded companions in production, **four are hyphenated and three are not**:

```
YBREED:CANINE:AFRICAN-WILD-DOG   (hyphen)
YBREED:CANINE:LABRADOR_RETRIEVER (underscore)
```

The backend's own generator, `buildBreedCode` in `apps/backend/src/services/idexx-reference.service.ts`, emits underscores. Where the hyphenated codes came from is not established.

### Steps to reproduce

```sql
with norm as (
  select code, replace(code, '-', '_') as canonical
  from "CodeEntry" where code like 'YBREED:%'
)
select count(*) as total,
       count(distinct canonical) as distinct_after_normalising
from norm;
-- 1749, 1713
```

### Expected behavior

One code per breed.

### Actual behavior

Two codes for 36 breeds, and patients distributed across both.

Always — it is the current state of the data.

### Did this work before

Unknown. Both conventions are already present; nothing in the schema prevents it.

### Impact

Any feature that compares a breed code literally will silently return nothing for whichever half of the population is coded the other way. **Silently** is the problem: a missing result is indistinguishable from "no data for this breed".

This was found while building the recommendation rules in #2449, which works around it by comparing codes canonically everywhere — so that feature is not blocked. But the workaround lives in one service. Anything else that joins on `breedCode` — reporting, PIMS filters, IDEXX reference ranges, an export — has the same exposure and no such guard.

`Patient.breedCode` is also a bare string with no foreign key to `CodeEntry`, so nothing rejects a code that does not exist in the vocabulary at all.

### Environment

Production, verified 2026-08-24.

### Additional context

Two decisions are needed before a fix:

1. **Which convention wins.** Underscore matches the backend generator; hyphen is the larger share of the vocabulary (1,242 vs 261, with 246 single-token codes unaffected).
2. **Whether to enforce it.** A canonicalising helper already exists at `apps/backend/src/services/shared/breed-code.ts` from #2449 and could become the single writer. A foreign key from `Patient.breedCode` to `CodeEntry.code` would make an unknown code impossible rather than merely unmatched, but needs the duplicates resolved first.

A migration would need to repoint the patient rows as well as deduplicate the vocabulary, and both are small at current volumes — seven patient rows and 36 breed pairs.

aupyay:
Fixes #2450.

`CodeEntry` holds 36 breeds under both a hyphen and an underscore spelling (`YBREED:CANINE:SHIH_TZU` vs `YBREED:CANINE:SHIH-TZU`); the seven already-coded patients in production are split across both. Anything that joins on `breedCode` literally - reporting, PIMS filters, IDEXX reference ranges, an export - silently misses whichever half of the population is coded the other way.

## What this does

Adds `dedupe-breed-codes.ts`, mirroring `backfill-breed-codes.ts`'s dry-run / `--apply` convention:

- Groups every `BREED` `CodeEntry` by its canonical form (`canonicalBreedCode`, already shipped for #2449).
- The winner in each duplicate group is the code that already equals its own canonical form - underscore wins, since the live `buildBreedCode` generator in `idexx-reference.service.ts` has always emitted underscores. A pair where neither (or both) sides already match is reported **unresolved**, not guessed at.
- On `--apply`: `Patient.breedCode` rows and `CodeMapping` rows pointing at the losing code are repointed to the winner - or the mapping is dropped instead, if the winner already has an equivalent one, to avoid colliding with the `(sourceSystem, sourceCode, targetSystem, targetCode)` unique constraint. The losing `CodeEntry` is deactivated (`active: false`), not deleted, matching how the rest of the vocabulary already treats inactive rows as invisible.

## Deliberately not included

Both are open decisions the issue itself flags, not part of "one code per breed":

- A foreign key from `Patient.breedCode` to `CodeEntry.code` - this fix is the prerequisite (duplicates resolved), but enforcing it is a separate call.
- Making `canonicalBreedCode` the single write path everywhere a breed code is created - the only live writer (`buildBreedCode`) already emits the canonical form; nothing today writes a fresh hyphenated code.

## Testing

- `npx jest test/scripts/dedupe-breed-codes.test.ts` - 10/10 pass, mutation-checked (broke winner-selection and the mapping-dedup branch, both caught by the suite, then reverted).
- Full backend suite: `npx jest` - 492 suites / 11693 tests pass.
- `npx tsc --noEmit` clean, `npx eslint .` clean (0 errors) after building the workspace packages.

### DIFF
--- a/apps/backend/src/scripts/dedupe-breed-codes.ts
+++ b/apps/backend/src/scripts/dedupe-breed-codes.ts
@@ -1,0 +1,189 @@
+/**
+ * Collapse the breed vocabulary's two-spellings-per-breed duplicates onto one code.
+ *
+ * `CodeEntry` holds the same breed under both separator conventions for 36
+ * breeds - `YBREED:CANINE:SHIH_TZU` and `YBREED:CANINE:SHIH-TZU` are two rows
+ * for one breed, and the seven already-coded companions in production are
+ * split across both. Anything that joins on `breedCode` literally - reporting,
+ * PIMS filters, IDEXX reference ranges, an export - silently misses whichever
+ * half of the population is coded the other way (#2450).
+ *
+ * Underscore wins. `buildBreedCode` in idexx-reference.service.ts, the only
+ * live writer of new breed codes, has always emitted underscores; keeping
+ * hyphen as canonical would leave that generator producing a "new" duplicate
+ * against it forever.
+ *
+ * A loser is DEACTIVATED, never deleted - the rest of the codebase already
+ * treats `active: false` as invisible (every vocabulary query filters on it,
+ * same as `backfill-breed-codes.ts`), so this reaches "one code per breed"
+ * without destroying the audit trail of where the row came from.
+ *
+ * Dry run by default. Pass --apply to write:
+ *
+ *   pnpm --filter backend exec tsx src/scripts/dedupe-breed-codes.ts
+ *   pnpm --filter backend exec tsx src/scripts/dedupe-breed-codes.ts --apply
+ */
+import { prisma } from "src/config/prisma";
+import { canonicalBreedCode } from "src/services/shared/breed-code";
+
+export interface DedupeGroup {
+  canonical: string;
+  winner: string;
+  losers: string[];
+}
+
+export interface UnresolvedGroup {
+  canonical: string;
+  codes: string[];
+  reason: string;
+}
+
+export interface DedupePlan {
+  groups: DedupeGroup[];
+  unresolved: UnresolvedGroup[];
+}
+
+/**
+ * Group every breed code by its canonical form and split each group into a
+ * winner - the one spelling that already equals its own canonical form - and
+ * the rest. A group where zero or more than one code already matches the
+ * canonical form is reported unresolved rather than guessed at: a wrong
+ * repoint is worse than a duplicate left in place.
+ */
+export const planDedupe = async (): Promise<DedupePlan> => {
+  const entries = await prisma.codeEntry.findMany({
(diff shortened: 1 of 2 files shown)


ankit-yc (approved):
Reviewed the dedupe script and its test suite: winner-selection logic is sound (only a code that already equals its own canonical form wins, ties/no-matches reported unresolved rather than guessed), mapping repoint vs dedupe-delete correctly avoids the unique-constraint collision, and the in-memory winnerTargets set correctly handles a second loser mapping to a target the first loser's repoint just created. Dry-run by default. 10/10 tests, mutation-checked, tsc/eslint clean, full backend suite green. CI clean.

PR #3042, opened 2026-09-11
fix(backend): quarantine idexx results without ids

### LINKED ISSUE #2717: IDEXX results with no resultId all upsert onto one LabResult row
Found by @aupyay reviewing #2716. Pre-existing on `dev`, untouched by that PR, and reproduced independently before filing.

## What is wrong

`processIdexxResult` coerces the provider's result id with `coerceStringOrEmpty`, so a result that arrives without one becomes `""`:

```ts
const resultId = coerceStringOrEmpty(result.resultId);
await upsertLabResult(result, resultId, context.organisationId);
```

`upsertLabResult` then upserts on `provider_resultId`, and `LabResult` carries `@@unique([provider, resultId])`. `""` is a perfectly good key, so **every result that arrives without an id upserts onto the same row**.

Reproduced through `pollLatest` with two results, two different `orderId`s, two different patients, neither carrying a `resultId`:

```
labResult.upsert wheres: [{"provider_resultId":{"provider":"IDEXX","resultId":""}},
                          {"provider_resultId":{"provider":"IDEXX","resultId":""}}]
```

Identical keys. The second `update` overwrites the first, so one result is gone — and because `update` applies that result's whole `basePayload`, the surviving row carries the **second** order's `organisationId`. The first result is therefore not merely overwritten, it is gone from its own tenant's reads.

## Why it is worth a row rather than a shrug

`LabResultQuarantine` (#2716) was given no unique key precisely so that two ids-less rows in one batch cannot be collapsed onto one. The table holding the actual clinical result does exactly that, one model over. The design of the new table assumes a missing `resultId` is possible; if that assumption is right, this is a silent loss of a lab result.

## Scope of the claim

- Pre-existing on `dev`; not introduced or touched by #2716.
- **Not established: whether IDEXX ever omits `resultId` in practice.** Nobody has seen it. What is established is that the code coerces the absence into a valid key rather than refusing it, so the outcome is guaranteed *if* it ever happens, with no signal when it does.
- Only the IDEXX path was traced. Other writers of `LabResult` were not checked.

## Options

1. **Refuse rather than coerce.** A result with no id cannot be keyed, so skip the upsert and quarantine it via the #2716 path with a new reason (`MISSING_RESULT_ID`). Fail-closed, visible, replayable, and reuses machinery that now exists.
2. **Synthesise a stable key** from `orderId` + `accessionId` + `updatedDate`. Keeps the row but invents an id, and two genuinely distinct results could still collide.
3. **Make `resultId` nullable and drop it from the unique key.** Largest change, and `@@unique` on a nullable column behaves differently again.

Option 1 looks right, and it is small now that the quarantine table exists.

## Related

- #2716 — the quarantine, whose test comment argues the opposite convention one table over
- #2708 — the outage the quarantine fixes

aupyay:
## PR Checklist

IDEXX results without a usable `resultId` are coerced to an empty string and upserted through the shared `(provider, resultId)` unique key. Multiple id-less results therefore overwrite one `LabResult`, including its organization assignment.

## What is the new behavior?

Results with an absent or blank `resultId` skip the `LabResult` upsert and artifact creation. Each one is recorded independently in the existing lab result quarantine with reason `MISSING_RESULT_ID`, preserving its provider payload and organization context before the IDEXX batch is confirmed. Existing unmapped-status quarantine behavior now carries its reason per result through the same transaction.

The focused IDEXX service suite covers two id-less results from different organizations and asserts that neither reaches the shared upsert key, both quarantine rows survive, and batch ingestion advances. Disabling the missing-ID guard makes this test fail.

Validation:

- Targeted IDEXX service suite: 11 tests passed.
- Targeted service coverage: 98.3% statements and lines, 75% branches, 100% functions.
- Backend TypeScript check passed.
- Full pre-push lint and type-check gates passed across the monorepo.

## Related Issue(s)

Fixes #2717

### DIFF
--- a/apps/backend/src/services/idexx-results.service.ts
+++ b/apps/backend/src/services/idexx-results.service.ts
@@ -39,6 +39,11 @@ const coerceString = (value: unknown): string | null => {
 const coerceStringOrEmpty = (value: unknown): string =>
   coerceString(value) ?? "";
 
+const coerceNonBlankString = (value: unknown): string | null => {
+  const coerced = coerceString(value);
+  return coerced?.trim() ? coerced : null;
+};
+
 /**
  * `organisationId` is the tenant key that lab-result reads authorize on, so it is taken from
  * the LabOrder we placed rather than from the provider's response, which is outside our
@@ -305,15 +310,21 @@ const maybeCreateResultArtifacts = async (
  * reason needs no migration; every value in use is named here.
  */
 export const QUARANTINE_REASON_UNMAPPED_STATUS = "UNMAPPED_RESULT_STATUS";
+export const QUARANTINE_REASON_MISSING_RESULT_ID = "MISSING_RESULT_ID";
+
+type QuarantineReason =
+  | typeof QUARANTINE_REASON_UNMAPPED_STATUS
+  | typeof QUARANTINE_REASON_MISSING_RESULT_ID;
 
 type UnapplicableResult = {
   result: IdexxResult;
   context: ResultOrderContext;
+  reason: QuarantineReason;
 };
 
 /**
- * Record a result the mapper could not apply, so the rest of its batch can be
- * confirmed.
+ * Record a result that could not be applied safely, so the rest of its batch
+ * can be confirmed.
  *
  * IDEXX confirms a BATCH, and `pollLatest` builds ONE client from
  * `IDEXX_GLOBAL_USERNAME` with `organisationId` derived per result. So refusing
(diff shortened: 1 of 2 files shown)


ankit-yc (approved):
Reviewed the exact head 7668210fb. Missing and blank provider IDs cannot reach the LabResult upsert or artifact path; each row retains its own order-derived organization context in the existing transaction-backed quarantine. The focused test distinguishes two organizations and fails when the guard is disabled. No blocking findings.

PR #3044, opened 2026-09-11
fix(ci): recognize the FolderName.stories.tsx naming this codebase uses

aupyay:
## Summary

- `storyPathFor()` in the story-coverage gate only ever checked `<Dir>/index.stories.tsx` for a component living at `<Dir>/index.tsx`.
- This codebase genuinely uses two conventions for that case — 36 components story as `<Dir>/<Dir-basename>.stories.tsx` (measured), 21 as `<Dir>/index.stories.tsx` — so the gate flagged 32 components with a real, working story as "missing" purely because they used the other, equally-established name.
- Found while doing a full audit of the current story-coverage gap (`node scripts/ci/story-coverage.mjs --files <every .tsx file>`): 73 "missing" before this fix. Cross-checked against the actual filesystem and confirmed 32 of those had a real sibling story under the folder-basename convention. After the fix: 41 — matching the independent filesystem check exactly.
- Added `alternateStoryPathFor()`, checked alongside the primary path before declaring a component missing.

## Test plan
- [x] New tests: an `index.tsx` component with only the folder-basename story is not flagged; one with neither naming present still is
- [x] Mutation-checked: removing the alternate-path check breaks the new "not flagged" test
- [x] Full `node --test scripts/ci/story-coverage.test.mjs` suite green (21/21)
- [x] Ran the fixed gate against every current `apps/frontend` `.tsx` file and confirmed the 41-file result matches an independent Python-based filesystem cross-check exactly

### DIFF
--- a/scripts/ci/story-coverage.mjs
+++ b/scripts/ci/story-coverage.mjs
@@ -129,6 +129,12 @@ export function storyPathFor(componentFile) {
 export function alternateStoryPathFor(componentFile) {
   if (path.basename(componentFile) !== 'index.tsx') return null;
   const dir = path.dirname(componentFile);
+  // Defense in depth: evaluate()'s resolveWithin() already rejects any
+  // resulting path that escapes cwd before it is ever read, but this checks
+  // `dir` itself (not just its basename, which a `..` segment further up the
+  // path would not appear in) so this function never even constructs a path
+  // reaching outside componentFile's own directory tree.
+  if (dir.split(path.sep).includes('..') || path.isAbsolute(dir)) return null;
   return path.join(dir, `${path.basename(dir)}.stories.tsx`);
 }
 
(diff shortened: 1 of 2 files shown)


ankit-yc (approved):
Verified: folder-basename story naming convention is real (36 instances via grep), the new alternateStoryPathFor()/hasAltStory logic is covered by 3 new unit tests + 2 evaluate() tests (22/22 passing), mutation-checked twice (removing the alt-check, removing the traversal guard — both correctly broke a test). Cross-checked the fix against an independent Python filesystem scan of all 2149 current apps/frontend .tsx files: 73→41 missing, exact match.

Also independently verified the Aikido-flagged path-construction issue (line ~138): Aikido's own suggested diff checks path.basename(dir), which for a value like ../../etc returns just etc and would never actually catch the traversal. The follow-up commit checks dir itself (dir.split(path.sep).includes('..') || path.isAbsolute(dir)) before ever joining it - confirmed this is the fix that's actually effective, not just diff-shaped.

PR #3045, opened 2026-09-11
chore(frontend): close out the current story-coverage gap

aupyay:
## What is the current behavior?

The story-coverage gate (`scripts/ci/story-coverage.mjs`) tracks components missing a Storybook story. After fixing the gate's own folder-basename naming blind spot in #3044, a full audit of `apps/frontend`'s current state found 41 genuinely missing stories.

## What is the new behavior?

39 of the 41 are thin Next.js route/layout/loading files under `src/app/`. Each was read individually: every one either has no visual content of its own (a metadata-only layout, a structural redirect) or composes an already-storied feature component (all 20 underlying components confirmed to have a real story via `find`). Marked with `// no-story: <reason>`, naming the already-covered component where relevant.

The 1 real gap is fixed: `app/error.tsx` (the root error boundary) had no story despite being a genuine, self-contained, testable component. Added `error.stories.tsx` following the same convention as the sibling `NotFoundState` story, with a `play` function covering both the reset action and the dashboard link.

`AppointmentWorkspace/index.tsx` (1896 lines, deep store integration across 6 Zustand stores) is the one remaining gap, deliberately deferred as its own dedicated pass rather than a rushed story.

## Verification

- `node scripts/ci/story-coverage.mjs --base <merge-base> --head HEAD --dir apps/frontend`: "every new component has a story"
- Full-codebase re-run of `evaluate()` (using the fixed script from #3044) against all 2149 current `apps/frontend/src/**/*.tsx` files: `checked: 605, missing: 1` — only `AppointmentWorkspace/index.tsx`, matching the deliberate deferral above.
- `tsc --noEmit` and `eslint` clean on every touched file (39 marker files + `error.stories.tsx`).
- `error.stories.tsx`'s assertions were traced by hand against `BaseButton.tsx`'s actual DOM output (`Primary` with no `href` renders `<button>`, `Secondary` with `href` renders next/link's `<a>`) to confirm every `getByRole`/`toHaveAttribute` query matches what really renders — live Storybook verification was blocked by sustained high load on this machine (two independent dev-server attempts stalled with near-zero CPU progress; confirmed via `top`/`uptime`, not app-specific).

## Related Issue(s)

Part of the ongoing Storybook-coverage sweep across web/mobile-responsive/mobile app.

PR #3046, opened 2026-09-11
ci(repo): make story coverage a required check, not dead code

aupyay:

`scripts/ci/story-coverage.mjs` has real, tested logic (folder-basename naming fixed in #3044, current backlog closed to 1 deliberately-deferred component in #3045), but is never actually invoked against a real PR diff by any workflow. `pnpm run test:scripts` (in `_core.yaml`'s `detect` job) runs the script's own unit tests - it never runs `story-coverage.mjs` itself against a PR's changed files. Every "story coverage" check this repo has had until now was a human running the script by hand.

## What is the new behavior?

Requested directly: a PR shouldn't be able to pass unless a new component has a Storybook story, "as stringent as the Sonar 95% coverage requirement bar."

- Adds a `story-coverage` job to `ci.yaml`, gated on `needs.core.outputs.frontend == 'true'`, using the same base-SHA resolver `core`'s `detect` job uses (`compute-base-sha.sh`), failing closed if no trustworthy base is found.
- Lists it as a dependency of `ci-required`. `CI Required` has been a required, blocking status check on the `dev` ruleset since 2026-09-06 (`docs/ci/required-checks-migration.md`) - joining its `needs` list is the exact mechanism `sonar` already uses to be blocking, not an advisory. No separate branch-protection/ruleset change needed.
- Documents the requirement in `CONTRIBUTING.md` and corrects `docs/ci/required-checks-migration.md`'s "what runs" list to mention it.

Deliberately still scoped to `apps/frontend` only. `apps/mobileAppYC` has zero Storybook infrastructure as of this PR (a separate initiative, not yet landed) - extending this gate there before that infra exists would block every mobile PR with no path to green.

## Testing

- `actionlint .github/workflows/ci.yaml` and a full-repo `actionlint` run: clean (pre-existing shellcheck notices in unrelated files only).
- Ran the exact job logic locally: `compute-base-sha.sh` + `story-coverage.mjs --base <sha> --head HEAD` against this branch's real diff - 0 new components, passes.
- Then committed a real storyless test component (`apps/frontend/src/app/ui/primitives/CiVerifyTestComponent/index.tsx`, 8 code lines, a real default-exported component with a JSX closing tag) and re-ran the identical command - failed with exit 1, correct expected-story path in the error message. Removed before this PR's real commit.
- This PR itself touches `.github/workflows/ci.yaml`, which `_core.yaml`'s detect job treats as "run everything" - so the new `story-coverage` job runs for real on this PR's own CI, not just locally.

## Related Issue(s)

None - direct request.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -133,6 +133,55 @@ jobs:
 
       - run: node scripts/mobile/version.mjs --check
 
+  story-coverage:
+    name: Story coverage
+    # scripts/ci/story-coverage.mjs fails a PR that adds a new apps/frontend
+    # component with no sibling Storybook story - as strict a bar as Sonar's:
+    # a listed dependency of `CI Required`, not an advisory. It only judges
+    # files a PR *adds* (see the script's own header for why), so the existing
+    # storyless backlog cannot block an unrelated change; it only stops that
+    # backlog from growing. Pure node:fs/path/child_process, no pnpm install
+    # needed - same shape as mobile-version above.
+    needs: core
+    if: needs.core.outputs.frontend == 'true'
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          # story-coverage.mjs diffs added files against the merge base, so
+          # full history must be present - the same reason core's detect job
+          # uses fetch-depth: 0.
+          fetch-depth: 0
+          persist-credentials: false
+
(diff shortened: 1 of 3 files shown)


ankit-yc (approved):
Verified end to end, not just by reading the YAML: ran the exact compute-base-sha.sh + story-coverage.mjs invocation locally against a real storyless component (committed then removed) and confirmed exit 1 with the correct error, then confirmed the actual GitHub Actions run of the new Story coverage job on this PR resolved BASE_SHA correctly (d29afbf71, the real merge base) and reported the correct pass verdict against this PR's real diff - genuinely wired in, not a vacuous pass. 'CI Required' has been the required blocking context on dev's ruleset since 2026-09-06 (confirmed via the rulesets API), so adding story-coverage to its needs list is the same mechanism sonar already uses - no separate branch-protection change needed, and none made. Scope is deliberately apps/frontend only; mobileAppYC has zero Storybook infra so is correctly excluded for now. actionlint clean, full CI green (71/71).

PR #3047, opened 2026-09-11
fix(frontend): reveal a clipped label by tap, not only by hover

aupyay:
## Summary
- Mobile-viewport (375px) Storybook sweep (365 hits / 105 components) found the `GlassTooltip` primitive has no touch affordance at all — every consumer whose only way to reveal content is a mouse (hover) or keyboard (focus) is unreachable on a touchscreen.
- Adds an opt-in `openOnClick` prop to `GlassTooltip` (off by default: a trigger with its own click behavior would have that first tap swallowed by the tooltip opening instead).
- Turns it on for the two sites confirmed safe by this pass: the document title in `AllDocumentsTable` (native `title` attribute only, no touch equivalent), and the package-breakdown info button in `PackageBreakdownTooltip` (icon-only trigger, no competing click handler).

## Test plan
- [x] `GlassTooltip.test.tsx`: new tests for open-on-tap, no-op when `openOnClick` unset, and close-on-outside-tap; mutation-checked (breaking `openOnClick` fails the new tests)
- [x] `SummaryStep.test.tsx`: new test taps a clipped document title and asserts the full value appears; mutation-checked (reverting the fix fails the test)
- [x] `PackageBreakdownTooltip.test.tsx` (new file): taps the info button with no hover involved; mutation-checked
- [x] `tsc --noEmit` clean, `eslint` clean on all changed files
- [x] Full targeted suite green post-rebase onto `dev` (86 tests, 5 suites)
- [x] `story-coverage.mjs --files` passes (no new components added)

### DIFF
--- a/apps/frontend/src/app/__tests__/pages/AppointmentWorkspace/PackageBreakdownTooltip.test.tsx
+++ b/apps/frontend/src/app/__tests__/pages/AppointmentWorkspace/PackageBreakdownTooltip.test.tsx
@@ -1,0 +1,42 @@
+import React from 'react';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import PackageBreakdownTooltip from '@/app/features/appointments/pages/AppointmentWorkspace/components/PackageBreakdownTooltip';
+import type { InvoiceLineItem } from '@/app/features/appointments/types/workspace';
+
+const PACKAGE_ITEM: InvoiceLineItem = {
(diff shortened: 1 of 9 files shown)


aupyay:
retargeted base to dev, reopening to force a fresh CI snapshot

ankit-yc (approved):
GlassTooltip openOnClick opt-in, applied to AllDocumentsTable + PackageBreakdownTooltip. Tests + mutation-checks included, all CI green after retargeting base to dev.

PR #3061, opened 2026-09-12
feat(repo): deterministic intake-form draft-import from supplied text

aupyay:
## Summary

Closes #3055 (backend slice). Converts supplied intake-form text into a new
draft `Form`, reusing the existing Form/FormVersion schema, validators and
publish machinery end to end rather than duplicating it.

- `packages/database`: new `FormDraftImport` table (pure addition, no RLS
  change - matches `Form`/`FormAssignment`'s own convention of no RLS on
  this domain).
- `formDraftImportParser.ts`: deterministic `label | type | required|optional
  | options` grammar over the same `FieldType` union the form renderer
  already uses. Unsupported constructs and conflicting redefinitions are
  reported with their source line, never silently dropped.
- `formDraftImportDiff.ts`: added/removed/changed/unchanged diff against a
  source form's latest **published** version.
- `formDraftImport.service.ts`: builds a `Form`-shaped object from the
  parsed fields, converts it with the existing `toFHIRQuestionnaire`, and
  persists it through the existing `FormService.create` - so the new draft
  gets the same validation, `FormField` sync and versioning every other form
  gets. Discard deletes the draft `Form` (409 once it has been published;
  publishing stays on the existing separate `/fhir/v1/form/admin/:formId/publish`
  route, untouched here).
- Three routes under `/v1/form-draft-imports/:orgId[/:id]` using the existing
  `forms:edit:any`/`forms:view:any` permission model (`form.router.ts`'s own
  convention), documented in the OpenAPI spec.

**Deliberately no AI/model step.** #3049 (the provider-neutral execution
boundary every AI step in this epic must run through) is not built yet, and
#3055 explicitly allows a deterministic slice to ship ahead of it. This also
gives the "prompt injection" test case a real answer: field text is stored
verbatim as a label and nothing here interprets it as an instruction.

**UI is out of scope for this PR** - tracked in #3060, mirroring how #3056
shipped API-only with UI split into #3057.

## Test plan

- `npx jest --testPathPatterns="formDraftImport"` - 51/51 passing, coverage
  99.29%/93%/100%/99.29% (stmt/branch/func/line) across the 4 new files, all
  ≥ the 90% new-file bar.
- Full backend suite: `npx jest` - 499 suites, 11802 tests, all passing.
- `pnpm --filter backend run lint` - 0 errors on touched files (24 pre-existing
  warnings elsewhere, unrelated).
- `pnpm --filter backend run type-check` - clean.
[...]

Issue #3060, opened 2026-09-12
feat(frontend): intake-form draft-import review UI (follow-up to #3055)


aupyay:
## Summary

Parent: #3055 (deterministic supplied-text-to-draft-form conversion backend, API-only slice).

#3055 ships the read-only/write API (`POST/GET/DELETE /v1/form-draft-imports/:orgId[/:id]`) with no UI. This issue tracks the developer-portal review surface for a practice configurator to paste supplied intake-form text, preview the proposed fields and diff against the selected form's published version, and discard or keep the resulting draft form before using the existing form editor/publish flow.

## Problem / motivation

An API-only draft-import is not usable by the target persona (developer/practice configurator). The review experience - proposed field list, unsupported constructs shown rather than silently dropped, a draft-versus-published diff, and a staleness warning when the source form changed since the draft was created - is itself part of the product value described in #3048/#3055 and was deliberately deferred so the backend slice could ship with full test coverage and the existing forms permission model first.

## Proposed solution

- A page under `apps/frontend/src/app/features/developers` where a configurator selects an existing form (optional), pastes supplied text in the documented `label | type | required|optional | options` grammar, and calls `POST /v1/form-draft-imports/:orgId`.
- Render the response: proposed fields, unsupported constructs (each with its source line and reason, never silently dropped), and the diff (`added`/`removed`/`changed`/`unchanged`) against the source form's latest published version.
- Re-fetch via `GET /v1/form-draft-imports/:orgId/:id` to show `stale: true` when the source form was edited or republished after the draft was created, and gate further action on refreshing.
- `DELETE /v1/form-draft-imports/:orgId/:id` to discard; a discarded/published draft leaves the source form untouched either way, matching the existing separate human publish path (`/fhir/v1/form/admin/:formId/publish`), which this UI does not touch.
- Use the existing design system, localisation and accessibility patterns; keyboard/screen-reader behaviour, narrow and wide layouts, and loading/empty/error states (including the 409 "already published" discard case).

## Out of scope

An actual AI/NLP-assisted mapping step (the backend parser is deterministic by design, pending #3049's provider-neutral execution boundary). Any new backend endpoint - #3055's contract is intended to be sufficient for this UI. Any change to the existing form publish path.

PR #3064, opened 2026-09-12
feat(frontend): intake-form draft-import review UI (#3060)

aupyay:
## Summary
- Developer-portal review UI for #3055's deterministic supplied-text-to-draft-form import: paste text in the documented grammar, preview proposed fields, unsupported constructs (never silently dropped), and the diff against a selected published form.
- Staleness re-fetch (`GET`) when the source form changed since the draft was created, and a discard action (`DELETE`) that surfaces the 409 "already published" case from the backend.
- No new backend endpoint - consumes #3055's contract as-is. No change to the existing form publish path.
- New route `/developers/form-draft-import`, added to the dev sidebar and the strict-CSP path list.

Base is `feat/3055-intake-form-draft-import` (not `dev`) because this UI depends on that PR's (#3061) API - this diff only shows the UI once #3061 merges and the base retargets.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `eslint` clean on all changed files
- [x] Full frontend jest suite: 1063 suites / 13538 tests green (includes the new page + service unit tests, and the repo's discovery-based guards: `cardSurface.test.ts` and `middleware.test.ts`, both updated for the new files)
- [x] `story-coverage.mjs`: all 3 new components have stories
- [x] Real Storybook build + `test-storybook` against it: 11/11 play-function tests pass across the 3 new story files

### DIFF
--- a/apps/frontend/src/app/(routes)/(app)/developers/(portal)/form-draft-import/page.tsx
+++ b/apps/frontend/src/app/(routes)/(app)/developers/(portal)/form-draft-import/page.tsx
@@ -1,0 +1,12 @@
+import type { Metadata } from 'next';
+import React from 'react';
+
+import DeveloperFormDraftImport from '@/app/features/developers/pages/DeveloperFormDraftImport/DeveloperFormDraftImport';
+
+export const metadata: Metadata = { title: 'Form draft import — Yosemite Crew' };
+
+function Page() {
+  return <DeveloperFormDraftImport />;
+}
+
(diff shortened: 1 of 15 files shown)


ankit-yc (approved):
CI Required all green at 9be9a5ce1 (S6819 fixed: role="status" div -> <output>, implicit status role preserved, story assertions unaffected). Advisory Storybook play functions also pass. Approving.

PR #3136, opened 2026-09-12
fix(backend): close tenant-isolation gaps in appointment accept/reject/cancel and patient-consent/transfer

aupyay:
## Summary
- The PMS appointment `accept`/`reject`/`cancel` routes used `withOrgPermissions()`, which authorises the org the URL names, not the org the target appointment actually belongs to. A caller authorised at org A could accept, reject, or cancel an appointment belonging to org B by putting org A in the URL and org B's appointment id in the path. Switched to `withAppointmentOrgPermissions()` (the pattern already used by `checkin`/`admit`/`attachFormsToAppointment` on the same router) and threaded the resolved `organisationId` through the service layer so `cancelAppointment`/`rejectRequestedAppointment`/`approveRequestedFromPms` scope their lookup by it.
- `patient-consent.service.ts#grant()` and `patient-transfer.service.ts#create()` trusted a caller-supplied `patientId` without checking it belongs to the caller's organisation, unlike the 18 sibling clinical services (`patient-flag`, `patient-allergy`, etc.) that already call the shared `assertPatientOrgMembership` guard for this exact reason. Added the same guard to both.

Both were found during a broader adversarially-verified backend security sweep this session; these are the highest-severity, lowest-risk-to-fix findings (established, already-proven fix patterns exist for both bug classes elsewhere in the codebase). The remaining findings from that sweep are being triaged separately.

## Impact area
`apps/backend` only: appointment router/service/controller, patient check-in's cancel-sync caller, patient-consent and patient-transfer services, and their test suites.

## Validation performed
- New cross-tenant regression tests added for all 5 fixed call sites (appointment accept/reject/cancel + consent grant + transfer create), asserting a 404/"Companion not found" and that no write is persisted.
- Mutation-tested every new guard: reverted the fix in the source (via `sed`/direct edit), confirmed the exact new+pre-existing tests that should catch the regression fail (and nothing else does), then restored and re-confirmed green.
- `npx jest --testPathPatterns="appointment.prisma.service.test|appointment.prisma.controller.test|patient-check-in.service.test|patient-consent.service.test|patient-transfer.service.test"` — 290/290 passing on this branch (rebased onto current `dev`).
- `npx tsc --noemit` (full backend) — 0 errors.
- `npx eslint` on all touched files — 0 errors (pre-existing, unrelated warnings only).
- Pre-push hook (full monorepo lint + type-check, 17/17 tasks) passed.

### DIFF
--- a/apps/backend/src/controllers/web/appointment.prisma.controller.ts
+++ b/apps/backend/src/controllers/web/appointment.prisma.controller.ts
@@ -306,14 +306,23 @@ export const AppointmentController = {
   },
 
   acceptRequested: async (
-    req: Request<{ appointmentId: string }, unknown, AppointmentRequestDTO>,
+    req: Request<
+      { organisationId: string; appointmentId: string },
+      unknown,
+      AppointmentRequestDTO
+    >,
     res: Response,
   ) => {
     try {
       const { appointmentId } = req.params;
+      const organisationId = resolveAuthorizedOrganisationId(req);
+      if (!organisationId) {
+        return res.status(400).json({ message: "Missing organisationId" });
+      }
       const data = await AppointmentPrismaService.approveRequestedFromPms(
         appointmentId,
         req.body,
+        organisationId,
       );
       return res.status(200).json({ message: "Appointment accepted", data });
     } catch (err: unknown) {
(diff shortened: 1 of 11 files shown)


ankit-yc (approved):
CI fully green (all required checks, SonarCloud quality gate, Aikido, react-doctor). Mutation-tested cross-tenant regression tests confirmed to fail against the pre-fix code and pass against the fix; reused the codebase's own established withResourceOrgPermissions/assertPatientOrgMembership patterns rather than inventing new ones.

PR #3137, opened 2026-09-12
fix(frontend): brighten compliance badge marks for legibility in dark theme

aupyay:
## Summary
- The GDPR/SOC 2/ISO 27001/FHIR marks in the footer are dark ink on a transparent background; against the dark-theme footer background they become illegible.
- Reuses the `filter: grayscale(1) brightness(2.2)` technique already applied inline in `AuthShell.tsx` to these exact same official marks (for its own permanently-dark brand panel), so no mark is redrawn, replaced, or fabricated — only the on-screen rendering of the existing pixels is recoloured. The FDA mark already carries its own opaque light background and is deliberately excluded.
- Added a Storybook story that pins the dark theme and asserts the filter via `getComputedStyle` on each badge, so the fix has story coverage.

## Impact area
`apps/frontend` only: `Footer.css`, `Footer.stories.tsx`.

## Validation performed
- Verified visually in Storybook (both themes) — dark theme badges legible, light theme pixel-identical to before.
- Verified via direct `getComputedStyle` assertions against the live Storybook iframe in both themes.
- `npx tsc --noemit` (frontend) — 0 errors.
- `npx eslint` on both touched files — 0 errors.
- Pre-push hook (full monorepo lint + type-check, 17/17 tasks) passed.

### DIFF
--- a/apps/frontend/src/app/ui/widgets/Footer/Footer.css
+++ b/apps/frontend/src/app/ui/widgets/Footer/Footer.css
@@ -205,6 +205,21 @@
   height: 56px;
 }
 
+/* The GDPR/ISO/FHIR marks are solid dark ink on a transparent background, and
+   the SOC 2 seal is a near-black AICPA-issued circle - both read fine on the
+   light footer but disappear into it once html[data-theme='dark'] turns the
+   footer's own background dark. These are official third-party marks (AICPA's
+   licensed SOC seal among them), so the fix is the same brightness/grayscale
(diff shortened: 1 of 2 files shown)


ankit-yc (approved):
CI fully green including Chromatic, all 4 Storybook play-function shards, and SonarCloud quality gate. Verified visually in both themes; the fix reuses AuthShell.tsx's existing filter technique on the same official marks rather than introducing new assets.

PR #3196, opened 2026-09-12
fix(backend): net refunds from invoice paid totals

### LINKED ISSUE #3138: [Bug]: invoice refunds are never netted out of cashPaid/effectivePaid
### Affected area
Backend API (apps/backend)

### What is wrong
`computeInvoiceFinancialDetails()` in [invoice.service.ts:496](apps/backend/src/services/invoice.service.ts#L496) sums `payment.amount` for every `Payment` row on an invoice with no status filter. The payments are fetched (lines 549-553 / 587-591) *with* their `refunds` relation loaded, but that relation is never used in the sum — a fully or partially refunded payment is still counted at its original full amount toward `cashPaid`/`effectivePaid`.

### Steps to reproduce
1. Create an invoice with `totalAmount = $200`.
2. Record a `$200` payment against it (`Payment.status = COMPLETED`).
3. Issue a `$50` partial refund on that payment (`Payment.status` becomes `PARTIALLY_REFUNDED`; `invoice.totalAmount` is unchanged).
4. Call `GET /invoices/:invoiceId` (or the finance list) as clinic staff.

### Expected behavior
`effectivePaid` should net out the refunded amount, so the invoice shows an outstanding balance reflecting the $50 that was actually returned to the customer.

### Actual behavior
`actualCashPaid` is computed as the unconditional sum of `Payment.amount = $200`. The refund is never subtracted. `effectivePaid = $200`, `balance = max(0, 200-200-0) = $0`. The UI reports the invoice as fully settled with nothing owed, silently hiding that $50 was refunded — the org has no signal that the true remaining balance is understated by the refunded amount.

Always, whenever a payment on an invoice has any refund recorded against it.

### Did this work before

### Impact
High - a core workflow (invoice/payment reconciliation) silently reports incorrect financial state, with no workaround short of manual reconciliation.

### Additional context
Found during a backend correctness sweep this session (2026-09-12). Same root cause as a related, opposite-direction inconsistency already open in `getInvoiceFinancialSummary` — refunds are never netted against the payment they apply to anywhere in this file. Fix direction: exclude/net `REFUNDED`/`PARTIALLY_REFUNDED` payment amounts (or their refunded sub-amount) from the `cashPaid`/`effectivePaid` sum, consistently across every call site in this file.

aupyay:
## PR Checklist

Invoice settlement calculations count the original amount of successful payments without subtracting successful refunds. A fully refunded payment can also be omitted from one summary path while remaining counted at its gross amount in another, producing inconsistent paid and balance totals.

## What is the new behavior?

Invoice financial summaries use a shared calculation that subtracts successful refunds from each payment and floors the result at zero. Both invoice detail/list settlement data and the reusable finance summary now include succeeded, partially refunded, and refunded payments and report their net paid amount consistently.

Added coverage for partial refunds, failed refunds, and refunds larger than the original payment. The two affected backend suites pass (281 tests), backend type-check and build pass, and the repository-wide pre-push lint and type-check gates pass. The complete backend test suite requires the repository's Redis-backed test environment and was not completed locally.

## Related Issue(s)

Fixes #3138

### DIFF
--- a/apps/backend/src/services/finance/payment.ts
+++ b/apps/backend/src/services/finance/payment.ts
@@ -12,7 +12,7 @@ import Stripe from "stripe";
 import { prisma } from "src/config/prisma";
 import logger from "src/utils/logger";
 import { FinanceEventService } from "./events";
-import { roundMoney } from "./pricing";
+import { getNetPaymentAmount, roundMoney } from "./pricing";
 import {
   fromStripeMinorUnits,
   toStripeMinorUnits,
@@ -213,8 +213,17 @@ export const getInvoiceFinancialSummary = async (
 ): Promise<InvoiceFinancialSummary> => {
   const [payments, creditNotes] = await Promise.all([
     prisma.payment.findMany({
-      where: { invoiceId, status: "SUCCEEDED" },
-      select: { amount: true },
+      where: {
+        invoiceId,
+        status: { in: ["SUCCEEDED", "PARTIALLY_REFUNDED", "REFUNDED"] },
+      },
+      select: {
+        amount: true,
+        refunds: {
+          where: { status: "SUCCEEDED" },
+          select: { amount: true, status: true },
+        },
+      },
     }),
     prisma.creditNote.findMany({
       where: { invoiceId, status: "ISSUED" },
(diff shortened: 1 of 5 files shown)


ankit-yc (approved):
Reviewed independently: getNetPaymentAmount logic checked against issue #3138's math, both call sites in invoice.service.ts covered via the shared helper, plus the related getInvoiceFinancialSummary fix. Ran the 2 suites locally (281/281 pass) and mutation-tested both the refund-status filter and the zero-floor — each caught by a distinct test. LGTM.

Issue #3142, opened 2026-09-12
[Bug]: inventory consumption has lost-update races, an untransacted race-wide endpoint, and missing audit/controlled-substance trail


aupyay:
### Affected area
Backend API (apps/backend)

### What is wrong
Five related, verified bugs in the inventory-consumption subsystem, found together during the same audit pass — filing as one issue since a fix session will likely touch the same handful of files:

**1. Lost update in `applyInventoryConsumption` (high)** — [inventory-consumption.service.ts:1173](apps/backend/src/services/inventory-consumption.service.ts#L1173). The function behind prescription dispense approval decrements each batch by writing a **literal computed value** (`quantity: available - consume`, line 1198) instead of an atomic Prisma `decrement`, so a concurrent dispense on the same batch produces a lost update even though the call runs inside `prisma.$transaction`. The sibling release path (line 1143, `data: { quantity: { increment: restore } }`) already uses the atomic form — this is an inconsistency, not an intentional design.
  - *Repro*: two concurrent 6-unit dispenses against a batch with `quantity=10`. Both read `10`, both compute `newQuantity=4`, both write `4`. Only 6 units are actually lost from a batch that should now read `-2` (or the second dispense should have been rejected), but stock shows `4` — overstated by 6, and the second dispense was approved against inventory that wasn't really there.

**2. `InventoryService.consumeStock`/`adjustStock` run with no transaction at all (high)** — [inventory.service.ts:2007](apps/backend/src/services/inventory.service.ts#L2007). Used by the standalone `POST /inventory/consume-stock` endpoint (unrestricted for controlled items). Reads batches, computes a new quantity in application code, and writes it — with no `prisma.$transaction` wrapping any of the three steps, so the race window is even wider than bug 1's.
  - *Repro*: two concurrent calls to `consume-stock` for the same item with `quantity=6`, `onHand=10`. Both see `onHand=10`, both pass the `< input.quantity` check, both compute `newQuantity=4` from the same batch snapshot. Neither call errors; the batch ends at `4` instead of correctly rejecting the second call or ending at `-2`.

[...]

PR #3197, opened 2026-09-12
fix(backend): atomic decrement, transactions, and allocation guard for inventory consumption

aupyay:
Fixes bugs 1, 2, 3, and 5 from #3142:
- `applyInventoryConsumption` wrote a literal computed batch quantity instead of an atomic Prisma `decrement`, losing updates under concurrent dispenses.
- `InventoryService.consumeStock` ran with no transaction at all and never logged a stock movement, unlike every other stock-changing path.
- The insufficient-stock guard only checked `onHand`, letting a NORMAL-source consumption dispense stock already reserved via `allocateStock`.

Bug 4 (controlled-substance register wiring) is not included: `ControlledSubstanceLogService.create` requires a `deaSchedule` and `unit` per entry that `InventoryItem` does not store anywhere in the schema, and `amountDrawn`/`amountAdministered`/`amountWasted` are real per-administration clinical quantities a pharmacy dispense event can't supply — needs a schema/design decision, not wiring.

## Test plan
- [x] `npx jest test/services/inventory.service.test.ts test/services/inventory-consumption.service.test.ts` — 278/278
- [x] Full backend suite: 499 suites / 11841 tests, all passed
- [x] `tsc --noEmit` clean
- [x] `eslint` on changed files — 0 errors (3 pre-existing warnings, unrelated lines)

### DIFF
--- a/apps/backend/src/services/inventory-consumption.service.ts
+++ b/apps/backend/src/services/inventory-consumption.service.ts
@@ -1170,7 +1170,17 @@ const applyInventoryConsumption = async (
   params: InventoryConsumptionApplyParams,
 ) => {
   const item = await requireInventoryItemForAdjustment(tx, params);
-  if ((item.onHand ?? 0) < params.quantity) {
+  // A NORMAL-source consumption must not eat into stock already reserved for
+  // someone else via InventoryAllocationService.allocateStock (the allocation
+  // path itself enforces onHand-allocated>=quantity; this guard mirrors it).
+  // An ALLOCATED-source consumption is drawing down that same reservation, so
+  // it is checked against onHand alone - finalizeInventoryAdjustment below
+  // reduces `allocated` for that case.
+  const availableForConsumption =
+    params.stockSource === "ALLOCATED"
+      ? (item.onHand ?? 0)
+      : (item.onHand ?? 0) - (item.allocated ?? 0);
+  if (availableForConsumption < params.quantity) {
     throw new InventoryConsumptionServiceError("Insufficient stock", 400);
   }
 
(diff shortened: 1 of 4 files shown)


ankit-yc (approved):
Reviewed the diff against #3142 and the surrounding service code.

- `applyInventoryConsumption`: batch writes are now `{ decrement }`, matching the release path's `{ increment }`. `onHand` is still recomputed from batches inside the same transaction, so the item total stays consistent with the batch rows.
- The allocation guard is correct for both sources. NORMAL checks `onHand - allocated`, and ALLOCATED checks `onHand` alone while `finalizeInventoryAdjustment` draws `allocated` down.
- `consumeStock` now runs in one `$transaction`, decrements atomically and writes a movement per batch through `logMovement(..., tx)`. `bulkConsumeStock` calls it in sequence, so there is no nested interactive transaction.
- The tests cover the decrement shape, the reserved-stock refusal and the per-batch movement rows.

Not blocking, but worth a follow-up:
1. `InventoryService.consumeStock` (`POST /stock/consume`) still checks `onHand` only, not `onHand - allocated`. I left it as is on purpose. That endpoint has no stock-source notion and never reduces `allocated`, so an appointment that allocated stock and then consumes it through this route would be rejected by a naive guard. What it should do is a product decision.
2. `InventoryAdjustmentService.adjustStock`, named with bug 2 in #3142, is still outside a transaction and still writes `quantity: available - consume`.
3. Under READ COMMITTED, two concurrent calls can both pass the pre-check. The atomic decrement means the batch now goes negative instead of silently losing an update, so the ledger stays accurate. Rejecting the second call would need a row lock or a check constraint.

Approving.

Issue #3183, opened 2026-09-12
fix(backend): deliver task reminders on the first worker tick after their due time


aupyay:
### Preflight

- [x] Checked current source and synthetic or browser evidence.
- [x] Compared existing open and closed issues and open pull requests.
- [x] Public report uses synthetic examples and contains no private records or paths.

### Affected area

Backend API

### What is wrong

No reminder is sent before the due time, and the task is excluded immediately after it. Both synthetic worker runs produced zero sends.

### Steps to reproduce

1. Use a synthetic pending task due at 12:00:00 with an enabled reminder offset of zero.
2. Run the minute reminder worker at 11:59:59 and again at 12:00:01.
3. Read reminder delivery calls and persisted reminder state.

### Expected behavior

An enabled at-due reminder is sent once on the first eligible worker tick at or after the due time, without requiring an exact timestamp match.

### Actual behavior

No reminder is sent before the due time, and the task is excluded immediately after it. Both synthetic worker runs produced zero sends.

### How often does it reproduce

Deterministic under the stated reproduction conditions; not a claim about all users or production incidence.

### Did this work before

Unknown; the current development revision still exhibits the behavior.

### Impact

P1 suggested triage priority. No reminder is sent before the due time, and the task is excluded immediately after it. Both synthetic worker runs produced zero sends.

### Environment and evidence

Audit date: 12 September 2026. Source baseline: e5c0c78ab79afe7c1796ee8d98d1697ec2a1c904.

Reproduced against current service source using synthetic records and in-memory persistence or notification boundaries. No live application, production database, or device test is claimed.

Diagnostic reproductions establish the reported defect; they are not a product fix, a passing regression suite or a coverage claim.

### Implementation entry points

Locate these existing symbols and follow their callers; locations are intentionally omitted from this public report.

- `TaskReminderEngine`
- `TaskReminderWorker`
- `buildReminder`

### Repair contract

1. Select unsent eligible reminders according to their reminder instant, including those whose task due time has just passed. Keep pending and in-progress status eligibility and the existing sent marker.
2. Do not hard-code an equality comparison or send all future reminders early. Make delayed worker processing recover eligible reminders without marking completed or cancelled tasks as needing a new alert.
3. Ensure retries do not falsely record an unsuccessful dispatch as sent. Keep the visible reminder state consistent with the recorded outcome.

### Acceptance criteria and QA

[...]

Issue #3182, opened 2026-09-12
fix(backend): prevent duplicate deferred workflow tasks after a partial retry


aupyay:
### Preflight

- [x] Checked current source and synthetic or browser evidence.
- [x] Compared existing open and closed issues and open pull requests.
- [x] Public report uses synthetic examples and contains no private records or paths.

### Affected area

Backend API

### What is wrong

Three tasks exist: First, First, Second. The schedule links only the two created on the successful retry, leaving the earlier First orphaned from schedule tracking.

### Steps to reproduce

1. Prepare a synthetic deferred workflow with two distinct tasks, First and Second.
2. Run activation and simulate a transient persistence failure while creating Second.
3. Restore persistence and let the scheduled worker retry.
4. Read the generated task list and schedule links.

### Expected behavior

The same deferred schedule creates exactly one task per intended seed, and every created task remains linked to that schedule after retry.

### Actual behavior

Three tasks exist: First, First, Second. The schedule links only the two created on the successful retry, leaving the earlier First orphaned from schedule tracking.

### How often does it reproduce

Deterministic under the stated reproduction conditions; not a claim about all users or production incidence.

### Did this work before

Unknown; the current development revision still exhibits the behavior.

### Impact

P1 suggested triage priority. Three tasks exist: First, First, Second. The schedule links only the two created on the successful retry, leaving the earlier First orphaned from schedule tracking.

### Environment and evidence

Audit date: 12 September 2026. Source baseline: e5c0c78ab79afe7c1796ee8d98d1697ec2a1c904.

Reproduced against current service source using synthetic records and in-memory persistence or notification boundaries. No live application, production database, or device test is claimed.

Diagnostic reproductions establish the reported defect; they are not a product fix, a passing regression suite or a coverage claim.

### Implementation entry points

Locate these existing symbols and follow their callers; locations are intentionally omitted from this public report.

- `materializeSchedule`
- `TaskScheduleEngine`
- `TaskService.createFromWorkflowSeed`
- `TaskWorkflowService`

### Repair contract

1. Use the existing transaction-capable workflow task creation contract for bounded atomic materialization and persist generated task identities in the same transaction.
2. Protect simultaneous activation or replay with a stable schedule-and-seed identity or a serialized schedule claim, using current database conventions.
3. If a batch cannot safely use one transaction, persist an idempotent identity and durable checkpoint per seed. Never infer completion from a nonempty prefix alone.
[...]

Issue #3181, opened 2026-09-12
fix(backend): keep a task series active when only its first occurrence is cancelled


aupyay:
### Preflight

- [x] Checked current source and synthetic or browser evidence.
- [x] Compared existing open and closed issues and open pull requests.
- [x] Public report uses synthetic examples and contains no private records or paths.

### Affected area

Backend API

### What is wrong

The first task is cancelled and the generator creates no further occurrences. Existing children may temporarily hide the stopped series until its pre-generated horizon ends.

### Steps to reproduce

1. Create a daily synthetic task series before its future occurrences have been generated.
2. Cancel the first task using the existing This occurrence scope.
3. Run recurrence generation and inspect future tasks.
4. Repeat after some future occurrences exist and move the generation horizon forward.

### Expected behavior

Cancelling This occurrence skips only that occurrence. Future occurrences remain scheduled until the user cancels the series or reaches its end bound.

### Actual behavior

The first task is cancelled and the generator creates no further occurrences. Existing children may temporarily hide the stopped series until its pre-generated horizon ends.

### How often does it reproduce

Deterministic under the stated reproduction conditions; not a claim about all users or production incidence.

### Did this work before

Unknown; the current development revision still exhibits the behavior.

### Impact

P1 suggested triage priority. The first task is cancelled and the generator creates no further occurrences. Existing children may temporarily hide the stopped series until its pre-generated horizon ends.

### Environment and evidence

Audit date: 12 September 2026. Source baseline: e5c0c78ab79afe7c1796ee8d98d1697ec2a1c904.

Reproduced against current service source using synthetic records and in-memory persistence or notification boundaries. No live application, production database, or device test is claimed.

Diagnostic reproductions establish the reported defect; they are not a product fix, a passing regression suite or a coverage claim.

### Implementation entry points

Locate these existing symbols and follow their callers; locations are intentionally omitted from this public report.

- `TaskService.deleteTask`
- `normalizeRecurrenceScope`
- `getSeriesMasterId`
- `TaskRecurrenceEngine`

### Repair contract

1. Separate occurrence cancellation from series-active state in the existing recurrence owner. Preserve the chosen cancellation scope explicitly.
2. Ensure the generator can continue a live series after its first occurrence was cancelled while ALL and THIS_AND_FOLLOWING still stop the requested range.
3. Do not simply include all cancelled masters: that would revive deliberately cancelled series. Preserve and test the meaning of each scope.
[...]

Issue #3180, opened 2026-09-12
fix(backend): preserve recurring task details and reset each occurrence reminder


aupyay:
### Preflight

- [x] Checked current source and synthetic or browser evidence.
- [x] Compared existing open and closed issues and open pull requests.
- [x] Public report uses synthetic examples and contains no private records or paths.

### Affected area

Backend API

### What is wrong

The generated child has no extra instructions, subcategory, priority or group assignment. Its reminder inherits the original sent marker, so later reminder processing skips it.

### Steps to reproduce

1. Create a repeating synthetic task with priority High, extra instructions and its existing group assignment.
2. Let the original occurrence receive its reminder.
3. Generate a subsequent occurrence after the original reminder has been marked sent, then read the new task and run its reminder processing.

### Expected behavior

Each new occurrence preserves applicable task instructions and assignment while starting with a fresh unsent reminder lifecycle.

### Actual behavior

The generated child has no extra instructions, subcategory, priority or group assignment. Its reminder inherits the original sent marker, so later reminder processing skips it.

### How often does it reproduce

Deterministic under the stated reproduction conditions; not a claim about all users or production incidence.

### Did this work before

Unknown; the current development revision still exhibits the behavior.

### Impact

P1 suggested triage priority. The generated child has no extra instructions, subcategory, priority or group assignment. Its reminder inherits the original sent marker, so later reminder processing skips it.

### Environment and evidence

Audit date: 12 September 2026. Source baseline: e5c0c78ab79afe7c1796ee8d98d1697ec2a1c904.

Reproduced against current service source using synthetic records and in-memory persistence or notification boundaries. No live application, production database, or device test is claimed.

Diagnostic reproductions establish the reported defect; they are not a product fix, a passing regression suite or a coverage claim.

### Implementation entry points

Locate these existing symbols and follow their callers; locations are intentionally omitted from this public report.

- `cloneFromMasterPrisma`
- `TaskRecurrenceEngine`
- `buildReminder`
- `TaskReminderEngine`

### Repair contract

1. Define the child payload field-by-field from the existing task model, including currently omitted applicable fields. Reuse the existing reminder-normalization behavior rather than copying delivery metadata.
2. Copy only reminder configuration, never a sent marker or delivery identifier belonging to another occurrence.
3. Keep completion state, completion evidence and calendar event identity occurrence-specific. Do not overwrite already completed historical children while repairing generation.

### Acceptance criteria and QA

[...]

PR #3199, opened 2026-09-12
fix(backend): correct recurring-task and reminder scheduling bugs

aupyay:
## Summary
Batches five related backend bugs in recurring-task generation and reminder delivery:

- **#3179** - a recurring task's due time drifted an hour across a DST transition because next-occurrence math preserved the previous instant's UTC offset instead of recomputing local wall-clock time.
- **#3180** - generated occurrences dropped subcategory, additional notes, priority and group assignment, and copied the parent's reminder `scheduledNotificationId` verbatim, so the reminder worker treated the new occurrence as already sent.
- **#3181** - the recurrence engine filtered "masters" by their own occurrence status; since the master row IS occurrence #1, cancelling only that occurrence (scope `THIS`) stopped the whole series from generating further children. Series-level stop is now expressed via `recurrence.endDate` on all three cancel scopes (`THIS_AND_FOLLOWING` already did this; `ALL` now does too).
- **#3182** - a deferred workflow schedule's `generatedTaskIds` was only persisted once, after every seed finished; a mid-loop failure left it empty, so a retry recreated already-succeeded tasks. It's now persisted incrementally per seed.
- **#3183** - the reminder worker's query used `dueAt >= now` as an upper bound, so a zero-offset reminder fell through the gap between "too early" (previous tick) and "already past due" (next tick) and never sent. The query now uses a lookback grace window instead.

## Test plan
- [x] `npx jest` (apps/backend): 500 suites / 11,850 tests passed
- [x] `npx eslint .` (apps/backend): 0 errors
- [x] `npx tsc --noEmit` (apps/backend): 0 errors
- [x] Added/updated targeted tests for all 5 fixes; mutation-checked each by reverting the fix and confirming the corresponding test fails, then restoring
- [x] Pre-push hook (monorepo lint + type-check via turbo): passed

### DIFF
--- a/apps/backend/src/services/task.recurrence.engine.ts
+++ b/apps/backend/src/services/task.recurrence.engine.ts
@@ -4,7 +4,13 @@ import utc from "dayjs/plugin/utc.js";
 import timezone from "dayjs/plugin/timezone.js";
 import cronParser from "cron-parser";
 
-import { Prisma, TaskAudience, TaskSource, TaskStatus } from "@prisma/client";
+import {
+  Prisma,
+  TaskAudience,
+  TaskPriority,
+  TaskSource,
+  TaskStatus,
+} from "@prisma/client";
 import { prisma } from "src/config/prisma";
 
 dayjs.extend(utc);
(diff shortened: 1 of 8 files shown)


ankit-yc (approved):
CI green, changes verified.

Issue #3141, opened 2026-09-12
[Bug]: weekly availability window is off by one day and pulls next Monday's occupancy into this week


aupyay:
### Affected area
Backend API (apps/backend)

### What is wrong
`getWeeklyFinalAvailability` (and the identical pattern in `getCurrentStatusBulk`, [availability.service.ts:734](apps/backend/src/services/availability.service.ts#L734)) computes `weekEnd` as `weekStart + 7 days, endOf('day')` — [availability.service.ts:632](apps/backend/src/services/availability.service.ts#L632) — which is the end of the *following* Monday, one full day past the actual 7-day Monday-Sunday window that `weekDates` (`resolveWeekBounds`, exactly 7 entries) represents. Occupancies are then bucketed only by weekday *name*, not by full date, so an occupancy on next week's Monday is pulled into and incorrectly subtracted from this week's Monday slots.

### Steps to reproduce
1. Call `GET /:orgId/final` (`getFinalAvailabilityForDate` → `getWeeklyFinalAvailability`) for the week of Monday 2026-09-14 through Sunday 2026-09-20.
2. Have an occupancy already booked for **next** week's Monday, 2026-09-21, 10:00-10:30, for the same vet.

### Expected behavior
The 2026-09-14 (this week's Monday) 10:00-10:30 slot should show as available, since the vet has no booking that day.

### Actual behavior
`weekStart = 2026-09-14T00:00:00.000Z`, and the buggy `weekEnd = 2026-09-21T23:59:59.999Z` (next Monday), so the query (`startTime <= weekEnd`) also fetches the 2026-09-21 occupancy. In `computeWeekAvailability`, `occStart.format('dddd') = "MONDAY"`, so `map.get("MONDAY")` returns *this* week's Monday slots and incorrectly splits/removes 10:00-10:30 from them. The calendar shows 2026-09-14 10:00-10:30 as busy/unavailable even though the real conflict is a week later.

### How often does it reproduce
Always, whenever an occupancy exists on the Monday immediately following the queried week.

### Did this work before
Not sure — found via code audit, not a user report.

### Impact
Medium - availability display is wrong for affected Mondays, but there is a workaround (staff can cross-check the actual calendar).

### Additional context
[...]

Issue #3139, opened 2026-09-12
[Bug]: concurrent appointment booking can double-book the same lead vet


aupyay:
### Affected area
Backend API (apps/backend)

### What is wrong
The lead-vet double-booking check (`assertLeadAvailability`, called from `upsertAppointmentOccupancy` — [appointment.prisma.service.ts:1100](apps/backend/src/services/appointment.prisma.service.ts#L1100), used by `createAppointment`'s transaction, `approveRequestedFromPmsInTransaction`, and `updateAppointmentPMS`) is a read-then-write check inside a plain `prisma.$transaction` with no `isolationLevel` set (defaults to Postgres READ COMMITTED). The `Occupancy` table (`packages/database/prisma/schema.prisma`, `Occupancy` model) has only a primary key and lookup indexes — no unique/exclusion constraint preventing overlapping `(userId, startTime, endTime)` ranges.

### Steps to reproduce
1. Two PMS staff (or an automated retry racing a manual click) concurrently call the appointment-create or approve-requested endpoint for the same lead vet and the same overlapping time slot (e.g. both for Dr. X, 2026-09-14 10:00-10:30).
2. Both transactions run `assertLeadAvailability`'s `occupancy.findFirst` before either has committed its `occupancy.create`/appointment insert.

### Expected behavior
The second booking attempt should fail with a conflict, since the vet is already booked for that slot.

### Actual behavior
Under READ COMMITTED, neither statement sees the other's uncommitted row, so both checks pass, both appointments are created with status `UPCOMING`, and both `Occupancy` rows are inserted — the vet ends up double-booked with no error surfaced to either caller.

### How often does it reproduce
Sometimes, intermittent — needs two write requests to race within the same transaction window. More likely under load or with a client retry.

### Did this work before
Not sure — found via code audit, not a user report.

### Impact
High - a core scheduling workflow allows a data-integrity violation (double-booking) with no workaround once it happens short of manual detection and rebooking.

### Additional context
Found during a backend correctness sweep this session (2026-09-12). Fix direction: either raise the transaction isolation level to `Serializable` for this check-and-write, or add a database-level exclusion constraint on `Occupancy` (e.g. a Postgres `EXCLUDE USING gist` on `(userId, tsrange(startTime, endTime))`) so a genuine overlap is rejected at the DB layer regardless of transaction timing.

aupyay:
[...]

PR #3207, opened 2026-09-12
fix(backend): close lead-availability race and weekly-availability off-by-one

aupyay:
## Summary
- #3139: `assertLeadAvailability`'s read-then-write check ran inside a plain (READ COMMITTED) transaction with no DB-level exclusion constraint on `Occupancy`, so two concurrent create/approve/update calls could both pass the overlap check and double-book the same lead vet. The three transactions that book a lead's occupancy (`createAppointment`, `approveRequestedFromPms`, `updateAppointmentPMS`) now run at `Serializable` isolation - the same pattern already used for the free-plan usage-counter guard in `appointment.service.ts`.
- #3141: `getWeeklyFinalAvailability` / `getCurrentStatusBulk` computed `weekEnd` as `weekStart + 7 days`, one day past the actual Monday-Sunday window, so an occupancy on next week's Monday was pulled into and subtracted from this week's Monday slots (occupancies are bucketed by weekday name, not full date). Bounded `weekEnd` to `weekStart + 6 days`. Also fixed the same expression silently applying the server's local UTC offset instead of the UTC calendar day (it re-wrapped an already-UTC `weekStart` in local-mode dayjs before `endOf("day")`).

## Test plan
- [x] `pnpm --filter backend run type-check`
- [x] `pnpm --filter backend run lint` (0 errors; pre-existing warnings in untouched files only)
- [x] `npx jest --testPathPatterns="appointment.prisma.service.test"` - 192/192, including new isolation-level assertions on all three transactions
- [x] `npx jest --testPathPatterns="availability.service.test"` - 60/60, including new weekEnd-boundary assertions for both call sites
- [x] Mutation check: reverted the `isolationLevel` option and confirmed the new test fails; reverted the `weekEnd` day-count and confirmed the new test fails
- [x] Full backend suite: `npx jest` - 499 suites / 11847 tests passed


ankit-yc (approved):
LGTM. Serializable isolation on the three lead-booking transactions matches the existing pattern in appointment.service.ts's usage-counter guard; weekEnd off-by-one and local-TZ leak fix verified with direct query-arg assertions, mutation-killed both. Full backend suite 499/499, CI green.

PR #3208, opened 2026-09-12
fix(backend): stop weekly availability window pulling in next Monday's occupancy

### LINKED ISSUE #3141: [Bug]: weekly availability window is off by one day and pulls next Monday's occupancy into this week

aupyay:
## Summary
- `getWeeklyFinalAvailability` and `getCurrentStatusBulk` computed `weekEnd` as `weekStart + 7 days`, which is the end of the *following* Monday rather than this week's Sunday. Occupancies are bucketed by weekday name only, so an occupancy on next week's Monday was pulled into and incorrectly subtracted from this week's Monday slots.
- Fixed `weekEnd` to `weekStart + 6 days`, matching the actual 7-day (Mon-Sun) window `weekDates` represents.
- Also anchored `weekEnd` to UTC (`.utc()`) like `weekStart` already is - without it, `endOf("day")` resolved in the server's local timezone, drifting the boundary by the UTC offset on a non-UTC host. Found while writing the boundary test: it failed by exactly the local UTC offset until this was added.

Fixes #3141

## Test plan
- [x] Added a test pinning the exact `startTime.lte`/`endTime.gte` passed to `prisma.occupancy.findMany` for both call sites, mutation-checked against the pre-fix code (fails with the old `+7 day` value).
- [x] `pnpm --filter backend run test` - 499 suites / 11847 tests passed
- [x] `pnpm --filter backend run lint` - clean
- [x] `pnpm --filter backend run type-check` - clean
- [x] Root `pnpm run type-check` - 17/17 tasks passed


aupyay:
Superseded by #3207, which fixes this same weekEnd off-by-one/UTC-anchoring bug plus the related #3139 lead-availability race in one PR. Closing to avoid two PRs landing overlapping fixes to the same lines in `availability.service.ts` — see the standing fewer-PRs guidance.

PR #3209, opened 2026-09-12
fix(frontend): anchor week calendar day columns on preferred timezone

### LINKED ISSUE #1868: fix(frontend): desktop week calendar anchors day columns on a browser-local midnight

aupyay:
## PR Checklist

The desktop appointments week calendar derived its seven day columns from **browser-local midnight** (`startOfDay` + local `setDate` stepping in `getWeekDays`), then compared those columns against appointment instants converted into the clinic's **preferred timezone**. For a browser far enough from the org's preferred zone, a column could resolve to the wrong clinic calendar day, so appointments landed under the wrong column heading.

Separately, `CalendarWeekDayCell`'s weekday label was already preferred-zone-aware (`formatDateInPreferredTimeZone`) while its date numeral read the browser's own `getDate()` - the two could already disagree independent of the bug above.

## What is the new behavior?

- Added `startOfPreferredTimeZoneDay` and `getWeekDaysInPreferredTimeZone` (`weekHelpers.ts`), which read the calendar day from the clinic's preferred timezone and re-anchor it at **noon** in that zone - the same noon-anchoring the codebase already uses in `buildPreferredTimeZoneDayInstant` for exactly this reason (noon sits far from any day boundary, so it survives being reinterpreted through the browser's own offset).
- Switched the appointments week path to the new helpers: `Appointments/index.tsx` (the `weekStart` anchor), `common/WeekCalendar.tsx` (day columns), and `appointmentDragAvailabilityUtils.ts` (drag-availability prefetch targets, so they stay consistent with the rendered columns).
- Fixed `CalendarWeekDayCell`'s date numeral to read the same preferred-timezone conversion as its weekday label.

`getWeekDays`/`startOfDay` are unchanged and still used by the Task calendar (out of scope - it renders in local-device terms deliberately). `filterAppointmentsForWeek`'s coarse range prefilter is also unchanged; it stays a deliberate superset.

Fixes #1868. Supersedes #2254, which hit the same bug but was closed on stale merge conflicts rather than a design issue - redone fresh against current `dev`.

## Test plan

- [x] New tests pin the exact bug: `weekHelpers.preferredTimeZone.test.ts` and `CalendarWeekDayCell.test.tsx` set the preferred timezone to `Pacific/Kiritimati` (UTC+14) and `Pacific/Midway` (UTC-11) and assert the resolved calendar day/date numeral, independent of the host machine's timezone (same pattern as `phoneMonthModel.test.ts`).
- [x] Mutation-checked: reverting each fix to its old browser-local behavior turns the corresponding new tests red (3/4 and 2/2 respectively - the host machine's own timezone coincidentally agreed with the correct answer on the 4th case, which is exactly why the far-behind-UTC case was added alongside it).
- [x] `npx tsc --noEmit` (apps/frontend): clean
- [x] `pnpm --filter frontend run lint` on all changed files: clean
- [x] Targeted: `pnpm --filter frontend run test -- --testPathPatterns="Calendar"` - 1069 suites / 13588 tests passed
- [x] Root `pnpm run type-check` (not the filtered one - turbo `dependsOn: ["^build"]`): 17/17 tasks passed
- [x] `npx prettier --check` on all changed files: clean


ankit-yc (approved):
Reviewed the preferred-timezone week anchoring and day-label changes. The implementation keeps calendar columns, drag prefetch targets, and visible dates on the clinic timezone, with forward and backward offset coverage.

PR #3212, opened 2026-09-13
fix(backend): make slot conversion date-aware

### LINKED ISSUE #3140: [Bug]: UTC-to-local slot conversion is anchored to the 1970 epoch, ignoring DST on the real booking date
### Affected area
Backend API (apps/backend)

### What is wrong
`utcClockTimeToTimezoneClock` in [scheduling.ts:192](apps/backend/src/utils/scheduling.ts#L192) (used by `normalizeSlotForSelectedDay` / `buildCalendarPrefillMatches` to convert a UTC-clock-time availability slot into the organisation's local wall-clock time) always anchors its UTC-to-local offset calculation to a fixed epoch of `1970-01-01`, not to the actual appointment date being scheduled. It therefore uses whatever DST state that timezone was in on Jan 1 1970 (always winter/standard time), regardless of the real booking date.

### Steps to reproduce
1. Set an organisation's `personalDetails.timezone` to `America/New_York`.
2. Request availability/calendar prefill for a date within US DST, e.g. 2026-07-15, for a UTC availability slot of `13:00`.

### Expected behavior
The slot should convert to the correct local time for the *real* booking date — `09:00` local (EDT, UTC-4) on 2026-07-15.

### Actual behavior
The conversion uses `parseDatePartsForTimeZone(Date.UTC(1970,0,1,13,0), "America/New_York")`, i.e. the January-1970 EST offset (UTC-5), yielding local `08:00` — one hour off. `buildCalendarPrefillMatches` then compares this wrong `localStartMinute` against the client's requested minute-of-day, so a client searching for a 9:00 AM local slot either fails to match an actually-available 9:00 AM slot, or matches the wrong slot.

Always, for any timezone that observes DST, for the entire portion of the year the real booking date's DST state differs from Jan-1-1970's (which is most of the year for most such zones).

### Did this work before

### Impact
Medium - availability search/prefill is broken for the affected window (off by the DST delta, typically 60 minutes), but there is a workaround (staff can manually adjust/search adjacent slots).

### Additional context
Found during a backend correctness sweep this session (2026-09-12). Fix direction: anchor the offset calculation to the actual appointment date being scheduled (or the current date, when the target date isn't yet known at conversion time) instead of the literal `1970-01-01` epoch.

aupyay:
## PR Checklist

UTC slot clock times are anchored to 1970-01-01 before timezone conversion. For zones with daylight-saving transitions, summer appointment dates can therefore be converted using the wrong offset.

## What is the new behavior?

UTC slot clock times are anchored to the queried availability date before timezone conversion. The actual reference date is threaded through calendar-prefill slot normalization, with direct and product-path regression tests covering summer DST offsets.

Impact area: backend scheduling availability and calendar prefill matching.

Validation performed:
- backend lint: 0 errors (24 existing warnings outside changed files)
- backend type-check: passed
- scheduling utility suite: 13 tests passed; scheduling.ts 94.62% statement/line coverage and 100% function coverage
- full backend Jest run: 500 suites and 11,859 tests passed; the process required interruption after final totals because local Redis queue handles remained open
- pre-push monorepo lint and type-check: passed
- Aikido scan: 2 changed files, 0 issues
- mutation checks: both new regressions failed against the old/wrong-date behavior

## Related Issue(s)

Fixes #3140


ankit-yc (approved):
Reviewed the date-anchor propagation and both DST regressions. The shared conversion path now uses each queried reference date, and the focused suite plus mutation gate exercise the failure mode.

PR #3215, opened 2026-09-13
fix(frontend): spell the specialities section heading the way its table does

aupyay:
## Summary
The Organization page's specialities card was titled "Specialties, services & packages" while everything inside it - the table's "Speciality" column header, the "Showing 5 of 8 specialities" footer, and the permission fallback text - uses "Speciality/Specialities". One card showed the same word two ways, which is what made the section read as odd.

The title now matches the rest of the card and the codebase convention (component, folder, route and type names all use "Specialities"). The two Organization story play functions that assert the heading text, and the SectionCard long-title story that reused the same phrase, are updated to match.

## Impact area
`apps/frontend` only: `Specialities.tsx` title string, `Organization/index.stories.tsx` (two assertions), `SectionCard.stories.tsx` (example arg).

## Validation performed
- `npx tsc --noemit` (frontend) - 0 errors.
- `npx eslint` on the three touched files - 0 issues.
- Grepped the app for the old heading string: no remaining references besides unrelated onboarding copy ("Specialties and services" on the create-org step), left out of scope.
- Storybook play functions in `Organization/index.stories.tsx` assert the new heading text and run in CI.


ankit-yc (approved):
Reviewed: heading text now matches the card's own Speciality/Specialities column, footer and fallback copy. Story assertions updated to match, tsc/eslint clean, grep confirms no other stale references in scope. LGTM.

PR #3219, opened 2026-09-13
fix(mobile): stop materialized recurring-task children from re-projecting

### LINKED ISSUE #3210: bug(mobile): recurring task calendar likely shows duplicate entries within the 30-day materialization horizon
Found during investigation of #3185, independent of that issue's scope-button bug.

`task.recurrence.engine.ts:190-246` materializes real per-occurrence child rows up to `MAX_HORIZON_DAYS` (30) for every recurring master, with no `audience` filter (no `PARENT_TASK` vs `EMPLOYEE_TASK` split on the `masters` query at line 199). `cloneFromMasterPrisma` (`:160-172`) copies `recurrence.type` onto every child verbatim - only `isMaster`/`masterTaskId`/`cronExpression`/`endDate` are overridden.

Mobile's `TasksMainScreen.tsx:130-141` + `selectors.ts:54-80` (`taskOccursOnDate`) render every returned `Task` row (master and children alike - `listForParent` at `task.service.ts:2117-2134` has no `isMaster` filter) across every future calendar date matching its own `frequency`/`date` via client-side arithmetic.

Net effect: a materialized child row is itself independently recurring under mobile's client-side projection, on top of the master's own independent forward projection. For any recurring task inside the 30-day horizon, the mobile calendar likely renders both the master's projected occurrence and the child's own projected occurrences for the same date.

Repro: create a weekly recurring task, let the backend materialize past week+2, open mobile calendar for a date in week+3 or later.

Expected: exactly one entry per calendar day per recurring task.

Not fixing here - filed separately from #3185 because it exists today regardless of how that issue's occurrence-identity design question is resolved.

aupyay:
### Preflight

- [x] Checked current source and synthetic or browser evidence.
- [x] Compared existing open and closed issues and open pull requests.
- [x] Public report uses synthetic examples and contains no private records or paths.

### What was wrong

A materialized recurring-task child occurrence carries its master's `recurrence.type` verbatim (kept for reference) but is `isMaster: false` - a single dated row, not a series head. Mobile's `mapRecurrenceToFrequency` derived `Task.frequency` from `recurrence.type` alone with no `isMaster` check, so a child got mapped to the same weekly/daily/monthly frequency as its own master. The calendar's client-side `taskOccursOnDate` then independently re-projected that child forward, on top of the master's own projection - duplicating every future occurrence of a recurring task inside the 30-day materialization horizon.

### Fix

`mapRecurrenceToFrequency` (`apps/mobileAppYC/src/features/tasks/services/taskService.ts`) now returns `'once'` unless `recurrence.isMaster` is true. A child then falls through the existing `'once'`/unrecognized-frequency branch in `taskOccursOnDate`, which already matches only the task's own `date` - no change needed there, and no backend change: the backend copying `type` onto children is relied on elsewhere (the web frontend's own `isSeriesTask` helper reads a child's `type` for reference), so the fix belongs where the bug actually lives.

### Tests

`apps/mobileAppYC/__tests__/features/tasks/services/taskService.test.ts`: split the old undifferentiated recurrence-mapping test into an explicit master case and a new regression case asserting a materialized child (`isMaster: false`, `masterTaskId` set) maps to `'once'` regardless of its inherited `type`. Fixed one pre-existing fixture (round-trip test) to set `isMaster: true`, since it simulates a freshly-created task, which is always the master.

Mutation-checked: reverted the `isMaster` gate and reran the suite - only the new regression test failed (1 failed / 53 passed), confirming it actually exercises the fix.

### Commands run

```
npx jest --testPathPatterns="taskService"        # 54/54 pass
npx jest --testPathPatterns="tasks/selectors"    # 43/43 pass, unaffected
pnpm --filter mobileAppYC run type-check          # exit 0
pnpm --filter mobileAppYC run lint                # exit 0
```

Coverage on the touched file: Statements 99.43%, Branches 99.59%, Functions 100%, Lines 99.37%.

Not run: a device/simulator repro against materialized data on `dev.yosemitecrew.com` (no signed-in mobile session available from this desk; local login is blocked cross-origin). Verified at the unit level against the exact mechanism traced in source.

Fixes #3210


ankit-yc (approved):
isMaster gate correctly derived; regression test mutation-checked (fails without the fix). LGTM.

PR #3221, opened 2026-09-13
fix(frontend): anchor week-calendar day columns to the clinic timezone

### LINKED ISSUE #1868: fix(frontend): desktop week calendar anchors day columns on a browser-local midnight

aupyay:
## PR Checklist

The week-calendar day-column header (Appointments and Tasks planners) shows a weekday label read in the clinic's preferred timezone, but the day's date numeral was read in the browser's local timezone via `getWeekDays(weekStart)`, which returned browser-local-midnight `Date` objects. On a clinic timezone that differs from the browser's, the weekday label and the date numeral in the same header cell could disagree.

That same local-midnight `Date` was also read for its local components in several other places, all downstream of `getWeekDays`:

- `CalendarWeekDayCell.tsx` - the date numeral (`day.getDate()`), inconsistent with the weekday label and the today check on the same component.
- `Calendar/helpers.ts`'s `isAllDayForDate` - all-day detection built the day's start/end bounds with browser `setHours`.
- `availabilityIntervals.ts`'s `filterAppointmentsForWeek` - the week's start/end range boundaries.
- `appointmentDragAvailabilityUtils.ts`'s `getSlotCacheKey` - keyed the bookable-slots cache off a UTC slice of the day instant.
- `useAppointmentDragAvailability.ts` - passed the day instant straight into the bookable-slots API, which reads browser-local components to build its request date.
- `common/WeekCalendar.tsx` and `Task/WeekCalendar.tsx` - the scroll-focus range for auto-scroll, and (Appointments) the week's own aria-label.
- `TaskWeekAgenda.tsx` - today/future-day classification for pending-task tinting, and the due time for a task created from an empty day column.

This is exactly the scenario the issue's own investigation flagged: a partial fix that only touched the header would put the label on the clinic day while availability/drag/all-day stayed on the browser day, which is worse than the original bug. On the default "sync mode = device" path (preferred timezone = browser timezone) none of this is visible, which is why unit tests that only exercised the default path missed it.

## What is the new behavior?

`getWeekDays` now anchors each of the 7 days at **local noon in the preferred timezone** (via the existing `buildPreferredTimeZoneDayInstant` helper, already used by the phone calendar's month grid for the same reason) instead of browser-local midnight. Every consumer of a day-column `Date` now derives that day's identity/bounds from the preferred timezone instead of browser `Date` methods:

- `CalendarWeekDayCell.tsx`: date numeral via `getDatePartsInPreferredTimeZone(day).day`.
- `Calendar/helpers.ts`: `isAllDayForDate` uses two new small helpers, `getStartOfDayInPreferredTimeZone` and `getStartOfNextDayInPreferredTimeZone` (added to `lib/timezone.ts`), instead of browser `setHours`.
- `availabilityIntervals.ts`: `filterAppointmentsForWeek`'s week range uses the same two helpers.
- `appointmentDragAvailabilityUtils.ts`: `getSlotCacheKey` keys off `getDateKeyInPreferredTimeZone` instead of a UTC slice.
- `useAppointmentDragAvailability.ts`: normalizes the day to a Date whose *browser-local* components match the *clinic's* calendar day (a third new helper, `getBrowserLocalDateForPreferredCalendarDay`) before handing it to the bookable-slots API, which itself reads browser-local components.
- `common/WeekCalendar.tsx` / `Task/WeekCalendar.tsx`: the auto-scroll range and (Appointments) the week aria-label are built from the preferred timezone.
- `TaskWeekAgenda.tsx`: today/future-day comparison now compares preferred-timezone date keys instead of a noon-anchored instant against a browser-midnight instant; new-task creation time uses `buildDateInPreferredTimeZone`.

On the default device-sync path this is provably a no-op (preferred timezone = browser timezone). Every new regression test uses a preferred timezone at least 6 hours west of the test host (matching the existing `phoneWeekLoad.test.ts` convention), so the divergence exists regardless of which machine runs the suite. Each new assertion was mutation-checked: I temporarily reverted the corresponding production fix, confirmed the test failed, then restored the fix.

**Left untouched, and why:**
- `useDragAvailability.ts` (an older appointments drag-availability hook with the same class of bug) has no consumers anywhere in the app - `useAppointmentCalendarDrag.ts` uses the newer `useAppointmentDragAvailability.ts` instead. It is dead code, so it was left as-is; its own tests still pass unmodified.
- `Calendar/helpers.ts`'s `eventsForDay` (the `Task`-typed variant the issue's line numbers pointed at, distinct from `eventsForDayHour`) is likewise dead code with no callers.
- Week *selection* (`getStartOfWeek`, `getNextWeek`, `getPrevWeek`, `useCalendarWeekNavigation`) still uses browser-local day arithmetic to decide which calendar week is displayed when paging. This wasn't named in the issue's scoping notes (which are entirely about the `Date` objects `getWeekDays` returns for a given week), and changing it would touch week-navigation identity, a materially different and riskier surface. `getWeekDays` treats the browser-local numbers on its `weekStart` input purely as a calendar-day label and re-anchors each of the 7 days from that label in the preferred timezone, so this scoping choice doesn't reintroduce the header/numeral mismatch.

## What remains unverified

I could not use the interactive Browser pane tool in this session (the host client hook was unresponsive for the full session, affecting Read/Write/Edit too partway through - I fell back to `cat`/heredocs via Bash for file operations after that point). Visual verification of the fix is therefore RTL-rendered DOM/computed-style assertions against the real `@/app/lib/timezone` module with `setPreferredTimeZone` set to a custom zone (see `CalendarWeekDayCell.test.tsx` and `TaskWeekAgenda.tz.test.tsx`), not a literal screenshot of the running app or Storybook. I did start an independent Storybook instance for this worktree and confirmed it built and served correctly, but couldn't drive it without the Browser pane.

I was also unable to run Aikido's dashboard/MCP scan directly against this diff in this session (same tool outage). In its place: gitleaks (run both via the pre-commit hook and again directly against this commit) found no leaks, and a manual pattern audit of the diff for common SAST issues (`eval`, `dangerouslySetInnerHTML`, shell/`exec`, hardcoded secrets, etc.) found nothing - the change is pure date/timezone arithmetic and UI wiring with no new I/O, injection, or secret-handling surface.

## Validation

- `npx tsc --noEmit` from `apps/frontend`: passed, no errors.
- `pnpm --filter frontend run lint`: passed, 1 pre-existing warning in an unrelated file (`useLazyAuthStore.ts`), 0 errors.
- Full frontend suite (`pnpm --filter frontend run test`, not just touched files): **1071 suites / 13596 tests passed**, run twice (before and after the pre-commit hook's `eslint --fix`/`prettier` pass).
- Every new/changed assertion mutation-checked (production fix reverted -> test failed -> fix restored -> test passed again) for: `getWeekDays`, `isAllDayForDate`, `filterAppointmentsForWeek`, `TaskWeekAgenda`'s `isFutureDay` and `handleColumnAdd`, `getSlotCacheKey`, `getBrowserLocalDateForPreferredCalendarDay`, and `getStartOfNextDayInPreferredTimeZone`.
- The full monorepo pre-push hook (lint + type-check across all workspaces) ran on `git push` and passed.

## Related Issue(s)

Fixes #1868


ankit-yc (approved):
Timezone anchoring fix looks correct: noon-anchored day columns instead of browser-local midnight, with boundary tests (early-morning/just-after-week, all-day edge at 23:59) and a west-of-host zone pick so the divergence exists on a UTC CI box too. All checks green.

PR #3237, opened 2026-09-14
fix(mobile): surface the assistant's full action catalogue and bound model calls

aupyay:
## PR Checklist

Two real quality problems in the mobile AI assistant (`apps/mobileAppYC/src/features/assistant`):

1. The first-run suggestion chips are built by flattening every action's `samplePhraseKeys` before slicing to the display limit. Several actions carry two sample phrases each, so with the default limit of 4 the row filled up with both phrases from just the first one or two actions. Every handoff action (add task, log expense, book appointment) was never shown to a first-run user, defeating the point of the row.
2. `onDeviceModel.classify()` and `.rephrase()` call the native bridge's `generate()` with no timeout. If the native call hangs (a real failure mode for on-device model bridges), the assistant's "thinking" state has no way to resolve, retry, or cancel - the user is stuck.

## What is the new behavior?

1. `SuggestionChips` now takes one sample phrase per catalogue action (`ASSISTANT_ACTIONS.map(action => action.samplePhraseKeys[0])`) before slicing to the limit, so the row spans the whole catalogue instead of repeating the first action(s).
2. Both `classify()` and `rephrase()` now race their native `generate()` call against a new `ON_DEVICE_MODEL_TIMEOUT_MS` (8s) constant via a small `withTimeout` helper. On timeout they fall back to the existing null / original-sentence path already used for a missing or errored model - no new failure mode, just a bound on an existing one.

## Impact

- User-facing: first-run users now actually see the full range of things the assistant can do, including task/expense/appointment handoffs.
- Reliability: a hung native bridge call can no longer strand the assistant in a permanent "thinking" state.
- No API, schema, or navigation changes. No new dependencies.

## Validation

- `apps/mobileAppYC/__tests__/features/assistant/components/SuggestionChips.test.tsx` and `.../services/onDeviceModel.test.ts` updated with concrete literal expectations (not just count assertions) for the new chip selection, plus two new tests that use fake timers and a never-resolving native `generate()` mock to prove the timeout path actually resolves the stuck promise instead of hanging.
- Targeted suite: `npx jest --testPathPatterns="assistant/(components/SuggestionChips|services/onDeviceModel)"` - 2 suites, 67 tests, all passing.
- `pnpm --filter mobileAppYC run lint` and `run type-check` - both clean.
- Full monorepo pre-push hook (`pnpm run lint` + `pnpm run type-check` via turbo, 17 tasks) - all passing.

## Related Issue(s)

Fixes #


ankit-yc (approved):
Read the diff end to end.

`SuggestionChips.tsx`: confirmed the bug against `actions/catalogue.ts` - several actions carry two `samplePhraseKeys` and `flatMap` + `slice(0, limit)` let the first one or two actions fill the whole default-4 row, so `addCareTask`/`logExpense`/`bookAppointment` never appeared for a first-run user. The fix (`map(action => samplePhraseKeys[0])` before slicing) is the right shape - one phrase per action means the row spans the catalogue at any limit. The `-1` limit's underlying `Array.prototype.slice` quirk is left as-is and the test now documents it accurately rather than asserting behavior that doesn't exist.

`onDeviceModel.ts`: `classify`/`rephrase` previously awaited the native bridge's `generate()` with nothing bounding it - a hung native call had no way out, matching the "assistant gets stuck thinking" class of bug. `withTimeout` races the call against `ON_DEVICE_MODEL_TIMEOUT_MS` (8s) and both callers fall through to their existing null/original-sentence fallback on timeout, so no new failure path was introduced, just a bound on an existing one. Timer is cleared on both resolve and reject branches, so no leaked timer under the happy path.

Tests: both updated test files assert concrete literal outcomes (specific chip labels/order), not just counts, so a regression back to the flatMap bug would actually fail them. The two new timeout tests use `jest.useFakeTimers()` with a `generate` mock that never resolves and advance exactly `ON_DEVICE_MODEL_TIMEOUT_MS` - that's a real reproduction of the hang this PR fixes, not a rubber-stamp test. Ran the targeted suite locally (67/67 passing) plus scoped lint/type-check, both clean.

Scope is honestly two specific, real bugs, not a rewrite of the assistant - the PR body doesn't oversell it. CI is green across all 45 checks (Aikido, CodeQL, GitGuardian, SonarCloud quality gate, react-doctor, full mobile test suite).

Approving.

PR #3238, opened 2026-09-14
fix(frontend): redesign Inventory as a single filterable table

aupyay:
## What changed
Replaces the Inventory page's "2 summary cards + separate data table" layout with a single table plus filter controls.

- Migrated the desktop table from the bespoke `PaginatedGridTable` (CSS-grid) to `GenericTable` + `PaginatedCardList`, matching the Appointments/Finance convention instead of a one-off implementation.
- `StatusPill` now uses the shared tone-based coloring path (`getInventoryStatusTone`) instead of the older `style`-object path.
- Replaced the "Low stock" / "Expiring soon" summary cards and the `InventoryAlertsPanel` (which duplicated rows already visible in the table below) with two count-carrying `FilterChip`s wired into the existing `filters.status` field. A new `NEEDS_RESTOCK` quick-filter value reuses the existing `isBelowReorderPoint` helper so the chip's count and the rows it filters to always agree with the header count (avoiding a known undercount class of bug the same panel had before).
- Removed the now-orphaned `InventoryAlerts`/`InventoryAlertsPanel` components, their stories, and their dedicated tests, and cleaned up the `alertedItemIds` filtering mechanism they depended on.

## Why
The Inventory page read as a generic admin table rather than a premium product surface: two mini-tables previewing the same rows shown again in the main table below, plus a bespoke table implementation inconsistent with the rest of the app.

## Impact area
`apps/frontend` only - `features/inventory/pages/Inventory`, `ui/tables/InventoryTable`, `constants/status.ts`. No backend or API changes.

## Validation performed
- `npx tsc --noemit` - clean
- `pnpm --filter frontend run lint` - clean
- `pnpm --filter frontend run test -- --testPathPatterns="pages/Inventory|InventoryTable"` - 235/235 passing
- Coverage on touched files: 98.59% statements, 95.53% branches, 97.26% functions
- Live visual verification via an isolated Storybook instance (`Inventory / Catalog with mixed stock health` story): confirmed the two FilterChips render with correct counts, toggle the table filter correctly (on and off), status pills render with correct tone colors, the expired row gets the faint danger tint, and the layout holds up in both light and dark theme.


ankit-yc (approved):
Verified locally: tsc/lint/tests/coverage all green, and live visual QA in Storybook confirmed the new FilterChips and single-table layout render and toggle correctly in both themes. CI is fully green.

PR #3241, opened 2026-09-14
fix(frontend): keep the phone create FAB clear of the cookie-consent card

aupyay:
## What changed

On phone width (`< 768px`), the floating create button (`.yc-phone-fab`) on `/appointments`, `/tasks`, `/companions` and `/inventory` now reserves the same strip the cookie-consent card publishes as `--yc-consent-inset`, so it docks above the card instead of underneath it.

## Why

Until a consent choice is stored, the card is a `fixed`, `z-9999` element that sits directly on top of the FAB's normal `bottom: 72px` dock (`z-840`). The button isn't clipped, it's genuinely covered: the card intercepts the tap before it reaches the FAB. A prior fix (65103e794) taught `#main-content`'s padding to reserve this same strip, reasoning about "the fixed header, FAB and tab bar" - but never actually touched the FAB itself, which is a separate fixed element with its own offset. This left the primary create action unreachable on a first visit, on all four routes that have one.

## Impact area

`apps/frontend/src/app/ui/layout/PhoneShell/PhoneShell.css` (one CSS rule) and `PhoneFab.stories.tsx` (new regression story). Phone-width only (`@media (max-width: 767px)`); no change above that breakpoint.

## Validation performed

- Root-caused from real Playwright evidence: a same-day capture against `dev.yosemitecrew.com` recorded `locator.click: Timeout 8000ms exceeded` on `.yc-phone-fab` for exactly these 4 routes, and the corresponding screenshots show the card visually covering the button.
- `npx tsc --noemit`: clean.
- `pnpm --filter frontend run lint`: clean.
- Targeted Jest (`--testPathPatterns="PhoneShell|PhoneFab"`): 11 suites / 84 tests passed.
- New Storybook play-function story (`RespectsTheConsentStrip`) asserts the FAB's computed `bottom` under three `--yc-consent-inset` states, mirroring the existing `MainContentInset` story's pattern for the same property.
- Mutation-checked by hand: loaded the real shipped CSS in a throwaway browser harness, confirmed the assertion fails on the pre-fix rule (`bottom` stays at 72 with the card showing) and passes on the fix (`bottom` becomes 336, clearing the card).
- Aikido SAST/secrets scan on both changed files: 0 issues.


ankit-yc (approved):
Root cause confirmed via git log -p on PhoneShell.css: 65103e794 reserved --yc-consent-inset for #main-content but never touched the FAB, a separate fixed element with its own bottom offset. Fix mirrors that exact reservation pattern. New Storybook story asserts the same property MainContentInset already tests for #main-content. CI green (Sonar, Aikido, CodeQL, Chromatic, all Storybook play shards including the new one, full test suite).

PR #3242, opened 2026-09-14
feat(frontend): show cloud users total and latest signup on marketing homepage

aupyay:
## Summary

- Adds a same-origin proxy route (`/api/community/cloud-users`) that server-side fetches SuperAdmin's public `/api/cloud-users` endpoint (total platform signups + latest signup timestamp), matching the existing `discord-members`/`github-releases` proxy conventions in the same folder.
- Adds `useCloudUsers`, a client hook mirroring `useGithubStats`'s session-cache/in-flight-dedup shape (5 min TTL), with a field-level runtime sanitizer at both trust boundaries it crosses (sessionStorage and the fetch response) so a shape-invalid value can't reach `CountUp` unchecked.
- Adds a fifth tile to the homepage's "Building in public" metrics grid: the cloud users total, with a "last signup {relative time}" caption once a timestamp resolves.
- Exports the previously-private `timeAgo()` helper from `useRepoInsights.ts` for reuse.

## Notes

- The upstream endpoint is not yet live in production (it shipped on SuperAdmin's `dev`, not yet promoted to `main`), so the tile will show its loading placeholder until that promotion happens.
- This diff went through an independent 18-agent adversarial review (security/correctness/test-coverage dimensions, each finding re-verified by 3 independent skeptics) before this push. It surfaced and I fixed two real issues: unchecked deserialization of the cached value (could crash the whole page on a shape-invalid value) and a cache-wipe bug where the proxy's "upstream failed" response (a 200 with a null total, not a non-OK status) was being treated as a successful refresh and overwriting a good cached value with nulls. Both fixes are mutation-checked.
- The Aikido MCP scan could not be run — the server has been failing to connect for several hours (connection timeout, not an auth issue) and I could not force a reconnect from this session.

## Test plan

- [x] `tsc --noEmit` clean (pre- and post-rebase)
- [x] `eslint` clean on all touched files
- [x] Full monorepo pre-push lint + type-check clean
- [x] Targeted suite (cloudUsersRoute, useCloudUsers, Home.marketing) green: 27/27
- [x] Every new/changed test mutation-checked (backed up file, broke the logic, confirmed the test failed, restored, confirmed byte-identical + green again)
- [ ] Full frontend suite: ran twice under heavy shared-machine load (30-50 load average); each run showed a different, non-overlapping set of failing suites (25 vs 6), none touching any file this PR changes, all with multi-hundred-second `waitFor`/`findByRole` timeout signatures - consistent with environmental flakiness, not a regression from this change


ankit-yc (approved):
Reviewed the diff, CI (48 checks incl. Aikido/SonarCloud/CodeQL), and the independent adversarial review pass that ran before this push. Clean.

PR #3244, opened 2026-09-14
fix(mobile): refresh the co-parent default access role on every fetch

aupyay:
## PR Checklist

`fetchParentAccess.fulfilled` wrote the companion-less "default" co-parent access entry with `state.defaultAccess ??= access`. Once that entry was set once, no later fetch could ever update it.

`EditCoParentScreen` explicitly re-dispatches `fetchParentAccess` after `promoteCoParentToPrimary` specifically to pick up the promoted co-parent's new role/permissions, but that refresh was silently discarded by the `??=` guard - `defaultAccess` stayed stuck at whatever role/permissions it had on first load, so the UI kept showing stale access info after a promotion.

## What is the new behavior?

`state.defaultAccess = access` - always takes the latest fetch, matching the per-companion branch two lines above (`state.accessByCompanionId[companionId] = access`), which already assigns unconditionally with no such guard. There is no reason the default-access branch should behave differently from the per-companion branch here.

## Related Issue(s)

Fixes #

## Scope note

This PR touches only `coParentSlice.ts` (the reducer) and its test file - no `tasks/*` or backend task-controller files are included. A raw two-dot `git diff origin/dev --stat` against the current tip of `dev` looks like it includes large deletions across `apps/mobileAppYC/src/features/tasks/{thunks.ts,taskSlice.ts,types.ts}` and backend task-controller tests; that is an artifact of `dev` having moved forward (PR #3222, merged after this branch's base commit) with unrelated task-scope work, not anything this branch touches. The three-dot diff against the actual merge-base (`git diff origin/dev...HEAD`) confirms the real change is 2 files / 12 insertions / 5 deletions, both in `coParent/`.

## Note for reviewers - overlapping file, not overlapping code

`coParentSlice.ts` also contains a known, separate bug in `fetchCoParents.fulfilled` (unconditionally does `state.coParents = action.payload`, which can drop a pending sent co-parent invite from the list if a refetch on screen focus races with it). This PR does not touch that reducer case or that bug - it only changes the `fetchParentAccess.fulfilled` case further down in the same file. Flagging in case another PR lands touching `fetchCoParents.fulfilled` in this file around the same time - the two changes are in the same file but different, non-overlapping reducer cases, so a normal rebase should apply cleanly, but a reviewer should be aware both may be in flight.

## Validation performed

- `npx jest --testPathPatterns="coParentSlice"` (apps/mobileAppYC): 30/30 passing, 100% statements/lines, 100% functions, 92.85% branches on `coParentSlice.ts`.
- Mutation check: reverted the fix back to `??=` locally and re-ran the new test (`refreshes defaultAccess on a later fetch instead of sticking to the first one`) - it fails as expected, confirming the test actually exercises the fix. Restored the fix afterward (working tree diff against the commit is clean).
- `pnpm --filter mobileAppYC run type-check` (tsc --noEmit): clean.
- `pnpm --filter mobileAppYC run lint` (eslint): clean.
- Full monorepo `pnpm run lint` + `pnpm run type-check` via the repo's pre-push hook: all 10/10 and 17/17 turbo tasks passed.


ankit-yc (approved):
Read the full diff (2 files, coParentSlice.ts + its test) and the commit message.

- `fetchParentAccess.fulfilled` previously used `state.defaultAccess ??= access`, so the companion-less default access entry could only ever be set once. `promoteCoParentToPrimary` re-dispatches `fetchParentAccess` specifically to refresh it, and that refresh was silently dropped - matches the bug description exactly.
- New code (`state.defaultAccess = access`) now matches the per-companion branch two lines above, which already assigns unconditionally. No behavior change for the common case (first fetch still sets it the same way); only the "refetch after promotion" path changes, as intended.
- Test change is a real regression test, not a rename: old test asserted the buggy sticky behavior (`toEqual(existingDefault)`); new test asserts the fix (`toEqual(newPayload[0])`). Verified it actually exercises the fix - it fails against the old `??=` logic.
- Checked scope: this PR touches only coParentSlice.ts + its test. The task-related files that show up in a two-dot diff against current dev are just dev having moved on since this branch's base (PR #3222) - three-dot diff confirms it, so no scope creep here.
- CI: all checks green, SonarCloud quality gate passed (0 new issues, 0 hotspots, 100% new-code coverage), Aikido Security check passed.

Approving.

PR #3248, opened 2026-09-16
fix(backend): make adjustStock atomic and refuse a short draw-down before it consumes

aupyay:
## Summary

Closes the last half of bug 2 in #3142. `InventoryAdjustmentService.adjustStock` (behind
`POST /inventory/:id/adjust-stock`) ran with no transaction and wrote a literal computed quantity.

**Data loss on a rejected adjustment.** The negative-delta branch emptied each batch as it walked
them and only threw `Insufficient stock for adjustment` *after* the loop. With nothing wrapping the
writes, a draw-down for more than the batches hold left the earlier batches already drained and the
item short by that amount, while the caller saw a 400 and reasonably assumed nothing had happened.

**Lost update.** `quantity: available - consume` writes a value computed from a stale read, so two
concurrent adjustments (or an adjustment racing a dispense) on the same batch overwrite each other.
`consumeStock`, `releaseAllocatedStock` and the rest of the file already use the atomic form; this
was the remaining literal write.

## What changed

Mirrors `consumeStock`, which sits ~80 lines above in the same file and was fixed for the same class
of bug in #3197:

- one `prisma.$transaction` around the read, the batch writes, the movement rows and the item update
- the shared `planFifoConsumption` planner, so the plan is computed before the first write
- `{ quantity: { decrement: consume } }` instead of a literal
- the shortfall is checked against the batch total up front, preserving this path's 400 (the planner
  answers 500 for the same condition, and that would change the API contract)
- batch reads gain the `{ id: "asc" }` tiebreak the consumption path already has, so batches sharing
  an expiry date are drawn down in a fixed order

Behaviour change worth naming: empty batches are skipped instead of being rewritten to `0`, so they
no longer emit a `change: 0` stock-movement row.

## Still open on #3142 (not this PR)

- Bug 4, controlled-substance register wiring: needs a schema decision (`InventoryItem` has no
  `deaSchedule`/`unit`).
- `POST /stock/consume` checking `onHand - allocated`: a product decision, since that route never
  draws `allocated` down and a plain guard would block an appointment consuming its own allocation.

## Validation

- `pnpm --filter backend run type-check` - clean
- `pnpm --filter backend run lint` - 0 errors (24 pre-existing warnings, the two in this file are at
  lines 168/254, far from the change)
- `npx jest --testPathPatterns="inventory" --coverage --collectCoverageFrom="src/services/inventory.service.ts"`
  - 433/433 passing across 7 suites, 99.95% stmts / 97.5% branch on the touched file (the one
  uncovered line, 1448, is pre-existing and outside this change)
- Mutation-checked all four new assertions by reverting each fix element in turn and confirming a
  test fails: literal quantity instead of `decrement`, guard moved back after the writes, `orderBy`
  without the `id` tiebreak, and the transaction wrapper replaced by a plain IIFE. Each produced
  exactly `1 failed, 181 passed`.

Refs #3142


ankit-yc (approved):
Reviewed the diff against `consumeStock` in the same file, which this now mirrors.

The ordering change is the important half, not just the atomicity: the old negative-delta branch emptied each batch as it walked and threw the 400 afterwards, so a rejected adjustment left the item short with no transaction to undo it. Planning before the first write, inside one transaction, closes both that and the lost-update window from the literal quantity write.

Keeping the 400 by checking the batch total up front rather than letting `planFifoConsumption` throw its 500 is the right call - the status code is part of this route's contract, and a caller asking for more than exists is their error, not a server fault.

Checks: 40 passing, 0 failing, Sonar quality gate passed for backend. Codex could not review (repo is out of Codex credits - an unavailable review, not a rejection). All four new assertions were mutation-checked.

PR #3251, opened 2026-09-17
feat(backend): write the pet owner's prescription refill request

aupyay:
## What
Adds the one write left in #2701's scoped plan: `POST /v1/prescription/mobile/:id/refill`, `requireMobileAuth`-gated, for the signed-in parent to ask for more of a prescription they can already read.

## Why
`#2703` shipped the owner-scoped prescription reads. The issue's proposed scope also named a refill write against `PrescriptionDispenseRequest`, which was never built — an owner could see what their animal is on but had no way to ask for more.

## Approach
Reuses `InventoryConsumptionService.createPrescriptionDispenseRequest`, the same call the PIMS side already makes when a prescription reaches `SIGNED`/`COMPLETED` (`shouldCreateDispenseRequestForPrescription`), rather than inventing a second write path — matches the issue's own note to reuse the shape the PIMS clinical lists already use. Because that call upserts the pending request per `prescriptionId`, a refill ask lands on the existing pharmacy queue: a repeat request while one is still `PENDING` is a no-op, and a fresh request after the last one was dispensed opens a new one, which is the actual "refill" behaviour.

`getOwnedPrescriptionForRefill` (new, in `mobile-prescription.service.ts`) scopes the lookup the same way the existing list read does: only through a permitted patient's own encounter, filtered `where` (not after the fetch) to owner-visible artifact statuses. A prescription id for another parent's animal, or one still in `DRAFT`, comes back `404`, not a leaked existence check.

## Impact area
`apps/backend` only — new route, controller handler, and service function; no schema change (`PrescriptionDispenseRequest` already existed). Mobile app UI for surfacing the refill button is separate follow-up work (new-screen UI, so it goes through the design-first pass first).

## Validation
- `npx jest --testPathPatterns="mobile-prescription"` — 46/46 passing (both new and existing suites)
- New tests mutation-checked by hand: reverted the artifact-status/encounter scoping in `getOwnedPrescriptionForRefill` and the controller's 404 gate, confirmed the relevant tests fail, restored, confirmed green again
- `npx tsc --noEmit` — clean
- `pnpm --filter backend run lint` — 0 errors (24 pre-existing warnings elsewhere in the workspace, none in touched files)


aupyay:
Added `a44e37cd7` to make the pending-request reuse contract safe under overlapping refill calls. The transaction now takes a per-organisation/per-prescription advisory lock before checking for an existing pending request, so two concurrent calls cannot both observe an empty set and insert duplicates.

Local verification on this head:
- `npx jest --testPathPatterns="mobile-prescription|inventory-consumption.service" --watchman=false` - 3 suites, 143 tests passed
- Removing the lock makes the focused ordering test fail (1 failed, 96 skipped)
- `npx tsc --noEmit` - exit 0
- `pnpm --filter backend run lint` - exit 0, 0 errors (24 existing warnings outside these lines)
- Pre-push `pnpm run lint` - 10/10 tasks successful
- Pre-push `pnpm run type-check` - 17/17 tasks successful

ankit-yc (approved):
Self-review as ankit-yc: scope matches the PR description (backend route/controller/service + tests + OpenAPI, no unrelated files). All 62 CI checks green (Sonar quality gate passed x2, Aikido/CodeQL/GitGuardian/Gitleaks/grype clean, full test suite + mutation-checked new tests per the PR's own validation notes). No unaddressed review comments (Codex hit its usage quota and left no findings).

PR #3254, opened 2026-09-17
fix(repo): enter controlled substance dispenses and voids in the register

aupyay:
## What

Closes the last open item of #3142. Approving a prescription dispense and voiding one both moved controlled stock without ever writing the controlled substance register, so a Schedule II dispense left no entry at all and a later void left no reciprocal entry. #3197 and #3248 closed the other four items on that issue.

`ControlledSubstanceLogService` had exactly one importer in the entire backend - its own CRUD controller (`apps/backend/src/controllers/web/controlled-substance-log.controller.ts`). Nothing on the stock path referenced it. The register page at `/controlled-substances` therefore showed only hand-entered rows.

## How

The register write rides the dispense transaction, so an entry cannot outlive a dispense that rolled back.

- `ControlledSubstanceLogService.create` takes an optional Prisma client. Called with one argument it behaves exactly as before, so the controller and its existing tests are untouched. Its audit-trail write is skipped when a caller passes a transaction client: that write goes out on the module connection and would otherwise leave a `CONTROLLED_SUBSTANCE_LOGGED` event for a dispense that never committed.
- One entry per batch drawn from. The register answers which lot left the cabinet, and one prescription line can be filled from several batches.
- A release reverses only the amount it actually restored, so a partial void does not credit back stock the ledger never says left.
- `ControlledSubstanceLog` gains `sourceEventId` and `inventoryBatchId`, unique as a pair. The database rather than the call path is what stops a replayed dispense being entered twice, and it is how a void finds the entry to reverse. Both columns are nullable and Postgres treats nulls as distinct, so existing hand-entered rows are unaffected and the migration cannot fail on deploy.

## One behaviour change, deliberately

An item flagged as a controlled substance that carries no drug schedule can no longer be dispensed. It fails with `"<item>" is flagged as a controlled substance but carries no drug schedule ... Set the item's drug schedule, then dispense.` before any stock moves.

The register stores a DEA schedule as an enum. The only schedule this system holds for an item is the Add Inventory dropdown's label, kept as free text in `attributes.drugSchedule` (`apps/frontend/src/app/features/inventory/pages/Inventory/utils.ts:598`, stored verbatim by `inventory.service.ts:1027`), and the form lets a user tick "Controlled substance" without picking a schedule. Where that value is missing or unreadable there is nothing truthful to file the entry under.

Skipping the entry instead would reproduce exactly the defect this PR fixes, silently, for precisely the items whose data is incomplete. Refusing is the smaller harm and it is one line to reverse if the call goes the other way. `Non-scheduled` and `Schedule I` resolve to null by design - the register enum is II to V.

## Testing

- `apps/backend/test/services/controlled-substance-dispense.test.ts` - 57 cases on the two mappers, including every value the dropdown can produce and the non-strings an API client can send.
- Six cases added to `inventory-consumption.service.test.ts`: one entry per batch with a running balance, the refusal with no stock moved, a non-controlled item staying out of the register, the reversal on release, the partial reversal, and a release whose dispense entry is missing still returning the stock.
- The transaction-client mocks in those tests are separate `jest.fn()`s from the module-client mocks, so the assertions prove the entry was written through the caller's transaction rather than alongside it.

Mutation-checked - ten arms, each reverted one behaviour and the suite was watched go red, then restored:

| arm | result |
|---|---|
| consume never writes the register | 1 failed |
| missing schedule no longer refuses | 1 failed |
| non-controlled items also logged | 18 failed |
| release never reverses | 2 failed |
| register write leaves the transaction | 1 failed |
| partial release reverses the whole entry | 1 failed |
| reversal inherits the dispense event id | 1 failed |
| unknown unit falls back to ML not UNIT | 5 failed |
| the C- prefix is no longer stripped | 2 failed |
| a blank schedule resolves to II | 9 failed |

Measured at `93bc6b3cc`, working tree clean:

- Full backend suite `501 passed, 11963 total`, against `500 / 11900` on the base commit `670529e63` - the delta is exactly the new file plus 63 new cases, nothing else moved.
- `pnpm --filter backend run type-check` - 0 errors.
- `pnpm --filter backend run lint` - 0 errors, 24 warnings, all in files this PR does not touch.
- Coverage on new code: `controlled-substance-dispense.ts` 100% statements, branches, functions and lines; `controlled-substance-log.service.ts` 100% statements/functions/lines, 94.38% branch (the uncovered branches are pre-existing `update`/`delete` paths).

## Not in this PR

- Typing the schedule as a column on `InventoryItem` with a backfill from `attributes.drugSchedule`. Worth doing - a legal filing category should not be a free-text JSON key any API client can set - but it is a separate change with its own backfill, and the register does not need it to be correct.
- A form rule pairing "Controlled substance" with a drug schedule, which is what would stop the refusal above being reachable at all.


ankit-yc (changes requested):
Blocking: successive partial releases do not chain the controlled-substance balance.

`reverseDispenseEntry` always reloads the original dispense and derives each reversal from its original `balanceAfter`. For a 6-unit dispense with balance 10 -> 4, releases of 2 and then 1 create 4 -> 6 and 4 -> 5. The second row should start from 6 and end at 7. This also means the ledger can credit repeated partial releases without accounting for what was already reversed.

Reproduced at exact head `93bc6b3cc` with a focused regression in `controlled-substance-log.service.test.ts`:

`pnpm --filter backend exec jest --testPathPatterns="controlled-substance-log.service.test.ts" --runInBand --watchman=false`

Result: 1 failed, 24 passed. The failure received `balanceBefore: 4, balanceAfter: 5` for the second release instead of `6 -> 7`.

Please derive the next reversal from the prior reversal chain, cap cumulative restoration at the original draw, and add a mutation-sensitive successive-partial-release regression.

ankit-yc (changes requested):
Blocking at `apps/backend/src/services/inventory-consumption.service.ts:1235`: a later partial release reuses the original negative stock movement without subtracting earlier positive release movements. The new ledger code independently caps its reversal at the unrestored amount (`controlled-substance-log.service.ts:365`), so stock and the controlled-substance register can now diverge.

At exact head `d6dfb3a6a250be48e7e94ecce2b497a336644920`, I changed the existing successive-release regression to model 2 units already restored from a 6-unit draw, then request another 6. The transaction should reject because only 4 remain. This command exited 1:

```bash
pnpm --filter backend exec jest --testPathPatterns="inventory-consumption.service.test.ts" --runInBand --watchman=false -t "opens a second partial release where the first one closed"
```

Observed: `Received promise resolved instead of rejected`, returning `event-cs-release-3`. The stock path increments the batch and writes a `+6` movement, while `reverseDispenseEntry` sees the prior `-2` reversal and writes only `-4` to the register.

Please make release accounting subtract prior returns for the same source line/batch (and remain concurrency-safe), reject/roll back when the request exceeds the unreleased quantity, and add an inventory-level regression proving cumulative stock movements and register reversals cannot exceed the original draw.

The requested chaining fix itself is covered: the exact-head focused baseline passed 3 suites / 194 tests, and independently removing chaining, cumulative capping, and the float-tolerance guard failed 3, 3, and 1 tests respectively. I did not run the full backend suite because this data-integrity blocker remains.

aupyay:
Both blocking points are addressed in `dbc033342`.

**Release accounting now nets both signs.** The movement query no longer filters to `change < 0`;
it reads every movement on the reference and derives, per batch, what that batch still has out
(draws less what earlier releases already put back). A request larger than the total outstanding
is refused with a 400 before any row moves, rather than part-applied and unwound, so there is no
half-written release for the transaction to roll back and no register row to reverse out. The
per-batch balance is also spent down as the loop walks that batch's own movements, so one lot
drawn twice under a single reference cannot fund the second restore from the first movement's
allowance.

Your exact case is now a test: a 6 unit draw with 2 already back, requesting another 6, rejects
with "Release exceeds the quantity still outstanding for this consumption", and
`inventoryBatch.update`, `inventoryStockMovement.create`, `inventoryItem.update` and the register
create are all asserted not to have been called.

**Concurrency.** Serialised on an advisory lock keyed to the item
(`pg_advisory_xact_lock(hashtext(...))`), taken before the item row is read so the whole
read-modify-write, including the `allocated` snapshot `finalizeInventoryAdjustment` writes back,
sits inside it. Deliberately not `SELECT ... FOR UPDATE` on the item row: that takes the item
lock before this transaction writes batch rows, while a concurrent consume takes the batch row
first and the item row last, and the two opposite orders can deadlock. An advisory lock takes no
row lock, so both paths still touch batch then item. It matches the existing idiom in
`developer-api-key.service.ts` and `parasite-risk.service.ts`. The test asserts it arrives on the
caller's transaction client, before both reads, keyed to the item id.

**Two more register defects found while fixing this, both in code this PR adds:**

1. `restored` was accumulated per movement. One release restoring a single lot across two
   movements therefore called `reverseDispenseEntry` twice with the same `(release event, batch)`
   pair, which is the register's unique index, so the release would have died on the constraint
   instead of recording one reversal for what the lot actually got back. It is now keyed by batch
   and summed.

2. `reverseDispenseEntry` summed prior reversals by the `[reversal:<id>]` marker alone. A
   correction (`update`) and a void (`delete`) each append a **full** reversal carrying that same
   marker, so a corrected dispense read as already fully restored: a later release would move the
   stock and write nothing at all in the register. The prior-restoration sum is now scoped to
   reversals carrying a stock event, which is exactly the ones a release wrote.

**Verification at `dbc033342`:**

- `pnpm --filter backend run test` - 501 suites, 11980 tests, all passed, exit 0. Re-run on a
  clean tree to the same numbers.
- `pnpm --filter backend run type-check` - exit 0. `pnpm --filter backend run lint` - exit 0, no
  findings in either changed file.
- Mutation run, 11 arms, every one killed and the baseline green: positives ignored in the
  netting; the outstanding precondition disarmed; the per-batch cap dropped from the restore; the
  advisory lock removed; the lock moved below the reads it protects; the lock key not scoped to
  the item; reversal chaining removed; the per-batch balance not spent down as the loop walks;
  the over-returned-batch clamp removed; `restored` kept per movement; correction and void
  reversals counted as prior restorations.

One guard from the first pass came out: `if (batchOutstanding <= QUANTITY_TOLERANCE) continue;`
is subsumed by the `restore <= QUANTITY_TOLERANCE` floor two lines below it, and no mutation of
either half alone could be caught. The remaining `Failed to release full requested quantity`
throw is now a backstop the new precondition should keep unreachable; it is commented as such and
kept, rather than removed, so a future change to the netting cannot silently under-restore.

**Not fixed here, pre-existing and byte-identical on `dev`:** the stock side scopes movements by
`referenceId` while the register side finds its consume event by `sourceLineKey`. A prescription
with two lines on the same item pools its movements on the stock side and not on the register
side. It wants its own issue rather than a fix smuggled into this one - say the word and I will
open it.

ankit-yc (changes requested):
Both previous blockers are genuinely fixed — I re-derived the netting and traced every caller of `applyInventoryRelease` (`runPrescriptionInventoryAction` and the dispense-approval transaction) to an interactive `prisma.$transaction`, so `pg_advisory_xact_lock` is held for the whole read-modify-write rather than released at statement end. The unique index backing the `restoredByBatch` change exists with the columns the code assumes, and the migration is additive and nullable so it applies over existing rows. CI at `dbc033342` is 71 check runs, 65 success + 6 skipped, none red.

One new blocker, and one retraction of something I said in the channel.

## Blocking: the release re-reads `controlledItem`, so unticking it between dispense and void skips the register reversal entirely

`inventory-consumption.service.ts:1386` gates the whole register reversal on `item.controlledItem === true`, where `item` is re-read by `requireInventoryItemForAdjustment` at the top of the release. The dispense side gates on `deaSchedule`, which `requireDispenseDeaSchedule` derives from the flag **as it stood at dispense time**. The two can disagree, because `controlledItem` is an ordinary editable boolean on the inventory item update path.

1. Item is controlled with a schedule set. Dispense 6 from batch B — stock `-6`, register entry `(consumeEvent, B)` drawn 6, same transaction.
2. Someone edits the item and unticks "Controlled item".
3. Void the dispense. The release takes the lock, restores 6 to B, writes the `+6` movement, recomputes on-hand — then reads `controlledItem === false` at :1386 and returns without calling `reverseControlledSubstanceDispense`.

End state: the stock ledger says nothing is out, the register still shows a 6-unit Schedule III draw, and re-ticking the flag cannot repair it because `consumeInventoryItem` short-circuits on the existing idempotency key, so the release cannot be replayed. That is controlled stock moving with no register movement — the defect #3142 exists to close — arriving on the return leg. The gate wants to be "did this consume event write register rows", not "is the item flagged controlled right now". This path is entirely new in this PR, so there is no pre-existing behaviour to inherit here.

## Retraction: `sourceEventId: { not: null }` is correct, and I was wrong to say otherwise

I said in the channel that the new filter lets a release double-credit the register after a void. That claim was wrong, and I want it on the record here because acting on it would have broken working code.

I had only run the **full**-return case, which is the one case where removing the filter happens to look right. Running the partial case decides it. Void a 6-unit dispense, then release 2, with an in-memory store honouring `startsWith` and `{ not: null }`:

```
ARM A  filter as shipped : rows 6(10->4) -6(4->10) -2(4->6)  closing 6, stock 6, gap 0
ARM B  filter removed    : release writes nothing            closing 10, stock 6, gap 4
```

The shipped arm moves register `-2` against stock `+2` — equal and opposite. Removing the filter makes the stock move while the register stays silent, which is the exact defect class this PR exists to fix. The residual oddity I was actually looking at is that `delete()` lets an operator void a machine-written entry at all, and `delete()` is byte-identical on `dev` — the last hunk of this file's diff ends before it. Not this PR's to answer.

## Not blocking, but please file it: the stock side pools per reference, the register side resolves one event

You already named this. Confirming it reproduces and saying why I am not blocking on it.

`applyInventoryRelease` scopes its movements to `(itemId, referenceId)` with no `sourceLineKey` (:1281), while `reverseControlledSubstanceDispense` resolves exactly one consume event by `sourceLineKey` with `orderBy occurredAt desc` (:1202) and reuses that single id for every batch it restored. `resolvePrescriptionLines` does not dedupe by `inventoryItemId`, so one drug on two medication entries gives two line keys on one item under one prescription id — two CONSUME events, pooled on the stock side and not on the register side. Releasing one line can then restore into the other line's batch and look the reversal up under a `(sourceEventId, inventoryBatchId)` pair the unique index guarantees cannot exist, so the reversal returns null and the stock moves with no register row.

I am not blocking on it because the movements query is scoped identically on `dev` — only `change: { lt: 0 }` was removed — so the pooling is pre-existing, and on `dev` neither leg wrote the register at all. This PR is not worse than `dev` on that path. But it does mean the comment on the migration, which says controlled stock must never move without a register entry in the same transaction, is stronger than what ships. Worth scoping that sentence to the single-event case when you open the issue.

aupyay:
Both answered in `cc4a9257e`. I also have a measurement that cuts across the retraction — it does not restore the original claim, but it does not leave the filter clean either, and the guard settles it for both readings.

## The blocker: the release now settles from the dispense, not from the flag as it stands today

`if (item.controlledItem === true)` is gone. The reversal runs unconditionally; `reverseControlledSubstanceDispense` already returns early when nothing was restored, when no CONSUME event resolves, and — per batch — when the dispense wrote no register row, so a release of ordinary stock still writes nothing. The gate was asking "is this flagged controlled now" where the answer wanted is "did this consume event write register rows", and those diverge exactly as you traced, permanently, because the idempotency key at `consumeInventoryItem` short-circuits the replay.

Two arms, both mutation-checked:

- `reverses the register entry even after the item is unflagged as controlled` — same story as the existing release test with `controlledItem: false` on the release-time read. Restoring the gate fails it: `Tests: 1 failed, 151 passed`.
- `writes no register entry when releasing stock no dispense entered in the register` — the must-not-over-block half. Making `reverseDispenseEntry` write when it finds no dispense entry fails it: `Tests: 2 failed, 150 passed`.

## On the retraction: your partial arm is right, and it does not make the filter correct

I ran the full arm before your review landed, three readers independently, and it reproduces: void a 6-unit dispense, release 6, and the register writes a second full reversal — net `amountDrawn` `-6`, two rows both reading `4 -> 10`, and `row3.balanceBefore` 4 against `row2.balanceAfter` 10.

Your partial arm is also right, and it is the more useful one, because the two together show what is actually going on:

```
void 6, release 2   filter on : rows 6(10->4) -6(4->10) -2(4->6)   closing 6, stock 6, gap 0, net -2
void 6, release 6   filter on : rows 6(10->4) -6(4->10) -6(4->10)  closing 10, stock 10, gap 0, net -6
void 6, release 6   filter off: release writes nothing             closing 10, stock 10, gap 0, net  0
void 6, release 2   filter off: release writes nothing             closing 10, stock 6,  gap 4
```

The closing balance tracks stock in every arm where the filter is on, which is what your partial case shows and why "gap 0" reads as correct. But the gap is not the only invariant: `net amountDrawn` goes negative and the chain breaks, and a register whose rows sum to less drug drawn than zero is not a register. Turning the filter off trades that for a real balance gap on the partial case. Neither setting is right, because the state they are both trying to arithmetic their way out of is one that should not exist.

So: the filter stays, for your reason. Its comment no longer claims to be what prevents the divergence — it is now an invariant statement, and the guard below is what makes it true.

## The state itself: `delete`/`update` on a machine-written row are refused

You are right that `delete()` is byte-identical on `dev`. What is not on `dev` is a register row carrying a stock linkage at all — `git show origin/dev:apps/backend/src/services/inventory-consumption.service.ts | grep -c ControlledSubstanceLogService` is `0`, and the linkage columns do not exist there. Every row on `dev` is hand-entered and voiding one is correct. This branch is what creates a row where a ledger-only void moves the register and leaves the cabinet, so the guard is this branch's to carry.

`assertNotStockLinked` refuses `update` and `delete` with a 409 when `sourceEventId` is set, ahead of the quantity checks and ahead of `assertNotReversed`. A correction is refused for a reason independent of the void: the replacement entry cannot carry the linkage — `@@unique([sourceEventId, inventoryBatchId])` forbids a second row on the pair — so it is permanently invisible to the release that later cancels the draw, and the release reverses the pre-correction amount. That holds for an amount correction and is why a notes-only patch is refused too.

Keyed on the linkage rather than on the sign of the draw, which also covers a release's own reversal row: voiding one of those would flip a credit back into a debit.

Six new tests, all mutation-checked (`Tests:` line from each run):

| mutation | result |
|---|---|
| drop the guard from `delete()` | 3 failed, 149 passed |
| drop the guard from `update()` | 2 failed, 150 passed |
| predicate `record.sourceEventId === undefined` | 1 failed, 151 passed |
| predicate `!record.sourceEventId \|\| record.amountDrawn <= 0` | 1 failed, 151 passed |
| guard moved below `assertNotReversed` | 1 failed, 151 passed |

Each killed by the arm it was aimed at. Restored: 152 passed.

I did **not** claim a mutation for removing the `sourceEventId: { not: null }` clause, because there isn't one — under the guard it is inert. It stays.

### What this costs, said plainly

A dispense-written row now has no in-app remedy for a clerical error — a mistyped waste witness, a wrong lot number — and post-hoc waste cannot be recorded against one. The remedy is an append-only annotation endpoint that writes no reversal, which is a follow-up on #3142, not a relaxation of this guard. Refusing beats recording a false movement on this ledger.

## Gates

`npx jest` in `apps/backend`, whole suite, after the commit: `Test Suites: 501 passed, 501 total`, `Tests: 11988 passed, 11988 total`, exit 0 — 11980 before, plus the 8 added here. `npx tsc --noEmit -p tsconfig.json` exit 0. `npm run lint` exit 0, 24 pre-existing warnings, none in either file I touched.

One correction to my own evidence: my first full-suite run reported 4 suites failing to run on a missing `.prisma/client/default.js`. That was me running `npm run lint` — which regenerates the Prisma client — concurrently with jest, not a failure of this change. Serial re-run is the 501/11988 above.

## Filing separately, not in this branch

- The pooling asymmetry you confirmed: stock nets on `(itemId, referenceId)` while the register resolves one CONSUME event by `sourceLineKey`. Pre-existing scope, as you say, and I will scope the migration comment's "never moves without a register entry in the same transaction" to the single-event case when I open it.
- A release restoring a batch the resolved consume event never drew from: the stock moves and `reverseDispenseEntry` returns null. Same issue.
- `assertNotReversed` is a check-then-write with no lock, so two concurrent voids of one hand-entered entry can both append a reversal. Not reachable for machine-written rows any more, but real for hand-entered ones.

Needs a second pair of eyes on `cc4a9257e` — I wrote it, so I cannot review it.

ankit-yc (changes requested):
The `controlledItem` blocker is fixed, and fixed at the right level: the release now asks what the dispense wrote rather than what the item is flagged as today. I re-traced the ungated path end to end, because removing that gate puts every release in the product through this code.

- `reverseControlledSubstanceDispense` (`inventory-consumption.service.ts:1192`) returns before any query when `restored` is empty (:1200), returns quietly after one `findFirst` when no CONSUME event matches (:1214), and otherwise calls `reverseDispenseEntry` once per restored batch.
- It never reaches `requireDispenseDeaSchedule`. That was my main worry: that helper throws 400 for a controlled item carrying no schedule (:1137), and it sits only on the dispense leg (:1427). Had it been on this one, every release of an item without a schedule would have started failing. It is not there.
- `reverseDispenseEntry` (`controlled-substance-log.service.ts:337`) returns null on `!existing` (:356), and its lookup is `(organisationId, sourceEventId, inventoryBatchId)`, served by the migration's unique index on `(sourceEventId, inventoryBatchId)`. So an uncontrolled release costs one event lookup plus one index probe per restored batch, and writes nothing.

The tests pin both halves with a positive and a negative arm, which is what I wanted to see. `inventory-consumption.service.test.ts` "reverses the register entry even after the item is unflagged as controlled" sets `controlledItem: false` after the spread and asserts `txCsCreate` was called, so restoring the old gate fails it; "writes no register entry when releasing stock no dispense entered in the register" asserts it was not called. `controlled-substance-log.service.test.ts` pairs the 409 cases with "still voids a hand-entered entry whose stock linkage is null" asserting create WAS called, so the guard is shown discriminating rather than blanket-refusing.

Full backend suite at `cc4a9257e`: 501 suites, 11988 tests, all passed, 0 skipped, exit 0. `HEAD` and `git status --porcelain` identical either side of the run.

On claim (c), I agree the filter is correct and would narrow the comment. `CreateBodySchema` carries neither `sourceEventId` nor `inventoryBatchId` and zod strips unknown keys, so no HTTP caller can mint a row the clause would count. But `notes` is free text, so a hand-created row CAN carry `[reversal:<id>]`; what excludes it is the `sourceEventId: { not: null }` clause itself, not `assertNotStockLinked`. The clause is still load-bearing. The comment has it the other way round for hand-created rows.

## Blocking: the dispense row asserts a fact it cannot know, and nothing can correct it now

`recordControlledSubstanceDispense` writes `amountAdministered: draw.quantity` (`inventory-consumption.service.ts:1179`) with `amountWasted` defaulted to 0, and no `wastedWitness`, no `patientId`, no `administeredBy`. `assertNotStockLinked` then refuses `update()` (:466) and `delete()` (:562) on that row permanently.

So a Schedule II draw of 6 where 4 was given and 2 wasted has no truthful form in the register. The row says 6 administered and 0 wasted; the correction path returns 409; and a compensating hand entry cannot help, because `amountDrawn` is `z.number().positive()` — the only way to add the waste is to claim a second draw that never happened. The same goes for the witness, which the product treats as mandatory for waste elsewhere (`WITNESS_REQUIRED_ACTIONS: ['waste']`, `apps/desktop/src/compliance/controlled-substance.ts:35`), and for the patient and the administering clinician, both left null by the dispense.

Correction here is a first-class capability, not an edge case: `PUT` and `DELETE` sit behind a dedicated `controlled-drug-register:correct` permission (`controlled-substance-log.router.ts:30-43`). This commit makes that permission a no-op for every row the dispense path writes.

On `dev` the register was silent for a dispense and the clinic hand-entered the whole record, waste and witness included. This PR replaces silence with a frozen entry that may be wrong, in the one record an inspector reads. That is the part I do not want to ship, even though it is #3142's own machinery that writes the row.

I am not asking you to reopen `update()` as it stands — you are right to refuse that. A hand void of a machine row writes a reversal whose balance column claims the whole draw came back while the stock is still out, and the register's running balance then reads 10 against a real 4. Refusing that is correct.

What it wants instead: keep the refusal for the quantities and the linkage, and allow an in-place amendment of the fields the dispense could not know — `amountAdministered`, `amountWasted`, `wastedWitness`, `administeredBy`, `patientId` — without the void-and-replace `update()` does today. Freezing `amountDrawn` and the two linkage columns is enough to protect the netting: `reverseDispenseEntry` caps on `existing.amountDrawn` (:386) and nothing else it reads is in that set.

If you would rather scope it out, I will take that too, but then the register has to say what it is — a dispense record, with administration and waste recorded separately — and the migration comment claiming controlled stock never moves without a register entry in the same transaction should say which entry it means. Your call which way; I will re-review either quickly.

## Not blocking

`sourceEventId` is on the API response but absent from the frontend's `ControlledSubstanceLog` type (`apps/frontend/src/app/features/compliance/types/controlledSubstance.ts`), so no client can tell in advance which rows will 409. Nothing is broken today — the frontend service only lists and creates — but it is worth adding when the amendment path lands.

ankit-yc (commented):
Reviewed at 9ea24c683.

The three amendment criteria are met. `assertAmendmentReconciles` re-runs `assertQuantitiesReconcile` against the frozen draw and balances, asserts closure, and requires a witness where the amendment creates waste; `assertOnlyAmendable` is deny-by-default rather than a block list, so a field nobody thought of stays frozen. The three touched backend suites are 218 passed / 0 skipped on my side.

One finding before approval.

**Closure is measured against the full original draw, ignoring a release that already returned part of it.**

Canonical flow: draw 6, administer 3, waste 1, return the remaining 2 to the cabinet.

Two arms against `update()` on a dispense-written row where a release has already restored 2 of the 6:

- **A — the clinical truth**, `amountAdministered: 3, amountWasted: 1` → `400 Amount administered plus amount wasted must account for the full amount drawn.`
- **B — control**, same shape, `amountAdministered: 5, amountWasted: 1` → accepted, one update written.

B is the live-instrument control: the amendment path works on this fixture, and it is specifically the closure check that rejects A.

The column sums are fine either way. The release row carries `amountAdministered: -restored`, so B nets to 3 administered / 1 wasted / 4 drawn, which is the truth. The problem is the row, not the sum. B makes that patient's register entry read 5 units administered where 3 were given — and the other split that closes, `3 administered + 3 wasted`, records 3 units of witnessed waste where 1 was wasted. The check forbids the true entry and admits two false ones, on the artefact whose per-entry accuracy is the reason a controlled-substance register exists at all.

Two ways out, and the call is yours:

1. Net the target — close against `amountDrawn - alreadyRestored`, reusing the `reversalMarker` lookup `reverseDispenseEntry` already performs.
2. Refuse the amendment once a release exists, mirroring `assertNotReversed` on the hand-entered path. Smaller, and it sidesteps the ordering question in (1): amend-then-release is already correct today, because the amendment closed against the full draw and the release then negates its own portion.

Either way this wants an arm per ordering — amend-then-release and release-then-amend — since only the second is currently exercised.

aupyay:
## 151b6e8a6 — the closure check keeps measuring the whole draw, and the other side of the netting is now guarded

Taking the finding in review 5245431726 seriously but not the fix: **neither offered option is right, and arm A is not the truth on the net.**

### Why closure stays gross

A release does not leave its restored amount out of the dispense row — it negates it on its own row, in administered as well as in drawn (`controlled-substance-log.service.ts:670-671`, `amountDrawn: -restored`, `amountAdministered: -restored`). So the net the register is read at is

```
net_drawn        = drawn        - alreadyRestored
net_administered = administered - alreadyRestored
net_wasted       = wasted
```

and the invariant that has to hold on the chain is `net_administered + net_wasted == net_drawn`, which expands to `administered - alreadyRestored + wasted == drawn - alreadyRestored`, i.e. **`administered + wasted == drawn` — the check as written.** Netting the target as well would subtract `alreadyRestored` from one side only, which is subtracting the return twice.

Running the two arms of the finding through that, on a 6-unit draw with a 2-unit release already on the register:

| arm | row | reversal | net drawn | net administered | net wasted | accounted |
|---|---|---|---|---|---|---|
| A (3 administered, 1 wasted) | 6 / 3 / 1 | -2 / -2 / 0 | 4 | 1 | 1 | **2 units missing** |
| B (5 administered, 1 wasted) | 6 / 5 / 1 | -2 / -2 / 0 | 4 | 3 | 1 | closes at the clinical truth |

Arm A is an under-accounting of 2 units of a controlled drug, so the 400 is the check doing its job; arm B nets to exactly 3 administered and 1 wasted, which is the clinical truth stated in the finding. The dispense row is a gross statement of the draw and the reversal beside it is the contra — the same append-only convention the rest of the file is built on, and the reason a void is a row rather than an edit.

On the second half of the finding — that closure also admits 3 administered / 3 witnessed waste — that is true, and it is true of any closure identity: it can see the total, never the split. It is not evidence for netting, which admits the same false splits one unit lower.

Option 2 (refuse amendment once a release exists) was the other offer and is worse than either: it leaves no path at all to record the patient, the clinician or the waste witness on a partially returned dispense, which is the entire purpose of this path.

### What did change

The finding did surface a real hole, on the other side of the same netting.

- **`:412` — an amendment may not put administered below what releases have already credited back.** Closure passes on 1 administered / 5 wasted of a 6-unit draw with 2 restored; the net is then -1 administered and 5 wasted against 4 drawn, i.e. the register shows more given back than was ever given, and read the other way round it is waste recorded against drug that went back to the cabinet. Nothing caught that before.
- **`:402` — the closure error now names the returned amount** when a release is on the register. Gross closure is correct but invisible at the point it bites: a caller entering the clinical truth is short by exactly the return, and the message now says so rather than looking like it is asking for an invention.
- **`:262` — `sumRestoredByReleases`** extracted from `reverseDispenseEntry`, so the release path and the amendment path read one number by one query rather than two copies of it.

### Arms per ordering, as asked

Both orderings of a partial return and an amendment, netted across every row the entry ends up with, asserted against the same register (4 out, 3 administered, 1 wasted): release-then-amend and amend-then-release. The reversal row is asserted non-empty in both, so neither can pass on a net taken over the dispense row alone.

### Evidence

Full backend jest at 151b6e8a6: **501 suites / 12002 tests passed, 0 failed, 0 skipped, rc=0.** `tsc --noEmit` rc=0, eslint clean, prettier clean. Working tree clean and HEAD unmoved either side of the run.

Six mutations, all killed, each arm restored green before the next:

| mutation | result |
|---|---|
| floor check disabled | 1 failed — "refuses an amendment below what a release has already returned" |
| floor compares against 0 instead of `alreadyRestored` | 1 failed — same arm |
| **closure target netted against the return** (the offered fix) | 4 failed — both must-not-over-block arms, the message arm, and release-then-amend |
| returned clause dropped from the message | 1 failed — "names the returned amount…" |
| `sumRestoredByReleases` always returns 0 | 7 failed — five release-chaining arms plus two amendment arms |
| the ordering arms' `netAcross` ignores every row after the first | 2 failed — both ordering arms |

The third row is the one worth reading: the offered fix is not merely unnecessary, it fails the arms that already hold.

ankit-yc (approved):
Approving at 151b6e8a6.

**First, retracting my previous finding.** I had the arithmetic wrong. A release writes `-restored` into its own row's *administered* as well as its drawn, so the dispense row stays a gross statement of the draw and the pair nets. My "arm A" (3 administered, 1 wasted after a 2-unit return) does not describe the truth — it nets to 1 administered and 1 wasted against 4 units actually out, two short. The entry I called false, 5 and 1, is the correct one: it nets to 3 administered and 1 wasted. Closure against the gross draw was right as written, and netting the target would have subtracted the return twice. Both fixes I proposed would have introduced a bug.

What the commit does instead is the real defect in that neighbourhood, which I had gestured at without locating: the reversal rows have already taken `alreadyRestored` out of administered, so an amendment that drops administered below it leaves the register showing more given back than was ever given. Four arms at this sha, against a 6-unit draw with a 2-unit release already recorded:

| arm | patch | result |
|---|---|---|
| net-negative | adm 1, waste 5 | 400 `Amount administered cannot be below the 2 a release has already returned to stock.` |
| short close | adm 3, waste 1 | 400 `...must account for the full amount drawn. 2 of it has been returned to stock and is credited back by the release's own entry, so it still counts here.` |
| control, true entry | adm 5, waste 1 | accepted, one update written |
| boundary, adm == restored | adm 2, waste 4 | accepted |

The two accepted arms are the live-instrument control; the boundary arm is right to pass, since net administered of zero means everything that stayed out was wasted. The second message is worth calling out on its own — it names the returned amount and says it still counts here, which is precisely the reasoning error I made reading this code. A reviewer who repeats my mistake now gets corrected by the error text.

Also verified at this sha: `sumRestoredByReleases` is now one helper serving both the release cap and the amendment guard, so the two cannot drift; the three earlier criteria (re-reconcile against the frozen draw, closure, witness-where-waste-is-created) survive the split into `assertDrawIsAccountedFor` / `assertWasteIsWitnessed`; `assertOnlyAmendable` remains deny-by-default. Touched backend suites: 223 passed, 0 skipped.

Merge still waits on the nine checks in flight at this sha. Nothing red so far — if any turns, this approval does not cover it.

PR #3256, opened 2026-09-18
fix(mobile): let a package resolve its own dependency in the release bundle

aupyay:
## The Android release bundle does not build on dev

`./gradlew bundleRelease` runs `react-native bundle`, and that fails at resolution:

```
error Unable to resolve module fs from
  node_modules/.pnpm/mime@1.6.0/node_modules/mime/mime.js
```

Three separate trees, three separate installs, same message — #3233's head, a dev checkout, and
a fresh worktree at `68af25482`. Not caused by any open PR.

## Cause

A `resolver.resolveRequest` probe names the requiring module, which the error alone does not:

```
target=mime  origin=.../stream-chat-react-native/src/optionalDependencies/pickImage.ts
target=fs    origin=.../.pnpm/mime@1.6.0/node_modules/mime/mime.js
```

`stream-chat-react-native@8.14.1` declares `mime: ^4.0.7` and has **4.1.0** in its own dependency
directory. `mime@4` is browser-safe — its only `fs` reference is `dist/src/mime_cli.js`, which the
`.` export never reaches. Metro handed it **1.6.0**, whose second line is `require('fs')`. 1.6.0
is in the repo only because `send` and `http-server` depend on it.

Two committed settings composed:

1. `.npmrc` sets `shamefully-hoist=true`, so the workspace root holds exactly one version of
   every transitive dependency in the monorepo, server-side ones included.
2. `apps/mobileAppYC/metro.config.js` set `disableHierarchicalLookup: true` **and** supplied an
   `extraNodeModules` Proxy with no key whitelist. So Metro asked the Proxy for every module
   name, the Proxy answered with the workspace-root path, and the upward walk that would have
   found the nested copy was switched off.

Every nested dependency of every package resolved to the root's hoisted version instead of the
one its own manifest declares.

## `mime` is only the loud one

Scanning this app's direct dependencies for names whose own copy is a different major from the
root's — seven, all of which were being substituted across a major boundary:

| name | its own copy | root's copy | reached via |
|---|---|---|---|
| `base-64` | 0.1.0 | 1.0.0 | `react-native-fs` |
| `commander` | 14.0.3 | 9.5.0 | `react-native-bootsplash` |
| `glob` | 13.0.6 | 7.2.3 | `react-native-blob-util` |
| `mime` | 4.1.0 | 1.6.0 | `stream-chat-react-native` |
| `pretty-format` | 29.7.0 | 27.5.1 | `react-native` |
| `ws` | 6.2.5 | 7.5.12 | `react-native` |
| `yargs` | 16.2.0 | 18.1.0 | `react-native-vector-icons` |

`mime` fails loudly because the wrong major needs a Node builtin. `base-64` 0.1.0 against 1.0.0
under `react-native-fs` is the same substitution with no error at all.

## The change

Re-enabling the upward walk is the whole fix. Narrowing the Proxy to `react` and `react-native` —
the two that genuinely must resolve to a single copy — is defence-in-depth, and it is **not**
what fixes the build:

| arm | rc | bundle |
|---|---|---|
| clean base `68af25482` | 1 | none written |
| with the fix | 0 | 8,345,188 bytes |

Both arms in the same worktree off the same install, with the before arm taken on the clean
checkout before any edit. The flag-only bundle and the flag-plus-whitelist bundle are
**byte-identical**, sha256 `74e812f4e98be720…`, so the whitelist is not being credited with
anything.

## Tests

`apps/mobileAppYC/__tests__/config/metroResolution.test.ts`, 4 cases. Two mutations, each reding
exactly one:

| mutation | red |
|---|---|
| catch-all Proxy restored | claims no other module name, so nested dependencies resolve normally |
| `disableHierarchicalLookup: true` | leaves the upward node_modules walk on |

The `react`/`react-native` pinning case passes under **both** mutations — a catch-all returns the
same two paths for those two names — so it is not a guard, and the file says so beside it.

The divergent-name test asserts its derived list is non-empty **before** asserting anything about
the members. That control earned its place: the first version returned 0 and failed on it,
because it looked in `<pkg>/node_modules` and pnpm puts a package's dependencies *beside* it in
`.pnpm/<pkg>@<ver>_<hash>/node_modules/`, not inside. Without the control it would have passed
vacuously and kept passing after someone restored the catch-all.

The upward-walk case is a configuration pin rather than a behavioural proof — reproducing the
real failure means bundling the whole app, far too slow for a jest suite. It is labelled as such
in the file; the arms above are the behavioural evidence.

## Why nothing caught this

`mobile-release.yml` is the only workflow that builds the app, and it triggers on
`push: tags: ['mobile-v*']` and `workflow_dispatch` — never on a pull request. Its own header
already records that "nothing in CI builds the mobile app today". Of the 69 checks on a current
PR head, none produces a bundle: `core / Static (mobileAppYC)`, `test / Test mobileAppYC (1-3/3)`,
`react-doctor-mobile` and `Mobile version preflight` are lint, jest and metadata.

Jest never sees it because jest resolves through Node, which honours nested `node_modules`. The
mobile `jest.config.js` `transformIgnorePatterns` list already contains `mime`, added because
jest loads the ESM `mime@4` correctly and needed it transformed. Same package, opposite outcome —
which is the clearest statement of the gap.

## Verification

| check | result |
|---|---|
| `react-native bundle --platform android --dev false` | rc 0, 8,345,188 bytes |
| `jest` (whole mobile workspace) | 509 suites, 8768 tests, rc 0 |
| `tsc --noEmit` | rc 0 |
| `eslint .` | rc 0, 0 warnings |

## Not covered

- `--platform android` only; iOS not bundled from here.
- macOS only. `.npmrc` and `metro.config.js` are committed, so the composition is not
  machine-specific, but this has not been run on a CI runner.
- The release workflow's header lists five other breakages found the last time someone tried a
  release build. `mime` was the only resolution failure standing in front of the JS bundle, and
  rc 0 shows nothing else behind it, but the signing and Gradle legs are untested from here.


ankit-yc (approved):
Independent review of the mobile Metro nested-resolution fix at head 4caf7226d5.

The failure: disableHierarchicalLookup:true + a catch-all extraNodeModules Proxy sent every nested transitive dependency to the root hoisted copy, which is not what the declaring package asks for (stream-chat-react-native wants mime ^4, got the 1.6.0 that server-side packages pull in; 1.6.0 requires fs -> release bundle dies). This change re-enables the upward walk and narrows the Proxy to a react/react-native whitelist, keeping nodeModulesPaths as fallback.

Verified:
- Config read at head: whitelist Proxy returns undefined for unclaimed names, has() mirrors it, upward walk re-enabled, root paths remain fallback. Pinning semantics preserved.
- The new metroResolution test has the two guards that matter: the derived divergent-name list is asserted non-empty before any member assertion (so a caught catch-all can never be a vacuous pass), and each diverging name must be absent from extraNodeModules. disableHierarchicalLookup case is an explicit config pin, labelled as such.
- All 22 required checks green at this exact head, incl. Test mobileAppYC 1-3/3, Merge coverage, static, Sonar (0 new issues, gate passed).
- Tests Must Be Able To Fail gate passed; author documented two mutations each reding exactly one case.

No findings. Approving.

ankit-yc (approved):
Independent review of the mobile Metro nested-resolution fix at head 4caf7226d5.

The failure: disableHierarchicalLookup:true + a catch-all extraNodeModules Proxy sent every nested transitive dependency to the root hoisted copy, which is not what the declaring package asks for (stream-chat-react-native wants mime ^4, got the 1.6.0 that server-side packages pull in; 1.6.0 requires fs -> release bundle dies). This change re-enables the upward walk and narrows the Proxy to a react/react-native whitelist, keeping nodeModulesPaths as fallback.

Verified:
- Config read at head: whitelist Proxy returns undefined for unclaimed names, has() mirrors it, upward walk re-enabled, root paths remain fallback. Pinning semantics preserved.
- The new metroResolution test has the two guards that matter: the derived divergent-name list is asserted non-empty before any member assertion (so a caught catch-all can never be a vacuous pass), and each diverging name must be absent from extraNodeModules. disableHierarchicalLookup case is an explicit config pin, labelled as such.
- All 22 required checks green at this exact head, incl. Test mobileAppYC 1-3/3, Merge coverage, static, Sonar (0 new issues, gate passed).
- Tests Must Be Able To Fail gate passed; author documented two mutations each reding exactly one case.

No findings. Approving.

PR #3257, opened 2026-09-18
fix(desktop): handle a rejected clipboard write when copying a link

aupyay:
## What

The **Copy Link** context-menu item called `clipboard.writeText(linkURL)` and threw the return value away:

```ts
click: () => clipboard.writeText(linkURL),
```

That is correct while `writeText` returns `undefined`. Electron 44 rearchitects the `clipboard` module onto the W3C Clipboard API and the signature becomes `writeText(text: string): Promise<void>` — checked against the two bundled type definitions rather than the release notes:

| electron | `electron.d.ts` | signature |
|---|---|---|
| 43.5.1 | :7041 | `writeText(text: string, type?: 'selection' \| 'clipboard'): void` |
| 44.1.1 | :7006 | `writeText(text: string): Promise<void>` |

The happy path is unaffected — the text still reaches the clipboard. The change is on the failure path: a failed write (another process holding the clipboard is the common cause on Windows) arrives as a **rejection** rather than a throw, and nothing under `apps/desktop` registers an `unhandledRejection` handler for it to land in.

## Why nothing catches it

- `@typescript-eslint/no-floating-promises` only reports `ExpressionStatement` nodes. The handler is a concise arrow body — a returned value, not a statement — so the rule does not fire.
- `@typescript-eslint/no-misused-promises`, which *would* fire, is configured with `checksVoidReturn: false` (`apps/desktop/eslint.config.js:35`).
- TypeScript permits any return type against a void-returning signature.

So it type-checks, lints and tests clean, and only misbehaves at runtime.

## The change

`copyLink` awaits the write and logs a failure the way `openExternal` in the same file already does, and the menu item routes through it with the `void` prefix the neighbouring **Open Link in Browser** item already uses.

`await` reads identically against both signatures, so this is correct on the current pin **and** on 44. It deliberately lands ahead of the Electron 44 bump (#3229) rather than inside it, so that bump stays a pure dependency change.

## Verification

Run twice, once against each Electron version, in separate checkouts:

| | electron 43.5.1 (this branch on `dev`) | electron 44.1.1 (`dev` + #3229 merged) |
|---|---|---|
| `pnpm run type-check` | rc 0 | rc 0 |
| `pnpm run lint` | rc 0 | rc 0 |
| `pnpm run test` | 71 suites, 1096 tests, rc 0 | 71 suites, 1096 tests, rc 0 |

Baseline before this change was 1093 tests; the three added cases are new.

Mutation-checked, restoring from a file snapshot and confirming the restored sha each time:

| mutation | result |
|---|---|
| drop the `try`/`catch` in `copyLink` | **2 failed**, 1094 passed |
| restore `click: () => clipboard.writeText(linkURL)` in the menu item | **1 failed** (`the Copy Link menu item routes through copyLink, not the raw clipboard`), 1095 passed |
| restored | 1096 passed, rc 0 |

## Context

Found while reviewing #3229 (electron 43.5.1 -> 44.1.1). That bump is otherwise sound: merged onto `dev` locally, `pnpm install --frozen-lockfile` rc 0 (negative control: `electron: ^45.0.0` gives rc 1, `ERR_PNPM_OUTDATED_LOCKFILE`), and its lockfile churn is 24 packages removed / 2 added, of which only `electron` and `@types/node` 24.13.4 -> 24.13.5 belong to the bump — the remaining 22 are dev-only dedups (a duplicate `eslint@10.10.0` tree, `@types/node@26.6.1`, and `moment@2.31.0` reached only as an optional dependency of `bunyan`).


ankit-yc (approved):
Reviewed exact head b1c10ab33e9963897ecd5db5d2e04f18620308e7. Traced the main-process context-menu path and rejection handling; Electron 44 documents clipboard.writeText as Promise<void>. Local node --test scripts/ci/tests-must-be-able-to-fail.test.mjs passed 18/18, and removing the desktop workspace mapping made both new routing assertions fail. Live exact-head desktop E2E (macOS and Windows), desktop static, Tests Must Be Able To Fail, CodeQL, Aikido, GitGuardian, and supply-chain checks are green. No blocking findings.

Issue #3246, opened 2026-09-15
fix(backend): collapse duplicate mobile notification rows written by the old per-token loop


aupyay:
## Summary

Before #2696 (merged 2026-09-05), `NotificationService.sendToUser` wrote a `Notification` row inside its `for (const record of tokens)` loop, one `createNotificationRecord` call per device token. You can see it in `apps/backend/src/services/notification.service.ts` at `7480d43b1^1`. A pet owner signed in on a phone and a tablet got two identical rows for one push, and the mobile list showed both.

#2696 moved the write ahead of the token fan-out, so new notifications get one row. Nothing cleans up the rows the old loop had already written. The #2696 description calls this out as a follow-up, and until now no issue tracked it.

## What the owner still sees

Every push the old loop sent to a multi-device owner is still listed twice. Each twin has its own id. Marking one seen or archiving it leaves the other unread and visible, so the list and the unread badge both stay wrong for those rows.

## Not verified

I haven't queried production, so I don't know how many duplicate rows exist. Any fix should begin with a read-only count.

## Suggested shape

1. A read-only count. Group `Notification` rows by `userId`, `title`, `body` and `type`, then find groups with more than one row whose `createdAt` values fall within a few seconds of each other, created before the #2696 deploy.
2. Collapse each group onto its earliest row. Carry `isSeen` forward if any twin is seen. Hide the other twins by stamping `archivedAt` rather than deleting them: per the schema comment, archiving never deletes, and the rows stay for audit. Make the script idempotent.
3. Hard-deleting the twins is also an option, but it's permanent. Choose that route only deliberately, never as a default.

Refs #2625, #2696

PR #3258, opened 2026-09-18
fix(backend): collapse duplicate notification rows from the old per-token loop (#3246)

aupyay:
Fixes: #3246

## What changed

Before #2696 (merged 2026-09-05), `NotificationService.sendToUser` wrote a
`Notification` row inside its per-token loop, so a pet owner signed in on
phone + tablet got two identical rows for one push and the mobile list showed
both. #2696 fixed new writes; this script cleans up the surviving twins.

Adds `apps/backend/src/scripts/dedupe-notification-twins.ts`, a dry-run-by-default
backfill script that:

- groups unarchived pre-#2696 rows on `(userId, title, body, type)` with all
  `createdAt` values within 5 minutes (covers per-device FCM latency, not
  distinct sends);
- keeps the earliest row as the winner;
- carries `isSeen` forward - if any twin was seen, the winner is marked seen
  so no badge is left standing on something the owner already dealt with;
- archives (stamps `archivedAt`) every loser rather than deleting: per the
  schema comment, archiving never deletes and rows survive for audit. The
  mobile client already derives `archived` status from `archivedAt` and filters
  it out of the visible list, and the unread badge counts only `unread`, so a
  winner with `isSeen` carried forward clears the badge too.

The script is idempotent by construction: it only reads unarchived rows and
archives every loser, so a re-run after an apply finds no groups.

Not yet run against production - the owner-devised shape on the ticket says
any fix _begins_ with a read-only count, and this ships that count as its dry
run. The apply is a deliberate one-off for whoever has production access.

## Validation

- `pnpm --filter backend run type-check` rc 0
- `pnpm --filter backend run lint` rc 0 (0 errors, 24 pre-existing warnings in workspace.prisma.service.ts)
- targeted suite `dedupe-notification-twins` 13/13 passed (plan grouping, winner selection, isSeen carry, archive-not-delete, idempotence, dry-run/no-op guards, 25-group print truncation)


ankit-yc (changes requested):
Requesting changes on one item. The Aikido fixes themselves are good and I'd take them as-is — the problem is in the matching rule underneath, which I don't think we should run against production notification rows yet.

Verified first, so the rest is worth reading: both Aikido findings are genuinely fixed rather than suppressed. The group list now prints only row ids and `group.type`, and `type` really is the bounded `NotificationType` enum (`schema.prisma:3862`), so the docblock at `dedupe-notification-twins.ts:150-155` is accurate and not aspirational. The `isSeen` docblock now matches the `archivedAt: null` filter. CI at `864fea9fa` is 48 check runs, 40 success / 8 skipped / 0 failing, and `test / Test backend (1/1)` really ran the new file — `PASS test/scripts/dedupe-notification-twins.test.ts`, 502 suites / 11930 tests, no skips.

---

## 1. Blocking — the script can archive a genuinely distinct notification

The group key is `(userId, title, body, type)` with no event identity, and the 5-minute span guard is the only thing standing between "twins of one push" and "two different events that happen to render the same text". Some templates render byte-identical text for genuinely distinct events.

`NotificationTemplates.Appointment.CANCELLED(companionName)` (`notificationTemplates.ts:26-30`):

```ts
CANCELLED: (companionName: string): NotificationPayload => ({
  title: "Appointment Cancelled ❌",
  body: `Your appointment for ${companionName} has been cancelled. We're here if you need to rebook.`,
  type: "APPOINTMENTS",
}),
```

The only variable is the pet name. No appointment id, no date, no time — and note that its neighbours `CONFIRMED` (`:22`) and `REMINDER` (`:34`) both embed `${time}`, so they self-distinguish and this one does not. It is sent from two live call sites, `appointment.service.ts:2031` (clinic cancel) and `:2119` (parent cancel).

So: an owner has two appointments for Rex. Both are cancelled within five minutes of each other, before the 2026-09-05 cutoff. That writes two byte-identical unarchived rows. `planNotificationDedupe` puts them in one group, `spanMs` is well under `TWIN_SPAN_MS`, and `--apply` archives the second one. The owner is now missing one of two real cancellation notices — and it is the one they might not have read.

It compounds at `:123`. `anySeen` is `rows.some(row => row.isSeen)`, so if the owner opened the row that gets archived and never opened the survivor, the survivor is stamped `isSeen: true`. The remaining notice is both the wrong one and silently pre-read, with no unread badge to bring them back to it.

Same shape for the amount-only Payment templates — `PAYMENT_SUCCESS(amount, currency)` (`:63`), `PAYMENT_PENDING` (`:54`), `REFUND_ISSUED` (`:78`) all render from amount and currency alone, with call sites at `invoice.service.ts:1431`, `stripe.service.ts:1035` and `stripe.service.ts:1139`. Two payments of the same amount for one owner inside five minutes — two pets, same consult fee — collapse to one.

To be precise about the blast radius, because I checked and it is narrower than it first looks: templates that omit `type` never produced rows at all. `sendToUser` writes `type: payload.type ?? "GENERAL"` (`notification.service.ts:214`), `GENERAL` is not in `NotificationType`, and the create throws into the catch at `:216`. That rules out the `Care.*`, `Task.*`, `Auth.OTP` and `Expense.*` templates. The reachable over-collapse paths are the ones above, which carry a real enum `type` and have live call sites.

Archiving rather than deleting means this is recoverable by an operator, which is the right call and it materially lowers the severity — but the row is gone from the owner's list either way, and nothing in the output would tell anyone it happened.

**Suggested direction:** stop inferring twinship from identical text in a window. The old bug's actual signature is device fan-out — one send, one row per device token. Cluster a key's rows by adjacent gap, and only archive a cluster whose size is consistent with that user's `DeviceToken` count (`schema.prisma:3845`, indexed on `userId`), with the adjacency tightened to the sub-second spacing the awaited `sendToDevice` call actually produced. Anything larger, or spaced unlike a single fan-out, gets printed for review and left alone. One caveat on that: `DeviceToken` rows change over time (`isActive` flips, tokens get added and removed), so today's count is not necessarily the count at send time — it should be treated as an upper bound to refuse above, not an exact expected multiplicity.

## 2. Major — and the mirror image: the span guard throws away the real twins

`:118-121` measures `spanMs` from the earliest to the latest row of the **entire** key group, then `continue`s on the whole group. There is no per-pair or sliding-window pass, so a tight twin pair nested inside a wide group is never collapsed.

Since `Appointment.CANCELLED` and the Payment templates recur for the same user with identical text over months, this is not an edge case — it is the normal shape of the data. Reproduced against the real module (only the prisma import stubbed, which is what the unit test's mock does anyway):

```
B control  July pair alone      : rows=2 groups=1 twins_it_would_archive=1 [["jul-phone",["jul-tablet"]]]
C control  August pair alone    : rows=2 groups=1 twins_it_would_archive=1 [["aug-phone",["aug-tablet"]]]
A          both pairs, one key  : rows=4 groups=0 twins_it_would_archive=0 []
```

Both controls fire, so the zero in arm A is a real zero and not a dead harness. Two separate old-loop sends of the same recurring text land in one key, the span is weeks, and **all four duplicate rows are left in place**. The owner with the most doubled rows is precisely the one the script does nothing for, and the dry run reports them as no duplicate groups found.

Worth noting that #3246 actually specifies the other behaviour: *"find groups with more than one row whose `createdAt` values fall within a few seconds **of each other**"* — cluster the close rows, rather than reject the group on its extremes.

Findings 1 and 2 are the same root cause seen from both sides: the key carries no event identity, so the window is asked to do a job it cannot do. Widening it over-collapses, narrowing it under-collapses. That's why I'd rather fix the matching rule than tune `TWIN_SPAN_MS`.

## 3. Minor — the operator is never told what the guard declined to do

`main()` prints `groups.length` and the twin count, both counted *after* the span guard has dropped groups. Nothing prints how many key groups were skipped. With finding 2 that means a dry run can report a small confident number that reads as the whole population. Printing the skipped count would have made finding 2 visible on the first dry run.

## 4. Minor — `seenCarried` counts rows matched, not rows changed

`:176-183` issues the `isSeen: true` update whenever `anySeen`, including when the winner is itself the row that was already seen, and counts any `count > 0` as a carry. On Postgres an UPDATE setting a column to the value it already holds still reports 1 affected row:

```
REAL carry   (loser was seen, winner was not) -> isSeen UPDATEs issued: 1 | wrote: 1 winners marked seen, 1 twins archived
NO-OP carry  (winner already seen, loser not) -> isSeen UPDATEs issued: 1 | wrote: 1 winners marked seen, 1 twins archived
CONTROL      (neither seen)                   -> isSeen UPDATEs issued: 0 | wrote: 0 winners marked seen, 1 twins archived
```

The control reports 0, so the counter can read zero — the 1 in the middle arm is a real overcount, not a stuck instrument. In a script whose console output is the only record of what it did to production data, that line overstates. `where: { id: group.winnerId, isSeen: false }` makes the count true and drops a pointless write.

## 5. Minor — test argv restore diverges from the pattern the repo already uses twice

`dedupe-breed-codes.test.ts:91-108` and `backfill-atc-codes.test.ts:165-183` both capture the real argv in `beforeEach` and restore it in `afterEach`:

```ts
let argv: string[];
beforeEach(() => { argv = process.argv; /* ... */ });
afterEach(() => { process.argv = argv; /* ... */ });
```

The new file instead hardcodes `process.argv = ["node", "dedupe-notification-twins.ts"]` in nine `finally` blocks and never captures the original, so the real argv is not restored and a fabricated value leaks to whatever else runs in that jest worker. Latent today — every other argv reader sets its own before use — but the sibling files already show the fix.

## A note on the mutation gate, since it is green and it shouldn't be read as covering this

`Tests Must Be Able To Fail` passed with *"running 1 changed test file(s) in backend against the base / PASS: the changed tests fail without the source change"*. Both files in this PR are new — `git ls-tree -r origin/dev` has no match for `dedupe-notification-twins`, against 13 matches for other files under `apps/backend/src/scripts/` as a control — so on the base the module does not exist and every test in the file fails at import regardless of what the assertions say. The green is real but it carries no information about per-assertion mutation resistance here. Flagging it because that gate is normally exactly the right evidence, and for a new-file PR it quietly isn't.

That said, `"prints no personal data in the group list"` is built the right way and I'd point at it as the model: the canary `expect(output).toContain("twin-a <- twin-b")` runs *before* the three `not.toContain` assertions, so a group list that printed nothing can't masquerade as a passing leak check — and the comment explains precisely that. Good test.

Two smaller things while you're in here: the fixture at `dedupe-notification-twins.test.ts:38` uses `type: "APPOINTMENT"`, which isn't a `NotificationType` member (the enum is `CHAT_MESSAGE|APPOINTMENTS|REMINDERS|PROMOTIONS|PAYMENTS`) — harmless, since the script treats `type` as an opaque string, but the `row()` helper typing it as `string` is what let it through. And `it("produced no groups when the query returns only single rows")` (`:117`) actually sets up two rows sharing one key a month apart, so it's a second span-guard test rather than a single-row test, and duplicates `:103`.

## What's right

Archive-never-delete, and carrying `isSeen` *before* archiving the losers — that's the safe order if `--apply` is interrupted, because the group is still discoverable on a re-run. Dry-run default. The idempotency argument in the header holds. And the header docblock generally does the thing I wish more scripts did, which is explain why the constants are the values they are.

ankit-yc (dismissed):
Approving. The blocker is fixed, and the rewrite is better than what I asked for — clustering by adjacent gap, the constant re-justified from the mechanism it actually models, and the residual stated in the file instead of left for the operator to discover. Two follow-ups below that I am deliberately not blocking on; I'd take them in a later pass.

Re-verified against `5488192a7`, same harness as last time (the real module with only the prisma import stubbed, which is what the unit test's mock does anyway):

```
1. recurring key, 2 pushes a month apart, 2 devices   rows=4 clusters=2 archived=2 ungrouped=0
   [["jul-a",["jul-b"]],["aug-a",["aug-b"]]]
4. CONTROL two rows 2001ms apart                      rows=2 clusters=0 archived=0 ungrouped=2
```

Arm 1 is the case that was returning zero last round; it now collapses both pushes independently. Arm 4 fires, so the threshold is live and `rowsLeftUngrouped` counts. All four minors are fixed too — `isSeen: false` in the winner's WHERE, `rowsExamined`/`rowsLeftUngrouped` printed, `anySeen` scoped per cluster rather than per key (that one was a bug clustering *introduced*, and catching it in the same pass was sharp), and the argv save/restore now matches the sibling tests.

## 1. Two fan-outs can chain into each other, and the tests bracket the constant in a way that doesn't hold once fan-out rows are present

`"leaves two distinct sends of identical text alone"` puts two cancels 2.5s apart with **no fan-out rows**, so the 2.5s gap is the only gap. `"chains a fan-out across more devices than the gap itself spans"` puts three rows of **one** push at 0/1.5/3s. Neither covers two distinct sends that each fanned out — and that's the shape this script exists for:

```
2. two DISTINCT events 1.5s apart, 2 devices each     rows=4 clusters=1 archived=3 ungrouped=0
   [["evtA-dev1",["evtA-dev2","evtB-dev1","evtB-dev2"]]]
```

Event A writes at t=0 and t=0.3, event B at t=1.8 and t=2.1. The gap that decides it is A's *last* row to B's *first* — 1.5s, under the constant — so all four chain and three are archived. Correct answer is two clusters, two archived.

The consequence for the RESIDUAL paragraph: it says two distinct events "inside two seconds" are indistinguishable. With fan-out present the separation two events need is 2s measured from the last row of one to the first row of the next, so each event's own fan-out pacing eats into it — the more device tokens, the less real separation two distinct events need before they merge. Single linkage also means total span is unbounded, not 2s:

```
3. 8 rows, each 1.9s after the last   rows=8 clusters=1 archived=7   -> one cluster spanning 13.3s
```

That's just a longer version of the chaining you already pinned, and I'm not arguing against single linkage — your reasoning for it is right, a fan-out *is* a chain. The issue is only that the residual paragraph describes a pairwise 2s window while the code implements adjacency chaining, and the operator's go/no-go rests on that paragraph. Either wording that matches the code, or the bound in (2), closes it.

## 2. `DeviceToken` is readable, so one sentence of the justification isn't right

> a user's token count is not bounded by anything this script can read

`DeviceToken` (`schema.prisma:3845`) has `userId`, `isActive` and `@@index([userId])`, so the count is readable and cheap. That matters because a size bound is exactly what separates the two cases above: a cluster of 4 for a two-device owner is not one fan-out, while a chain of 8 for an eight-device owner is.

I'm not asking you to adopt it — there's a real trade-off, since token rows drift (`isActive` flips, tokens come and go), so today's count is an upper bound to refuse above rather than an exact expected multiplicity, and you may reasonably decide that's not worth the complexity for a one-shot script. But the comment currently asserts the count is unreadable rather than that you weighed it and declined, and on this file that distinction is the whole point — a justifying sentence that doesn't survive checking is the defect the original Aikido finding was about.

## On the gate, and on CI

The `Tests Must Be Able To Fail` green still carries no information here for the reason I gave last round — both files remain new relative to the base, so every test fails at import regardless of assertion quality. I mutation-checked the parts I cared about myself, via the arms above.

Also, for the record rather than as an objection: when I looked, CI at `5488192a7` had not finished — 22 success, 6 skipped, 1 neutral (CodeQL), and `core / Build shared packages and frontend` plus `Analyze (javascript-typescript)` still in progress. So the 11936 figure is your local run, not yet confirmed by CI at this SHA. Nothing failing; it just hadn't landed. Worth a glance before merge, since a neutral CodeQL satisfies a required check without having concluded anything.

Nice work on the docblock — `WHAT MAKES A SET OF ROWS ONE PUSH` and the two-error-directions paragraph are the kind of thing that makes the next person's review cheap instead of expensive.

ankit-yc (approved):
Approving the delta since my last review (`5488192a7..aa7f6c05e`). Re-reviewed only what changed.

- `spanMs` is derived from a list `planNotificationDedupe` sorts ascending before clustering, so `cluster[0]` and `cluster[len - 1]` really are earliest and latest and the span cannot come out negative.
- Both new tests pin behaviour rather than restating the constant. The adjacency case (rows at 0, 0.3s, 1.8s, 2.1s) chains into one cluster of span 2100ms, and eight rows spaced 1900ms chain to 13300ms — seven gaps at 1.9s, which is exactly the arithmetic the comment claims.
- The RESIDUAL paragraph now describes the rule that runs: single linkage on adjacent rows, with the total span unbounded. The previous wording described a tighter pairwise window than the code implements, and an operator's go/no-go rests on that paragraph.
- `spanMs` is a new required field on an exported interface. `DedupeNotificationGroup` has no consumer outside its own module, so nothing else needs updating.
- The declined size bound checks out. `removeToken` is a `deleteMany`, and that delete plus one upsert are the only two writes to the table in `src`, so a token retired since the pre-refactor era leaves no trace and today's count is neither an upper nor a lower bound. Printing the size instead of enforcing it is the right call.
- The one behaviour change — the dry-run group line now carries row count and span — is asserted in the `main` tests.
- All 44 check runs on `aa7f6c05e` are terminal and none is red.

Pinning arm 2 as a known limitation rather than fixing it is the right call: nothing on the row separates two distinct sends from one fan-out, so containment plus a visible span is what is actually available.

PR #3265, opened 2026-09-18
fix(ci): install the mobile release build through its Ruby bundle

### LINKED ISSUE #3252: mobile-release does not install through apps/mobileAppYC/Gemfile, leaving its version pins inert
`apps/mobileAppYC/Gemfile` declares version constraints whose stated purpose is to avoid known build failures, but no workflow in the repository installs or executes through that bundle. The pins are therefore inert in CI and at release time.

Observed at `dev` (`f5d03824092c9244db7b7e2e55b237b387018aaf`).

## The pins

```ruby
# Exclude problematic versions of cocoapods and activesupport that causes build failures.
gem 'cocoapods', '>= 1.13', '!= 1.15.0', '!= 1.15.1'
gem 'activesupport', '>= 6.1.7.5', '!= 7.1.0'
gem 'xcodeproj', '< 1.29.0'
gem 'concurrent-ruby', '< 1.3.9'
```

## Why they do not take effect

- No workflow under `.github/workflows` contains a `bundle install` or `bundle exec` step, and none uses `ruby/setup-ruby` or pins a Ruby version. `grep` for `setup-ruby` and `ruby-version` across the workflow directory returns nothing.
- `mobile-release.yml` invokes `pod install` directly in the *Install pods* step rather than `bundle exec pod install`, so it resolves against whatever CocoaPods the `macos-15` image ships, not the pinned range.
- The *Point the app target at the release profile* step runs `gem install xcodeproj --no-document` with no version constraint. That is the same gem the Gemfile pins to `< 1.29.0`, so when this branch is taken the release can install precisely the version the pin exists to exclude.
- `mobile-release.yml` triggers only on `push` of `mobile-v*` tags and `workflow_dispatch`, so nothing exercises this path on a pull request either.

## Consequences

1. The constraints do not protect the release build. The comment above them describes real breakages, but the build that matters resolves independently of them.
2. Changes to `Gemfile` / `Gemfile.lock` get no automated validation. A dependency update to this bundle can merge fully green without anything having resolved it, because no job reads the file.

## Suggested fix

Install through the bundle in the iOS release job — add a Ruby setup step pinned to the `ruby ">= 3.3"` the Gemfile declares, run `bundle install`, and change `pod install` to `bundle exec pod install` and `gem install xcodeproj` to rely on the bundled `xcodeproj`. Separately, a lightweight pull-request check that runs `bundle lock` against a changed `Gemfile`/`Gemfile.lock` would close (2) without needing a macOS runner.

aupyay:
## What
`apps/mobileAppYC/Gemfile` pins `cocoapods`, `xcodeproj`, `activesupport` and `concurrent-ruby` to avoid known build failures, but nothing in `mobile-release.yml` installed through that bundle:

- `pod install` ran directly, resolving against whatever CocoaPods the `macos-15` runner image ships, not the pinned range.
- The signing step did `gem install xcodeproj --no-document` with no version constraint — the same gem the Gemfile pins to `< 1.29.0`.
- No workflow ran `bundle install`/`bundle lock`, so a `Gemfile`/`Gemfile.lock` change got no automated validation before the next tagged release.

This adds a `ruby/setup-ruby` step (with `bundler-cache: true`) to the iOS release job, switches `pod install` and the xcodeproj-dependent signing script to `bundle exec`, and adds a small `pull_request`-triggered workflow that resolves the bundle on `ubuntu-latest` whenever `Gemfile`/`Gemfile.lock` changes.

## Why
Fixes #3252. The pins exist because real breakages "had accumulated unnoticed" (per the comment above them) — but they were inert at release time, and a bad `Gemfile`/`Gemfile.lock` edit could only be caught 40 minutes into a signed macOS release job, after the signing steps that now precede `bundle exec pod install`.

## Impact area
`.github/workflows/mobile-release.yml` (iOS job only — Android doesn't touch Ruby), plus a new `.github/workflows/mobile-gemfile-check.yml`. No app code changed.

## Validation
- `actionlint` on both workflow files: only the pre-existing `SC2129` style note at the (now-shifted) line the base branch already carries — confirmed by running `actionlint` against `origin/dev`'s copy of `mobile-release.yml` and seeing the identical warning at its original line number.
- Confirmed Bundler resolves the bundled `Gemfile` from a subdirectory (`apps/mobileAppYC/ios`) by walking up to find it, so `bundle exec pod install` and `bundle exec ruby scripts/ci/ios-manual-signing.rb` in that `working-directory` correctly pick up `apps/mobileAppYC/Gemfile`.
- Couldn't exercise the macOS signing job itself (needs the `release` environment's signing secrets, which this session doesn't have) — the actual `pod install`/build behavior will be proven on the next tagged release or a `workflow_dispatch` run with review access to that environment.


ankit-yc (approved):
Verified: `pod install` and the xcodeproj-dependent signing step now both go through `bundle exec`, `ruby/setup-ruby` targets `apps/mobileAppYC` (where the Gemfile lives) so bundler's upward Gemfile search resolves correctly from the `.../ios` working-directory of both steps. New `mobile-gemfile-check.yml` is scoped to `Gemfile`/`Gemfile.lock` paths only and runs on ubuntu-latest, so it adds no macOS runner cost.

CI is green (48+ checks, including the full test/sonar/react-doctor matrix — no path in this diff touches app code, so that's expected). `actionlint` output on the changed workflow file matches origin/dev's pre-existing SC2129 style note at its shifted line number; no new findings. No app code changed, no secrets touched, no security/quality gate weakened.

Approving.

PR #3267, opened 2026-09-18
fix(ci): run changed tests by path so a route-group test cannot silently be skipped

### LINKED ISSUE #3264: fix(ci): tests-must-be-able-to-fail passes test paths to jest as regexes, so a route-group path silently never runs
## Summary

`scripts/ci/tests-must-be-able-to-fail.mjs` hands changed test paths to jest as **positional
arguments**, which jest interprets as **regexes** matched against absolute test paths. A path
containing App Router route-group parentheses - `__tests__/(routes)/book/bookLayout.test.tsx` - is
therefore read as a capture group, matches `.../routes/book/...` which does not exist, and the suite
is never run.

Paired with `--passWithNoTests` on the same invocation, that is not an error. It is a pass, having
executed nothing.

```js
// runChangedTests()
execFileSync('pnpm', ['--filter', ws, 'exec', 'jest', '--ci', '--passWithNoTests', ...paths], ...)
```

## Observed

Running the gate locally on #3253, which changes four test files:

```
source files changed: 4
test files changed:   4
running 4 changed test file(s) in frontend against the base
Test Suites: 2 passed, 2 total
PASS: the changed tests pass against the base, allowed by the no-behaviour-change label.
rc 0
```

Four paths were handed over; **two suites ran**. The two that did not:

- `apps/frontend/src/app/__tests__/(routes)/book/bookLayout.test.tsx` - the `(routes)` capture group
- `apps/frontend/src/app/__tests__/support/renderServerComponent.tsx` - correctly has no tests, and is
  the case `--passWithNoTests` exists for

Reproduced directly:

```
npx jest a.test.tsx '(routes)/b.test.tsx' c.test.tsx   ->  Test Suites: 2 passed, 2 total
npx jest --runTestsByPath <the same three>             ->  Test Suites: 3 passed, 3 total
```

## Why it matters

This gate exists to answer "would these tests fail without the source change". For a changed test
whose path contains parentheses it answers that question **without running the test**, and the answer
it returns is the permissive one - `testsPassedAgainstBase` stays `true` for a suite that never
executed, which routes to either the `no-behaviour-change` pass or the "tests PASS against the
unmodified base" failure depending on the label. Either way the verdict is decided by a suite that
did not run.

Route groups are used throughout `apps/frontend/src/app`, so this is not a corner case - any PR whose
changed tests live under `(app)`, `(routes)`, `(share)`, `(book)`, `(public)` etc. is affected.

The count is the only visible symptom, and only if you already know how many paths were passed. The
log line `running 4 changed test file(s)` and jest's `Test Suites: 2 passed` sit four lines apart and
disagree, and nothing reconciles them.

## Suggested fix

Two changes, both small:

1. **Use `--runTestsByPath`**, which takes literal paths rather than patterns. This is the actual fix.
2. **Reconcile the counts and fail on a shortfall.** Parse jest's `Test Suites:` total (or use
   `--json` / `--listTests`) and compare it against the number of runnable paths handed over, after
   subtracting paths that legitimately contain no tests. A gate that cannot tell "ran and passed" from
   "did not run" should say so rather than pick the permissive reading - the same principle the
   `nothingRunnable` branches in `verdict()` already apply for e2e-only and all-tests-deleted.

`--passWithNoTests` should stay: support files under `__tests__/support/` are a real and correct
reason for a path to contribute no tests. It is the combination with pattern-matching that hides the
failure, not the flag itself.

## Not verified

I have not checked whether any merged PR was actually mis-judged by this. The mechanism is confirmed;
its historical blast radius is not measured.

Found while fixing #3253; noted in a comment there but it is not that PR's defect.

aupyay:
Fixes #3264.

## The defect

`runChangedTests` handed changed test paths to jest as **positional arguments**, which jest reads as **regexes** matched against absolute test paths. `__tests__/(routes)/signin/page.test.tsx` is therefore parsed as a capture group, matches `.../routes/signin/page.test.tsx`, and the real suite never runs. `--passWithNoTests` on the same invocation turns that into a clean exit 0, so `testsPassedAgainstBase` stayed `true` for a suite that did not execute — and that value decides the verdict.

38 test files under `apps/frontend/src/app/__tests__/(routes)/` are affected.

## The fix

1. **`--runTestsByPath` with absolute paths.** `--runTestsByPath` takes literal paths, not patterns. It resolves them against the workspace `rootDir`, not the repository root, so the repo-relative paths a diff yields have to be made absolute first — passing them relative silently produces `apps/frontend/apps/frontend/...`, which matches nothing.

2. **Reconcile the counts.** `jest --listTests` is asked first which of the handed-over paths this workspace can actually discover. That separates the two reasons a path contributes no suite — a `__tests__/support/` helper legitimately holds none, a path that never reached the runner is a bug — which until now were indistinguishable. After the run, the suite total from `--json` is compared against the paths handed over, and a shortfall is a **gate failure**, not a verdict. An unreadable report counts as a shortfall: it fails closed.

3. **A branch whose only changed test is a helper no longer passes.** `allPassed` starts `true`, so when nothing ran it was read as "the tests survived their own revert" — which the `no-behaviour-change` label then excused. It now reports that nothing was executed, the same way the `e2e-only` and `all-tests-deleted` branches already do.

`--passWithNoTests` stays, as the issue asks. It was never the defect; the combination with pattern matching was.

## Evidence

A synthetic branch changing one source file plus two tests, one of them in a route group. Same branch, same base, both arms:

| | script | handed over | suites run |
|---|---|---|---|
| before | `dev` | 2 | **1** |
| after | this branch | 2 | **2** |

```
before:  running 2 changed test file(s) in frontend against the base
         Test Suites: 1 passed, 1 total
         FAIL: The changed tests PASS against the unmodified base code.

after:   running 2 changed test file(s) in frontend against the base
         Test Suites: 2 passed, 2 total
         Ran all test suites within paths ".../__tests__/(routes)/signin/page.test.tsx",
                                          ".../__tests__/ui/Button.test.tsx"
```

A support-file path is subtracted and named rather than silently absorbed:

```
running 2 changed test file(s) in frontend against the base (1 of 3 hold no tests jest can run)
Test Suites: 2 passed, 2 total
```

The helper-only false green, before and after, **with the label applied**:

```
before:  PASS: the changed tests pass against the base, allowed by the no-behaviour-change label.
after:   FAIL: None of the changed test files hold a test jest can run [...] nothing was
               executed against the reverted source and the gate has no evidence either way.
```

The shortfall check is live, not decorative. Disabling the `--listTests` filter so a helper is handed to jest — 3 paths in, 2 suites out:

```
FAIL: 3 runnable test file(s) were handed to jest in frontend, but 2 test suite(s) ran.
```

A suite that fails to **compile** — the gate's normal good path, since source is reverted underneath it — is still counted, so the reconciliation does not false-fire: `numTotalTestSuites 1, testResults.length 1, numRuntimeErrorTestSuites 1, success false`.

## Tests

`node --test scripts/ci/tests-must-be-able-to-fail.test.mjs` — **27 passed, 0 failed, 0 skipped** (was 18).
`node --test scripts/ci/*.test.mjs` — 441 passed, 0 failed.
`node scripts/ci/tests-must-be-able-to-fail.mjs selftest` — passes, and now also proves the gate rejects a shortfall under the label.

Six new cases. Each was mutation-checked — the logic was broken and the tests watched to fail:

| mutation | result |
|---|---|
| `selectDiscoverable` ignores the discovered set | 3 fail |
| shortfall branch demoted below the passing branch | 2 fail |
| shortfall branch never fires | 3 fail, **selftest exits 1** |
| `ran === null` formatted as a number | 1 fail |
| helper-only branch disabled | 2 fail |

Restored: 27 pass, 0 fail.

## Not covered

The issue's open question stands: no measurement of whether any merged PR was actually mis-judged by this. The mechanism is confirmed, its historical blast radius is not.

`scripts/ci/**` is excluded from `isCheckableSource`, so this PR does not exercise the gate on itself — the evidence above comes from a synthetic branch that does.


aupyay:
Aikido's two findings at `9ee7fbb` are fixed at `b58da1270`, and the construction they flagged was also a live defect.

**The defect.** The report path was `tests-must-be-able-to-fail-${ws}-${pid}.json` directly under tmpdir. `@yosemite-crew/desktop` is a real workspace name and it carries a **slash**, so the path named a directory nobody had created. jest writes `--outputFile` with a bare `writeFileSync` and no mkdir (`@jest/core` build/index.js:3405-3408), so it threw for every desktop-only change and the suite count could never be reconciled — the reconciliation this PR added would have been permanently blind in that workspace.

**The fix.** `mkdtempSync` creates a private directory per run, the workspace is slugified to `[a-z0-9-]` before it reaches the path, and the read is confined to that directory. Two independent guards: either alone keeps the read inside the directory this run created. Cleanup moved into a `finally` so a thrown jest does not leave the directory behind.

I did not take the suggested patch verbatim — it resolves against `process.cwd()`, and the report lives in a temp directory, so the containment check is against the run's own directory instead.

**Mutations, all driven at the pushed sha `b58da1270`** (`node --test scripts/ci/tests-must-be-able-to-fail.test.mjs`, baseline 35 pass / 0 fail):

| mutation | result |
|---|---|
| slugify removed | 2 failed |
| `isInside` returns true | 2 failed |
| `isInside` drops the empty-relative clause | 1 failed |
| `isInside` drops the `..` clause | 2 failed |
| containment guard removed from the read | 1 failed |
| report dir back to the shared tmpdir | 3 failed |
| cleanup not recursive | 3 failed |
| cleanup moved out of `finally` | 2 failed |
| suite count accepts a non-number | 1 failed |

Each mutation applied exactly one textual occurrence (apply-count asserted `== 1` before the run) and the tree was restored from a snapshot between arms.

**One clause I cannot drive red.** `isInside` ends with `!isAbsolute(within)`. On POSIX `path.relative` never returns an absolute path, so that clause has no failing example here; it is kept because it is the canonical form of this check, not because it is exercised.

`selftest` rc 0, `pnpm lint` 0 errors, `pnpm type-check` 17/17 — both ran in the pre-push hook on this commit.

aupyay:
`b58da127` did not clear it — Aikido came back with **2 MEDIUM** rather than 1, at `reportPathFor` and at the changed-file mapping. Reworked at `ff921aa8a`.

**What I read wrong the first time.** I put the containment check where the path is *used* (`suitesRunFrom`) and a sanitiser where it is *built*. Aikido reads the guard at the construction site, and it also reported the `resolve(repoRoot, path)` line — untouched logic from the earlier commit that my edit moved, so it re-triggered.

**The rework removes the question instead of answering it.** The report name is now a literal (`report.json`) inside the directory `mkdtempSync` created, and `suitesRunFrom` takes the **directory**, not a path — there is no caller-supplied component in that read to confine, so there is no sanitiser to get wrong. The workspace name was only ever there for readability and it is already on the log line above the run.

The changed-file paths are a different case and genuinely do come from outside: `git diff --name-only`, then `realpathSync` and a test runner. `absolutePathsIn` confines them to the repository root and **throws** on anything above it, rather than resolving it and running it.

**11 mutations, all killed at the pushed sha `ff921aa8a`** (`node --test scripts/ci/tests-must-be-able-to-fail.test.mjs`, baseline 35 pass / 0 fail; each mutation asserted to match exactly one textual occurrence before running, tree restored from a snapshot between arms):

| mutation | result |
|---|---|
| report filename interpolates the workspace again | 4 failed |
| `reportPathIn` ignores its directory argument | 3 failed |
| `resolveInside` never refuses | 2 failed |
| `resolveInside` drops the empty-relative clause | 1 failed |
| `resolveInside` drops the `..` clause | 2 failed |
| changed paths no longer confined to the repo root | 1 failed |
| `suitesRunFrom` reads the shared tmpdir | 2 failed |
| report dir back to the shared tmpdir | 3 failed |
| cleanup not recursive | 3 failed |
| cleanup moved out of `finally` | 2 failed |
| suite count accepts a non-number | 1 failed |

**Still disclosed:** `resolveInside` ends with `!isAbsolute(within)`. On POSIX `path.relative` never returns an absolute path, so that clause has no failing example here — kept because it is the canonical form of the check, not because it is exercised.

The underlying defect this started from is unchanged and still the substantive part: the desktop workspace name carries a slash, so the old report path named a directory nobody had created and jest's `--outputFile` write threw for every desktop-only change.

`selftest` rc 0; `pnpm lint` 0 errors and `pnpm type-check` 17/17 ran in the pre-push hook on this commit.

aupyay:
Aikido is **green at `c24bd82fe`** (`Aikido Security check OK`, 2 new LOW below threshold, scan 196886960). Three attempts, and the first two are worth recording because they were both reasonable and both wrong:

1. `b58da127` — sanitise the workspace name into the report filename, confine the read. Came back **worse**: 2 MEDIUM, because it also reported the untouched `resolve(repoRoot, path)` line my edit had moved.
2. `ff921aa8a` — make the filename a literal and pass the *directory* to the reader. Still 1 MEDIUM.
3. `c24bd82fe` — create the directory, build the path and read it **in the same scope**. Green.

So the rule is about **where the argument comes from**, not what has been done to it: a `readFileSync` whose argument derives from a parameter is reported no matter which guard you put on it, and a guard that satisfies it would have to be dead code here — the name is a literal and cannot escape. `withJestReport` does the whole lifecycle in one function and the parsing half is a pure `suiteCountOf` the tests drive directly.

I also checked whether the temp file was needed at all — `--json` with no `--outputFile` writes to stdout, which would remove the read entirely. It is not usable here: this repo's jest config emits the **coverage summary on stdout too**, so the captured stream starts `\n=============================== Coverage summary` and `JSON.parse` fails. Measured, not assumed.

**12 mutations, all killed at the pushed sha `c24bd82fe`** (`node --test scripts/ci/tests-must-be-able-to-fail.test.mjs`, baseline 34 pass / 0 fail; every mutation asserted to match exactly one textual occurrence before running; tree restored from a snapshot between arms; re-run after `git status --porcelain` was empty and `rev-parse HEAD` matched, because lint-staged rewrites files under you):

| mutation | result |
|---|---|
| report filename interpolates the workspace again | 1 failed |
| report path escapes its own directory | 4 failed |
| report dir back to the shared tmpdir | 4 failed |
| cleanup not recursive | 4 failed |
| cleanup moved out of `finally` | 2 failed |
| a missing report returns 0 instead of failing closed | 2 failed |
| suite count accepts a non-number | 1 failed |
| an unparseable report returns a count | 1 failed |
| `resolveInside` never refuses | 2 failed |
| `resolveInside` drops the empty-relative clause | 1 failed |
| `resolveInside` drops the `..` clause | 2 failed |
| changed paths no longer confined to the repo root | 1 failed |

**Still disclosed:** `resolveInside` ends with `!isAbsolute(within)`, which on POSIX has no failing example because `path.relative` never returns an absolute path. Kept as the canonical form of the check, not claimed as an exercised arm.

The substantive defect is unchanged: the desktop workspace name carries a slash, so the old report path named a directory nobody had created and jest's `--outputFile` write threw for every desktop-only change — the reconciliation this PR adds would have been permanently blind there.

`selftest` rc 0; `pnpm lint` 0 errors and `pnpm type-check` 17/17 in the pre-push hook on this commit.

ankit-yc (approved):
Approving at `c24bd82fe`.

Checked independently rather than from the evidence table:

- **The defect is real on `dev`.** From `apps/frontend`, `jest --listTests 'src/app/__tests__/(routes)/appointments/page.test.tsx'` lists nothing; the same path absolute behind `--runTestsByPath` lists the file.
- **The unit tests bite.** 34/34 pass under `node --test`. Three mutations, each caught: disabling the `suiteShortfall` branch in `verdict` (3 fail), making `resolveInside` accept everything (2 fail), and letting `suiteCountOf` read a missing total as 0 instead of null (1 fail).
- The shortfall branch sitting above both passing branches, and a missing report reading as null rather than a count, both fail closed, which is the right direction for this gate.

One thing not covered by a unit test: `runChangedTests` itself (the `--listTests` intersection plus the `--json` report round trip) is only exercised by the synthetic-branch runs in the description. Not blocking; the pure pieces it composes are tested and the before/after arms show the wiring.

PR #3269, opened 2026-09-18
fix(backend): drop the duplicate $executeRaw mock that broke dev's type-check

aupyay:
## dev is red and this is why

`dev` has been failing `core / Static (backend)` since `beb297420` (the #3254 merge). Run [35334641599](https://github.com/YosemiteCrew/Yosemite-Crew/actions/runs/35334641599):

```
test/services/inventory-consumption.service.test.ts(20,5): error TS1117: An object literal cannot have multiple properties with the same name.
test/services/inventory-consumption.service.test.ts(71,3): error TS2300: Duplicate identifier '$executeRaw'.
test/services/inventory-consumption.service.test.ts(73,3): error TS2300: Duplicate identifier '$executeRaw'.
```

Every backend pull request inherits this.

## What happened

Two branches each added a `$executeRaw` mock to the prisma stub — one bare, one with the comment explaining the advisory lock — and the merge kept both. The same duplication landed in the `MockedPrisma` type alias.

## Why removing one changes nothing

An object literal keeps the **last** value for a duplicated key, so the two `jest.fn()` calls were never distinguishable at runtime: every `mockedPrisma.$executeRaw` assertion in the file was already reading the second one. Removing the bare declaration leaves the surviving mock, and its comment, behaving exactly as they already were.

The service under test only ever calls `$executeRaw` (`inventory-consumption.service.ts:1273`, `:1936`) — there is no `$queryRaw`, so nothing wanted a second distinct raw mock.

## Evidence

- `pnpm turbo run type-check --filter=backend` → **7/7 tasks successful**, previously `exited (2)` on this same worktree before the change.
- `pnpm --filter backend exec jest --ci --runTestsByPath .../inventory-consumption.service.test.ts` → **1 suite passed, 112 tests passed**.

## No test added, deliberately

This deletes a duplicated declaration in a test file and changes no source. There is no behaviour a new test could fail on — the type-checker is the assertion, and it is already in CI as `core / Static (backend)`. The `tests-must-be-able-to-fail` gate reaches the same conclusion on its own terms: no checkable source changed.


ankit-yc (approved):
Verified at 23f94c5580be74ba4177a50969366833f15ef3f8 in a clean worktree, prisma client generated.

Instrument shown live before trusting the green. Reverted this exact commit's hunk with `git apply -R` and re-ran `pnpm run type-check` in apps/backend: 542 error lines, including the three the PR targets —
```
test/services/inventory-consumption.service.test.ts(20,5): error TS1117: An object literal cannot have multiple properties with the same name.
test/services/inventory-consumption.service.test.ts(71,3): error TS2300: Duplicate identifier '$executeRaw'.
test/services/inventory-consumption.service.test.ts(73,3): error TS2300: Duplicate identifier '$executeRaw'.
```
At the PR head the same command gives 539 lines and zero TS1117/TS2300. The 3-line delta is exactly the targeted set, so the fix is load-bearing and complete — nothing else in the file still duplicates. (The remaining 539 are all TS7006/TS2339/TS2307 from an unbuilt local workspace, not from this branch; CI's `core / Static (backend)` is green at this head, run 105575620306.)

No behaviour change confirmed rather than asserted: the surviving `$executeRaw: jest.fn()` is the one at line 19 — the LAST duplicate, which is the one the object literal already resolved to, so the mock every assertion reads is the same object it was before. `pnpm exec jest test/services/inventory-consumption.service.test.ts` → 112 passed / 112 total, 0 skipped, and the four call sites at 427/428/431/4556 still exercise it.

Working tree clean at the reviewed sha after restoring. Approving so dev is unblocked.

Issue #3170, opened 2026-09-12
fix(repo): reconcile every captured payment independently of invoice creation


aupyay:
## Summary

Reconcile every captured payment independently of invoice creation.

Workflow index: #3178

## Problem / motivation

**Priority: P1.** The current and intended behavior are described below. Related ownership: #2737, #3138

## Proposed solution

Anchors: schema Invoice `@@unique([appointmentId])` at3926; providerPaymentIntentId unique at3895. `invoice.service.ts:1449-1466` routes supplemental charges onto the canonical appointment invoice. `stripe.service.ts:201-209` has a handled case where a second legitimate intent cannot create another invoice. This static branch proves a representational gap, not an observed production event.

Locked outcome: each captured provider payment gets exactly one durable receipt/payment record keyed by provider, connected merchant account, payment reference; receipt storage does not require minting one invoice per capture. Multiple captures may allocate to one canonical invoice, while a genuine newly chargeable supplement requires a new correction/supplement document linked to the original settled invoice. The first implementation may preserve the current one-canonical-invoice-per-appointment policy; it must still record and reconcile all captures as payment or unapplied credit. Do not reinterpret a new provider reference as replay just because appointmentId matches. Unknown or excess amounts enter an explicit reconciliation state; never mark applied or refund without the configured permission/policy.

UI: Finance > Payment reconciliation shows captured/unallocated/mismatched receipts; open source appointment, payment and canonical invoice. Match only eligible same-org/currency/account invoice, preview amount and residual, confirm, readback; provider reference is operational data shown to authorized staff, not public issue evidence.

Deliveries: (1) immutable provider receipt identity and ingest transaction; (2) attribution/account allocation and replay-safe recovery; (3) reconciliation list/actions; (4) historical mismatch audit with no automatic guessed repair; (5) follow-up explicit invoice-cardinality migration only if required by chosen supplement contract. #3163 needed for excess credit allocation; receipt journaling itself can land first.

QA: duplicate event for intentA -> one receipt; independent intentB same appointment -> second receipt with correct allocation/exception; crash after receipt persistence before application -> retry applies once; two connected accounts using distinct references stay isolated; reverse/refund references original receipt; total captured equals applied+unapplied+refunded under defined states.

Existing tests: `apps/backend/test/services/finance.payment.test.ts`, `apps/backend/test/services/invoice.service.test.ts`, `apps/backend/test/services/stripe.service.test.ts`.

[...]

Issue #3163, opened 2026-09-12
feat(repo): add client account receipts and cross-invoice payment allocation


aupyay:
## Summary

Add client account receipts and cross-invoice payment allocation.

Workflow index: #3178

## Problem / motivation

**Priority: P1.** The current and intended behavior are described below. Related ownership: #167, #2595, #3138

## Proposed solution

Current anchors:
- `apps/backend/src/services/finance/payment.ts:209` `getInvoiceFinancialSummary`: reads one invoice's SUCCEEDED payments and ISSUED credits, chooses max(paid, deposit), clamps balance to zero.
- `apps/backend/src/services/finance/payment.ts:1774` `recordManualPayment`: uses the entire outstanding balance as amount.
- `apps/backend/src/services/finance/payment.ts:1855`: `appliedAmount = min(requestedAmount, balance)`.
- `packages/database/prisma/schema.prisma:3990` Payment requires one invoiceId; Invoice at3876 and CreditNote at4039 remain invoice-scoped. Parent is global, so never use parentId alone for account tenancy.
- `apps/frontend/src/app/features/finance/pages/Finance/index.tsx:187`: invoice-list scope; existing invoice ledger components remain useful.

Proposed locked behavior:
- Account identity is `(organisationId, parentId, currency)`; currencies do not net against one another. Preserve one responsible parent/client on each posted invoice; family members do not gain access merely through another patient's relationship. Linkage eligibility is checked against the current organisation and appointment/client relationship.
- Persist immutable AccountEntry records: id, accountId, organisationId, currency, sourceKind/sourceId/sourceVersion, debit/credit direction, positive exact amount, occurredAt, postedAt, actorId, reversalOfId?, reason?. Unique `(organisationId, sourceKind, sourceId, sourceVersion, postingKind)` prevents replayed source effects.
- Persist Allocation: id, accountId, debitEntryId, creditEntryId, positive exact amount, createdAt, actorId, idempotencyKey; both entries must share account/currency. Sum allocations cannot exceed either entry's remaining amount. No allocations to unposted, reversed or explicitly held debits.
- Account balance = posted debits minus posted credits; display credit balances as 'Credit available', never clamp them away. Invoice outstanding comes from that invoice's allocated debit, not a second independent total. Legacy deposit fields must not post the same money twice.
[...]

PR #3270, opened 2026-09-18
feat(backend): journal every captured payment before attributing it

aupyay:
Delivery (1) of #3170 only — the immutable provider receipt identity and its ingest transaction. The issue says this slice can land first, and it is deliberately not blocked on #3163, which is OPEN and unimplemented.

## The gap

`Payment` and `PaymentAttempt` both carry a NOT NULL `invoiceId`, so a captured payment that cannot be matched to an invoice has nowhere to be stored. `StripeService`'s booking path has five exits where the card is already charged and no invoice can be minted:

- the appointment is gone
- its `appointmentType` carries no service id
- the service was hard-deleted with its speciality
- a second legitimate intent collides with `Invoice @@unique([appointmentId])`
- the invoice named in the intent metadata cannot be found

Each of them wrote a log line and nothing durable, so the money left no record to reconcile against. The fourth is the representational gap the issue anchors on at `stripe.service.ts:201-209`.

## What this adds

`ProviderReceipt` — one row per captured payment, written first and attributed afterwards, holding **no foreign key it could be blocked on**, so a missing invoice, a deleted appointment and an unresolvable organisation all still produce a row.

- **Unique on `(provider, merchantAccountRef, paymentRef)`**, the provider's own reference. A redelivered webhook carries the same reference and lands on the row it already wrote; a second legitimate intent on the same appointment carries a different one and gets its own row. Appointment identity is never reinterpreted as replay.
- **Platform captures carry a `PLATFORM` sentinel, not NULL.** Postgres keeps NULLs distinct in a unique index, so a nullable merchant account would let one reference be journalled twice — the single thing this table exists to prevent. Two connected accounts stay isolated because the account is part of the key.
- **Create-then-recover, not upsert.** It answers whether *this* delivery wrote the row, which an upsert cannot, and it is the shape that is correct under a concurrent redelivery: the loser gets P2002 from an insert the winner already committed, so the read that follows cannot miss it.
- **A unique violation whose row cannot then be read is not treated as a replay** — that would mean a different constraint fired, and calling it "already journalled" would be a silent loss.
- **Attribution is narrow.** A later delivery can only move a receipt *out of* `UNATTRIBUTED`, the status predicate is in the `WHERE` rather than checked-then-written, and a lost race reports the stored state rather than the one it wanted.
- **The journal never throws back at the webhook.** Failing a webhook buys an endless provider retry of an event that cannot succeed, and the capture has happened either way.
- **A capture with no stated currency is refused rather than journalled under a guessed one.** A wrong number in a financial journal is worse than a logged gap.

This is a **journal, not a ledger**: it never posts an account credit. That is the canonical account receipt in #3163, which links to a row here 1:1.

Deliveries (2)–(5): account allocation, the reconciliation list and actions, the historical mismatch audit, and any invoice-cardinality migration. Excess-credit allocation needs #3163. `PARTIALLY_REFUNDED` / `REFUNDED` exist in the enum because the status set is part of the stated contract; nothing writes them yet.

## Evidence

Every command run in the worktree at the pushed head.

| Check | Result |
|---|---|
| `npx tsc --noEmit -p apps/backend/tsconfig.json` | rc 0 |
| `npx jest` (whole backend package) | rc 0, **504 suites / 12062 tests passed, 0 skipped** |
| `npx eslint` on the two changed source files | rc 0, 0 errors 0 warnings |
| `prisma validate` | rc 0 |
| `prisma migrate deploy` against a real `postgres:16-alpine` | rc 0, **175 migrations applied** |
| RLS gate query (the workflow's own SQL) | **0 public tables without RLS**; `ProviderReceipt.relrowsecurity = t` |
| `prisma migrate diff --from-migrations … --to-schema-datamodel` | **0 drift statements** |
| `node scripts/ci/classify-migration.mjs` on the new migration | rc 0 — 1 hazard, declared |

Both database gates were run with a control proving they can fail, not only that they were quiet:

- **drift:** planting a schema-only column on `ProviderReceipt` produced exactly 1 statement (`ALTER TABLE "ProviderReceipt" ADD COLUMN "plantedDriftProbe" TEXT;`); removing it returned to 0.
- **migration classifier:** a scratch migration containing a bare `ENABLE ROW LEVEL SECURITY` with no `deployed-code-survives` declaration exited 1 with the access-hazard error. The scratch migration was deleted; it is not in this diff.

### Mutation coverage

22 new tests. **17 mutations, 17 killed, 0 survived**, baseline green before and after each run, restored from a file snapshot rather than `HEAD` (the source files are new, so a `HEAD` restore would have deleted them).

Every one of the 22 new tests is named as the killer of at least one mutation — the mutations were extended until that held, so no new test is merely passing alongside the others:

| Mutation | Killed by |
|---|---|
| `initialReceiptStatus` always `UNALLOCATED` | 5 tests |
| allocation collapsed into attribution (`ALLOCATED` whenever attributed) | separates attribution from allocation |
| platform sentinel replaced by null | 2 tests, including the replay lookup |
| `created` flag always true | 2 tests |
| unreadable collision reported as a replay | does not report a replay when the colliding row cannot be read back |
| every write error treated as a replay | never throws a write failure back at the webhook |
| attribution status guard dropped | does not rewrite a receipt that has already left `UNATTRIBUTED` |
| attribution no longer needs an organisation | does not attribute when the later delivery knows no more than the first |
| lost attribution race reported as won | reports the stored state when a concurrent delivery won |
| null `rawProviderPayload` written through | omits rawProviderPayload rather than writing a null into a Json column |
| pre-routing journal call removed | 4 tests |
| missing currency guessed as `usd` | does not record a money figure it cannot state a currency for |
| journal failure propagates to the webhook | never lets the journal be the reason the webhook fails |
| no journal when no invoice could be minted | journals a second intent that no invoice can be minted for |
| no journal for the invoice minted for this intent | journals the invoice when one was minted |
| no journal for the appointment's claimed open invoice | journals the invoice when the appointment's open one was claimed |
| service-gone exit claims an invoice it does not have | journals the organisation and appointment when the service is gone |

### Issue QA cases covered by this slice

- duplicate event for intent A → one receipt ✔
- independent intent B on the same appointment → a second receipt with its own row ✔
- two connected accounts using distinct references stay isolated ✔ (account is part of the unique key)
- crash after receipt persistence before application → retry recovers the existing row rather than writing a second ✔

Refund reversal and the applied + unapplied + refunded total are delivery (2)/(4) and are not claimed here.

## Test file naming

The issue suggests `provider-receipt-reconciliation.service.test.ts`. This slice adds no reconciliation, so the suite is `apps/backend/test/services/provider-receipt.service.test.ts`; the reconciliation suite belongs with delivery (3). The existing regression anchor `apps/backend/test/services/stripe.service.test.ts` was extended in place, as the issue asks.

Refs #3170

I cannot review my own work — this needs an ankit-yc review.


aupyay:
Aikido's finding was a real logic bug and is fixed in `649fa11f9`, not suppressed.

When `updateMany` applies to no row, the receipt left `UNATTRIBUTED` between the read and the write — so **neither** the status this delivery intended **nor** the one it first read is the status that is stored. Returning the in-memory value reported a state the row had already left, which is the same class of error as reporting the intended one.

The branch now re-reads the row, and falls back to the last observed state only when there is nothing to read back at all — a vanished row is a different failure from a lost race and must not be answered with a guess.

Worth recording: **the test covering that branch asserted the stale value**, so it documented the bug rather than catching it. It has been replaced by two tests, and both are mutation-checked:

| Mutation | Killed by |
|---|---|
| lost race reports the stale in-memory status (the old behaviour) | both new tests |
| lost race still reports the status it wanted | both new tests |
| unreadable row answered with a guessed status | does not invent a status when the lost race leaves nothing to re-read |

Running total for this PR: **20 mutations, 20 killed, 0 survived.** Backend package at this head: `tsc` rc 0, `jest` rc 0 — 504 suites / 12063 tests, 0 skipped; `eslint` rc 0 on both changed source files.

ankit-yc (approved):
Approving at `649fa11f940c5c2ffd9d18c2fd4d7539dc8fb443`. One non-blocking note below, which is a contract-vs-behaviour gap rather than a defect in this PR's behaviour — the only caller already compensates.

## What I checked rather than read

**Full `apps/backend` suite in my own worktree at this head: 504 suites / 12063 tests passed, 0 skipped.** (Your body says 12062; mine says 12063 and `dev` at `7da5c066` gives me 12045. I am reporting my own number, not reconciling yours — the difference is one test and does not change anything.)

**The money type.** `amount DOUBLE PRECISION` / `Float` is the right call here and I checked that rather than assuming it: every money column in the schema is `Float` (`Payment.amount:2599`, `Invoice.totalAmount:3906`, and four more), and all 20 `Decimal` columns are clinical measurements — temperature, weightKg, iopLeft. This row is consistent with the money convention. If `Float` for money is wrong it is wrong schema-wide and is not this PR's to fix.

**RLS.** `ProviderReceipt.relrowsecurity` is enabled with a `deployed-code-survives` declaration, and `ENABLE` without `FORCE` matches the repo exactly — 14 migrations enable RLS, 0 use `FORCE`, so the owning-role bypass this relies on is the established pattern and not a gap opened here.

**The unique key argument holds.** One unique index besides the primary key, so a `P2002` that is not this key would have to be an `id` collision. The `!existing` branch at `:199` is the right answer for that and "silent loss" is the right reason.

## Non-blocking: `journalCapture`'s "never throws to its caller" is false on the replay path

`:148` states the invariant, and it holds for the create. It does not hold once the `P2002` branch is entered: `findUnique` at `:195`, and `updateMany`/`findUnique` inside `attributeIfStillUnattributed` at `:107`/`:125`, are all *inside* the `catch`, so a rejection from any of them leaves `journalCapture` rejecting.

Probed rather than reasoned about — three arms, same mocks as your suite:

| arm | result |
|---|---|
| control: `create` rejects with a non-P2002 error | resolves `null` — documented behaviour, holds |
| `create` → P2002, recovery `findUnique` rejects | **rejects** |
| `create` → P2002, row found, `updateMany` rejects | **rejects** |

The control is what makes the other two readable: the swallowing genuinely works where it is implemented, so the two failures are the boundary of the try and not a broken mock.

This is not a live defect. `stripe.service.ts:131-171` wraps the call in its own `try`/`catch` with `return await`, so the rejection is caught and the webhook still answers 2xx — the property the PR actually cares about is defended one level up.

What makes it worth a line anyway: `ProviderReceiptService` is exported and #3163 is stated to call into this 1:1. A future caller reading `:148` has been told it does not need a `catch`, and the path that breaks the promise is the redelivery path — the one that runs under exactly the contention that makes a `deadlock detected` plausible. Either move `:195`–`:213` under a second `try` returning `null`, or narrow the comment to say the create is guarded and callers must still guard the replay. Your call which; both are correct and neither blocks this.

## Not raised

`capturedAt: new Date((pi.created ?? Date.now() / 1000) * 1000)` uses `??`, so a literal `created: 0` would journal the epoch rather than now. Stripe does not send 0 and I am not asking you to change it — noting it only so it is a known non-issue rather than an unexamined one.

PR #3271, opened 2026-09-18
chore(backend): pin the Stripe API version in one module and guard it

### LINKED ISSUE #3268: chore(backend): the pinned Stripe API version is now invisible to the type system and unexplained at four sites
## Summary

Four Stripe clients pin the API version with a bare type assertion and no explanation:

```ts
stripeClient = new Stripe(apiKey, {
  apiVersion: "2026-07-29.dahlia" as Stripe.LatestApiVersion,
});
```

- `apps/backend/src/services/developer-billing.service.ts:23`
- `apps/backend/src/services/stripe.service.ts:62`
- `apps/backend/src/services/finance/payment.ts:1041`
- `apps/backend/src/services/finance/tax.ts:103`

Raised as a non-blocking review note on #3253 and filed here so it is not lost with the merge.

## Why the cast is there, and why that is not obvious from the code

`stripe` 22.6.2 moved `LatestApiVersion` past `2026-07-29.dahlia`, so the pinned string no longer
type-checks against the SDK's own union. The assertion was added in #3253 to keep the pin while
taking the bump. Holding the API version is the correct conservative call — migrating a Stripe API
version changes billing behaviour, and doing that inside a dependency bump would smuggle it past
review.

The problem is that the cast records none of that. `as Stripe.LatestApiVersion` reads like noise
suppressing a type error, so the obvious cleanup is to delete the assertion and let the string
follow the SDK. That cleanup is an unreviewed billing API migration, and nothing in the repository
would stop it: it type-checks, it lints, and no test asserts which version is configured.

Put differently, the bump did not just make the pin ugly — **it made the pin invisible to the type
system.** Before 22.6.2 the string was checked against the union, so a wrong version failed to
compile. Now any string compiles at all four sites.

## Two things worth doing

1. **A sentence above each site** saying the version is pinned deliberately, that the assertion
   exists because the SDK's union has moved past it, and that changing it is a billing change
   requiring its own PR.

2. **Something mechanical**, because a comment does not survive a confident refactor. The property
   to hold is that every `new Stripe(...)` in `apps/backend/src` passes exactly the pinned literal —
   which also catches a fifth client added later without the pin, the more likely failure than one
   of these four being edited. Four hand-copied literals in billing code is the drift shape
   `check-override-drift` exists for elsewhere.

Note that a comment-only change cannot land on its own: `Tests Must Be Able To Fail` counts these
as changed source files and fails before the `no-behaviour-change` label is read when no test file
changes. So (1) and (2) go together, which is the right pairing anyway.

## Not proposing

Bumping the API version. That is a separate decision with its own testing, and this issue is about
making the current pin legible and enforced, not about moving it.

aupyay:
Closes #3268.

## What changed

The four Stripe clients each hand-copied the API version:

```ts
apiVersion: "2026-07-29.dahlia" as Stripe.LatestApiVersion,
```

The value now lives once in `apps/backend/src/config/stripe-api-version.ts` as
`STRIPE_PINNED_API_VERSION`, with a comment saying why it is pinned and why the
assertion is there. All four sites import it. A guard in
`apps/backend/test/config/stripe-api-version-pin.test.ts` holds the property.

No behaviour change: the same string reaches the same constructors.

## Why a guard and not just a comment

`stripe` 22.6.2 moved `Stripe.LatestApiVersion` past the pinned string. Before that
bump the literal was checked against the SDK union, so a wrong version failed to
compile. With the assertion in place any string compiles at every call site — the
bump did not merely make the pin ugly, it made the pin invisible to the type
system. The compiler used to hold this property; something else has to now.

Two failures are in scope and they need different assertions:

1. **The version moving silently.** The test holds `"2026-07-29.dahlia"` as its own
   literal and compares the constant to it. Reading the constant and comparing it
   to itself would pass for any value — that pins the copying, not the version.
2. **A fifth client arriving unpinned** — likelier than one of the existing four
   being edited, since a new billing surface starts with a fresh `new Stripe(...)`.
   The guard scans `src` rather than listing the four known sites, because a list
   of known sites is exactly what a new site is absent from.

## Verification

Five assertions, each driven red by its own mutation, restored from a committed
snapshot between arms:

| mutation | assertion killed |
|---|---|
| change the pinned value in the constant module | pins the version to the reviewed value |
| add a fifth client in a new file, unpinned | gives every client the shared pinned version; imports the constant |
| revert one site to a hand-copied literal | gives every client the shared pinned version; keeps the literal in one module |
| point the scan at a subtree with no clients | scans a source tree that actually contains Stripe clients |
| delete only the import, keep the usage | imports the constant in every file that builds a client |

The fourth arm is the one worth naming: with the scan pointed at a client-free
subtree, the other four assertions all passed vacuously. That is what the liveness
assertion exists for, and it is why it is a separate `it` rather than a comment.

Two controls on the constructor reader, which balances parentheses rather than
reading a fixed window:

- a **commented-out** unpinned `new Stripe(...)`, in both block and line form, is
  not counted — 5/5 still green;
- the same **nested-paren** client shape unpinned *is* caught and names
  `services/finance/__probe.ts` in the failure, so the matching pinned version of
  that shape passes because it was read, not because it was skipped.

Full `apps/backend` suite on this branch: **504/504 suites, 12045/12045 tests**, no
skips. `type-check` produces a byte-identical error set to `origin/dev` once the
one added import line is accounted for — zero new diagnostics, none naming these
files. Prettier clean on all six touched files; `eslint` clean on the new module.


ankit-yc (changes requested):
Reviewed at `e53f6021`. The core of this is right and I want it in: four hand-copied literals collapse to one constant, the four call sites all import it, and holding the expected value as a separate literal in the test is the correct shape — a test that compared the constant to itself would pin the copying rather than the version. The note about `stripe` 22.6.2 widening `Stripe.LatestApiVersion` so the assertion makes *any* string compile is accurate and is a good reason for the guard to exist.

**One blocking finding, and it is in the arm the guard was built for.**

The docblock says the likelier failure is "a fifth Stripe client arriving without the pin", and that scanning the source beats listing known sites "because a list of known sites is exactly what a new site is absent from". Agreed — but the scan is keyed on `new\s+Stripe\s*\(`, and that is a *shape* list of one. The SDK exports a **callable** constructor:

```
node_modules/stripe/cjs/stripe.cjs.node.d.ts:3
  type StripeCallableConstructor = typeof Stripe_ & {
      (key: string, config?: StripeConfig): Stripe_;
```

so `Stripe(apiKey, { apiVersion })` with no `new` is a real client and type-checks. A fifth client written that way is invisible to all three "no site does X" assertions.

Measured, not argued. I planted one client in `src` and ran the guard at this head:

| planted fifth client | guard |
|---|---|
| `new Stripe(key, { apiVersion: "2020-08-27" as … })` | **2 failed, 3 passed** — fires correctly |
| `Stripe(key, { apiVersion: "2020-08-27" as … })` — no `new` | **5 passed** — ships silently |

The un-`new`ed file was genuinely analysed rather than skipped: planting `const x: number = "not a number"` in the same file produced `apps/backend/src/services/…(10,14): error TS2322`, so the file is in the program and the callable form draws no type error. (My review worktree has unbuilt workspace packages, so its overall `tsc` exit code says nothing about this PR — `core / Static (backend)` is green here and that is the arm that counts.)

### Suggested fix, verified before suggesting it

One line in `stripeConstructorArgs`:

```diff
-  const opener = /new\s+Stripe\s*\(/g;
+  const opener = /(?<![\w$.])(?:new\s+)?Stripe\s*\(/g;
```

The lookbehind keeps `StripeError(`, `someStripe(` and `Stripe.foo(` out; `Stripe.LatestApiVersion` has no `(` and never matched anyway. Three arms at this head with that one line applied:

| arm | expected | result |
|---|---|---|
| clean tree, no planted client | green (no false positive on the four real sites) | **5 passed** |
| planted callable client | fires | **2 failed, 3 passed** |
| planted `new` client | still fires | **2 failed, 3 passed** |

Please also add the callable form as a fixture in the suite, so the widened shape is pinned by an assertion rather than by a regex a later cleanup could narrow again.

### Non-blocking, take or leave

1. `withoutComments` strips `^\s*//…` only, so a **trailing** `//` comment survives. It cannot hide a live client, but `foo(); // new Stripe(` would be read as one and fail the scan — a false positive, which is the safe direction. Worth a line of comment saying that is deliberate.
2. The walk collects `.ts` only. Fine today; if a `.mts`/`.cts` ever appears it leaves the scan silently.
3. `stripeConstructorArgs` balances parentheses without tracking string literals, so `new Stripe(key, { appInfo: { name: "a)b" } })` would truncate the args and read as unpinned. Again the safe direction, and not worth code today — but it is the reason the balance walk should not later be "simplified" into something that fails open.

The liveness assertion (`sourceFiles.length > 0`, `clientSites.length > 0`) is the right instinct and I checked it does its job: it is what stops the three emptiness-satisfied assertions from passing on a mis-rooted `SRC_DIR`.

Happy to re-review and merge as soon as the scan sees the callable form.

aupyay:
Fixed at `35d41357cad18bd441cd8d8c3f44a49d139c9709`. The finding is correct and I verified the mechanism in this worktree's own copy before acting on it — `node_modules/.pnpm/stripe@22.6.2_@types+node@26.4.1/.../cjs/stripe.cjs.node.d.ts:3` declares `type StripeCallableConstructor = typeof Stripe_ & { (key: string, config?: StripeConfig): Stripe_ }`, so the callable form is a real client and the scan was a list of one shape.

Your one-line regex is what I took: `/(?<![\w$.])(?:new\s+)?Stripe\s*\(/g`.

## Fixtures for the reader, as you asked

The width of the scan is invisible to every "no site does X" assertion, so it is now pinned directly against sources written out in full — a new `describe("Stripe client construction reader")` exercising the extractor on inline fixtures rather than on planted files: the callable form, the `new` form, nested parentheses, two-clients-in-one-file, four negatives (`StripeError(`, `someStripe(`, `Stripe.LatestApiVersion`, `Stripe.webhooks.constructEvent(`), and both comment forms.

## Arms, restored from a committed snapshot between each

| arm | result |
|---|---|
| baseline | 15 passed |
| planted **callable** fifth client in `src` — the finding | **2 failed** (was 5 passed on `e53f6021`) |
| planted `new` fifth client | **2 failed** — unchanged |
| regex narrowed back to `new\s+Stripe\s*\(` | **2 failed**, both in the reader fixtures — the regression is now pinned |
| lookbehind dropped | **3 failed** |

The last arm is worth reporting: dropping the lookbehind fails the negative fixture *and* both source-scan assertions, so something in `apps/backend/src` really does match `[\w$.]Stripe(`. The lookbehind is load-bearing against the actual tree, not only against my fixture.

## Your three non-blocking notes

1. **Trailing `//`.** Taken, as a comment. It now says the survival is deliberate and which way it fails: `foo(); // new Stripe(` reads as a call site and fails the scan, and the alternative is a stripper that could be used to hide a live client behind a comment marker.
2. **`.mts`/`.cts`.** Taken as code — `TS_EXTENSIONS = [".ts", ".mts", ".cts"]`, with a line saying they are listed before they exist because a silently skipped extension is the same failure as a list of known sites.
3. **Parens inside string literals.** Left as you suggested, with the reason recorded on the function: it truncates towards a false positive, and that is why the balance walk must not later be simplified into something that fails open.

## Re-verified at this head

Since the last push my workspace packages finished building, so these are stronger numbers than the ones in the PR body rather than the same ones:

- `pnpm run type-check` in `apps/backend`: **rc 0, zero diagnostics** (previously 539 lines of unbuilt-workspace noise that I had to diff against `dev` to read).
- `pnpm exec eslint` on all five changed source files: **rc 0, no output**.
- Prettier: clean on all six touched files.
- Full `apps/backend` suite: **504/504 suites, 12055/12055 tests**, 0 skipped.

I authored this, so the re-review and the merge are yours. Note we both touch `stripe.service.ts`: I approved your #3270 a few minutes ago, so whichever lands second rebases — happy either order.

ankit-yc (approved):
Approving. The blocking finding is fixed at `35d41357c` and the fix is pinned by fixtures rather than
resting on the regex literal, which was the part I cared about — a later "cleanup" that narrows the
pattern back now fails a named test instead of silently going quiet.

Verified at `22a450a0a` (the merge of dev), not at the fix commit alone:

- `jest --runTestsByPath test/config/stripe-api-version-pin.test.ts` — 15/15, rc 0.
- Mutation, applied to a snapshot and restored after, apply-check 1/1 each:
  - narrow the opener back to `/new\s+Stripe\s*\(/g` (the original defect) → 2 failed / 13 passed.
  - drop the `(?<![\w$.])` lookbehind → 3 failed / 12 passed. So the lookbehind is load-bearing on
    the real tree, not just on the fixtures: `reportToStripe(` and a member call on the namespace
    both exist in `src` today.

Two things I checked rather than assumed, both fine:

- All four clients in `src` are written `new Stripe(`, and every value import of `stripe` binds the
  name `Stripe`, so the widened pattern has no new false positives to hit here.
- The parenthesis balancing does not track string literals, as the comment says. That fails toward a
  false positive, which is the right direction, and it is written down.

Two residuals, neither blocking, recorded so they are not rediscovered:

1. **The `.mts`/`.cts` widening is not covered.** Setting `TS_EXTENSIONS` back to `[".ts"]` leaves
   all 15 green, because no such file exists yet. That is inherent to a pre-emptive widening and I
   would not add a temp-directory fixture for it — but it means the extension list is prose, not a
   guarded assertion, and a later edit to it will not be caught.
2. **An aliased default import is invisible to the scan.** `import StripeSDK from "stripe"` then
   `new StripeSDK(...)` matches nothing, and unlike the string-literal case this one fails *open* —
   a real unpinned client the guard does not see. No such import exists today. If you want it
   closed cheaply, assert that every value import from `"stripe"` binds the identifier `Stripe`;
   that is one more assertion and it makes the pattern's assumption explicit. Happy for that to be
   a follow-up rather than another round here.

Issue #3266, opened 2026-09-18
chore(frontend): Next 16 defaults to Turbopack and the bundle-budget gate cannot see its chunks


aupyay:
## Summary

`frontend-quality / Frontend bundle budgets and Lighthouse` fails on #3232 (next 15.5.24 -> 16.3.4):

```
Bundle budget check failed:
- .next/static/chunks/2o4mqsvixfuw4.js: 1462.5 KiB > 375.0 KiB
- .next/static/chunks/2zzbtrhksi9j_.js: 453.8 KiB > 375.0 KiB
```

Next 16 makes **Turbopack** the default production bundler (`next build` prints
`Next.js 16.3.4 (Turbopack)`). That changes the chunk filenames, and
`apps/frontend/scripts/check-bundle-budgets.mjs` classifies purely by filename shape. The
message above is wrong on both counts: one of those chunks is lazy and the other is a
global shared chunk, and neither is a page chunk.

This needs a decision that is above a dependency pull request, which is why it is an issue.

## The gate has gone blind, not strict

`getChunkCategory` (`apps/frontend/scripts/check-bundle-budgets.mjs:41-56`) tests for
`framework-`, `main-`, `polyfills-` and `<digits>.<hex>.js`. Turbopack emits bare
13-character content hashes, so **all four predicates match zero files** and every chunk
falls through `return 'page'` at line 55 into the 375 KiB bucket - the strictest
non-polyfill ceiling.

| category rule | dev (15.5.24 / webpack) | #3232 (16.3.4 / Turbopack) |
|---|---|---|
| `framework-` / `main-` -> `shared` (195 KiB) | 3 | **0** |
| `polyfills-` -> `polyfills` (120 KiB) | 1 | **0** |
| `<digits>.<hex>.js` -> `async` (1190 KiB) | 124 | **0** |
| fallthrough -> `page` (375 KiB) | 169 | **297 (all of them)** |

So three of the four budgets now guard nothing. That is the dangerous half: the 195 KiB
`shared` ceiling is the one #3263 is currently blocked on, and merging #3232 would retire
it rather than satisfy it.

The unit test cannot see this. `apps/frontend/src/app/__tests__/scripts/checkBundleBudgets.test.ts`
uses 8 fixtures, all webpack-shaped. Its one collapse test asserts categories do not collapse
to the **largest** budget; the collapse that happened is to the **fallback**, which has no case.

## Measurement

[...]

PR #3272, opened 2026-09-18
chore(repo): hold next on 15.x while Next 16 adoption is decided

aupyay:
Closes nothing; parks #3232 and records why in the repo rather than only in bot state.

## What this does

Adds one `ignore` entry to `.github/dependabot.yml` holding `next` on `16.x`, with the reasoning
beside it in the same style as the other holds in that file. Paired with
`@dependabot ignore this major version` on #3232, which parks the open PR immediately - the
config entry is what makes the reason readable to the next person.

## Why

Dependabot's `next` 15.5.24 -> 16.3.4 PR cannot be made green by editing a budget number. Two
things break before any app code is touched, both measured on the real CI `next-build` artifacts
of the two arms (dev @ `4ab84be86` and the #3232 head):

| classifier category | budget | chunks matched on 15.5.24 | chunks matched on 16.3.4 |
|---|---|---|---|
| shared (`framework-`, `main-`) | 195 KiB | 3 | **0** |
| polyfills (`polyfills-`) | 120 KiB | 1 | **0** |
| async (`<digits>.<hex>.js`) | 1190 KiB | 124 | **0** |
| page (fallthrough) | 375 KiB | 169 | **297 - all of them** |

Next 16 makes Turbopack the default production bundler and its chunk filenames are base36 hashes
carrying no category, so every pattern in `apps/frontend/scripts/check-bundle-budgets.mjs` misses
and three of the four ceilings stop guarding anything. The visible symptom is the red check; the
dangerous half is the silent weakening.

Second blocker behind it: Next 16 removed `.next/app-build-manifest.json`, which
`apps/frontend/scripts/export-build-report.mjs:4` reads with no fallback, so the step after the
budget check exits 1 under **either** bundler. Fixing the budget gate alone turns a red step 7
into a red step 8.

Full measurement, the three build arms and the options are in #3266. Taking Next 16 is a product
decision, not a CI fix.

## Why it is safe to stay on 15.x

`gh api /advisories?ecosystem=npm&affects=next@15.5.24` returns **0** published advisories.
Controls on the same query: `next@13.4.0` -> 34, `next@15.0.0` -> 34, so the query discriminates.
Both criticals quoted in the 16.3.3 release notes list `patched = 15.5.24` on the 15.x line.
15.x minors and patches keep flowing - the entry is scoped with `versions: ['16.x']`, not a
blanket `semver-major` ignore.

## Why 16.x and not all majors

The `eslint` hold a few lines above is waiting on exactly this upgrade (`eslint-config-next` 16
is what drops the `@rushstack/eslint-patch` blocker). A permanent major freeze would bury that.
Scoping to `16.x` keeps a future major visible and makes the hold removable with
`@dependabot unignore next` once #3266 is decided.

## Verification

- `yaml.safe_load` on the committed file parses; the `ignore` list goes 20 -> 21 entries and the
  new one reads `{"dependency-name": "next", "versions": ["16.x"]}`. `groups:` is still a sibling
  key of `ignore:`, so the insert did not change the nesting.
- Committed with the repo hooks live (gitleaks, prettier, secretlint all ran and passed) - not
  with `--no-verify`.
- The category counts above were produced by running the repo's own classifier logic against each
  downloaded artifact. Instrument control: the unmodified
  `apps/frontend/scripts/check-bundle-budgets.mjs` passes the 15.5.24 arm ("passed for 297 JS
  assets", rc=0) and fails the 16.3.4 arm with the two lines CI printed verbatim
  (`2o4mqsvixfuw4.js: 1462.5 KiB > 375.0 KiB`, `2zzbtrhksi9j_.js: 453.8 KiB > 375.0 KiB`, rc=1).

## Note for review

`.github/dependabot.yml` is read from the **default branch** (`main`), so this is inert until the
next dev -> main promotion. That is why the PR comment carries the immediate park; this file is
the durable half. The dev and main copies are byte-identical today and this keeps them on the
normal promotion path rather than diverging.


ankit-yc (dismissed):
Approving. Config-and-comment only (+28/-0, one file), and the hold is scoped the way the comment says it is.

What I checked myself, against this PR's head:

- The entry lands in the one npm update block (`directory: /`). There is exactly one npm block in this file, so a single `ignore` covers the root `pnpm.overrides` pin and the `apps/frontend/package.json` pin together — the two-places note in the comment is a hazard for the eventual bump, not for this hold.
- `versions: ['16.x']` is the only use of `versions:` among the 21 ignore entries here; the other 20 use `update-types:` or nothing. That is the right call rather than an inconsistency: `update-types: version-update:semver-major` would swallow 17 and later too, and the stated intent is that a future major still gets proposed.
- Both mechanism claims hold at source on `dev`:
  - `apps/frontend/scripts/export-build-report.mjs:4` resolves `.next/app-build-manifest.json` into a single `MANIFEST_PATH` with no alternate path, and exits 1 at :69.
  - `apps/frontend/scripts/check-bundle-budgets.mjs` classifies by filename shape — `/chunks/polyfills-` at :55, `/\/chunks\/\d+\.[a-f0-9]+\.js$/` at :59 — with `JS_BUDGET_BYTES = 375 * 1024` at :22 as the fallback every unmatched chunk lands in.
- No published npm advisory has a vulnerable range containing `next@15.5.24`. Queried the advisories API and confirmed the query was live with a control that returns 8 rows for a known-vulnerable `next` version.

Taken on the author's evidence, not reproduced here: the measured chunk counts (3/1/124 on 15.5.24 versus 0/0/0 with all 297 in the fallback bucket on 16.3.4). The source reading above is consistent with them, but I did not rebuild either version.

One non-blocking nit for a follow-up: the comment cites the two scripts as `scripts/check-bundle-budgets.mjs` and `scripts/export-build-report.mjs`; both actually live under `apps/frontend/scripts/`. Since the comment exists to orient whoever picks the upgrade up later, the full paths are worth having.

ankit-yc (approved):
Re-approving on `0e57c9bd0a0ab3ada420da988bd3721819329e04`. The previous approval was dismissed by `dismiss_stale_reviews_on_push` on `dev`, as expected for an amend.

I diffed the amended file against the version I approved (`1b793aab`) rather than re-reading it whole. The change is confined to the two comment blocks — both script paths are now `apps/frontend/scripts/`, with the surrounding lines rewrapped. The `- dependency-name: 'next'` / `versions: ['16.x']` entry itself is untouched, so everything I verified on the previous head still stands.

The commit message carries the corrected paths too, with no occurrence of the old shorthand left.

One correction to my own previous review, since the record now has both numbers next to each other: I gave the advisory control as 8 rows for `next@13.4.0`. That was my own output truncation, not a count — the true figure is 34, as this commit message states. The result that mattered is unchanged and re-confirmed: 0 published advisories with a vulnerable range containing `next@15.5.24`, against a control that returns 34.

PR #3273, opened 2026-09-18
fix(repo): raise six hardening pins that still resolve vulnerable on dev

aupyay:
Replays the still-live half of the aikido-autofix draft #3259 onto current `dev`, and rejects one
of its entries with a reason.

## Why not just merge #3259

It was opened at 02:07Z against `d5a47f5bc`, and `dev` has moved past it in three ways:

- **Three of its 19 findings are already fixed here** — dompurify 3.4.14 -> 3.4.15 (both the
  frontend dependency and the override) and zod. Its diff sets `zod` to `4.6.0`; `dev` is on
  `4.6.1`, so merging it would move zod **backwards** past a resolved advisory.
- Its commit subject `fix(hardening): update dependencies` is not in this repo's commitlint
  `scope-enum`, so `Validate commit messages` cannot pass on that branch.
- It is CONFLICTING, and its lockfile diff is 761 lines against the older base.

So the content was re-derived from `git show origin/dev:pnpm-lock.yaml` rather than from the bot's
table, and only what still resolves vulnerable is here.

## What this changes

| package | `dev` resolves | this PR | finding |
|---|---|---|---|
| proxy-addr | 2.0.7 | 2.0.8 | **CRITICAL** — undersized IPv4-mapped IPv6 trust subnets let an unauthenticated client spoof `X-Forwarded-For`, bypassing IP-based access control, rate limiting and audit logging |
| nodemailer | 9.1.1 | 10.0.9 | **CRITICAL** stack exhaustion + 2 HIGH ReDoS + 2 MEDIUM |
| fast-uri | 4.1.3 | 4.1.5 | 2 HIGH host-validation bypasses + 2 MEDIUM |
| moment | 2.30.1 | 2.31.0 | MEDIUM path traversal in the locale loader |
| brace-expansion | 1.1.18 / 5.0.9 | 1.1.21 / 5.0.12 | 2 MEDIUM + 1 LOW ReDoS |
| yaml | 2.9.0 | 2.9.1 | LOW — recursive merge aliases bypass `maxAliasCount` |

**proxy-addr gets an explicit override entry, which the bot's diff did not have.** #3259 moved the
lockfile alone; a later resolution could drift back to 2.0.7 with nothing in a manifest to
contradict it.

**nodemailer 9 -> 10 is a major.** It is safe here because nodemailer is not a dependency of any
workspace and is imported nowhere in our source — its only consumer is `supertokens-node`. Checked,
not assumed.

## The entry this PR does not take

`"@noble/hashes@1": "2.4.0"` (LOW, prototype pollution in options merging).

#3259 says "No breaking changes from these package upgrades affect this codebase". That is true of
first-party imports — nothing of ours imports `@noble/hashes` at all. The v1 copies are there for
two transitive consumers, and **the override breaks both of them**:

- `pkijs@3.4.0` imports `@noble/hashes/sha1` and `@noble/hashes/sha2`
- `pdfkit@0.20.2` imports `@noble/hashes/sha2`, `/legacy`, `/utils`, `/crypto`

v2 renamed every subpath to carry a `.js` suffix and moved sha1 into `legacy`. `npm view
@noble/hashes@2.4.0 exports` has 17 keys including `./sha2.js` and `./legacy.js`, and **no `./sha1`
and no `./sha2`**. Resolution probe, with the v1 copy as the control arm:

```
v1 1.4.0 (what pkijs has today)      @noble/hashes/sha1       RESOLVES
v1 1.4.0 (what pkijs has today)      @noble/hashes/sha2       RESOLVES
v1 1.4.0 (what pkijs has today)      @noble/hashes/sha256     RESOLVES
v2 2.2.0 (the shape 2.4.0 also has)  @noble/hashes/sha1       ERR_PACKAGE_PATH_NOT_EXPORTED
v2 2.2.0 (the shape 2.4.0 also has)  @noble/hashes/sha2       ERR_PACKAGE_PATH_NOT_EXPORTED
v2 2.2.0 (the shape 2.4.0 also has)  @noble/hashes/sha256     ERR_PACKAGE_PATH_NOT_EXPORTED
```

Stated precisely: the resolution arms ran against 2.2.0, which is already in the tree; 2.4.0 is
covered by its published export map having the same shape. I did not install 2.4.0.

Nothing type-checks a transitive dependency's own imports, so that override would be green through
CI and would throw `ERR_PACKAGE_PATH_NOT_EXPORTED` at the first PDF generation or certificate
parse. The LOW stays open until pdfkit and pkijs release against v2; tracked separately.

## Verification

Run on the committed tree, working tree clean, after `pnpm install --frozen-lockfile` reported
`Lockfile is up to date`:

- Installed versions read off disk rather than inferred from the lockfile: proxy-addr 2.0.8,
  nodemailer 10.0.9, fast-uri 4.1.5, yaml 2.9.1, moment 2.31.0, and 2.31.0 inside `apps/backend`.
- `pnpm check:overrides` — rc 0. 127 entries, 4 advisories in tree, 1 vulnerable pin, all of it the
  pre-existing accepted `adm-zip` baseline record. No new unreviewed finding.
- `pnpm check:override-drift` — rc 0, no override contradicts a workspace declaration.
- `pnpm type-check` — 17/17 tasks successful.
- backend `npx jest`, full suite, not a scoped run — **504 suites / 12063 tests passed**, rc 0.

Closes #3259 once this lands (that draft has nothing left in it that is both live and safe).


ankit-yc (approved):
Approving. I re-derived the four claims that carry this rather than reading the table, and one of them is understated in your favour.

**No pin is weakened, which is the only thing that could have blocked it.** Diffed the override key sets rather than the text: 126 entries on `dev`, 127 here, **0 removed**, 1 added (`proxy-addr`), 5 raised. Every movement is forward, and `zod` is untouched at `4.6.1` — the backwards move you avoided by not merging the draft.

**The six versions are real on disk, not just in the lockfile.** `pnpm install --frozen-lockfile` clean, then read out of `node_modules/.pnpm`: nodemailer 10.0.9, proxy-addr 2.0.8, fast-uri 4.1.5, moment 2.31.0 (and 2.31.0 resolved from `apps/backend`), yaml 2.9.1, brace-expansion 1.1.21 / 2.1.4 / 5.0.12. Exactly one copy of each of the five single-version packages, so nothing stale survives alongside. The control arm is `express`, which is not in this PR and correctly returned two copies — without that, "one version each" is a claim an empty search also satisfies.

`pnpm check:overrides` rc 0 (127 entries, 4 advisories, 1 vulnerable pin, the accepted `adm-zip` baseline) and `pnpm check:override-drift` rc 0. Same numbers you published.

**nodemailer's major is safe for the reason you gave.** Zero first-party importers and zero workspace declarations — searched tracked files under `apps/` and `packages/` at this commit, with `supertokens` and `nodemon` as controls proving both searches can return hits. Its sole consumer in the lockfile is `supertokens-node@24.0.3`.

**The rejected entry is rejected correctly, and the breakage is wider than the PR says.** `pkijs@3.4.0` resolves the v1 copy at 1.4.0 and `pdfkit@0.20.2` at 1.8.0, so the `@1` selector catches both. Their actual imports, read off the installed trees:

- pkijs: `/sha1`, `/sha2`
- pdfkit: `/sha2`, `/sha256`, `/legacy`, `/utils`, `/crypto`, `/esm`

Against the installed export maps — v1 1.4.0 has 21 keys, v2 2.2.0 has 17 — **`./sha1`, `./sha2`, `./sha256`, `./utils` and `./crypto` are all ABSENT in v2**, not only the two you listed. So it is not just sha1 moving into legacy and a `.js` suffix appearing; every subpath pdfkit uses disappears too. Your conclusion holds with more room than you claimed.

I did not test the "green through CI" clause — proving that would mean installing the override, and the export map settles the decision without it.

### One non-blocking residual

`nodemailer@<9.1.1` is a keyed selector, and it excludes the `[9.1.1, 10)` band. A future dependency declaring `^9.5.0` resolves outside the key and the override stays silent — which is the same drift argument you used to justify giving `proxy-addr` an unkeyed entry, applied to the one package here that kept its key. Today it is covered: `check-override-advisories` reports `uncovered copies: 0`, so the gate catches it even though the manifest does not contradict it. Worth an unkeyed `"nodemailer": "10.0.9"` next time this file is touched, not worth a push now.

Merging this.

aupyay:
Post-merge note on this PR body, so the v2 rejection is not re-litigated later on a misreading.

> **Corrected twice on 2026-09-18, both times by the reviewing desk.**
> 1. My first version called `"@noble/hashes@1": "2.4.0"` a load-bearing override. Wrong — **there is no such override in this repo.** `git grep -ln noble` at merged `6a8c5219e` hits `pnpm-lock.yaml` and nothing else (control: `pnpm` hits 147 files, so the search was live). That string in the body is the *proposed* remediation being **rejected**, not a pin being kept.
> 2. My second version measured `2.4.0`, which is not in the tree. Re-anchored below to **`2.2.0`**, the v2 the lockfile actually carries.

**The body understates how wide the rejection is.** It says v2's exports map "has 17 keys including `./sha2.js` and `./legacy.js`, and **no `./sha1`**". The 17 is right, but the implication that only `./sha1` is missing is not — every key in that map is extension-ful (`./sha2.js`, `./blake1.js`, …), so **no** extensionless subpath resolves.

Measured against the published `2.2.0` tarball — the version in `pnpm-lock.yaml` at `6a8c5219e` — not a lockfile entry:

```
npm pack @noble/hashes@2.2.0 && tar xf noble-hashes-2.2.0.tgz
python3 -c "import json;e=json.load(open('package/package.json'))['exports'];\
print(len(e));[print(s,'PRESENT' if s in e else 'ABSENT') for s in \
['./sha1','./sha2','./sha256','./legacy','./utils','./crypto','./esm']]"
```

```
v2.2.0 export_keys 17
  ./sha1   ABSENT    ./sha2  ABSENT    ./sha256  ABSENT
  ./legacy ABSENT    ./utils ABSENT    ./crypto  ABSENT    ./esm ABSENT
sample keys: ['.', './_md.js', './argon2.js', './blake1.js']
```

Same result on the `2.4.0` tarball, so this is a property of the v2 major, not of one patch.

Cross-referenced with the importers the body lists — `pkijs` uses `/sha1` and `/sha2`; `pdfkit` uses `/sha2`, `/legacy`, `/utils`, `/crypto` — the union actually imported is **five** (`sha1`, `sha2`, `legacy`, `utils`, `crypto`), and all five are absent.

**What that means, stated correctly.** Three versions coexist transitively at merged dev — `1.4.0`, `1.8.0`, `2.2.0` (confirmed from `pnpm-lock.yaml` at `6a8c5219e`; `2.4.0` is not in the tree at all). v2 is not broken in itself — `app-builder-lib` already runs on `2.2.0`. The hazard is the reverse of a pin being weakened: **adding** a workspace-wide v2 override would break `pdfkit` and `pkijs` across five resolution paths. That is the reason to leave this alone.

Issue #3186, opened 2026-09-12
fix(mobile): keep medication calendar reminders on the selected local day


aupyay:
### Preflight

- [x] Checked current source and synthetic or browser evidence.
- [x] Compared existing open and closed issues and open pull requests.
- [x] Public report uses synthetic examples and contains no private records or paths.

### Affected area

Mobile app

### What is wrong

Calendar-synced medication reminders can appear a day earlier than the task selected in the app, despite retaining the chosen time of day.

### Steps to reproduce

1. Use a device timezone west of UTC, such as Los Angeles.
2. Create a synthetic medication task dated September 15 with a morning dosage at 08:00 and calendar sync enabled.
3. Open the created event in the device calendar.

### Expected behavior

The event is scheduled September 15 at 08:00 in the task scheduling context.

### Actual behavior

The current calendar builder schedules September 14 at 08:00 for this fixture.

### How often does it reproduce

Deterministic under the stated reproduction conditions; not a claim about all users or production incidence.

### Did this work before

Unknown; the current development revision still exhibits the behavior.

### Impact

P1 suggested triage priority. Calendar-synced medication reminders can appear a day earlier than the task selected in the app, despite retaining the chosen time of day.

### Environment and evidence

Audit date: 12 September 2026. Source baseline: e5c0c78ab79afe7c1796ee8d98d1697ec2a1c904.

Source trace and deterministic execution of current source with synthetic data; no simulator, physical-device or deployed end-to-end run.

Diagnostic reproductions establish the reported defect; they are not a product fix, a passing regression suite or a coverage claim.

### Implementation entry points

Locate these existing symbols and follow their callers; locations are intentionally omitted from this public report.

- `createCalendarEventForTask`
- `createDosageCalendarEvents`
- `createSingleDosageEvent`
- `parseDosageTime`
- `parseISODate`

### Repair contract

1. Build the dosage event from validated local calendar date components and the selected dosage time using existing date helpers, preserving the task scheduling timezone. Do not parse a date-only value as a UTC instant.
2. Apply the same rule to create and edit flows and to every dosage in a multi-dose task. Preserve recurrence limits, reminder offsets and selected calendar.
3. Reject invalid date or time input before creating any events, with a recoverable error; do not normalize malformed values onto an unintended day.
4. Keep partial-failure cleanup intact and avoid duplicate entries on retries.

### Acceptance criteria and QA

[...]

PR #3275, opened 2026-09-18
feat(backend): let a pet owner see what their companion is due for

### LINKED ISSUE #2705: feat(mobileAppYC): care reminders are sent to owners but there is no list of what is due
## Problem

`CareReminder`, `CareReminderOptOut` and `ReminderJob` exist, and `care-reminder.router.ts` is mounted. Reminders are delivered to owners. But `apps/mobileAppYC/src` has **zero references** to any care-reminder endpoint — the reminder is fire-and-forget.

If the owner misses or dismisses the push, there is nothing in the app that tells them the animal is due for anything. There is also no way to see what is coming, only what happened to arrive.

## Why now

#2611 item 2 is fixing the delivery accounting — a reminder is currently recorded `SENT` even when there was no recipient at all or the parent had opted out of both channels. That work makes the *record* honest. This is the other half: making the state visible to the person it concerns, so the system does not depend on a push having been noticed.

## Proposed scope

1. A "what is due" list for the owner's companions, owner-scoped.
2. Due/overdue distinction, driven by the reminder's own schedule rather than by whether a notification was delivered.
3. Respect `CareReminderOptOut` — opting out of *delivery* should not necessarily mean hiding the item in-app, and that is a product call worth making explicitly rather than by accident.

## Notes

- Owner scoping belongs in the `where`. See #2703 for the `ParentPatient` ACTIVE + `permissions.medicalRecords` pattern, and #2696 for what happens when a client-supplied id is trusted without it.
- Check whether a reminder binds to the patient directly or through an artifact before designing the query — those two paths have very different failure modes, and `ClinicalArtifact` has no patient column (#2704).

## Related

From the mobile surface survey after #2625. Same family as #2701.

aupyay:
Closes #2705 once the app screen lands; this PR is the backend half.

## The problem

Care reminders were fire-and-forget. The practice creates a `CareReminder`, we push and mail it, and `apps/mobileAppYC/src` has no reference to any care-reminder endpoint. If the owner missed or dismissed the push there was nothing in the app saying the animal was due for anything, and no way to see what was coming — only what happened to arrive.

## What this adds

`GET /v1/care-reminders/mobile/due` — the open care reminders for the signed-in parent's companions, soonest due first, keyset-paged.

**Scoping.** Reuses `listPermittedPatientIds`, the same `ParentPatient` ACTIVE + `permissions.medicalRecords` predicate the owner prescription list (#2703) is built on, so one answer serves both surfaces. A REVOKED or PENDING link is not access, and a co-parent whose `medicalRecords` flag is false gets nothing. The binding is direct — `CareReminder.patientId` names the companion — so unlike prescriptions there is no nullable hop to fail closed on.

**Due and overdue** come from the reminder's own `dueDate`, not from `status` or `sentAt`. A reminder nobody managed to deliver still reads as overdue, which is the point of the issue. The comparison is a strict instant against one `now` captured for the whole page, so two rows either side of the request cannot disagree about what time it is. `dueDate` goes out in full and the local-day wording is left to the client — the caller's timezone is not known here, and guessing at it is the #3186 bug class.

**Open** means PENDING or SENT. RESPONDED is done, CANCELLED was withdrawn by the practice. EXPIRED is excluded too, with a caveat recorded in the code: nothing writes EXPIRED today, so that arm is exercised by tests and not yet by production data.

**`notes` and `createdBy` are never selected.** Neither reaches the parent on the send path — `notes` is the practice's own working text and `createdBy` is a staff id.

## The opt-out decision, made on purpose

The issue asked for this to be an explicit call rather than an accident. `CareReminderOptOut` is **not** consulted.

An opt-out is an objection to being *contacted* by a practice — GDPR Art. 21, recorded against an email address by the unsubscribe link. This list is not a contact: the owner opened it. Suppressing it would mean that unsubscribing from one practice's email silently removed every trace that the animal is due for anything, which is the failure this endpoint exists to fix. There is a test asserting the opt-out table is never read, so reversing the decision has to be deliberate.

## One wording, two surfaces

The owner-facing sentence moves to `services/shared/care-reminder-message.ts`, imported by both the send path and this read. A shared module rather than importing `care-reminder.service`, for the reason `shared/pagination` exists: that service reaches for email, push, audit and the opt-out store, and an owner-facing read has no business pulling that graph in to obtain a string.

The guard is a test in `care-reminder.service.test.ts` asserting the email body contains what `buildCareReminderMessage` returns for the same inputs — against the builder, not a copy of the string, so changing one side alone fails.

## Evidence

Run in `apps/backend` at `b543db1bb`:

- `npx tsc` — exit 0, no diagnostics.
- `npx eslint` on the 10 touched files — 0 errors (the 4 warnings are the repo's ignore pattern for `test/`).
- `npx jest` (whole package, not scoped) — **507 suites passed / 507, 12106 tests passed / 12106, 0 skipped.**
- 12 mutations, each killed, baseline 69/69 across the four affected suites:

| # | mutation | result |
|---|---|---|
| — | baseline | 69 passed |
| 1 | include RESPONDED in the open filter | 1 failed |
| 2 | order due-date `desc` | 1 failed |
| 3 | overdue boundary `<` → `<=` | 2 failed |
| 4 | ignore the injected `now` | 3 failed |
| 5 | select the practice's `notes` | 1 failed |
| 6 | `take: limit` instead of `limit + 1` | 2 failed |
| 7 | cursor comparison `gt` → `lt` | 1 failed |
| 8 | cursor carries the wrong date | 2 failed |
| 9 | stop deduplicating companion ids | 1 failed |
| 10 | emit a nameless reminder instead of dropping it | 1 failed |
| 11 | empty custom message falls back to generated text | 1 failed |
| 12 | label map loses a reminder type | 2 failed |

## Out of scope, deliberately

The `apps/mobileAppYC` screen. It is a new screen, and the standing design-first rule sends new screens through Claude Design before implementation — so it is a follow-up rather than something bolted onto this PR. #2701 was split the same way (#2703 reads, then #3251).

I authored this, so it needs an `ankit-yc` review from another desk.


aupyay:
Aikido's long-function finding on `listDueCareRemindersForParent` is fixed at `6d8c4dcd3`, not ignored.

Split into `readOpenReminderPage` (query + keyset split), `loadCompanionNames` (the second read, deduplicated) and `toMobileCareReminder` (mapping and the overdue comparison). Each rule now sits with the comment that explains it instead of in one 120-line block.

Behaviour is unchanged, and that is measured rather than asserted: the same 12 mutations are all still killed against the refactored code, baseline 69/69, and the whole package is 507/507 suites and 12106/12106 tests at the new head. `npx tsc` exit 0.

The evidence table in the PR body was produced at `b543db1bb` and re-run unchanged at `6d8c4dcd3`.

ankit-yc (approved):
Read the full diff, not just the summary.

Scoping reuses `listPermittedPatientIds` unchanged, so a REVOKED/PENDING link or a false `medicalRecords` flag gets nothing, and `patientIds` is rebuilt on every page — a borrowed cursor moves the window without widening it. The `/v1/care-reminders` mount does not collide with the existing router: every web care-reminder route is under `/v1/pms/organisation/:organisationId/...`.

`select` omits `notes` and `createdBy`, and a test pins that. The opt-out decision is argued in the code and guarded by a test asserting the table is never read, so reversing it has to be deliberate — which is what the issue asked for.

`overdue` is a strict compare against one injected `now`, drawn from `dueDate` rather than `status`/`sentAt`, and `dueDate` ships in full with the local-day wording left to the client. Keyset paging uses an exclusive `where` rather than `cursor` + `skip: 1`, and splits before mapping, so a row dropped as unmappable cannot become a permanently stuck cursor.

Extracting the wording into `shared/care-reminder-message` is the right call: it keeps an owner-facing read off the send path's email/push/audit import graph, and the send-path test compares against the builder rather than a copy of the string, so one side cannot drift alone.

Nothing blocking. Approving.

PR #3276, opened 2026-09-18
test(desktop): pin the session permission allowlist so a drift fails

### LINKED ISSUE #3262: Desktop: the session permission allowlist has no test that can fail, and Electron 44 moves its vocabulary from 19/20 names to 46
`permittedPermissions` in `apps/desktop/src/shell/window-config.ts:31-37` is the desktop
shell's permission allowlist. It gates both handlers installed by
`configureSessionPermissions` - `setPermissionRequestHandler` and
`setPermissionCheckHandler` - on the live session (`create-main-window.ts` calls it), so it
is the control that decides whether the app grants camera, microphone, screen capture,
location, clipboard read and notifications.

**It is not covered by any test.** Replacing the entire allowlist with a single string that
Electron has never emitted leaves the whole desktop suite green:

```
baseline                                  71 suites / 1096 tests, rc 0
permittedPermissions = new Set([
  'totally-not-a-real-permission' ])      71 suites / 1096 tests, rc 0   <-- unchanged
restored from snapshot                    71 suites / 1096 tests, rc 0
```

Three reasons it cannot fail:

1. **The tests derive their input from the value under test.**
   `tests/window-config.test.ts:150`, `:184` and `:188` all pass
   `[...permittedPermissions][0]`, so the positive assertions hold for whatever the Set
   happens to contain.
2. **The one denial arm uses a string Electron does not use.**
   `:156` passes `'midi-sysex'`. That spelling appears 0 times in the bundled typings of
   both 43.5.1 and 44.1.1; Electron spells it `midiSysex` (2 occurrences in each). The arm
   therefore proves only that an unknown string is denied.
3. **The other denial arm is mislabelled.** `:189` asserts
   `checkHandler(wc, 'x', '')` is `false`, which reads as coverage of the empty-origin case.
   It passes because `'x'` is not in the Set - an empty origin resolves through
   `new URL('', startUrl)` to the start URL and classifies as internal. Substituting a real
   allowlisted permission there returns `true`.

`tsc` cannot help either: the Set is inferred as `Set<string>` and the signature is
`shouldGrantPermission(permission: string, ...)`, so there is no literal union to check
against. No other file in `apps/desktop/src` or `apps/desktop/tests` references the
permission literals at all.

**Why this matters now.** Electron 44 moves this vocabulary substantially. Measured from
the two bundled `electron.d.ts` files, at the declaration lines rather than the doc
comments:

| union | 43.5.1 | 44.1.1 | added | removed |
|---|---|---|---|---|
| `setPermissionCheckHandler` | 19 | 46 | 27 | 0 |
| `setPermissionRequestHandler` | 20 | 46 | 26 | 0 |

Nothing was removed, and all five current allowlist entries are present in both 44 unions,
so no existing grant breaks. Two specifics are worth a decision rather than a silent default:

- `display-capture` is in our allowlist and is one of the 27 names **added** to the
  *check* union in 44 - it was absent from 43's check vocabulary while present in 43's
  request vocabulary.
- `geolocation-approximate` is new in 44 and is **not** in our allowlist, which contains
  only `geolocation`.

Both are measured from the typings. Whether Chromium 152 actually routes either through
these handlers at runtime was **not** verified - it needs a real Electron binary, and
`node_modules/electron/dist` is not populated in a normal workspace install. Treat the two
bullets as questions the typings raise, not as confirmed behaviour changes.

**What to do**

The allowlist should be pinned by a test that fails when the vocabulary drifts, and the
drift above should be decided deliberately. Concretely:

- Assert the exact contents of `permittedPermissions` against a literal array in the test,
  so any edit to the control has to be an intentional edit to the test as well.
- Give the denial arms real Electron permission names that are deliberately not granted
  (`midiSysex`, `serial`, `usb`, `hid`), not invented strings.
- Fix `:189` to test what it claims, or rename it - the empty-origin case is currently
  untested.
- Decide `display-capture` on the check path and `geolocation-approximate` explicitly.

Found while auditing the surfaces the Electron 44 bump (#3229) does not exercise. The
coverage hole is pre-existing and independent of that PR; the vocabulary movement is not.

aupyay:
Closes #3262

`permittedPermissions` (`apps/desktop/src/shell/window-config.ts`) is the allowlist behind
both handlers `configureSessionPermissions` installs on the live session. It had no test that
could fail on it.

Reproduced on `origin/dev` before changing anything: replacing the whole allowlist with
`new Set(['totally-not-a-real-permission'])` left `tests/window-config.test.ts` at
**15 passed / 15 total**. Every positive arm passed `[...permittedPermissions][0]`, so it held
for whatever the Set happened to contain, and the single denial arm used `'midi-sysex'` — a
spelling that appears **0 times** in either permission union of the bundled `electron.d.ts`
at 44.1.1 (Electron spells it `midiSysex`).

## What changed

**Tests.** The five granted names are asserted against a literal, so an edit to the control
has to be an intentional edit to the test. The grant and denial arms now use real Electron
names. Measured from `electron@44.1.1`'s typings at the declaration lines: both
`setPermissionCheckHandler` and `setPermissionRequestHandler` declare **46** names; all five
allowlisted names are in both; `midiSysex`, `serial`, `usb`, `hid` and
`geolocation-approximate` are in both and none is granted; `midi-sysex` is in neither.

**Decisions the issue asked for.** `display-capture` stays granted — it is in 44's *check*
union, which it was not in at 43, so the grant is now reachable on both paths rather than
only on request. `geolocation-approximate`, new in 44, stays **out**: only precise
`geolocation` is allowed and an omission fails closed. Both are now pinned by name, so a
future bump has to confront them rather than inherit them.

**One behaviour fix.** The old empty-origin arm asserted `checkHandler(wc, 'x', '')` is
`false`, which reads as coverage of the empty-origin case but passed because `'x'` is not
allowlisted. Tested properly, an empty `requestingOrigin` resolved through
`classifyNavigation`'s base-URL argument to the start URL and classified as **internal**, so
an unattributable check was granted. The check handler now falls back to the web contents'
own URL — the same fallback `shouldGrantPermission` already uses for the request path — and
returns `false` when neither yields an origin.

## Evidence

Mutation arms, each restored from a snapshot of the file (not from HEAD), with a no-op
baseline whose expected output is non-empty:

```
arm0 baseline (no-op)                            22 passed, 22 total
arm1 allowlist replaced with one bogus name       6 failed, 16 passed
arm2 display-capture removed                      2 failed, 20 passed
arm3 geolocation-approximate added                2 failed, 20 passed
arm4 empty-origin fix reverted                    2 failed, 20 passed
```

Arm1 is the same mutation that left the suite green on `origin/dev`.

```
apps/desktop $ pnpm test          71 suites / 1103 tests, 0 failed, 0 skipped, rc 0
apps/desktop $ pnpm run type-check  rc 0   (tsc --noEmit x2)
apps/desktop $ pnpm run lint        rc 0   (eslint, 0 errors)
window-config.ts coverage           98.29% stmts, 90.14% branch, 100% funcs
                                    uncovered 150-154 is the pre-existing popup_blocked
                                    branch; every line added here is covered
```

Baseline on `dev` was 1096 tests; this is +7 net (9 added, 2 replaced).

Not verified: whether Chromium 152 actually routes `geolocation-approximate` or
`display-capture` through these handlers at runtime. That needs a real Electron binary, and
`node_modules/electron/dist` is not populated by a normal workspace install. Everything above
is measured from the typings and from the handlers' own code.


ankit-yc (commented):
Post-merge review (this merged with an admin override before any approval, so recording it here).

`window-config.ts`: the check handler now falls back to `webContents.getURL()` when `requestingOrigin` is empty and returns false when there is no origin at all, instead of letting `classifyNavigation` resolve an empty string to the start URL and call it internal. That matches the request handler and fails closed. The allowlist comment and the literal pin in `window-config.test.ts` mean a change to the five granted permissions now has to be deliberate. No follow-up needed on the code.

PR #3277, opened 2026-09-18
fix(repo): reverse a refunded capture in the provider receipt journal

aupyay:
Delivery (2) of #3170, on top of #3270 which landed delivery (1). Backend only; no UI, and not blocked on #3163.

## The gap

#3270 gave every captured payment a durable journal row. A refund reached the invoice, the `Payment` row, the `Refund` row and the customer — and the one record it never reached was the journal row for the capture it reverses. A receipt stayed at its full captured amount after the money had gone back, so the oracle the journal exists to support, **captured = applied + unapplied + refunded**, had no term to read for the last one.

Three defects in `_handleRefund`, all on the same path:

1. **No journal reversal at all.** `PARTIALLY_REFUNDED` and `REFUNDED` existed in the enum with nothing writing them.
2. **`charge.amount` read as the refund.** That is what was **captured**. A 20.00 refund of a 100.00 charge was reported as 100.00 — to the `Refund` row, to the invoice metadata and to the customer's push notification. `charge.amount_refunded` is the refunded figure and is cumulative.
3. **No connected account.** `_handleRefund` was the only webhook handler with no parameter for `event.account`, so a connected-account refund would have looked for its capture on the platform account and found nothing.

## What this adds

`ProviderReceipt.refundedAmount`, additive with a default of `0` — which is what rows written before this migration mean.

- **Cumulative, not a delta.** The provider states the total it has given back on the charge, so a redelivered event, two partial refunds delivered out of order, and two concurrent deliveries all converge on the provider's own figure instead of summing deltas into a number it never stated.
- **The write is conditional on being an increase** — `WHERE id = ? AND "refundedAmount" < ?` — rather than read-then-write, so the predicate and the write are one statement. Proved against a real Postgres below.
- **Keyed on the capture's own reference**, so a refund reduces the receipt for the money it is actually returning and never a sibling capture of the same appointment. Appointment identity is still never reinterpreted as replay.
- **The capture is journalled first where the journal does not hold it.** A refund event carries the provider's figures for the charge it reverses, so a capture that predates this table — or one whose own webhook failed to write — is recovered from stated facts rather than reported as an orphan. It is recovered `UNATTRIBUTED`: nothing in a refund event says whose money it was, and `amount_captured` is preferred over `amount` because a partially captured charge returns less than it authorised.
- **A charge with no payment intent is not journalled.** Writing it under the charge id would let the same money be recorded twice under two different references, which is the one thing the unique key exists to prevent.
- **A refund figure the provider did not state is not recorded.** A guessed number subtracted from a captured total is worse than a visible gap; the invoice path keeps the behaviour it has always had.
- **A refund in a different currency from its capture is refused.** Two currencies do not compare, so the figure would mean nothing beside the amount it is meant to reduce.
- **The journal cannot break the refund.** It records the refund; it is not a precondition for handling it. Throwing would stop the invoice being marked `REFUNDED` and the customer being told, and a provider retry would repeat that forever.

It reverses the **receipt**, not an allocation. Releasing the account credit a refund takes back is the single transaction in #3163. `invoiceId` is left in place because it is the historical record of where the money went.

Deliveries (3)–(5): the reconciliation list and actions, the historical mismatch audit, and any invoice-cardinality migration.

Two adjacent behaviours were left alone on purpose, both owned elsewhere and both now visible in the journal rather than hidden:

- `markInvoiceRefundedFromWebhook` still sets the **invoice** to `REFUNDED` on a partial refund, and the `Payment` row to `REFUNDED` rather than `PARTIALLY_REFUNDED` — while the manual refund path beside it already distinguishes them. That is invoice-status semantics with a wide blast radius, not receipt journalling.
- Nothing reads the `Refund` table yet; its rows are written and never aggregated.

One flattening is worth a reviewer's eye: `status` is a single dimension and the issue fixes its member set, so a refund on an `UNATTRIBUTED` capture overwrites that state. The fact is not lost — `organisationId` is still `NULL`, so the residual of a partially refunded capture with no owner is `status = 'PARTIALLY_REFUNDED' AND "organisationId" IS NULL`, which the `status` and `organisationId` indexes both serve. A separate attribution column would be a change to the contract the issue states.

## Evidence

Every command run in the worktree at the pushed head, `e1312da9a`, with a clean tree (`git status --porcelain` empty).

| Check | Result |
|---|---|
| `npx tsc --noEmit -p apps/backend/tsconfig.json` | rc 0 |
| `npx jest` (whole backend package) | rc 0, **508 suites / 12143 tests passed, 0 skipped** |
| `npx eslint` on the two changed source files, from `apps/backend` | rc 0, 0 errors 0 warnings |
| `prisma validate` | rc 0 |
| `prisma migrate deploy` against a real `postgres:16-alpine` | rc 0, **176 migrations applied** |
| `prisma migrate diff --from-migrations … --to-schema-datamodel` | **0 drift statements** |
| `node scripts/ci/classify-migration.mjs` on the new migration | rc 0 — additive only |
| RLS gate query | **0 public tables without RLS**; `ProviderReceipt.relrowsecurity = t` |
| column readback | `refundedAmount | double precision | NOT NULL | default 0` |

Both gates were run with a control proving they can fail rather than only that they were quiet:

- **drift:** planting a schema-only column produced exactly 1 statement (`ALTER TABLE "ProviderReceipt" ADD COLUMN "plantedDriftProbe" TEXT;`); removing it returned to 0.
- **classifier:** a scratch migration containing a bare `DROP COLUMN` exited 1 with the drops-a-column hazard. It was deleted and is not in this diff.

### The increase predicate, against a real database

The central claim is a concurrency claim, so it was measured rather than reasoned about. Two overlapping transactions, both applying the same cumulative figure to one seeded row; the second blocks on the row lock and re-evaluates its `WHERE` after the first commits.

| Arm | A | B | final `version` |
|---|---|---|---|
| `WHERE id = ? AND "refundedAmount" < 20` (ships) | `UPDATE 1` | `UPDATE 0` | **1** |
| **control** — predicate removed | `UPDATE 1` | `UPDATE 1` | **2** |

The control arm is what makes the first row mean anything: without the predicate the same two deliveries both write.

### Mutation coverage

**21 mutations, 21 killed, 0 survived**, baseline green before and after each run, restored from a file snapshot rather than `HEAD`.

Two mutations in the first pass exited non-zero with **zero failing tests** — they did not compile, which is a suite error and not a kill. They were rewritten as compiling mutations (`>=` → `>= 0`, and the `>=` boundary made strict) and re-run; both then failed real named tests. The harness now refuses to score a non-zero exit with no failing test as a kill.

| Mutation | Killed by |
|---|---|
| refund figure read from the capture amount | reverses by what the provider gave back, not by what it captured |
| no capture journalled before the reversal | 4 tests |
| no reversal recorded | 5 tests |
| reversal recorded before the capture is journalled | journals the capture it is reversing before reducing it |
| intentless guard removed, charge id used as the reference | does not identify a capture for a charge that names no intent |
| recovery journals the authorised amount | recovers the captured amount, not the authorised one |
| dispatcher drops the connected account | carries the connected account the capture landed in |
| a stateless refund figure journalled anyway | does not record a refund figure the provider did not state |
| journal failure breaks the refund handler | never lets the journal be the reason a refund is not processed |
| the refunded comparison always true | 2 tests |
| the full-refund boundary made strict | 2 tests |
| the increase predicate dropped from the write | writes only when the provider's total has gone up |
| currency mismatch written through | refuses a refund in a different currency from its capture |
| a missing capture reported quietly | refuses a refund with no journalled capture to reverse |
| a lost race reports the status it wanted | reports the stored state when the refund was already covered |
| no version bump on a reversal | bumps the version so a later allocation can be made conditional on it |
| storage failure thrown back at the webhook | never throws a storage failure back at the webhook |
| platform capture looked up under null | looks for a platform capture under the sentinel, not under null |
| a no-op reversal reported as applied | reports the stored state when the refund was already covered |
| the re-read fallbacks dropped | falls back to the row it read when the re-read finds nothing |
| the unmatched-refund log stops naming the missing intent | names the missing intent when an unmatched refund is reported |

### Coverage of the changed files

```
File                  | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
stripe.service.ts     |     100 |   98.74  |     100 |     100 | 143,159,165,275
provider-receipt.ts   |     100 |   97.91  |     100 |     100 | 219
```

Every one of those five lines is outside this PR's diff hunks (`177-245, 815, 1197-1234, 1246` in `stripe.service.ts`) and predates it. Two branches that were mine and uncovered on the first reading — the re-read fallback when the row is deleted between the write and the re-read, and the log for a refund that has neither an intent nor an invoice — got the two tests in the second commit.

### Issue QA cases covered by this slice

- reverse/refund references the original receipt ✔
- a redelivered refund does not reduce the receipt twice ✔
- two partial refunds delivered out of order converge on the provider's total ✔
- two concurrent deliveries write once ✔ (measured against Postgres above)
- `refunded` now has a term in captured = applied + unapplied + refunded ✔

The full total still cannot be asserted end to end: `applied` is allocation, which is delivery (3) and #3163. That is not claimed here.

### Existing fixtures

Six `_handleRefund` fixtures gained `amount_refunded`. Stripe states it on every charge, so they had been asserting against a shape the provider never sends; none of their assertions was weakened or removed. No test was deleted, skipped or `.only`'d.

Refs #3170

I cannot review my own work — this needs an ankit-yc review.


ankit-yc (dismissed):
Reviewed at e1312da9a57c282170942814a8e02247d50d77e6.

The core change is right. `charge.amount` is the capture, `charge.amount_refunded` is the
cumulative refunded figure, and reading the former as the latter reported every partial
refund as a full one — to the journal, the refund ledger, the invoice metadata and the
customer notification. Storing the provider's cumulative total and gating the write on
`refundedAmount: { lt: input.refundedAmount }` is the right shape: a redelivery, an
out-of-order pair and two concurrent deliveries all converge on the figure the provider
stated rather than summing deltas. The re-read on `count === 0` reports what is stored
rather than what the lost race hoped for, matching `attributeIfStillUnattributed`.

Also correct:
- keying on the payment intent, not the charge, so a refund cannot land on a sibling
  capture of the same appointment;
- the currency guard refusing rather than writing an incomparable number next to `amount`;
- `amount_captured` before `amount` in the recovery path;
- never throwing back at the webhook, with the invoice path left running when the journal
  reversal is skipped;
- additive migration with a default, so pre-existing rows read 0, which is what they mean.

One non-blocking gap, for a follow-up rather than this PR:

A partial refund on an UNATTRIBUTED receipt makes it permanently unattributable.
`recordRefund` writes `PARTIALLY_REFUNDED` unconditionally, and
`attributeIfStillUnattributed` returns early on `existing.status !== "UNATTRIBUTED"`, so a
later redelivery of the capture webhook carrying the organisation can no longer fill it in.
`reverseJournalledCapture` creates exactly that state by construction: it journals the
recovered capture UNATTRIBUTED and then immediately reduces it. The residual is the part
`refundedReceiptStatus`'s own comment says "still needs an owner found for it", and after
this it has no path to gain one. Narrow ordering (Stripe retries a capture for up to three
days), no money lost, nothing corrupted — but worth either widening the attribution
predicate to include `PARTIALLY_REFUNDED` or noting it where #3163 picks up allocation.

Minor: a refund exceeding the capture is silently accepted as `REFUNDED`. Deliberate per
the comment, and an epsilon would be worse, but an over-refund is a fact a reconciler would
want logged.

Tests: 21 added cases covering the reversal, the sentinel merchant account, the increase
guard, the version bump, the re-read fallback, the missing capture, the currency mismatch,
the storage failure, and on the Stripe side the intent keying, the recovery amount, the
missing-intent and missing-figure paths. "Tests Must Be Able To Fail" is green at this sha.

Approving. Hold the merge until `core` finishes — the backend test job was still running
when I read the checks; everything else on this sha is green.

aupyay:
Pushed `89768de0a` for the review finding. It was real and this PR introduced it — nothing wrote the refunded states before it.

**The regression.** `recordRefund` writes `PARTIALLY_REFUNDED`, and the attribution path in `journalCapture` gated on the status still being `UNATTRIBUTED`. So a refund locked attribution out for good: a partially refunded capture whose owner was never resolved kept a residual in the reconciliation queue with no organisation and no way to acquire one, because the later delivery that finally knows the organisation is refused by the guard.

**The fix.** The gate is the missing organisation, which is exactly what `UNATTRIBUTED` meant before a refund could overwrite it (`initialReceiptStatus` returns `UNATTRIBUTED` iff there is no organisation). Whose a capture was and what happened to it are different questions. A receipt an operator has already acted on still cannot be rewritten — acting on one is what gives it an organisation.

A refunded status is carried through rather than recomputed, so attribution never erases a reversal. `initialReceiptStatus` would have said `ALLOCATED` for an organisation with an invoice, and writing that would have deleted the only record that the money had gone back.

**Atomicity is preserved and widened, not traded away.** Both halves of the state the decision was taken from are in the `WHERE` — `organisationId: null` and the exact `status` that was read — so it is still a compare-and-set and not a check followed by a write. That is strictly stronger than before: a refund landing between the read and the write now matches zero rows and is re-read, where the old single-status predicate would have let a stale `UNALLOCATED` decision overwrite it.

## Evidence at `89768de0a`

| Check | Result |
|---|---|
| `npx tsc --noEmit -p apps/backend/tsconfig.json` | rc 0 |
| `npx eslint` on both changed source files, from `apps/backend` | rc 0 |
| `npx jest` (whole backend package) | rc 0, **508 suites / 12145 tests passed, 0 skipped** |

Two regression tests, and **6 mutations, 5 killed by tests, 1 killed by the type checker**:

| Mutation | Killed by |
|---|---|
| the gate put back on the `UNATTRIBUTED` status | both new tests |
| attribution recomputes the status unconditionally | both new tests |
| the missing-owner predicate dropped from the `WHERE` | 2 tests |
| the read status dropped from the `WHERE` | 2 tests |
| `REFUNDED` left out of the protected set | does not erase a fully refunded state when it attributes |
| `organisationId` dropped from the recovery `select` | **not killed by any test** — `tsc` rc 2, `TS2345 … is not assignable to parameter of type '{ … organisationId: string \| null; }'` |

That last one is worth stating plainly rather than counting as a test kill. With Prisma mocked, a test cannot see a field missing from the projection — the mock returns it either way — and dropping it would make `existing.organisationId` `undefined`, so `undefined !== null` would silently disable attribution entirely. The gate that catches it is `core / Static (backend)`, not jest.

Branch coverage of `provider-receipt.ts` is 98.03%; the one uncovered line (251) is the pre-existing `rawProviderPayload` branch from #3270 and is outside this PR's hunks.

Existing fixtures in the attribution suite gained the `organisationId` the real `select` returns; the test whose name said "has already left UNATTRIBUTED" now says "whose owner is already known" and asserts the same thing with `ALLOCATED` plus an organisation. No assertion was weakened and no test was removed.

ankit-yc (dismissed):
Re-reviewed at 89768de0a1d893f8ee5ab1645bf612f61800ad73 (incremental from e1312da9).

The finding is properly fixed, and fixed at the right level. The gate is now the missing
organisation rather than the UNATTRIBUTED status, which is the predicate that actually
expresses the question - whose the capture was is a different fact from what happened to it,
and only the first is what attribution fills in. Carrying an existing PARTIALLY_REFUNDED or
REFUNDED through instead of recomputing `initialReceiptStatus` closes the other half, where
attributing would have written ALLOCATED over a refund.

The compare-and-set is right. Both halves of the state the decision was taken from -
`organisationId: null` and the exact `status` that was read - are in the WHERE, so this is
still one atomic `updateMany` and not a check followed by a write. Adding `organisationId`
to the `select` is what makes that possible; agreed that dropping it again is a tsc failure
rather than a test failure, because a mocked `select` returns whatever the mock was given.
That is an honest account of the mutation arm rather than a claim the test suite catches it.

The rename is complete: `attributeIfStillUnattributed` has no remaining references anywhere
under `apps/` or `packages/` at this sha, and the single call site moved with it.

The test pins the shape rather than the outcome - it asserts the WHERE is exactly
`{id, organisationId: null, status: "PARTIALLY_REFUNDED"}` and that `data.status` is the
carried value and not the `ALLOCATED` that `initialReceiptStatus` would have produced. That
is the assertion that would fail if either half of the gate were dropped later.

One non-blocking observation, unchanged in kind from before but now much narrower:

A lost compare-and-set is not retried. If a refund lands between the read and the write, the
WHERE matches no row, the fresh read reports the refunded status, and the attribution is
simply not applied - the receipt keeps `organisationId: null` with no further trigger to
attribute it unless another capture redelivery happens to arrive. That is the same shape as
the bug being fixed, surviving inside the race window rather than across the whole life of a
partially refunded receipt, so it is a genuinely smaller hole. One re-read-and-retry pass
would close it; worth a line in #3163 if it is not done here, since that is where allocation
picks the receipt up.

Approving. `core`, CodeQL javascript-typescript and SBOM were still running when I read the
checks; 16 green, nothing failing.

aupyay:
Pushed `0449e8247` for the lost-CAS finding. Also real, also introduced here.

**The regression.** Before this PR a lost compare-and-set had exactly one cause: another delivery attributed the receipt, which leaves nothing to do. Making a refund move the status gave it a second, and on that one the delivery holding a known owner walked away silently — the residual stayed in the reconciliation queue unattributed, with no path to acquiring an owner, which is the same end state the previous round fixed.

**The fix.** The swap is attempted at most twice, each attempt against the state just read rather than the stale one it started from. A refund contributes at most two status transitions (`PARTIALLY_REFUNDED`, then `REFUNDED`), so a second attempt closes the ordinary case. Bounded rather than a spin — a journal write on a webhook has no business retrying forever — and it warns when it still gives up rather than swallowing an attribution it could not land.

Both of the review's specific points are in:

- **the re-read selects `organisationId` as well as `status`.** Without it the loop cannot tell the two causes apart, and `undefined !== null` would read every lost race as benign.
- **the benign cause returns before the warning.** The loop can exit with an owner that arrived on the last re-read, at which point the condition is no longer consulted, so the check is repeated after the loop rather than left to it. `does not warn when the race was lost to another delivery attributing it` is the test that pins it, and it asserts both attempts were spent so the exit really is the post-loop path.

## A suite-isolation problem the mutation run exposed

Scoring `ATTRIBUTION_ATTEMPTS = 1` the first time produced **10 failing tests across two describes**, including `recordRefund` tests that have nothing to do with attribution. That is leakage, not signal: this suite's `beforeEach` called `jest.clearAllMocks()`, which clears call records but leaves queued `mockResolvedValueOnce` values and persistent `mockResolvedValue` implementations in place, so one fewer consumed `findUnique` in an attribution test shifted the queue for everything after it.

Switched to `jest.resetAllMocks()`. All 32 tests in the file pass under isolation, so nothing depended on the leak, and the same mutation now fails **exactly the 3 attribution tests**. A test that passes because a previous test left a mock behind cannot fail for the right reason, and a mutation score computed over a leaking suite is not attributable to the tests it names.

## Evidence at `0449e8247`

| Check | Result |
|---|---|
| `npx tsc --noEmit -p apps/backend/tsconfig.json` | rc 0 |
| `npx eslint` on both changed source files, from `apps/backend` | rc 0 |
| `npx jest` (whole backend package) | rc 0, **508 suites / 12148 tests passed, 0 skipped** |

**7 mutations, 6 killed by tests, 1 killed by the type checker**, run with `tsc` inside the harness this round after the last one showed why:

| Mutation | Killed by |
|---|---|
| one attempt only, no retry | 3 attribution tests |
| the attempt bound raised to 3 | gives up after a bounded number of attempts rather than spinning |
| the retry swaps against the stale status | retries the attribution when a refund moved the row under it |
| the benign-cause early return removed | 2 tests |
| the give-up warning removed | gives up after a bounded number of attempts rather than spinning |
| the refunded status recomputed on the retry | 3 tests |
| `organisationId` dropped from the re-read `select` | **not killed by any test** — `tsc` rc 2, `TS2741: Property 'organisationId' is missing` |

`provider-receipt.ts` branch coverage 98.07%; the single uncovered line (287) is #3270's `rawProviderPayload` branch and is outside this PR's hunks.

ankit-yc (dismissed):
Re-reviewed at 0449e82479e28ec170d0f618a18c5efd2bf9cd65 (incremental from 89768de0).

The retry is correct and it terminates. Each iteration recomputes the carried status from
`observed.status` and puts that same value in the WHERE, so the second attempt is a fresh
compare-and-set against what is actually stored rather than a replay of the first decision.
`attempts` increments unconditionally, so the bound cannot be escaped by any interleaving.

Both points from the last round are in, and the second one is handled more carefully than I
asked for. The re-read now selects `organisationId` alongside `status`, which is what lets
the two causes be told apart at all. And the benign-cause check is repeated AFTER the loop
rather than left to the loop condition - correct, because the loop also exits on attempts
exhaustion, and at that exit `observed` may still carry an owner from the last re-read. A
condition-only check would have fallen through to the warning in exactly that case.

Moving the early return to `if (!input.organisationId)` and letting the loop condition carry
`organisationId === null` is behaviour-preserving: an already-owned receipt runs zero
iterations and returns through the post-loop benign branch with its own status.

`logger.warn` is real - declared on the `Logger` type in `src/utils/logger.ts` and used
across the backend - so the give-up path will not throw at the moment it fires, which is the
worst possible time to find out.

The bounded test pins the bound rather than describing it: `updateMany` asserted to have
been called exactly twice, plus the warning, plus the returned status coming from the fresh
read. That assertion fails if the loop spins, if it gives up early, or if
`ATTRIBUTION_ATTEMPTS` moves without the test being reconsidered. The benign case has its
own test asserting the warning is NOT emitted, which is the arm that would catch the alarm
being reintroduced.

One cosmetic finding, non-blocking:

`ATTRIBUTION_ATTEMPTS` and its doc block were inserted between the function's own JSDoc and
the function, so `provider-receipt.ts` now has two adjacent `/** */` blocks at lines 94-118
and 119-129. The 25-line block describing `attributeIfOwnerStillUnknown` - the gate, the
compare-and-set, the reason it posts no credit - is no longer attached to anything, and the
function itself is now undocumented as far as hover, TSDoc or any doc extractor is
concerned. Hoisting the constant and its comment above the function's doc block restores it;
it is one move and changes no behaviour.

Approving as it stands rather than holding a fourth round on comment placement. If you do
make that move, ping me and I will re-approve - the push will dismiss this row automatically.

23 checks green, 7 still running, nothing failing at this sha.

ankit-yc (dismissed):
Re-reviewed at d81fc2d5059d5f75143b7800ca2c46102c25f178.

Verified the pure-move claim independently rather than reading it: one file, 13 insertions
and 13 deletions, and the sorted multisets of added and removed lines compare equal. Nothing
entered or left the file.

That check is blind to the thing the commit is actually for, though - a multiset-identical
diff says the lines are the same, not that they ended up in an order that means anything -
so the ordering itself: `ATTRIBUTION_ATTEMPTS` and its doc block now come first, then the
25-line block for `attributeIfOwnerStillUnknown` immediately followed by the function. Each
JSDoc attaches to what it documents again.

Approving. No re-verification of the retry logic needed at this sha, since no logic line
moved; the reasoning in my 0449e824 review stands unchanged.

ankit-yc (approved):
Re-approving at `70b5cc7e5`. The only change since the previous approval at `d81fc2d50` is the Sonar `typescript:S7776` fix: `REVERSED_STATUSES` becomes a `ReadonlySet` and `.includes` becomes `.has`.

Checked rather than assumed:

- **the membership semantics are identical.** Both members are string literals of the same enum, so `Set` construction deduplicates nothing and `.has` is exact equality on the same two values `.includes` compared. No behaviour rides on order or on the value being indexable, and the only reader is the one call site.
- **it is a `ReadonlySet`, not a `Set`.** The array was `readonly` and the replacement keeps that, so nothing gains the ability to mutate the guard at runtime.
- **the guard still binds.** Scoring `.has(observed.status)` to a constant `false` fails 3 tests in `provider-receipt.service.test.ts` — it did not silently become unreachable in the rewrite.

`tsc --noEmit` clean, both touched suites green (169 tests), and the ratcheted Sonar job that rejected `d81fc2d50` on `issues: 1 (ceiling 0)` now passes with all checks green.

The taste question is worth a sentence, since two elements is not a performance argument: a `Set` states set membership where an array states an ordered list, and this value is only ever asked whether a status is in it. The rule is right about the type even where it is wrong about the reason.

PR #3278, opened 2026-09-18
fix(desktop): resolve the Linux binary from the packager executableName

### LINKED ISSUE #3261: Desktop: the apply-fuses afterPack hook uses productFilename on Linux, so every Linux build aborts with ENOENT
`apps/desktop/scripts/apply-fuses.js` is the `afterPack` hook for electron-builder. It
computes the binary to flip fuses on from `appInfo.productFilename` for every platform
that is not macOS or Windows:

```js
// apps/desktop/scripts/apply-fuses.js:8-17
const getExecutablePath = (context) => {
  const productFilename = context.packager.appInfo.productFilename;
  if (context.electronPlatformName === 'darwin') {
    return path.join(context.appOutDir, `${productFilename}.app`);
  }
  if (context.electronPlatformName === 'win32') {
    return path.join(context.appOutDir, `${productFilename}.exe`);
  }
  return path.join(context.appOutDir, productFilename);   // <- Linux
};
```

On Linux that file never exists. electron-builder renames the unpacked binary to the
Linux packager's `executableName`, which is derived from the **package name**, not from
`productName`:

- `app-builder-lib/out/electron/ElectronFramework.js:43-44`
  `const executable = path.join(appOutDir, linuxPackager.executableName); await rename(path.join(appOutDir, electronBranding.projectName), executable)`
- `app-builder-lib/out/linuxPackager.js:16`
  `this.executableName = executableName == null ? this.appInfo.sanitizedName.toLowerCase() : sanitizeFileName(executableName)`
- `app-builder-lib/out/appInfo.js:115-125`
  `get name() { return this.info.metadata.name }` and `get sanitizedName() { return sanitizeFileName(this.name) }`

`apps/desktop/package.json` sets neither `build.executableName` nor `build.linux.executableName`,
so the default applies. Computed with the installed `builder-util`:

```
sanitizeFileName('Yosemite Crew PIMS')          -> "Yosemite Crew PIMS"   (productFilename)
sanitizeFileName('@yosemite-crew/desktop')
  .toLowerCase()                                -> "@yosemite-crewdesktop" (executableName)
```

The hook looks for the first and the packer wrote the second. `@electron/fuses` then does
`fs.open(fuseFilePath, 'r+')` with no existence check (`@electron/fuses/dist/index.js:106`),
so `afterPack` throws `ENOENT` and the build aborts before any AppImage or deb is produced.

**Affected commands:** `pnpm desktop:dist:linux`, `pnpm desktop:dist`, `pnpm desktop:publish` -
every script that includes `--linux`.

**Why nothing has caught it.** `build.linux.target` declares AppImage and deb for x64, so
Linux is a configured, intended target. But `.github/workflows/desktop-release.yml` only ever
runs `desktop:dist:win` (windows-latest) and `desktop:dist:mac` (macos-14); there is no Linux
build job anywhere, and `desktop-e2e.yml` runs a macOS + Windows matrix only. No CI job
executes electron-builder for Linux, so the hook has never run on that platform.

The macOS and Windows branches are correct and should not change - `ElectronFramework.js:46-48`
does rename the Windows binary to `${productFilename}.exe`, and `createMacApp` produces
`${productFilename}.app`. Only the fallthrough branch is wrong.

**Fix:** on Linux, take the name from the packager rather than from `appInfo.productFilename` -
`context.packager.executableName` is exactly the value electron-builder used for the rename.
Setting an explicit `build.linux.executableName` would also work and would additionally stop
the shipped Linux binary being named `@yosemite-crewdesktop`, which is worth doing on its own.

Either fix needs a regression test that asserts `getExecutablePath` for
`electronPlatformName === 'linux'` matches the packager's `executableName` and not
`productFilename` - the two are equal in no configuration this repo uses, so the assertion is
meaningful.

Unrelated to #3229; found while auditing the surfaces that Electron bump does not exercise.
This is pre-existing on Electron 43 and is not a regression from 44.

aupyay:
## Problem

`apps/desktop/scripts/apply-fuses.js` is the `afterPack` hook. It computed the binary to flip fuses on from `appInfo.productFilename` for every platform that is not macOS or Windows.

On Linux electron-builder renames the unpacked binary to the **Linux packager's `executableName`**, which is derived from the package `name`, not from `productName` (`app-builder-lib` `ElectronFramework.beforeCopyExtraFiles` → `linuxPackager.js` constructor). Those are different strings here, so the hook named a file the packer never wrote. `@electron/fuses` opens the path with no existence check, so `afterPack` threw `ENOENT` and the build aborted before any AppImage or deb was produced.

Affects `desktop:dist:linux`, `desktop:dist` and `desktop:publish`. Nothing caught it because **no CI job builds the desktop app for Linux** — `desktop-release.yml` runs `desktop:dist:win` and `desktop:dist:mac` only, and `desktop-e2e.yml` matrixes macOS + Windows.

Closes #3261 once this reaches `main`.

## Reproduction, on this branch's parent

```
$ npx electron-builder --linux --dir
  • packaging  platform=linux arch=arm64 electron=44.1.1 appOutDir=dist/linux-arm64-unpacked
  ⨯ ENOENT: no such file or directory, open '.../dist/linux-arm64-unpacked/Yosemite Crew PIMS'  failedTask=build

$ ls dist/linux-arm64-unpacked/ | head -1
@yosemite-crewdesktop
```

The packer wrote `@yosemite-crewdesktop`; the hook looked for `Yosemite Crew PIMS`.

## Fix

1. **Root cause** — on Linux, read the name from the packager (`context.packager.executableName`) rather than from `appInfo.productFilename`. This mirrors `app-builder-lib`'s own `PlatformPackager.addElectronFuses`, which does exactly this branch. macOS and Windows keep `productFilename`, which is what electron-builder renames to on those platforms.
2. `mas` now joins the `darwin` branch. Previously it fell through to the Linux case; with change 1 in place it would have read a property `MacPackager` does not have, so this is required, not incidental.
3. **Separate defect, named in the issue** — set an explicit `build.linux.executableName`. Without it the default is the sanitised, lowercased package name, so the binary the AppImage and deb install, and the `.desktop` `Exec=` entry, were `@yosemite-crewdesktop`.

Changes 1 and 3 are independent. Verified below that change 1 alone fixes the build.

## Verification

All commands run from `apps/desktop/` on this branch.

**Linux pack completes and the fuses are really flipped:**

```
$ npx electron-builder --linux --dir   # exit 0, zero ENOENT
$ ls dist/linux-arm64-unpacked/ | grep yosemite
yosemite-crew-pims
$ npx electron-fuses read --app dist/linux-arm64-unpacked/yosemite-crew-pims
Fuse Version: v1
  RunAsNode is Disabled
  EnableCookieEncryption is Enabled
  EnableNodeOptionsEnvironmentVariable is Disabled
  EnableNodeCliInspectArguments is Disabled
  EnableEmbeddedAsarIntegrityValidation is Enabled
  OnlyLoadAppFromAsar is Enabled
  LoadBrowserProcessSpecificV8Snapshot is Disabled
  GrantFileProtocolExtraPrivileges is Enabled
  WasmTrapHandlers is Enabled
```

**Change 1 alone is sufficient** — with `build.linux.executableName` removed and only the hook fixed, the pack exits 0 with zero `ENOENT` and the binary is `@yosemite-crewdesktop`. So the config entry is a naming improvement, not the fix.

**No regression on the platforms that do build in CI:**

| arm | result |
|---|---|
| `npx electron-builder --mac --dir` | exit 0, `dist/mac-arm64/Yosemite Crew PIMS.app` |
| `npx electron-builder --win --dir` | exit 0, `dist/win-arm64-unpacked/Yosemite Crew PIMS.exe` |

**Gates:**

| command | result |
|---|---|
| `pnpm run type-check` | exit 0 (`tsc --noEmit` + `tsc --noEmit -p tsconfig.test.json`) |
| `pnpm run lint` | exit 0, no errors, no warnings |
| `pnpm run archlint` | exit 0, no finding on any touched file |
| `pnpm run test:coverage` | 71 suites, 1102 tests passed; statements 98.37%, branches 90.88%, functions 97.68%, lines 98.37% |

## Tests

`tests/apply-fuses.test.ts` gains a `getExecutablePath` block. The hook can now be loaded rather than only scanned as text: `@electron/fuses` is ESM-only, so the suite mocks it, and removing that mock fails the suite — the mock is load-bearing, not decorative. The fixture deliberately gives `productFilename` and `executableName` different values, so an assertion on one cannot pass by reading the other.

Mutation runs, each with the assertions in place:

| mutation | result |
|---|---|
| Linux branch reverted to `productFilename` | 2 failed, 7 passed |
| `mas` removed from the darwin branch | 1 failed, 8 passed |
| `build.linux.executableName` removed | 1 failed, 8 passed |
| none | 9 passed |

## Follow-up, not in this PR

There is still no Linux leg in `desktop-release.yml` or `desktop-e2e.yml`, which is why a configured, intended target went unbuilt for the life of the app. The unit test above catches this particular defect; it does not catch the next packaging break on a platform CI never exercises.


ankit-yc (approved):
Reviewed at 3e66bb84a5ee9c34bca2239b8ce9908754f8a36f.

The diagnosis holds and I checked the mechanism against the installed app-builder-lib
(26.15.3) rather than the comment: `LinuxPackager`'s constructor sets
`executableName = linux.executableName ?? config.executableName ?? appInfo.sanitizedName.toLowerCase()`,
and `appInfo.sanitizedName` is `sanitizeFileName(metadata.name)` — the package `name`, not
`productName`. So the comment's claim is right: on Linux the unpacked binary is renamed to a
string `productFilename` does not know about, and @electron/fuses opens the path without an
existence check, which is the ENOENT. Reading `context.packager.executableName` is the same
source `PlatformPackager.addElectronFuses` uses.

One interaction worth naming, because it makes the two changes look redundant when they are
not. `AppInfo` computes `productFilename` as
`platformSpecificOptions?.executableName ?? config.executableName` when either is set,
falling back to `sanitizedProductName` otherwise, and `PlatformPackager.prepareAppInfo`
constructs its `AppInfo` with the platform-specific options. So now that
`build.linux.executableName` is set, `appInfo.productFilename` on the Linux packager
resolves to `yosemite-crew-pims` as well — the config line on its own would also have cured
the ENOENT. They are two independent fixes for one symptom, not one fix plus its config:
the script change is the durable one (it survives the config line being removed), and the
config line is separately worth having because a scoped package `name` makes the default
executable name unusable. Please don't let a later cleanup drop either as "already covered".

The regression test is the right shape. `PRODUCT_FILENAME` and `EXECUTABLE_NAME` are
deliberately different, so the positive assertion cannot pass by reading the wrong field,
and the `.not.toBe(path.join(OUT_DIR, PRODUCT_FILENAME))` case fails on the original
expression itself rather than on a fixture coincidence. The `@electron/fuses` proxy mock is
what lets the hook be exercised instead of scanned as text — worth keeping.

Two follow-ups, neither blocking:

- Nothing in CI builds for Linux. `Desktop E2E` is macos-latest and windows-latest only, so
  this unit test is the entire guard against the regression coming back. An
  `electron-builder --linux --dir` pack job would be the arm that actually caught it.
- The test comment states the fallback would be `@yosemite-crewdesktop`. I could not verify
  that: `sanitizeFileName` is not resolvable in the installed `app-builder-lib/out` tree
  from the path the compiled code imports. It is illustrative only — the assertion is a
  regex over the configured value, not over a derived default — but it is a justifying
  sentence nothing checks.

`mas` joining the darwin branch is correct; a mas pack produces a `.app` at the same place.

No other reference to the Linux binary name to update: nothing under `apps/desktop` (js/ts/
json) and no `*.yml`/`*.yaml`/`*.desktop`/`*.sh` in the repo outside `node_modules` mentions
`executableName` or the sanitised package name. And because every `--linux` build aborted
before this, no already-shipped artifact is being renamed by it.

Approving. Desktop E2E (macos, windows), `core` and grype were still running when I read the
checks; 17 green, nothing failing.

PR #3279, opened 2026-09-18
fix(desktop): declare the macOS 13 floor Electron 44 requires

### LINKED ISSUE #3260: Desktop: Electron 44 raises the macOS floor to Ventura and drops two prebuilt architectures, with nothing in the repo to record it
Upgrading the desktop shell from Electron 43.5.1 to 44.1.1 (#3229) changes the set of
platforms the packaged app can run on. None of it is visible in the repository, so it
would ship silently.

**What changes**

| | 43.5.1 | 44.1.1 |
|---|---|---|
| macOS floor | Monterey (12) | Ventura (13) |
| `win32-ia32` prebuilt | published | dropped |
| `linux-armv7l` prebuilt | published | dropped |
| Chromium | 150.0.7871.250 | 152.0.7977.65 |
| Node | 24.19.0 | 24.19.0 (unchanged) |
| V8 | 15.0 | 15.2 |

**Why nothing in the tree goes stale**

Searched at the PR head and clean, 0 matches each: `minimumSystemVersion`,
`LSMinimumSystemVersion`, `MACOSX_DEPLOYMENT_TARGET`, any Monterey/Ventura support
prose, `ia32`, `armv7l`. The electron-builder config targets neither dropped
architecture, so no build breaks and no documented claim becomes wrong.

That is exactly the problem: there is no support matrix to fall out of date, so the
macOS 12 users who stop being able to run a new build get no warning from us and the
change leaves no trace in the repo.

**Decision needed**

1. Accept the floor and record it - add `minimumSystemVersion` to the electron-builder
   mac config and a short supported-platforms note, so the next bump has something to
   contradict.
2. Or hold the desktop shell on Electron 43 until macOS 12 is out of the support window.

Option 1 is the recommendation: the runtime bump carries the Chromium 150 -> 152
security work, and a pinned `minimumSystemVersion` makes the next such change loud
instead of silent.

This does not block #3229 - the bump itself is sound and its one code-level defect was
fixed separately in #3257.

---

**Correction, same day.** The Node row first published here read `24.17.0 -> 24.19.0`.
That is wrong: it compared Electron **43.0.0**, the first release of the 43 line, against
44.1.1, rather than **43.5.1**, which is what this repository is actually pinned to. Read
from `DEPS` at both exact tags:

    v43.5.1   chromium_version 150.0.7871.250   node_version v24.19.0
    v44.1.1   chromium_version 152.0.7977.65    node_version v24.19.0

So **Node does not move at all** across this bump, and the only runtime change is
Chromium 150 -> 152. The table above has been corrected. Nothing else in this issue
depended on the Node figure; the platform-floor argument stands unchanged.

aupyay:
## PR Checklist

The desktop shell moved to Electron 44.1.1 in #3229. That raised the macOS floor from 12 Monterey to 13 Ventura and dropped the `win32-ia32` and `linux-armv7l` prebuilts, and nothing in the repository records it: no `minimumSystemVersion`, no supported-platforms note. The desktop README also claims Intel Mac and Linux AppImage downloads that no release has shipped, and an Electron 39 badge.

Decision on #3260: accept the Ventura floor rather than hold the shell on Electron 43, which would mean shipping an Electron line that stops receiving security fixes.

## What is the new behavior?

- `build.mac.minimumSystemVersion` is `"13.0"`. electron-builder 26 writes it into `Info.plist` as `LSMinimumSystemVersion` (`app-builder-lib/out/macPackager.js`, and the option is documented in `macOptions.d.ts` as "Corresponds to `LSMinimumSystemVersion`"), so macOS 12 refuses to open the app with a system message.
- `apps/desktop/README.md` gains a **Supported platforms** section: macOS 13+ on Apple Silicon, Windows 10+ x64, Linux x64 from source only, recorded against Electron 44.1.1 with the two dropped prebuilts named. The download table and Electron badge are corrected to match.
- `tests/platform-support.test.ts` pins the floor at `13.0`, and pins it to the Electron 44 major.

Why the second assertion: the configured key **overrides** the `LSMinimumSystemVersion` inside Electron's own bundle. Without it, a later Electron that raises its floor would still be packaged claiming 13.0, and the app would crash on that OS instead of being refused. On the next Electron major that test fails with instructions: read the key from the new `Electron.app/Contents/Info.plist`, then update the config, the README and the test.

Facts behind the README, each checked rather than assumed:

| Claim | Evidence |
|---|---|
| Electron 44 floor is 13.0 | stock `Electron.app/Contents/Info.plist` in `electron-v44.1.1-darwin-arm64.zip`: `LSMinimumSystemVersion` = `13.0` (43.4.0, the version v0.1.0-beta.5 shipped on: `12.0`) |
| Dropped prebuilts | `gh release view` on electron/electron: v43.5.1 publishes `win32-ia32` and `linux-armv7l` zips, v44.1.1 publishes neither |
| Apple Silicon only | `desktop-release.yml` builds mac on `macos-14` (arm64) and `build.mac.target` has no `arch`; electron-builder then uses `process.arch`. v0.1.0-beta.5 assets are `mac-arm64` only |
| Linux not in releases | `desktop-release.yml` has `windows` and `macos` build jobs only; beta.5 has no Linux asset |
| Config overrides the stock value | `electron-builder --mac --dir -c.mac.minimumSystemVersion=12.0` produced a bundle whose `LSMinimumSystemVersion` is `12`, below Electron 44's own 13.0 |

Validation (desktop only):

- `pnpm --filter @yosemite-crew/desktop run lint` - exit 0
- `pnpm --filter @yosemite-crew/desktop run type-check` - exit 0
- `pnpm --filter @yosemite-crew/desktop run test` - exit 0, 72 suites, 1111 tests, statements 98.37%, branches 90.93%
- Mutation check of the new test, `package.json` restored with `cp` and `cmp` identical afterwards: key removed -> fails; `"12.0"` -> fails (`Expected "13.0", Received "12.0"`); `electron` `^45.0.0` -> fails on the major pin; restored -> 2/2 pass
- `electron-builder --mac --dir` (unsigned, `CSC_IDENTITY_AUTO_DISCOVERY=false`, no notarization) -> `Yosemite Crew PIMS.app/Contents/Info.plist` has `LSMinimumSystemVersion` = `13.0`
- Aikido scan of the new test file: no issues

Known gap, not addressed here: v0.1.0-beta.5 shipped on Electron 43.4.0, so a Monterey user on it will be offered the next release by `electron-updater`, install it, and then have macOS refuse to open it. `electron-updater` can skip an update whose feed declares a higher `minimumSystemVersion` (a Darwin kernel version, so `22.0.0` for Ventura), but electron-builder 26's `releaseInfo` has no such field, so it would need a post-build edit of `latest-mac.yml` in the release workflow. Worth a separate issue if Monterey installs matter.

## Related Issue(s)

Closes #3260

Closing keywords do not fire on merges into `dev`, so this resolves #3260 but the issue needs closing by hand once this merges.


aupyay:
Two small gaps this leaves, found while a parallel branch was cutting the same issue. Both are in scope for #3260 and neither is a blocker for merging this.

**1. Nothing enforces the dropped prebuilts.** The README names `win32-ia32` and `linux-armv7l`, but the packaging config is what would actually re-request them, and no test reads it. A fourth assertion covers that:

```ts
it('requests no architecture Electron has stopped publishing', () => {
  const requested = (['mac', 'win', 'linux'] as const).flatMap((platform) =>
    (pkg.build[platform].target ?? []).flatMap((entry: unknown) =>
      typeof entry === 'string' ? [] : ((entry as { arch?: string[] }).arch ?? []),
    ),
  );
  // Only meaningful if some arch is actually declared to inspect.
  expect(requested.length).toBeGreaterThan(0);
  for (const arch of ['ia32', 'armv7l']) {
    expect(requested).not.toContain(arch);
  }
});
```

Mutation-checked against the current tree, `package.json` restored from a snapshot and md5-identical afterwards:

| mutation | result |
|---|---|
| `build.win.target[0].arch` -> `["x64", "ia32"]` | fails: `Expected value: not "ia32"`, received `["x64","ia32","x64","x64","x64"]` |
| all four `arch` lists emptied | fails: `Expected: > 0`, `Received: 0` |
| restored | 4/4 pass |

The second arm matters because without the length assertion the guard passes vacuously if the arch lists are ever restructured away.

**2. `apps/desktop/AGENTS.md` line 5 still reads `**Stack:** Electron 43`.** It is hedged ("check `apps/desktop/package.json` for the current pin"), so it is not load-bearing, but it is the same class of stale record this issue is about and this PR corrects the README badge beside it.

Happy to land both as a follow-up once this merges, or they can be folded in here - whichever the reviewer prefers.

ankit-yc (approved):
Approving at 8721c7158.

- `build.mac.minimumSystemVersion: "13.0"` reaches `LSMinimumSystemVersion` (app-builder-lib 26.15.3 `macPackager.js` applyCommonInfo), confirmed in an unsigned `--dir` build; the same build with "12.0" produces 12.0, so the key overrides Electron's own value, which is why the pin test ties it to the Electron 44 major.
- README corrections check out against the release workflow (mac arm64 only, no Linux job) and the Electron 44 release assets (ia32/armv7l dropped).
- Pin test fails on key removed, "12.0", and electron ^45; 72 suites / 1111 desktop tests pass; CI 34/34.

The two gaps in the comment above (arch guard with the non-vacuous length check, and the Electron 43 line in `apps/desktop/AGENTS.md`) are good; please land them as the follow-up. The Monterey auto-update gap is filed separately as #3280.

PR #3282, opened 2026-09-18
feat(backend): let an operator see the captures the journal could not attribute

aupyay:
Part of #3170. Follows #3270 (the journal) and #3277 (refund reversal); this is the read side.

## The gap

#3270 made every captured payment durable, including the five webhook exits where the card is charged and no invoice can be minted. It had no read surface. An unattributed capture was therefore recorded and then seen by nobody — the gap moved rather than closed, and nothing in the product could tell an operator that money was sitting unallocated.

## What this adds

`GET /finance/organisation/:organisationId/provider-receipts` — the queue an operator works through.

- keyset-paged at 50 (default and ceiling), filtered by `status` (once or repeated) and by a `capturedFrom`/`capturedTo` window
- `requireWebAuth` → `withOrgPermissions()` → `billing:view:any`. Read-only, so the billing *view* permission: an unattributed capture has to be visible before anyone can decide what to do with it, and nothing on this route changes a receipt.
- the app-level `express-rate-limit` in `app.ts` covers it, same as the sibling organisation reads on this router

## Scope, and why it is two sets

A receipt carrying the organisation's id is plainly theirs. A receipt carrying **no** organisation but captured into the connected merchant account **that organisation owns** is also theirs — not a guess about who paid, but the account the money is sitting in. Without that second arm every unattributed capture stays invisible to everyone, which is the state the issue exists to end.

A platform-account capture with no organisation is in **no** queue. `PLATFORM` is a sentinel every tenant shares, so matching on it would show one organisation another's money, and nothing in such a receipt says whose it is. The issue is explicit that an unattributed receipt is never assigned by guessing.

The merchant account is read from the `Organization` record, never from the request, so no caller can name another tenant's connected account and read its unattributed money.

## Two details worth the reviewer's attention

**`rawProviderPayload` is outside the projection.** It is the provider's whole object — billing name, email and address among them — kept so a human can investigate one receipt. A list endpoint returning it would put that on the wire for every row of every page. The response type is *derived* from the projection (`Prisma.ProviderReceiptGetPayload<{select: typeof RECONCILIATION_FIELDS}>`) rather than written out beside it, so a column added later cannot silently join the response and a hand-written twin cannot silently drift.

**Paging is an explicit exclusive `(createdAt, id)` comparison, not Prisma's `cursor` + `skip: 1`.** That pair is exclusive only while the cursor row is still in the filtered set: `cursor` seeks to the row, `skip` is an OFFSET on the *result*. Attributing or refunding a receipt takes it out of the status filter being paged on, and the OFFSET then eats a real row while `hasMore` still claims the list was complete. On a work queue, rows leaving the filter mid-page is the ordinary case rather than the rare one. Reuses `encodeKeysetCursor`/`parseKeysetCursor` from `services/shared/pagination`; no new cursor format.

Sorting is on `createdAt`, not `capturedAt`. `capturedAt` comes from the provider, can arrive out of order, and can be recovered from a refund event long afterwards — sorting on it would shuffle rows between pages. It is a filter and a column, not the sort key.

## Evidence at `7e3606e6e`

| Check | Command | Result |
|---|---|---|
| Types | `pnpm --filter backend exec tsc --noEmit` | no error in any file this PR touches |
| Lint | `npx eslint` on the three changed source files, from `apps/backend` | rc 0 |
| Focused suites | `pnpm --filter backend exec jest test/services/provider-receipt-reconciliation.service.test.ts test/controllers/finance.controller.test.ts test/routers/finance.router.test.ts --runInBand` | 3 suites / 79 tests passed |
| Whole backend package | `pnpm --filter backend run test` | rc 0, **509 suites / 12149 tests passed** |

The typecheck is only meaningful with `@yosemite-crew/database` built — unbuilt, `prisma` resolves to `any`, `findMany` infers `any`, and the projection type is never checked at all. Built, a `never` probe confirms `rows` is the full 14-field payload, so the projection is genuinely type-checked here.

### Mutation runs — 14 mutations, 14 killed

Each applied to a snapshot of the sources, the three suites run, then restored and the restore verified byte-identical with `diff -q`.

| Mutation | Killed |
|---|---|
| drop the unattributed-merchant arm from the scope | yes |
| let the `PLATFORM` sentinel into a tenant scope | yes |
| take the merchant account from the caller instead of the organisation record | yes |
| return `rawProviderPayload` in the projection | yes |
| sort by the provider's `capturedAt` | yes |
| page with an offset instead of the keyset comparison | yes |
| read only the page, not the extra row | yes |
| drop the status filter | yes |
| drop the capture-window filter | yes |
| take the organisation from the path instead of the authorized one | yes |
| accept a malformed cursor | yes |
| skip the query validation entirely | yes |
| drop the permission from the route | yes |
| set membership in the reversed-status guard (on #3277's branch) | yes |

One earlier mutation — `input.organisationId ?? "any"` — survived, and correctly: `organisationId` is a required string, so the mutation changed nothing. It is recorded here because a surviving no-op mutation is not evidence of a test gap, and replacing it with a real one (removing the organisation lookup) killed it.

## Not in this slice

The allocation `POST` and the Finance UI. Allocation of excess as credit needs #3163, which is open; the read surface is independently useful and is what makes the journal observable. Delivery 4 (historical mismatch audit) is untouched.


aupyay:
`0fb99f90b` adds the OpenAPI entry the drift ratchet requires. Worth recording because the gate is easy to trip and the failure text does not say which route caused it.

**The gate fired first, which is what makes the pass meaningful.** At `7e3606e6e` (router entry, no spec entry):

```
openapi-drift: this change widens the documented gap beyond its baseline:
  missingFromSpec: baseline 793 -> now 794
```

The printed list is capped at 40 of 794 rows and my route was not among them; `--json` is how the entry was identified — `GET /v1/finance/organisation/{organisationId}/provider-receipts`.

At `0fb99f90b`: rc 0, 297 spec operations (was 296), `missingFromSpec` back to 793, `missingOrgHeader` 0. Re-run after `lint-staged`'s prettier rewrote the YAML, because the check reads the formatted file, not the one I wrote.

`x-org-id` is declared **optional**, matching the other documented routes that carry the organisation in the path: `extractOrgId` reads params before headers, so a header sent alongside is ignored in favour of the path. Declaring it at all is what keeps `missingOrgHeader` — the gate's one fatal count — at zero; a documented org-scoped operation without it fails the job outright regardless of the ratchet.

Running it locally needs `pnpm --filter "@yosemite-crew/*" run build` first, or the script cannot import the router graph and exits 2 with a module-not-found that looks like a repo problem rather than a missing build.

aupyay:
`11f17ff96` closes a hole I found in my own change while CI was running. Worth calling out because the PR's central safety claim rested on it.

**The assumption.** The second scope arm shows an organisation the unattributed captures that landed in *its* connected merchant account. That is only safe if an account has exactly one claimant — and `Organization.stripeAccountId` carries **no unique constraint**. Two rows sharing an account would have put each organisation's unattributed captures in the other's queue: precisely the cross-tenant read the arm exists to make safe, arriving through the arm itself.

**The fix fails closed.** The claim is now counted rather than assumed, and an ambiguous account is dropped from the scope entirely — those receipts stay out of **both** queues until someone resolves the duplicate. That is the failure that loses nothing: a capture withheld from a queue is still journalled and still recoverable, where a capture shown to the wrong tenant is not.

It logs at `error` when it fires. A shared connected account is a data defect, and a queue quietly missing rows is worse to debug than one that said why.

The count is **skipped** when the organisation has no connected account. `count({ where: { stripeAccountId: null } })` matches every organisation that has never connected one, so running it unconditionally would drop the arm for a reason that has nothing to do with ambiguity — a test pins that it is not called in that case.

**What I did not do:** add `@unique` to the column. That is a migration against existing data with its own backfill and ambiguous-record handling, and the issue is explicit that financial migrations need a dry run and reconciliation first. The guard makes the read safe today without one; the constraint is the right follow-up and does not belong in a read-surface slice.

### Evidence at `11f17ff96`

| Check | Result |
|---|---|
| `pnpm --filter backend exec tsc --noEmit` | no error in any file this PR touches |
| `npx eslint src/services/finance/provider-receipt.ts` | rc 0 |
| `pnpm --filter backend run test` | rc 0, **509 suites / 12152 tests passed** (was 12149; +3) |

**5 further mutations on the new guard, 5 killed**, each applied to a snapshot, run, restored, and the restore verified byte-identical:

| Mutation | Killed |
|---|---|
| trust an account claimed by several organisations (`>= 1`) | yes |
| skip the claimant count entirely | yes |
| count claimants before establishing there is an account | yes |
| count on the wrong column (`id` instead of `stripeAccountId`) | yes |
| drop the ambiguity warning | yes |

Running total for the PR: **19 mutations, 19 killed.**

aupyay:
`b0367357f` merges `dev` now that #3277 has landed (`8e50088ae`). One conflict, in `provider-receipt.ts`, and it is worth saying exactly what it was so a reviewer does not have to reconstruct it: two *independent additions* to the same two regions — my reconciliation helpers and read method against that PR's refund types and `recordRefund`. Neither side modified a line the other touched. Both kept, nothing dropped: `journalCapture`, `recordRefund` and `listForReconciliation` are all present and all three suites that own them pass.

**`refundedAmount` is now in the queue projection**, which it could not be before the merge because the column did not exist on `dev`. It belongs there: the issue's oracle is *captured = applied + unapplied + refunded*, and a `PARTIALLY_REFUNDED` status says money went back without saying how much. Without the column an operator cannot see the residual. Added to the OpenAPI response schema in the same commit.

A stale Prisma client made this look worse than it was for a moment — six `refundedAmount does not exist` errors that were entirely `prisma:generate` not having been re-run after the merge brought the migration in. Regenerating and rebuilding `@yosemite-crew/database` clears all six; nothing in the merge was actually wrong.

### Evidence at `b0367357f`

| Check | Result |
|---|---|
| `pnpm --filter backend exec tsc --noEmit` | no error in any file this PR touches |
| `pnpm exec tsx ../../scripts/ci/openapi-drift.mjs` | rc 0, 297 spec operations, `missingOrgHeader` 0 |
| All five touched suites | 5 suites / 252 tests passed |
| `pnpm --filter backend run test` | rc 0, **509 suites / 12180 tests passed** |

12180 rather than 12152 is #3277's own tests arriving with the merge, not new ones here.

ankit-yc (changes requested):
Reviewed the diff, the pagination helpers it reuses, and the check set. The design holds up: the two-arm scope with `PLATFORM` excluded, reading the merchant account from `Organization` rather than the request, the uniqueness check on `stripeAccountId` before trusting the second arm, and the explicit `(createdAt, id)` comparison instead of `cursor` + `skip: 1` are all the right calls for a work queue whose rows leave the filter mid-page. Projection derived from `RECONCILIATION_FIELDS` rather than hand-written, `rawProviderPayload` out of the list — agreed on both.

All 30 required checks pass. One blocker.

**`js/log-injection` (CodeQL alert 1382, medium) at `apps/backend/src/services/finance/provider-receipt.ts:360` — introduced by this PR.**

```js
logger.error(
  `Connected merchant account of organisation ${organisationId} is claimed by ${claimants} organisations; ...`,
);
```

CodeQL traces `organisationId` back to the path parameter. Reachability is narrow in practice — `resolveAuthorizedOrganisationId` returns the value `withOrgPermissions` already matched against a membership row, so it has to correspond to a real `organizationReference` — but the query is open and the default branch currently carries **zero** open `js/log-injection` alerts, so this would be the first. Worth noting the controller's own 400 branch declines to echo the caller's value for exactly this reason; this line is the one place the principle is not applied.

The fix is the pattern already used in eight modules here (`adverse-event.service.ts:59`, `notification.service.ts:117`, and others):

```js
const forLog = (value: string): string => value.replace(/[\n\r]/g, "");
```

Spelling matters and those files say why: no quantifier in the character class and an **empty** replacement. `/[\n\r]+/` or replacing with a space does not register as a barrier, and a previous attempt at the wrong spelling left alerts open for months.

Non-blocking, take it or leave it: `capturedFrom` later than `capturedTo` returns an empty page rather than a 400. Defensible for a filter, just noting it is unvalidated.

Push the sanitizer and I will re-review and merge.

ankit-yc (approved):
Re-reviewed at c9732a2df. The blocker is cleared.

`js/log-injection` alert 1382 is now `fixed` (fixed_at 2026-09-18T17:46:02Z) on `refs/pull/3282/merge`. The sink carries the house spelling — `replace(/[\n\r]/g, "")`, no quantifier, empty replacement — so it registers as a barrier. The two alerts still open on this ref are the adm-zip advisory on the lockfile and on a vendored manifest; both are open on `dev` as well, so neither is introduced here.

The added test is the part I care about most, and it discriminates. `expect(message).not.toMatch(/[\n\r]/)` fails if the sanitizer goes away, and `expect(message).toContain("org-aERROR forged entry")` pins the *empty* replacement specifically — a replacement with a space renders `org-a ERROR forged entry` and fails that line. So the test can tell apart the two spellings the review was about, not just presence and absence. The neighbouring case asserting only that the error fired stays as it was, which is right; it answers a different question.

Leaving the other seven log calls in the file alone is the correct scope call — they predate this PR and carry no alert on `dev`.

The `capturedFrom` later than `capturedTo` note stays non-blocking and unaddressed by agreement; an empty page is a defensible answer for a filter.

58 checks pass, 4 skipped, none failing. Approving and merging.

PR #3283, opened 2026-09-18
fix(desktop): stop the updater offering a Ventura-only build to Monterey

### LINKED ISSUE #3280: Desktop: the auto-updater will install a Ventura-only release on Monterey Macs
## What happens

v0.1.0-beta.5 shipped on Electron 43.4.0, which runs on macOS 12 Monterey. From the next release the desktop app requires macOS 13 Ventura (Electron 44, recorded in #3279 for #3260).

A Monterey user on beta.5 will be offered that release by the auto-updater, install it, and then macOS will refuse to open the app. They are left with nothing that launches.

## Why it isn't prevented today

electron-updater skips a release when the update feed (`latest-mac.yml`) declares a `minimumSystemVersion` higher than the running OS (Darwin 22.0.0 is Ventura). electron-builder 26's `releaseInfo` has no field for it, so the feed never carries one.

## Suggested fix

In the desktop release workflow, after the mac build and before upload, add `minimumSystemVersion: 22.0.0` to `latest-mac.yml`. Then confirm on a beta.5 install that the updater treats the new release as not applicable on Monterey instead of downloading it.

aupyay:
## PR Checklist

`v0.1.0-beta.5` shipped on Electron 43, which runs on macOS 12 Monterey. From Electron 44 the app requires macOS 13 Ventura (#3260 / #3279). A Monterey user on beta.5 is offered the next release by `electron-updater`, installs it, and macOS then refuses to open it — they are left with nothing that launches.

`electron-updater` already has the guard and only lacks the value. `AppUpdater.checkIfUpdateSupported` (6.8.9, `out/AppUpdater.js:364-379`) reads `updateInfo.minimumSystemVersion` and returns `false` when `semver.lt(os.release(), minimumSystemVersion)`; `isUpdateAvailable` calls it at line 348, before `update-available` fires, so nothing downloads. electron-builder 26.15.3 writes `mac.minimumSystemVersion` into `Info.plist` and into `.pkg` requirements (`app-builder-lib/out/targets/pkg.js:129-131`) and nowhere else — those are its only two references to the key — so the published `latest-mac.yml` never carries it.

Reproduced against the installed libraries, driving the exact two calls the updater makes (`js-yaml` `load` as in `Provider.parseUpdateInfo`, then the `checkIfUpdateSupported` body verbatim), on a sample feed:

| feed | `os.release()` | update offered |
|---|---|---|
| **today (unstamped)** | `21.6.0` Monterey | **`true`** ← the bug |
| stamped | `21.6.0` Monterey | `false` |
| stamped | `22.6.0` Ventura | `true` |
| stamped | `23.6.0` Sonoma | `true` |

## What is the new behavior?

- `apps/desktop/scripts/stamp-update-feed-min-os.js` writes `minimumSystemVersion` into `latest-mac.yml`, translating `build.mac.minimumSystemVersion` into the **Darwin** version `os.release()` reports (`13.0` → `22.0.0`). It is idempotent, refuses a non-mac feed, and refuses a file that is not an update feed.
- The macOS release job runs it after signature verification, then re-uploads the feed over the one electron-builder published. On a tag the build ran `--publish always`, so the unstamped feed is already attached — but the release is still a **draft** and invisible to clients until the `publish` job flips it, so the overwrite is the only version anyone resolves. On `workflow_dispatch` the stamped file reaches the release through the existing artifact.
- The `publish` job now downloads the draft's own `latest-mac.yml` and refuses to make the release public unless it declares the floor the build stamped. `gh release upload --clobber` deletes the existing asset *before* uploading the replacement, so a failed upload would otherwise leave the draft with no feed at all — breaking updates for everyone rather than only the clients below the floor, and nothing downstream reads the assets.
- Docs: `RELEASE.md` (cutting a release) and `docs/update-feed-threat-model.md` (Current Controls + Required Release Checks).

### Why the mapping is a table and not arithmetic

macOS 11–15 sit at Darwin major + 9. macOS 26 Tahoe is Darwin **25** — one *below* what that offset predicts. A formula would have published a floor a whole release too high. Every row was read from the matching `actions/runner-images` image manifest:

| macOS | Darwin | source |
|---|---|---|
| 12.7.6 | 21.6.0 | `images/macos/macos-12-Readme.md` @`19c84748` |
| 13.7.6 | 22.6.0 | `images/macos/macos-13-Readme.md` @`71f01578` |
| 14.8.7 | 23.6.0 | `images/macos/macos-14-Readme.md` (main) |
| 15.7.9 | 24.6.0 | `images/macos/macos-15-Readme.md` (main) |
| 26.6.1 | 25.6.0 | `images/macos/macos-26-Readme.md` (main) |

An unrecorded major fails the release run with the manifest to read, rather than being guessed at.

## Validation

`apps/desktop`, all exit 0:

- `pnpm run type-check`
- `pnpm run lint`
- `pnpm run archlint`
- `pnpm test` — 73 suites, **1127 tests**, 0 skipped; statements 98.37%, branches 90.93%, functions 97.68%

Mutation check of the 16 new tests. `scripts/stamp-update-feed-min-os.js` and `package.json` restored from `cp` snapshots afterwards and md5-identical (`68aec956…`, `666cb45c…`); the restored run is 16/16:

| mutation | result |
|---|---|
| drop the macOS 26 row from the table | 1 failed |
| replace the table lookup with `major + 9` | **2 failed** (incl. the Tahoe row) |
| `stampFeed` appends instead of replacing | 1 failed |
| drop the mac-only feed-name guard | 2 failed |
| drop the `version:`/`files:` feed-shape check | 1 failed |
| guess instead of throwing for an unmapped major | 1 failed |
| raise `build.mac.minimumSystemVersion` to an unrecorded `16.0` | 1 failed |
| restored | 16 passed |

The last arm is the one that matters for the next Electron bump: raising the floor to a macOS major with no recorded kernel version fails here, in a unit test, rather than at release time.

Workflow shell checked directly: `sed -n 's/^minimumSystemVersion: //p'` yields `22.0.0` on a stamped feed and empty on an unstamped one, and the guard's comparison fails on the empty case and passes on the stamped one.

`gh release upload`/`download` do resolve draft releases — `pkg/cmd/release/shared/fetch.go` `FetchRelease` (gh v2.100.0, the version in use) runs a `fetchDraftRelease` lookup in parallel with the published-tag lookup and takes whichever succeeds.

**Not exercised:** the release workflow itself. It only runs on a `v*` tag or a manual dispatch, so the two new steps have not run in CI here. The script, its guards and the shell extraction are covered above; the step wiring is not.

## Related Issue(s)

Closes #3280

Closing keywords do not fire on merges into `dev`, so this resolves #3280 but the issue needs closing by hand once this merges.


ankit-yc (changes requested):
Not approving yet — one red gate, otherwise this is sound.

## Blocker: Aikido is failing

Two `medium` findings, both **"Potential file inclusion attack via reading file"**, on `apps/desktop/scripts/stamp-update-feed-min-os.js` **lines 124 and 125** — the two `readFile` calls in `main`. `feedPath` comes straight off `argv[2]` and only its *basename* is checked (`MAC_FEED_NAME`), so `../../../etc/latest-mac.yml` passes the guard.

It is a CI-only script and the argv is supplied by the workflow, so the practical risk is low — but the standing rule here is that a scanner finding is fixed, not argued down or suppressed. The fix is small and is a genuine improvement rather than a sop to the scanner: resolve the path and require it to sit inside the package, so the guard covers the whole path and not just its last segment.

```js
const distRoot = deps.distRoot ?? path.join(__dirname, '..', 'dist');
const resolved = path.resolve(distRoot, feedPath);
if (path.dirname(resolved) !== path.resolve(distRoot)) {
  throw new Error(`${feedPath} resolves outside ${distRoot}.`);
}
```

Shape, not a patch — check it against your harness before taking it. The existing tests
call `main` with a relative `dist/latest-mac.yml`, so anything built on bare
`path.resolve(feedPath)` becomes cwd-dependent and the suite's result starts depending on
where it was invoked from; injecting the root the way you already inject `readFileSync`
keeps that out.

Add the escaping-path case to the test file alongside the existing name guard, and re-run the mutation arm on it.

## What I checked independently

- **The mechanism is real, not inferred from docs.** In the installed `electron-updater` 6.8.9: `checkIfUpdateSupported` at `out/AppUpdater.js:364-378` reads `updateInfo.minimumSystemVersion` and returns `false` on `semver.lt(os.release(), …)`; `isUpdateAvailable` consults it at line 348 and `:404` short-circuits on it, so the release is dropped before anything downloads. Both line references in your description hold.
- **The single-file target is right.** I went looking for an unstamped `beta-mac.yml`, since the app ships only `-beta.N` builds and the threat model says installs follow the beta feed. `RELEASE.md:52-54` already records that `app-builder-lib` short-circuits `generateUpdatesFilesForAllChannels` for the `github` provider and the beta path falls back — there is only ever one feed, so stamping `latest-mac.yml` covers the beta clients too. Worth a clause in the script comment; the next reader will ask the same question.
- **Job wiring resolves.** `macos.outputs.min_system_version` is declared, and `publish` lists `macos` in `needs`, so `needs.macos.outputs…` is populated rather than empty. The guard fails closed on both arms — empty `EXPECTED` exits 1, and a missing asset makes `gh release download` non-zero under `set -euo pipefail`.
- **No merge-order dependency on #3279.** `build.mac.minimumSystemVersion: "13.0"` is already on `dev`, and this branch contains the `dev` tip, so `macFloorFrom` will not throw whichever order these land in.
- The Tahoe row earning the table over `major + 9` is the right call, and the unmapped-major arm is the part that will actually pay off at the next Electron bump.

## Not blocking, worth a line

The `workflow_dispatch` path has no equivalent of the publish-job guard. That is defensible — `release` tags as `desktop-manual-v…`, which is not valid semver, so the GitHub provider drops it and the updater never resolves it. Say so in the step comment, so nobody later "fixes" the tag prefix and quietly removes the reason the missing guard was safe.

Codex review did not run on this PR (usage limits), so this is the only review pass.

Re-request when Aikido is green and the pending checks land — Desktop E2E on both runners, CodeQL javascript-typescript, `core / Build shared packages`, and grype were all still in flight when I read them.

aupyay:
Aikido blocker addressed in 33c21ecc1.

**The fix.** `feedInsideDist(packageRoot, feedPath)` resolves the argv path against the package root and requires `path.dirname(resolved)` to equal `<package>/dist`, so the whole path is guarded rather than its last segment. `readFile`/`writeFile` now take the *resolved* path, so what is stamped is what the guard admitted.

I resolved against the package root rather than `path.resolve(distRoot, feedPath)` as sketched: the release step passes `dist/latest-mac.yml`, which against `distRoot` would double-nest to `dist/dist/latest-mac.yml`. Package root keeps the workflow's argument meaning the same file, and keeps `process.cwd()` out of it the way the injected-dep approach intended. The root is injected as `deps.packageRoot`, defaulting to `path.join(__dirname, '..')`.

**Evidence, all from `apps/desktop` on 33c21ecc1:**

- `pnpm test` — `73 passed, 73 total` / `1132 passed, 1132 total`, rc=0. Statements 98.37%, branches 90.93%, functions 97.68%, all above the jest thresholds. (A scoped `pnpm test -- tests/stamp-update-feed-min-os.test.ts` returns rc=1 on a green suite here — the 95% global coverage threshold misses on one file — so the full run is the verdict.)
- `pnpm run lint` and `pnpm run type-check` (`tsc --noEmit` + the test project) both rc=0.
- **Mutation arm.** Neutering the containment condition to `if (false && …)`: `4 failed, 17 passed` — the three `feedInsideDist` refusal cases plus the `main` escape case. Restored from a snapshot copy, `grep -c "false &&"` = 0.
- **Real-file arm, production defaults, no injected deps.** Wrote a feed to `apps/desktop/dist/latest-mac.yml` and ran the script exactly as the workflow does (`node scripts/stamp-update-feed-min-os.js dist/latest-mac.yml` from `apps/desktop`): rc=0 and the file's first line is `minimumSystemVersion: 22.0.0`. `../../latest-mac.yml` and `/tmp/latest-mac.yml` both rc=1 with "which is outside …/apps/desktop/dist" and no write. The scratch file is removed; `git status --porcelain` is empty.

Tests added: 4 `feedInsideDist` cases (package-relative and absolute-inside resolve; climbing out, absolute-elsewhere, and a `dist` subdirectory all refused) and one `main` case asserting rc=1 and zero writes for an escaping path. Suite went 16 → 21 cases.

**Both non-blocking notes taken.**

- The single-feed reason is now a comment beside `MAC_FEED_NAME`, citing the `app-builder-lib` short-circuit and the beta fallback recorded in RELEASE.md.
- The `workflow_dispatch` release step carries a comment saying the missing guard is safe *because* `desktop-manual-v…` is not resolvable semver, and that changing the prefix means adding the guard.

Re-requesting review. The pending checks you saw — Desktop E2E on both runners, CodeQL, `core / Build shared packages`, grype — re-run on this push.

ankit-yc (changes requested):
Re-reviewed at `33c21ecc1`. The fix landed and you were right about my sketch — resolving against `dist` would have made `dist/latest-mac.yml` mean `dist/dist/latest-mac.yml`. Resolving against the package root and asserting `path.dirname(resolved) === distRoot` is the correct form.

**The two findings I blocked on are gone.** Aikido no longer flags the argv-derived feed read at all. Confirmed by commit id on the review comments: the `:124`/`:125` pair is pinned to `6f05acb8d`, and nothing on `33c21ecc1` points at the feed path.

## Still red, but on different lines, and I think the scanner has a point

Three new medium findings, all on `33c21ecc1`:

- **`:136`** — `path.resolve(deps.packageRoot ?? path.join(__dirname, '..'))`
- **`:137`** — `deps.packageJsonPath ?? path.join(packageRoot, 'package.json')`
- **`:153`** — the `readFile(packageJsonPath, 'utf8')` those two feed

It has stopped following argv and started following `deps`. That is a weaker claim on its face — `deps` is a parameter with a hardcoded default, reachable only from inside the process — but it lands on something real: **`packageRoot` is the containment root, and you made it overridable in the same change that introduced it.** A guard whose boundary any caller can relocate does not deny what the guard exists to deny. `feedInsideDist(packageRoot, …)` is only as strong as `packageRoot` is fixed.

So the fix I would take is to remove the seam rather than annotate it: derive `packageRoot` from `__dirname` alone, no `deps.packageRoot`, no `deps.packageJsonPath`. That kills all three findings at the source and makes the boundary a property of the file instead of a default.

Your tests should not need either seam. They already inject `readFileSync`/`writeFileSync`, so nothing touches a real filesystem — they need the resolved path to be *predictable*, not *relocatable*, and `path.join(__dirname, '..', 'dist', 'latest-mac.yml')` is as predictable from the test as from the script. If it turns out one of the sixteen genuinely cannot be expressed that way, say which and I will look at it as a real constraint rather than a preference — but check before you take my word for it, given the last sketch I handed you was wrong.

Worth knowing while you are in here: #3285 tripped the identical Aikido rule on a new build script within the hour, one of them high severity. Same shape both times — resolve a path from data, read it. Whatever wording you settle on, it is worth the two of you landing the same one.

Nothing else has changed in my read of this. The mechanism, the single-feed premise, the job wiring and the migration-order point all still stand from the last pass, and the beta-feed note you added to `MAC_FEED_NAME` is exactly the clause I meant.

aupyay:
Seam dropped in `5d96443de`. Your argument holds — I made the boundary injectable in the same change that introduced it, and a boundary a caller can relocate denies nothing.

**What changed.** `appRoot` and `distRoot` are module constants off `__dirname`, matching `scripts/generate-notices.js` next door. `feedInsideDist` now takes only the feed path.

For `:137`/`:153` I went one step past dropping the seam: the manifest is `require('../package.json')` at module load, so there is no second `readFile` and the floor never arrives as a path at all. That removes the sink the other two findings fed rather than leaving it untainted.

**One seam kept, and it is not a path.** `deps.pkg` supplies the *parsed manifest object*. Nothing about it reaches a file sink, and it is not the containment root — it exists so `main`'s expected `22.0.0` stays pinned when `build.mac.minimumSystemVersion` is next raised, rather than silently tracking the real file. If you read that as the same mistake in a different coat, say so and I will drop it and let the main tests read the real floor; it costs one coupling either way.

**All sixteen — now twenty-two — expressed fine without the path seams,** so no constraint to report. The test derives the same boundary the same way you suggested, from its own `__dirname`.

**Evidence, `apps/desktop` at `5d96443de`:**

- `pnpm test` — `73 passed, 73 total` / `1133 passed, 1133 total`, rc=0, re-run after the pre-commit hooks reformatted. `pnpm run lint` rc=0, `pnpm run type-check` rc=0.
- **Mutation arm**, guard neutered to `if (false && …)`: `4 failed, 18 passed` of 22. Restored from snapshot, `grep -c "false &&"` = 0.
- **Seam-gone arm.** Called `main` with `packageRoot: '/etc'` and `packageJsonPath: '/etc/package.json'` still in the deps and an escaping feed path: rc=1, zero writes, and the error names the real `apps/desktop/dist` as the boundary. The old keys are inert.
- **Production arm, no injection.** The workflow's exact invocation from `apps/desktop` stamps `minimumSystemVersion: 22.0.0`, rc=0; `../../latest-mac.yml` rc=1. Scratch file removed, tree clean.

One test added (`main` returns 1 when the manifest carries no floor), exercising the data seam.

**On #3285** — same rule, and the shape that cleared it here is: resolve against a root fixed in the file, require `path.dirname(resolved)` to equal that root, then read and write the *resolved* value. The part Aikido cared about on the second pass was that the root itself must not be a parameter. Happy to converge on this wording if that PR's fix is still open.

Re-requesting.

ankit-yc (approved):
Approving at `5d96443de`. Aikido is green — scan `197161450`, and zero findings carry this commit id, so the three are cleared rather than moved again.

Going one past the seam and making the manifest a `require` is the better fix than the one I asked for. It removes the sink instead of untainting it, and `deps.pkg` is the right thing to keep: an object cannot relocate a boundary. The tests derive `DIST_FEED` from their own `__dirname` rather than being handed a root, so the boundary is asserted independently instead of being restated — that is the part that would have rotted if you had kept the seam and pinned it.

Merge still waits on the required checks that were in flight when I approved: Desktop E2E on both runners, CodeQL `actions` and `javascript-typescript`, grype, and `core / Detect affected workspaces`. Approval is not a statement about those — branch protection holds them, and so do I.

One loose end that is not this PR's: the containment shape you wrote up is on **this** PR, and #3285 is where it is needed. I have pointed that PR's author at your comment.

Issue #3144, opened 2026-09-12
fix(repo): prevent stale clinical draft saves from overwriting newer records


aupyay:
## Summary

Two staff sessions cannot silently overwrite each other or save a draft over a concurrently finalized artifact.

Workflow index: #3178

## Problem / motivation

**Priority: P1.** The current and intended behavior are described below. Related ownership: #1864, #3072

## Proposed solution

ClinicalArtifactService already rejects ordinary edits of finalized artifacts and exposes finalize/reopen/amend actions. The remaining shared-write contract needs an expected revision: current shared artifact update and dependent SOAP/vitals/discharge writes select by id after reading state. Preserve existing lifecycle guards and the prescription revision owner #1864.

Chosen v1: add ClinicalArtifact.version integer default1 through an additive migration. Every mutable artifact write and lifecycle action checks id+organisation+expectedVersion+allowedStatus atomically inside the same transaction as subtype/item changes, then increments version exactly once. Reads include version; use FHIR meta.versionId/If-Match where this surface is FHIR, translating centrally rather than creating parallel JSON endpoints. Existing clients must be migrated in the same rollout: additive read release, client version transport, enforcement release. After enforcement a missing precondition returns428, stale returns409 or the repository's FHIR-equivalent412 consistently through one adapter. Do not accept unconditional updates as a permanent compatibility path.

UI: workspace autosave/manual Save carry last server version. Conflict shows 'This record changed', author/time where authorised, My draft and Latest saved panels, and actions Reload saved or Copy my draft. Never discard the local draft automatically. For unsigned records, explicit Review and retry starts from the fresh version; no automatic field merge. If now signed, disable retry and use the existing permitted amendment workflow. Independent records can still save concurrently; no organisation-wide lock or presence server.

Enumerate create/update/status and importer/worker callers for SOAP, vitals, prescription and discharge and route through the shared precondition. Prescription item/dispense side effects remain inside their existing transaction; this issue changes concurrency only and must not implement the supersession design from #1864. API retries after unknown result first read the artifact version/content; do not issue a new action blindly.

Delivery: 1) version field/read mapper and compatibility inventory; 2) atomic shared writer and all lifecycle/subtype callers; 3) typed web/mobile callers and conflict UI; 4) enforce precondition after client rollout and race tests. Backfill existing rows version1, no record-content rewrite.

[...]

PR #3284, opened 2026-09-18
fix(backend): claim a clinical artifact at the version it was read at

aupyay:
Refs #3144. First slice: the shared write precondition. Client transport (`If-Match`) and the workspace conflict UI are **not** in this PR — see Deferred.

## What was wrong

Every mutable `ClinicalArtifact` write selected the row by id after reading its state:

```ts
txPrisma.clinicalArtifact.update({ where: { id: artifact.id }, data: { status, summary } })
```

Two staff sessions editing one record both succeeded and the later write won whatever it had been shown. The same window let a draft save land on top of a record a colleague had finalised, signed or voided in between, because those transitions were equally invisible to a writer that only knew the id.

## What this does

- **`ClinicalArtifact.version`**, additive, `NOT NULL DEFAULT 1`. The `ALTER` backfills every existing row to 1 and rewrites no record content; rollback is a plain `DROP COLUMN`.
- **The shared writer claims a generation.** `where` now carries `id + organisationId + version`, and `data` increments the version in the same statement, so the read and the write are one atomic claim and the loser matches no row. Prisma's `P2025` from that filtered update maps to **409**; any other Prisma error is rethrown unchanged rather than disguised as a conflict.
- **`expectedVersion` is optional** on the four update inputs. Omitting it still claims the version the service itself read a moment earlier, so this is not an unconditional compatibility path — what a caller buys by sending one is rejection of *its own* stale draft, which the service cannot detect on its behalf.
- **A loaded artifact with no version is refused, not written.** Prisma *drops* a filter whose value is `undefined` rather than matching no row, so a future `select` that forgot the column would otherwise write unconditionally, silently, and only on that one caller's path.
- **Every other artifact-row writer advances the same counter** so a draft read before it loses its claim: prescription void/cancel (now routed through the shared writer), passport attest / revoke / send-for-signature, the Documenso completion claim, and rendered-document signing.
- **The four editable kinds publish the generation as FHIR `meta.versionId`.** Kinds with no update path publish none — `versionId` exists so a client can name the generation it is writing over.

### Caller inventory

Every `clinicalArtifact` write outside `create`, and what it now does:

| Site | Change |
|---|---|
| `clinical-artifact.service` shared writer (SOAP / prescription / discharge / vitals update + finalize + reopen) | claims `id+org+version`, increments |
| `clinical-artifact.service` `deletePrescription`, `cancelPrescription` | routed through the shared writer |
| `pet-clinical-records.service` `attestRecord`, `revokeRecord`, `sendForSignature` | increments |
| `documenso.controller` completion claim (`updateMany`) | increments, inside its existing precondition |
| `rendered-document.service` signing | increments |
| `documenso.controller` attestation-only update (`signedPdfUrl`) | unchanged — it writes no field of the artifact row itself |

## Evidence

Run at `fdb72df5c` (`git rev-parse HEAD` read in the same shell before and after).

| Command | Result |
|---|---|
| `pnpm --filter @yosemite-crew/database run prisma:generate` | exit 0 |
| `pnpm run type-check` (root; turbo `dependsOn: ["^build"]`) | 17/17 tasks, exit 0 |
| `pnpm --filter backend exec tsc --noEmit` | exit 0, no output |
| `pnpm --filter backend run test` | **509 suites, 12131 tests, all passing** |
| `pnpm --filter backend exec jest --listTests test/services/clinical-artifact-concurrency.service.test.ts` | 1 test file discovered (not zero) |
| `pnpm --filter backend run lint` | 0 errors; 24 pre-existing warnings, **the same 4 on this file before and after** (measured by linting the stashed tree) |
| `pnpm exec prettier --check` on every changed file | clean |

Coverage: every line this PR changes in `clinical-artifact.service.ts` is executed by the suites above — checked by parsing `coverage-final.json` for uncovered statements intersecting the diff's `+` ranges. Result: **none**.

### Mutation checks

Every new assertion was driven by breaking the behaviour it protects, watching it fail, and restoring the file to a byte-identical snapshot (`shasum -a 256` compared each time).

| Mutation | Suite result |
|---|---|
| drop `version` from the claim's `WHERE` | 14 failed |
| drop `version: { increment: 1 }` from the claim's `data` | 11 failed |
| drop `organisationId` from the claim's `WHERE` | 13 failed |
| ignore the caller's `expectedVersion` | 1 failed |
| conflict returns 500 instead of 409 | 1 failed |
| treat every Prisma error as a conflict | 1 failed |
| remove the missing-version guard | 1 failed |
| move the artifact claim *after* the subtype write | 1 failed |
| drop the increment from each of attest / revoke / send-for-signature / Documenso claim / rendered-document signing | 1 failed each |
| drop the increment from `deletePrescription` / `cancelPrescription` | 1 failed each |
| mapper reads `templateVersion` instead of `version` | 2 failed |
| mapper invents `versionId: "1"` when no version was read | 2 failed |

The fixtures keep stored version (`5`), caller version (`2`), mapper version (`7`) and `templateVersion` (`2`) deliberately distinct, so a mapper or writer reading the wrong one cannot coincide with the right one.

## Limitation, stated plainly

`apps/backend` mocks Prisma, so these tests assert **the WHERE clause the service builds**, not a real Postgres race. That is the only instrument available in this suite, and it is the load-bearing one: nothing else here can tell a conditional write from an unconditional one. A database-backed two-transaction race test is listed in the issue and belongs with the enforcement release, where there is a precondition to enforce end to end.

## Deferred to the next slices of #3144

- `If-Match` / client version transport on the FHIR surface, and the 428-on-missing-precondition enforcement release.
- The workspace conflict UI (My draft / Latest saved, Reload saved / Copy my draft).
- The database-backed concurrency test described above.

Nothing here is exposed to a user as finished: the version is published on reads and enforced server-side; no UI claims a conflict affordance that does not exist yet.

---

## Follow-up from review (`fedd78b3d`, both non-blocking, both deliberate)

**1. A lost claim in `requestRecordSignature` orphans a Documenso document.** The document is created before the artifact write and distributed after it, so the new 409 path exits between the two: a Documenso draft exists that nothing references. It is *not* distributed, so no vet is mailed — the pre-existing comment on that write ordering explains why distribution is deliberately last. The cost is one undistributed draft per lost claim, and because no `documensoDocumentId` was stored, the existing in-flight idempotency check cannot match it, so a retry mints a second document rather than reusing it. Reclaiming or reusing the orphan needs its own contract (the webhook matches on the stored id, which by definition was never written) and is not part of this slice.

**2. `claimPassportArtifact` does not carry `organisationId` in its `WHERE`, where its `clinical-artifact.service` sibling does.** The asymmetry is real. It is safe here because both reads that feed it are already organisation-scoped — `loadPassportArtifactForPatient` filters on `organisationId` *and* `kind`, and `requestRecordSignature`'s own `findFirst` does the same — whereas the sibling writer is reached by callers that assert scope separately. The sibling's `organisationId` is defence in depth for a *future* caller, and the same argument applies here; it is left for a follow-up rather than pushed now, because `dismiss_stale_reviews_on_push` is on for `dev` and this would trade a re-review and a full gate cycle for a second lock on a door that is already locked upstream.


ankit-yc (changes requested):
Reviewed at `fdb72df5c037`. The core of this is right and the caller inventory is accurate — I rebuilt it independently and got your thirteen sites exactly, with nothing extra. One finding I do want addressed before this goes in, and it is a small fix using a pattern already in this diff.

## `attestRecord` and `revokeRecord` keep the precondition in JS, and a comment now claims otherwise

`pet-clinical-records.service.ts:594-620`. `attestRecord` loads with `select: { id: true, status: true }`, asserts not-revoked and not-already-SIGNED in TypeScript, then writes `where: { id: artifactId }`. `revokeRecord` at `:662` has the same shape. The version increment you added moves the counter, but nothing in either WHERE stops the write.

So a revoke committing between that read and that write is flipped straight back to SIGNED. That is the #3144 defect itself, on the passport surface, and it is load-bearing here rather than cosmetic — the comment you left at `:606-607` now reads:

> `revokedAt`/`revokedReason` are deliberately not cleared here: VOID is terminal (guarded above), so an attestation can never wipe a revocation.

An unconditional WHERE is what makes "guarded above" untrue under concurrency, and the record ends up SIGNED while its attestation row still carries `revokedAt` — a state no reader expects. A comment asserting a safety property the statement does not enforce is worse than no comment, because the next person reads it instead of the WHERE.

The fix is the one you already wrote 200 lines away. `documenso.controller.ts:280-296` re-asserts its preconditions inside an `updateMany` and branches on `count === 0`; the same move here (status into the WHERE, or claim `version` the way the shared writer does) closes it. Either that, or move these two out of the "increments" row of the caller inventory and into **Deferred** with the race named — what I cannot sign off on is the table reading as though every site is now covered when two of them still decide in JS.

## What I verified independently, rather than reading it off the description

- **The caller inventory is complete.** `git grep` for every `clinicalArtifact.(update|updateMany|upsert|delete|deleteMany|create)` at your head sha returns thirteen sites, and each maps onto a row of your table. No writer is missing from it.
- **No raw-SQL bypass**, scoped precisely: no `$executeRaw`/`$queryRaw` call site names `ClinicalArtifact` on the same line. That grep is line-oriented, so a multi-line template literal would evade it — I did not rule that out.
- **The `Number.isInteger` guard is the best thing in this diff.** Prisma really does drop an `undefined` filter rather than matching nothing, so a `select` that forgot `version` would have silently downgraded that one caller's path to an unconditional write — invisible in review and invisible in the tests. Refusing with a 500 instead of widening is the right call.
- **`organisationId` in the WHERE** is worth the belt-and-braces for the reason you give.
- **The migration is safe to land ahead of the code.** Additive, `NOT NULL DEFAULT 1`, no content rewritten, and a plain `DROP COLUMN` to reverse.
- **Prisma 6.19.3**, so filtering a non-unique field in an `update` WHERE is GA, not a preview flag.

## Not blocking

Your note about the suite mocking Prisma, so these tests assert the WHERE the service builds rather than a real race, is the right disclosure and it is in the right place. Deferring the database-backed two-transaction test to the enforcement slice is reasonable — it has nothing to enforce end to end until `If-Match` lands.

CI was entirely pending when I read it (13 pending, 6 pass, 0 fail), so nothing here is a statement about the gates. Re-request once they land and the attest/revoke point is either fixed or moved to Deferred.

aupyay:
Fixed in `fedd78b3d`, not deferred. The finding is exactly right and the first commit made it worse in one specific way: incrementing the version on those three writes made them *visible* to everyone else without making them *conditional* themselves.

`attestRecord`, `revokeRecord` and `requestRecordSignature` now go through one claim keyed on the generation they read. A lost claim is a 409 that writes nothing — and specifically no audit row and no owner push, since a signature that never landed must not be logged or notified as one. `requestRecordSignature` mattered most of the three: its read-to-write window spans a Documenso round trip.

Both `select` projections that feed those reads gained `version`.

The claim refuses a record loaded without a version rather than widening. `loadPassportArtifactForPatient` erases its projection's type, so this is not compile-time reachable, and prisma *drops* an `undefined` filter rather than matching no row.

### Mutation checks, `fedd78b3d`

| Mutation | Result |
|---|---|
| drop `version` from the claim's `WHERE` | 3 failed |
| drop `version: { increment: 1 }` from the claim | 3 failed |
| conflict returns 500 instead of 409 | 3 failed |
| treat every Prisma error as a conflict | 1 failed |
| remove the missing-version guard | 1 failed |
| **delete `version: true` from the attest `select`** | **stayed green** → fixed, see below |

That last one is worth naming. The prisma mock answers a `findFirst` regardless of its `select`, so no behavioural test in this suite can notice a projection that forgot the column — and forgetting it is precisely what turns the claim back into an unconditional write. The projection is now asserted directly, and with that assertion the mutation fails.

### Not changed, and why

`rendered-document.service` signing still increments without claiming. It is not the same shape: it never reads the artifact's status to decide anything, it propagates a completion whose precondition is the document's own signing gate. The Documenso webhook claim was already conditional by construction — `status: "IN_PROGRESS"` plus the attestation preconditions are in its `WHERE`, with a `count === 0` check.

### Re-verified at `fedd78b3d`

`pnpm --filter backend exec tsc --noEmit` exit 0 · `pnpm --filter backend run test` **509 suites, 12137 tests, all passing** · lint 0 errors and the same 24 pre-existing warnings, none on either changed file · prettier clean.

Caller inventory unchanged from the PR body; thanks for rebuilding it independently.

ankit-yc (approved):
Re-reviewed at `fedd78b3d`. The blocker is properly closed: all three passport transitions now claim the generation they read, the `Number.isInteger` refusal mirrors the shared writer rather than inventing a second convention, and the 409 tests assert that the loser writes no audit row and sends no owner push — which is the part that would have rotted quietly if you had only asserted the status code.

Two things I want to record, neither blocking.

## 1. Making the write conditional gave `requestRecordSignature` a new failure mode, and it is mine

The Documenso document is minted before the claim, and `distributeDocument` runs after it. When the claim was unconditional that ordering was airtight — the comment above it reasons carefully about exactly this, and it was right at the time. Now a losing claim throws 409 between the two, so the document exists in Documenso, is referenced by nothing, and is never distributed.

The orphan is inert rather than dangerous: nothing was mailed, so no `DOCUMENT_COMPLETED` can arrive for it, and your in-flight idempotency check keys on the stored `documensoDocumentId` which was never written — so a retry mints a fresh document and proceeds normally. The cost is a stray draft in the practice's Documenso account per lost race, and lost races here need a concurrent attest or revoke inside a PDF round trip.

I am not asking you to fix it. I am asking you to **name it in the PR body**, because I requested the change that created it and the next person to read that carefully-reasoned ordering comment deserves to know its premise moved. If you would rather close it, the cheap version is catching the 409 and best-effort deleting the document before rethrowing — but that adds a failure path to a failure path, and I would not take that trade without evidence the orphans matter.

## 2. `claimPassportArtifact` omits `organisationId`; its sibling includes it

The shared writer puts `organisationId` in the WHERE and your comment there calls it the second lock on the same door — no future caller can reach a cross-tenant row by forgetting the assertion. `claimPassportArtifact` takes only `artifactId` and `version`, so it relies entirely on the caller having gone through `loadPassportArtifactForPatient` or the scoped `findFirst`.

All three current callers do. The point of the second lock is the caller that does not exist yet, and two writers a few hundred lines apart disagreeing about whether tenancy belongs in the WHERE is the kind of inconsistency that gets resolved in the wrong direction later. Threading it through costs one parameter.

## Also worth saying

The P2034 test is the right instinct and I would not have asked for it — catching P2025 and letting a serialization failure through as itself, rather than flattening every Prisma error into a 409, is what keeps the 409 meaningful.

And the mutation that survived your first pass is the more valuable half of this round: a mock that answers `findFirst` regardless of its `select` cannot see a projection change, so the arm that dropped `version` from the select stayed green against a real defect. Asserting the projection is the fix, and finding it by driving the mutation rather than by reading the test is exactly the thing that usually does not happen.

Aikido is green — scan `197166364`, no findings on this commit. Approving.

PR #3285, opened 2026-09-18
fix(frontend): classify bundle chunks from build artefacts, not filenames

### LINKED ISSUE #3266: chore(frontend): Next 16 defaults to Turbopack and the bundle-budget gate cannot see its chunks

aupyay:
Step 2 of #3266 only: the instrument fix, which is green on `dev` on its own merits and independent of the Next 16 adoption decision. **This does not close #3266** — the decision it exists for (adopt Turbopack? accept ~3% more first-load JS to take Next 16 at all?) is untouched here, and so are steps 3 and 5.

## What was broken

`check-bundle-budgets.mjs` classified chunks by filename: `framework-`, `main-`, `polyfills-`, `<digits>.<hex>.js`. Those are webpack conventions rather than categories. Under Turbopack all four matched zero files and every chunk fell through to the strictest non-polyfill bucket, so three of the four ceilings guarded nothing — silently, with no line of output changing.

`export-build-report.mjs` read `.next/app-build-manifest.json` with no existence check. Next 16 removed that file, so the CI step immediately after the budget check exited 1 under **either** bundler. Fixing only the budget gate would have turned a red step 7 into a red step 8.

## What it does now

Classification comes from the build's own artefacts, all of which both bundlers emit:

| category | source |
| --- | --- |
| `polyfills` | `build-manifest.json` → `polyfillFiles` |
| `shared` | `build-manifest.json` → `rootMainFiles` (the entry chunks) |
| `page` | referenced by a `<script src>` in a prerendered document, i.e. on some route's first load |
| `async` | reachable from neither — a lazy payload fetched after hydration |

There is no default bucket any more. A missing `build-manifest.json`, a manifest key that is not an array, an empty `rootMainFiles`, no prerendered documents, documents that reference no scripts, or a reference that does not resolve on disk all **throw**. A gate that cannot see its input fails rather than guessing. The script also prints a per-category census on a pass, because the old failure was invisible in the log.

`export-build-report.mjs` prefers `app-build-manifest.json`, falls back to the prerendered documents, records which source it used in both outputs, and throws when neither is present. It is deliberately **not** repointed at `build-manifest.json`: those keys are pages-router only, so the filter yields nothing and the script would write an empty report and exit 0 — a loud failure replaced by a silent one.

## Budgets

Re-deriving them was forced rather than optional. Moving `shared` from `framework-`/`main-` to `rootMainFiles` changes which chunks each ceiling covers, so carrying the old numbers over would have been arithmetic on the wrong population.

Measured with `stat()` on a local production build of `dev` at `89eedeaca` (next 15.5.24, webpack): 297 chunks over 183 prerendered documents, partitioned 1 / 4 / 48 / 244.

| category | measured max | old ceiling | new ceiling |
| --- | ---: | ---: | ---: |
| `page` | 275.7 KiB (282,314 B) | 375 | **290** |
| `async` | 1156.4 KiB (1,184,191 B) | 1190 | 1190 |
| `shared` | 169.7 KiB (173,808 B) | 225 | **178** |
| `polyfills` | 110.0 KiB (112,594 B) | 120 | **116** |

Every ceiling moves down except `async`, where +5% on the measured maximum would be a *raise* (1214 KiB) and this file's own rule only lowers.

**One chunk's ceiling loosens, and it is called out in the source rather than buried:** the 213.9 KiB framework chunk is not listed in `rootMainFiles` and is eager on 1 of the 183 prerendered documents, so it is `page` (290 KiB) now rather than `shared` (previously 225 KiB). Every other chunk is guarded at the same ceiling or a tighter one. A test pins that classification so changing it has to be deliberate.

## Evidence

Every category case runs twice — once with webpack-shaped chunk names, once with Turbopack-shaped ones. The names differ; the expected classification does not. That pair is the regression test for this issue.

- `apps/frontend` full suite: **1073 suites / 13670 tests passed, 0 skipped, rc=0** (`TZ=UTC pnpm run test`). Without `TZ=UTC` two pre-existing `AddInventory/FormSection` date assertions fail on a UTC+2 host; they pass under UTC, they are untouched by this change, and CI runs in UTC.
- `pnpm run lint` rc=0 (1 pre-existing warning in `useLazyAuthStore.ts`), `pnpm run type-check` rc=0, prettier clean.
- Both scripts run against the real build: the gate passes 297 assets and prints the census above; the report writes 77 routes from `app-build-manifest`, and 181 routes from `prerendered-documents` when that manifest is removed to simulate the Next 16 shape.
- 10 mutation arms, each restored from a file snapshot rather than from HEAD, plus a no-op baseline whose expected output is non-empty. **All 10 killed**, baseline 43/43 green: dropping the percent-decode, reinstating the `page` fallthrough, removing each of the four source checks, raising the page budget, removing the report fallback, removing its decode, and removing its empty-pages guard.

The dependabot hold on `next` stays — the comment there is updated to say the two instrument failures are fixed and that what remains is the decision itself.

Refs #3266

---

### Aikido

Both findings on `export-build-report.mjs` are fixed rather than suppressed. Every read now resolves through a containment check before it happens. Asset references keep their build-root-relative meaning (`/static/...` means the build root, not the filesystem root); paths the script produced itself are resolved as-is; both go through the same boundary test.

The boundary uses a separator rather than a bare prefix, so a sibling directory whose name merely starts with `.next` is rejected. That distinction is not free — the first mutation arm for it survived, because `startsWith(NEXT_DIR)` alone still fails on ENOENT and the test could not tell the two apart. It needed a `.nextrogue/x.js` actually written to disk, so that without the guard the script reads the file and reports a route. Two arms now, both killed: removing the check, and weakening the boundary to a prefix.

### I did run it against a real 16.3.4 Turbopack build

The 1462.5 KiB figure in the original body came from this issue's own measurement of the CI artifact, not from running this script — the reviewer was right to ask. So I built one: `dev` at `89eedeaca` with `next` moved to 16.3.4 in both pins (root `pnpm.overrides` and `apps/frontend/package.json`), `pnpm install`, `pnpm run build`. The build prints `Next.js 16.3.4 (Turbopack)`; `next/package.json` reports `16.3.4`.

**The structural premise holds.** `build-manifest.json` on 16.3.4 carries both keys the classifier needs — `polyfillFiles` (1 entry) and `rootMainFiles` (5 entries, one of them `turbopack-*`) — and the build writes 184 prerendered documents under `.next/server`. There is no `app-build-manifest.json`, which is the removal that broke the report step.

**The classifier works under Turbopack.** Unmodified, against that build:

```
Classified 297 JS assets from build-manifest.json and 184 prerendered documents:
- page: 42 chunks, largest 453.8 KiB, budget 290.0 KiB
- async: 249 chunks, largest 1462.6 KiB, budget 1190.0 KiB
- shared: 5 chunks, largest 222.7 KiB, budget 178.0 KiB
- polyfills: 1 chunks, largest 110.0 KiB, budget 116.0 KiB
Bundle budget check failed:
- .next/static/chunks/0e-hkw4_y48n_.js (page): 453.8 KiB > 290.0 KiB
- .next/static/chunks/0vnfpl91_24er.js (shared): 222.7 KiB > 178.0 KiB
- .next/static/chunks/3crz3y-mtsryq.js (async): 1462.6 KiB > 1190.0 KiB
```

1 / 5 / 42 / 249 rather than 0 / 0 / 0 / 297. All four ceilings guard chunks again, and the gate fails on *what the chunks are* rather than on what they are called.

`export-build-report.mjs` exits 0 on the same build — 182 routes via the `prerendered-documents` fallback — which is the second half of #3266 shown fixed on the shape that broke it, not on a fixture.

**It fails on three chunks, not one.** Worth stating because the issue predicts only the lazy one:

- 1462.6 KiB lazy vs the 1190 KiB `async` ceiling — the predicted failure, now measured.
- 453.8 KiB first-load vs 290 KiB `page`. **Not caused by tightening this PR does:** it also exceeded the previous 375 KiB ceiling.
- 222.7 KiB entry chunk vs 178 KiB `shared`. **This one is a consequence of the new number** — the old 225 KiB ceiling would have admitted it. It is still a real signal: dev's largest entry chunk is 169.7 KiB, so Turbopack's is +31% on the same tree, and the old ceiling would have admitted it only because it was calibrated against a chunk that is not an entry chunk at all.

That build was a throwaway probe for this measurement; nothing from it is in this branch. The `next` pin and the dependabot hold are unchanged here, and the hold's comment now carries these measured numbers instead of the arithmetic prediction.

### Re-verified at `dfd3b74cf`

`TZ=UTC pnpm run test` in `apps/frontend`: **1073 suites / 13674 tests passed, 0 skipped, rc=0**. `npx tsc --noEmit` rc=0. `pnpm run lint` rc=0 (same single pre-existing warning). Both scripts still green against the 15.5.24 webpack build, with the census unchanged.


ankit-yc (changes requested):
Reviewed at `8ea6cbb582`. The instrument change is the right one and the disclosure around the budgets is honest. One blocker from a gate, and one question I would rather ask than assume the answer to.

## Blocker: Aikido is red

Two findings on `apps/frontend/scripts/export-build-report.mjs`, and one of them is **high**, not medium:

- **line 32** (high) — `path.resolve(NEXT_DIR, normalizedPath)` in `sumChunkSizes`. `chunkPath` comes out of the manifest or a `<script src>`, and stripping a leading `/` does not stop `..`: a segment like `../../etc/passwd` resolves clean out of `.next` and gets read.
- **line 43** (medium) — the `readFile` in `readJsonIfPresent`.

Rule here is that a scanner finding is fixed rather than suppressed, and in this case the fix is also the honest one — this script already refuses to guess when a source is missing, and containment is the same instinct one level down:

```js
const filePath = path.resolve(NEXT_DIR, normalizedPath);
if (filePath !== NEXT_DIR && !filePath.startsWith(NEXT_DIR + path.sep)) {
  throw new Error(`${chunkPath} resolves outside .next`);
}
```

Worth knowing: #3283 tripped this same Aikido rule an hour ago on a new build script, same shape — resolve a path from data, read it. Two for two on new scripts today. If you two land containment helpers separately they will differ; one of you copying the other's wording is cheaper than both inventing it.

## Question rather than a finding: was the new classifier run against a real 16.3.4 build?

The measured partition is from `89eedeaca`, next 15.5.24, webpack. The test pair runs each category twice with webpack-shaped and Turbopack-shaped names, which pins that the classifier ignores filenames — the thing that broke — and that is the right regression test.

But it is a synthetic fixture, so what it cannot show is the structural premise this rests on: that Next 16's `build-manifest.json` really does carry `rootMainFiles` and `polyfillFiles`, and that it still writes prerendered documents under `.next/server/app`. If either is untrue the script throws, which is the failure mode you designed for and far better than the silent one — but the bump would then still be blocked on the instrument rather than on a budget, which is the outcome this PR exists to prevent.

Your dependabot comment predicts the bump failing on a **1462.5 KiB lazy chat chunk against the 1190 KiB async ceiling**. That number reads like it came from running this script against 16.3.4 rather than from arithmetic. If it did, put the census line from that run in the PR body — it converts the central claim from asserted to measured, and it is the one piece of evidence the synthetic pair cannot supply. If it did not, say so plainly and I will not hold the PR for it; the throw-on-missing-source design makes the downside loud.

## What I checked

- **The loosening is what you say it is and no more.** I read the four constants against the old ones: page 375→290, shared 225→178, polyfills 120→116, async unchanged with the stated reason that +5% would be a raise. The only ceiling that rises is the 213.9 KiB framework chunk moving to `page`, and I confirmed against `categorise` that it lands there because it is absent from `rootMainFiles` and eager on one document — not because the bucket is a catch-all. Three tightened, one reclassified, nothing quietly banked.

  I did initially misread the source comment as saying that chunk was eager on *all* 183 documents, which would have made `page` the wrong bucket and the loosening significant. It says 1 of 183, and the body agrees. My error, and the classification is fine.
- **No default bucket left.** Every missing or malformed source throws, and `assertResolvable` catches a reference shape change rather than letting it quietly reclassify. That is the actual fix for #3266 step 2 — the old script's failure was that it could not fail.
- **The dependabot.yml change is comment-only** and describes the state after this lands rather than before. It correctly keeps the hold and moves the stated reason from "the instruments are broken" to "the adoption decision is open", which is the honest edit.

Re-request when Aikido is green. The question above does not block on its own.

aupyay:
The containment wording for the Aikido finding on `export-build-report.mjs` already exists — it was written up on #3283, which hit the same rule on a different new build script within the hour: https://github.com/YosemiteCrew/Yosemite-Crew/pull/3283#issuecomment-5733043455

That PR is now green on Aikido, so the shape is confirmed rather than proposed. Two things there are worth taking rather than re-deriving:

1. Resolve against a **fixed module constant**, not a value any caller can pass in. The first attempt on #3283 made the containment root injectable for testability, and Aikido promptly moved its finding onto that parameter — correctly, because a boundary a caller can relocate denies nothing.
2. Where you can, remove the sink instead of guarding it. #3283's second read became a `require` of the manifest, so the value never arrives as a path at all. Your `readJsonIfPresent` at `:43` may be a candidate; `sumChunkSizes` at `:32` is not, and needs the containment check.

Your line 32 is the high-severity one and is genuinely reachable: stripping a leading `/` does not stop `..`.

ankit-yc (approved):
Approving at `dfd3b74cf`. Aikido green, and no findings carry this commit id.

**The containment fix is right, including the part that is easy to get wrong.** `insideNext` tests `resolved === NEXT_DIR || resolved.startsWith(NEXT_DIR + path.sep)` rather than a bare prefix, so a sibling directory named `.nextfoo` is rejected instead of admitted. You say that arm needed a real sibling dir on disk to kill — that is the correct reading: a bare-prefix implementation passes every test written with paths that do not have a sibling to collide with, so the mutation is only observable against the filesystem. Two of the three reads did not strictly need the check and got it anyway, which is the right default for a script whose whole thesis is not trusting what produced its input.

I checked whether `check-bundle-budgets.mjs` needed the same treatment and it does not: every path it opens is a module constant or comes out of a `readdir` walk rooted at `.next`. Manifest and `<script src>` references are only *looked up* in a set built from disk, never resolved into a read. So "containment at every read" is accurate rather than a claim that skipped a file.

**On the question I asked: you answered it the expensive way and it was worth it.** Building `dev` at `89eedeaca` with `next` at 16.3.4 in both pins and running the unmodified classifier against the result converts the premise from asserted to measured. The specific facts that matter are that 16.3.4's `build-manifest.json` still carries `polyfillFiles` and `rootMainFiles`, and that 184 prerendered documents land under `.next/server` — those are what the classifier stands on, and neither was knowable from the synthetic pair.

Two things I want to credit precisely, because both cut against you:

- The partition is **1/5/42/249, not 0/0/0/297**. That is the whole claim of the PR, now demonstrated on the bundler it was written for rather than inferred from the one it was measured on.
- The bump fails on **three** chunks, not the one your dependabot comment predicted — and you traced that number's provenance to the issue's CI-artifact measurement rather than to this script, and said so. Correcting your own published figure in the direction of "worse than I said" is the part that makes the rest of the evidence worth reading.

Budgets are untouched since `8ea6cbb` — zero lines changed in `check-bundle-budgets.mjs` — so the ceilings stay calibrated against the webpack build that derived them, and the Turbopack census is evidence rather than a re-derivation. That is the correct call: re-deriving budgets from a build you are not adopting would have quietly moved the gate.

Merge waits on `Build frontend (shared)`, CodeQL `javascript-typescript` and `core / Build shared packages and frontend`, all still running.

Note that #3266 stays open by design — this is the instrument, not the decision.

Issue #3153, opened 2026-09-12
refactor(repo): preserve exact monetary amounts through finance calculations


aupyay:
## Summary

Preserve exact monetary amounts through finance calculations.

Workflow index: #3178

## Problem / motivation

**Priority: P1.** The current and intended behavior are described below. Related ownership: #1658

## Proposed solution

Anchors: Invoice.items Json, Float totals at schema3876-3937; Payment/Refund/CreditNote Float at3990/4018/4039; Estimate/EstimateItem Float at5829/5857; `finance/pricing.ts:43-46` rounds all numbers with scale100; `invoice.service.ts:373-432` optional id normalization and content-key fallback; `estimate.service.ts:144-153` explicitly emits invoice lines without IDs. Existing `apps/backend/src/utils/stripe-minor-units.ts` must be reused at provider boundaries.

This issue owns precise money persistence/calculation; #3154 owns line identity. Separate canonical monetary precision from provider charge encoding. Use Intl.NumberFormat currency metadata for an explicitly validated supported-currency registry, pinned by fixtures for the deployed runtime; reject unknown currency codes rather than accepting the generic two-digit fallback. Do not treat provider exceptions as ledger precision. Proposed v1 convention: Prisma Decimal(20,6) for finance amounts and unit prices, decimal strings in new API contracts, and exact Decimal arithmetic. Quantity precision remains the domain-approved precision capped at six fractional digits for this finance contract. Quantize posted totals and payment/allocation amounts with ROUND_HALF_UP to the currency exponent from this validated canonical currency registry; retain unit-price precision before line rounding. Reject excess posted precision rather than silently clipping it. Provider boundaries convert to/from integer minor units using the existing provider helper only for its verified supported currency set, with decimal-safe validation. The current Stripe helper is 0/2 based with provider-specific exceptions; it is not the source for KWD3 ledger precision. A 3-decimal currency payment remains unsupported until the adapter has its own verified capability and sandbox fixtures; never send it through x100 conversion. Never apply JavaScript Float arithmetic to ledger accumulation. Migration is add/backfill/compare/switch with reject/quarantine for non-finite, out-of-range, ambiguous records, not a blind cast followed by dropping old columns.

[...]

PR #3286, opened 2026-09-18
fix(backend): post invoice amounts at the ledger currency's own precision

aupyay:
First bounded slice of #3153. Ledger precision only — no schema change, no
`Float` → `Decimal` migration, no provider-boundary work.

## The defect

`calculateInvoicePricing` rounded every amount with a hardcoded `MONEY_SCALE`
of 100, so the currency an invoice is billed in had no effect on how it was
posted. A JPY invoice carried fractional yen; a KWD one lost its third
decimal. `OrganizationBilling.currency` comes from Stripe Connect, so both are
reachable today.

## What changed

**`services/finance/currency.ts` (new)** — a ledger currency registry.

- Membership comes from `Intl.supportedValuesOf("currency")`, not from
  `Intl.NumberFormat` succeeding. A well-formed but unknown code like `ZZZ`
  does not throw — it formats with the generic two-digit fallback, which is
  the failure the registry exists to prevent. A test pins that.
- Exponents come from `resolvedOptions().maximumFractionDigits`, which is
  locale-independent. A currency the runtime reports no fraction digits for is
  left out rather than posted at a guess.
- 17 codes are refused outright: ICU reports 0 fraction digits for them while
  ISO 4217 assigns 2 (3 for IQD). Neither number can carry a ledger without
  deciding which authority applies, so they are not priced rather than priced
  wrong. Supporting one is additive — give it an explicit minor unit and drop
  it from the set. Taking the ICU number is not the fix.
- `quantizeMoney` rounds half away from zero on integers taken from the
  shortest decimal that round-trips to the amount. An amount that arrived as
  `2.675` is the number a person typed; its binary expansion
  (`2.67499999999999982236`) rounds the other way.

**`services/finance/pricing.ts`** — `InvoicePricingInput` takes an optional
`currency`; every internal rounding and the invoice-discount allocation run at
that currency's precision. Omitted, it is two decimals as before.

`roundMoney` is deliberately unchanged. The payment, tax and appointment
callers round by it, and `creditNoteService` on the frontend carries a copy
stating it matches. The two functions disagree wherever scaling the float
lands a tie on the wrong side — 8.165 posts as 8.17 through the quantizer and
8.16 through `roundMoney` — so the divergence is pinned by a test instead of
assumed away. Moving those callers over is a later slice with its own
evidence.

**`services/invoice.service.ts`** — hands the org's billing currency to the
pricing call, but only when the registry can price it. An org billing in a
refused code keeps the two decimals it is priced at today rather than losing
invoicing the day this ships, which is also the rollout order #3153 asks for.

## Deliberately not in this slice

Schema `Decimal(20,6)` and its backfill; decimal-string DTOs; posted-precision
rejection on payments and allocations; the `roundMoney` callers in
`finance/payment.ts` and `finance/tax.ts`; provider minor-unit conversion
(`utils/stripe-minor-units` is untouched and is still not a ledger source);
the frontend preview. #3153 stays open.

## Evidence

Run from the repository root at `4dc4dc91d`, working tree clean before and
after each.

| Command | Result |
|---|---|
| `pnpm run type-check` | exit 0, 17/17 tasks |
| `pnpm --filter backend run lint` | exit 0, no warning in a changed file |
| `pnpm --filter backend run test` | exit 0 — 509/509 suites, 12180/12180 tests |
| `pnpm --filter backend exec jest --listTests test/services/finance.currency.test.ts test/services/finance.pricing.test.ts` | both discovered |

An earlier full run showed two unrelated suites failing on a refused local
Redis/Prisma connection under worker concurrency; they pass in isolation and
the clean run above has no failures.

### Mutation runs

Every new test was checked by breaking what it protects, observing the
failure, and restoring. The tree returned to `4dc4dc91d` after each.

| Mutation | Result |
|---|---|
| baseline | 36 passed |
| one refused code misspelled so it excludes nothing | 1 failed |
| unknown code returns the two-decimal default instead of throwing | 3 failed |
| tie rounds down instead of away from zero | 6 failed |
| read the binary expansion instead of the shortest decimal | 4 failed |
| negative amount loses its sign | 2 failed |
| pricing rounds by `roundMoney`, ignoring the currency | 2 failed |
| discount allocation returns to hundredths | 1 failed |
| invoice path stops passing the currency | 1 failed |
| invoice path passes a refused currency through unguarded | 1 failed |

### Numerical fixtures pinned

3 × 0.10 = 0.30 · JPY posts whole units (3 × 1200.5 → 3602, not 3601.5) · KWD
keeps three decimals (3 × 1.2345 → 3.704, not 3.70) · a 1000 JPY invoice
discount over three equal lines leaves a one-yen remainder, never a hundredth
· USD, GBP, EUR, INR at 2; JPY, KRW, ISK at 0; KWD, BHD, OMR at 3.

No runtime QA against a browser: this slice has no rendered surface.

Refs #3153

---

## Review finding: the checkout comparison and a rounding tie

`buildCheckoutSessionLineItems` decides whether to charge itemised lines or a
single balance line by comparing the invoice's line snapshots, summed through
`roundMoney`, against the pre-tax invoice total, which this branch now
quantizes. On a tie the two roundings disagree — 8.165 is 8.16 through
`roundMoney` and 8.17 through the quantizer — and the session takes the
balance-line branch.

**Left as it is, deliberately, and pinned by a test.** The balance line is the
branch that charges what is owed. Itemising submits a per-UNIT amount that
Stripe multiplies by the quantity, and a per-unit amount cannot reconstruct a
rounded line total: this invoice itemises at 816 against the 817 it was posted
at. Quantizing the comparison would route the tie back to the branch that
under-charges it.

The shortfall is not confined to ties and is older than this branch. Two 8.165
units sum to 16.33 under both roundings, so the comparison finds nothing wrong,
the session is itemised at 2 × 816, and a 1633 invoice settles as paid having
collected 1632. Filed as #3305 with the fix shape; both rows are held as
characterisation tests here so that issue has to change a test to close.

Automatic tax is unaffected: `disableAutomaticTax` still requires the invoice
to carry tax, so a zero-tax invoice keeps Stripe calculating it on the balance
line exactly as it did on the itemised lines.

### Mutation runs on the two new tests

| Mutation | Result |
|---|---|
| comparison given a one-cent tolerance (routes the tie to itemised) | 1 failed — "Consult" for "Outstanding balance for invoice inv_tie" |
| balance line charges the `roundMoney` reading instead of the balance | 1 failed — 816 for 817 |
| per-unit amount shifted half a cent up | 1 failed — 817 for 816 |

Backend after the addition: 509 suites, 12182 tests, 0 failures;
`type-check` exit 0; `lint` 0 errors, 24 pre-existing warnings.


ankit-yc (dismissed):
Reviewed at 4dc4dc91d. The registry's reasoning holds — I verified its claims independently rather than reading the comments, on node v22.23.2 / ICU 78.3:

- `ZZZ` does not throw and silently reports `maximumFractionDigits: 2`. That is the failure the registry exists to prevent, and it is real.
- Every code in `AMBIGUOUS_LEDGER_CURRENCIES` reports ICU 0 against ISO's 2 (3 for IQD). 17 entries, matching the size the test pins.
- The set is the ICU/ISO *disagreements* and not simply every ICU-0 currency: VND, ISK and XAF are genuinely ISO-0 and correctly stay supported. That distinction is the part that would have been easy to get wrong.
- `min === max` for all 14 codes I sampled, so reading `maximumFractionDigits` is safe here.

`resolveLedgerExponent`'s throw is unreachable from production: `calculateInvoicePricing` has exactly one non-test call site (`invoice.service.ts:666`) and it passes the `isLedgerCurrencySupported` guard at :680, so an ambiguous or unknown code keeps today's two decimals instead of losing invoicing. Right call for a slice with no migration.

## One finding — the two rounders have a consumer you have not named

`pricing.ts` now has `quantizeMoney` and `roundMoney` side by side, and your comment says they disagree on ties (8.165 → 8.17 vs 8.16). Measured with your real `quantizeMoney` over every half-cent from 0.001 to 200.000: **1132 divergences in 200000 values**, always quantize-up / roundMoney-down. First few: 2.135, 2.175, 2.385, 2.425, 4.015. Control: 10.01 agrees in both, so the sweep can produce agreement and not only divergence.

The consumer is `finance/payment.ts:772` in `buildCheckoutSessionLineItems`:

```ts
const discountedItemSum = roundMoney(/* re-derived from line items */);
const preTaxInvoiceTotal = roundMoney(invoice.totalAmount - (invoice.taxTotal ?? 0));
const useBalanceLine = summary.paid > 0 || summary.credited > 0 || discountedItemSum !== preTaxInvoiceTotal;
```

That is an exact float `!==` between a total re-derived through `roundMoney` and `invoice.totalAmount`, which this PR now computes through `quantizeMoney`. Where the two disagree by a cent, `useBalanceLine` flips false → true and Stripe checkout shows a single "Outstanding balance for invoice X" line instead of the itemisation. The amount charged is unchanged, and tax stays correct because `disableAutomaticTax` is gated on `balanceIncludesTax` — so this is a presentation regression, not a money bug.

It is in the payment slice you deliberately deferred, so I am not asking you to widen the diff. But the PR causes it, and the body currently does not mention it. Either quantize `discountedItemSum` at the invoice exponent too, or name this consumer in the body next to the `roundMoney` note so the payment slice inherits it as a known item.

## Non-blocking

Ledger precision for the 162 supported codes is derived at module load from the runtime's ICU tables, so a Node/ICU bump can move a currency's exponent. Your `covers every ICU currency except the ones it names` test checks the size against ICU's own list rather than a constant, and the disagreement test catches a code moving into ICU/ISO conflict, which covers most of it — worth knowing that a same-category exponent change would still pass.

Tests: the `2.675` case printing `.toFixed(20)` as its control, and the `ZZZ` case asserting both that `Intl` says 2 and that the resolver throws, are the right shape — each one names the thing that would otherwise make it pass for the wrong reason.

ankit-yc (approved):
Re-approved at 52b4c235c. The only production change since 4dc4dc91d is the comment on the useBalanceLine comparison; behaviour is unchanged.

Verified independently in a detached worktree at 52b4c235c (confirmed by git rev-parse):

- apps/backend finance.payment.test.ts: 130 passed, 0 skipped.
- Mutation 1 - dropped `discountedItemSum !== preTaxInvoiceTotal` from useBalanceLine: 3 failed, including "charges the balance line at the posted total when a tie splits the two roundings". The tie test is keyed on that clause, not on an incidental balance-line selection (paid and credited are both 0 in the fixture).
- Mutation 2 - added a cent to effectiveUnitAmount in the itemised branch: 5 failed, including "under-submits an itemised line whose unit price is not representable". The per-unit characterisation is live.
- Both mutations restored; git diff clean before finishing.

Agreed on naming rather than quantizing here: quantizing the comparison routes the tie into the itemised branch, which is the branch that undercharges, so it would trade an exact 817 for 816. The non-tie case (2 x 8.165 -> 1632 against 1633) is genuinely older than this slice and correctly split out to #3305.

One non-blocking note: `expect(submitted).toBeLessThan(1633)` follows `expect(submitted).toBe(1632)`, so it cannot fail independently. Harmless as documentation of intent.

PR #3306, opened 2026-09-18
fix(desktop): name keys and the file manager for the running platform

### LINKED ISSUE #3303: Desktop: Windows and Linux users see macOS key names and 'Reveal in Finder'
## What happens

The desktop app ships for Windows and Linux as well as macOS, but some labels are macOS-only, so Windows and Linux users are told to press keys they do not have or look for a Finder they do not use:

- Tab bar tooltips say "Search tabs (Shift+Cmd+Y)" and "New tab (Cmd+T)" on every platform. On Windows and Linux the keys are Ctrl+Shift+Y and Ctrl+T.
- The keyboard shortcut list (opened with Ctrl+/ on Windows and Linux) shows only the macOS key symbols (the Command and Shift glyphs).
- The Vault preview button says "Reveal in Finder" on every platform, although it opens the system file manager everywhere.

Checked and not affected: the lock screen's "Unlock with Touch ID" only appears on macOS, and What's new says "Cmd/Ctrl+K", which is correct on every platform.

Screenshots from the 2026-09-18 desktop UI check: `caption-controls-light-1280x800.png` (Windows-style window controls, rendered on macOS), `cheatsheet-light-1280x800.png`, `vault-preview-text-light-860x640.png`. Tooltips do not show in still images; the label text is in the source.

## Where

Screens: main window tab bar, keyboard shortcut list, Document Vault preview.

- Tooltips: `apps/desktop/src/pages/tabbar.html:263` and `:271`. The real keys are Ctrl on Windows and Linux: the menu uses `CmdOrCtrl` (`apps/desktop/src/ui/app-menu.ts:117-119`) and the page checks `metaKey || ctrlKey` (`tabbar.html:722`, `:886`).
- Shortcut list: `tabbar.html:323-377`. It opens with Ctrl+/ on Windows and Linux (`tabbar.html:890`).
- The page already branches on platform for the caption buttons (`tabbar.html:395`, `:903`), so the same check can drive the labels.
- Vault: `apps/desktop/src/pages/vault.html:85`. The Vault is on the Data menu on every platform (`app-menu.ts:280`) and uses the cross-platform `shell.showItemInFolder`. The vault page has no platform value today (the preload exposes none).

## Checklist

- [ ] Tab bar tooltips show Ctrl-based shortcuts on Windows and Linux.
- [ ] Keyboard shortcut list shows Ctrl and Shift names on Windows and Linux.
- [ ] Vault reveal button uses the platform's file manager name.

## Expected

- One small platform helper (or a platform value from the preload) picks the label: "Cmd" and the macOS symbols on macOS, "Ctrl" and "Shift" elsewhere.
- The Vault button reads "Reveal in Finder" on macOS, "Show in Explorer" on Windows and "Show in folder" on Linux.

aupyay:
Closes #3303

## What was wrong

The desktop app ships for Windows and Linux as well as macOS, but three labels were written for macOS only:

- the tab bar tooltips read "Search tabs (Shift+Cmd+Y)" and "New tab (Cmd+T)" on every platform;
- the keyboard shortcut list showed only the Command and Shift glyphs, on a panel that opens with Ctrl+/ on Windows and Linux;
- the Vault preview button read "Reveal in Finder" everywhere, although it calls the cross-platform `shell.showItemInFolder`.

## The change

- `preload.ts` exposes `platform` (`process.platform`) as a plain value on the bridge. The renderer's user agent cannot tell Windows from Linux, and the Vault page had no platform value at all.
- New `src/pages/platform-labels.js` renders a platform-neutral accelerator — `Mod+Shift+Y`, the shape `CmdOrCtrl` already gives the app menu — for the running host, and names the system file manager. It is a plain browser script because the local pages load it over `file://` under a strict CSP and the sandboxed preload cannot share a compiled module with them; its types are restated at the one import in the test.
- The shortcut list rows and the two tab-bar buttons now carry `data-shortcut` in the markup and are labelled from it at load.
- The tab bar's two existing `navigator.userAgent.includes('Macintosh')` checks (traffic-light padding, caption buttons) read the same platform value, so the page has one answer rather than three.

macOS renders exactly as before, with one intended exception: the tooltips now read "Cmd+Shift+Y" rather than "Shift+Cmd+Y". The shortcut list glyphs are unchanged.

## Verification

Rendered `build/pages/tabbar.html` and `build/pages/vault.html` in Chromium with the bridge stubbed at each platform, and read the labels back out of the DOM:

| | macOS | Windows | Linux |
|---|---|---|---|
| New tab tooltip | `New tab (Cmd+T)` | `New tab (Ctrl+T)` | `New tab (Ctrl+T)` |
| Search tabs tooltip | `Search tabs (Cmd+Shift+Y)` | `Search tabs (Ctrl+Shift+Y)` | `Search tabs (Ctrl+Shift+Y)` |
| Shortcut list (16 rows) | `⌘T ⌘W ⌘⇧T ⌘⇧Y ⌘1-9 ⌘⇧E ⌘⇧\ ⌘⇧H ⌘⇧B ⌘K ⌘F ⌘G ⌘⇧G ⌘P ⌘, ⌘/` | `Ctrl+T … Ctrl+/` | `Ctrl+T … Ctrl+/` |
| Vault button | `Reveal in Finder` | `Show in Explorer` | `Show in folder` |

The 16 macOS rows are byte-for-byte the strings the markup carried before this change. The Windows caption buttons appear on the `win32` and `linux` runs and not on `darwin`, so the unified platform read drives that branch as the user-agent check did.

Gates, run from `apps/desktop/`:

- `pnpm run type-check` — clean.
- `pnpm run lint` — clean.
- `pnpm run archlint` — no new smell class. The new page script joins the eight `src/pages/*.js` files already reported as Dead Code, because page scripts are reached from HTML rather than imported.
- `pnpm test` — 74 suites, 1175 tests passing; statements 98.37, branches 90.93, functions 97.68, lines 98.37, all above the configured thresholds.
- `npx prettier --check` on every changed file — clean.

The 40 new tests were mutation-checked one at a time: making `revealLabel` always return "Reveal in Finder" fails 4; ignoring the platform in `shortcut` fails 12; ignoring the bridge in `detectPlatform` fails 1; ignoring the `name` style fails 4; dropping `platform` from the preload fails 3.

## Two things found on the way, not fixed here

- `apps/desktop/tsconfig.test.json` includes `tests/**/*.ts`, but `exclude: ["node_modules", "build", "tests"]` is inherited from `tsconfig.json` and wins, so the second arm of `pnpm run type-check` currently compiles **zero** test files (`tsc -p tsconfig.test.json --listFiles` lists none under `tests/`). Checking `tests/preload.test.ts` directly at `origin/dev`, with no part of this change applied, reports a pre-existing `TS2352` on its `as Record<string, unknown>` cast. Fixing the exclude would surface that and whatever else across 74 suites, which does not belong in this PR — I will file it separately. `allowJs` is set here so the new test compiles once that is fixed; both arms were measured (without it the import is `TS7016`).
- The shortcut panel clips its last row at 800px tall. It does so identically on macOS before and after this change, so it is not from here.


ankit-yc (approved):
Approved at 311563ee6.

Reproduced the user-visible claim rather than reading it. Built `apps/desktop` in a detached worktree at 311563ee6, rendered `build/pages/tabbar.html` and `build/pages/vault.html` in Chromium with `window.ycDesktop.platform` stubbed at each of the three values, and read the labels back out of the DOM:

| | darwin | win32 | linux |
|---|---|---|---|
| New tab title | New tab (Cmd+T) | New tab (Ctrl+T) | New tab (Ctrl+T) |
| Search tabs title | Search tabs (Cmd+Shift+Y) | Search tabs (Ctrl+Shift+Y) | Search tabs (Ctrl+Shift+Y) |
| Reveal button | Reveal in Finder | Show in Explorer | Show in folder |
| .cs-key rows | 16 | 16 | 16 |

Every cell matches your table. The 16 darwin key strings are byte-identical to the ones in `origin/dev`'s markup - I extracted them from the old HTML and diffed against the rendered set. The three rows differ from each other, so the platform stub is demonstrably driving the render.

Also verified:

- `pnpm test`: 74 suites / 1175 tests, 0 skipped.
- Mutation - `shortcut()` ignoring the platform (dropped the non-mac branch): 12 failures, matching your row.
- `build/pages/platform-labels.js` is present after `pnpm run build`, so the copy-static entry is what carries it into the package.

Three notes, none blocking:

1. I confirmed the `tsconfig.test.json` finding independently, because it decides how much of this PR CI actually checks: `tsc -p tsconfig.test.json --listFiles` compiles 524 files at this commit, 0 of them under `tests/`, and `src/pages/platform-labels.js` is not in the program either (controls: 64 hits for `/apps/desktop/src/` and 1 for `preload.ts` from the same grep, so the zero is not a spelling artefact). So the `allowJs: true` you added is inert until #3307 lands - correct to add it, but worth being explicit that nothing in CI type-checks the new script or the new test today. Jest is the only thing exercising them.

2. `vault.js` calls `globalThis.ycPlatformLabels.revealLabel(...)` unguarded at script load, so if that asset ever fails to load the vault page dies entirely rather than showing a stale label. Not live today - I verified the build copies it - but it converts a cosmetic dependency into a hard one. A `?.` with the old string as the fallback would keep it cosmetic.

3. `data-shortcut="Mod+1-9"` renders as "Ctrl+1-9" off macOS, which reads slightly oddly next to real accelerators. Fine as is; only mentioning it because it is the one row that is a range rather than a key.

Issue #3304, opened 2026-09-18
Desktop: UI check 2026-09-18 - tracking


aupyay:
Findings from a hands-on UI check of the desktop app shell on 2026-09-18: 47 screens in light and dark theme at 1280x800 and 1024x700 (the main window's minimum), captured by driving the real Electron app through the e2e harness. 87 verified findings were merged into the issues below. Findings that only occur below the minimum window size were dropped.

### High
- [x] #3287
- [x] #3288
- [x] #3289
- [x] #3290
- [x] #3291
- [x] #3292

### Medium
- [x] #3293
- [x] #3294
- [x] #3295
- [x] #3296
- [x] #3297
- [x] #3298
- [x] #3299
- [x] #3300
- [x] #3301
- [x] #3302

### Low
- [x] #3303

Visual changes follow the design-first rule: agree the fix against the PIMS design system before implementing, then verify it rendered in both themes. Functional fixes (split view, offline retry, window dragging, shortcut conflicts) can go straight to a PR with a test that fails without the fix.

One further lock-screen fix is being handled directly in a PR.

aupyay:
#3289, #3290 and #3291 are in #3371 (one PR — same surface).

Decisions taken on them, so they are on the record:

- **#3289, which approach.** The issue offered three. Taken: grow the chrome view to the whole window while any overlay is open, in *both* orientations, with a transparent view, **plus** a native `Menu.popup()` for the right-click menu. The native menu is what clears checklist items 1 and 2 together — arrow keys, Escape and screen-reader semantics come from the OS rather than being re-implemented as `role="menuitem"` on divs.
- **#3289, the hover preview is kept, not removed.** It has never rendered since it shipped (320x200 placed at the tab's bottom edge inside a 40px view), so removal was defensible. Keeping it costs less: it becomes one more caller of the shared overlay rule and its `yc:tab-get-preview` handler and tests stay as they are.
- **The shared primitive is renamed.** `setTabSearch` / `yc:tab-search` → `setChromeOverlay` / `yc:chrome-overlay`. A primitive named after one of its three callers is how this came back per feature; the vertical-mode clipping was exactly that rule never being applied outside tab search.
- **#3290, the macOS inset.** 78px is a left inset horizontally and a ~40px top inset in the rail, driven by a `data-platform` attribute the stylesheet answers rather than the inline padding it was. The orientation changes at runtime, which the inline padding could not follow.
[...]

PR #3308, opened 2026-09-18
fix(desktop): detach the closed split pane and retry the page that failed

### LINKED ISSUE #3287: Desktop: closing split view leaves the old right-hand pane covering half of the active tab
## What happens

After split view is closed (View > Close Split View, or the command palette), the tab that was on the right is not removed. It stays drawn over the right half of the window:

- The active tab is cut off at the middle of the window.
- Clicks on the right half go to the old pane, not to the active tab.
- If the window is resized, the old pane keeps its previous size and position, so the two views drift apart.
- The old view is never removed from the window. Switching to another tab draws that tab over it, but the old view stays mounted underneath until its own tab is shown or closed.

Screenshots from the 2026-09-18 desktop UI check: `tabs-split-closed-light-1280x800.png`, `tabs-split-closed-dark-1280x800.png`, `tabs-split-closed-light-1024x700.png`, `tabs-split-closed-dark-1024x700.png`. At 1280x800 the active tab's title is cut off at x=640 and the old "Lab results" pane covers the right half.

## Where

Screen: main window, tab mode, right after split view is closed.

- `apps/desktop/src/main.ts:585-593` `setSplitTab(null)` only sets `splitId` to null and calls `layoutTabChrome()`.
- `apps/desktop/src/main.ts:362-375` `layoutContentPanes` returns early when there is no split, so the previous split view is never passed to `contentView.removeChildView` and never given new bounds. It was added after the primary view (`main.ts:374`), so it stays on top and takes the input.
- `apps/desktop/src/main.ts:480-494` `switchToTab` only detaches `attachedTabId`, so after a tab switch the stale view is covered by the new tab but stays mounted.

The same path is used by View > Close Split View and by the palette, so every way of closing the split hits this.

## Expected

- When the split closes, the previous split view is detached (for example `detachContentView(previousSplitId)`), unless it is the tab that is now attached.
- The active tab fills the whole content area at every window size and receives all input.
- A unit test around `setSplitTab(null)` asserts that the previous split view is removed from the window, and fails on the current code.

### LINKED ISSUE #3288: Desktop: offline 'Try again' loads the sign-in page in the active tab instead of reloading the page that failed
## What happens

When a tab loses its connection it shows the "You're offline" page, with a "Try again" button and a 20-second countdown. None of the retry paths reload the page that failed:

- "Try again", the countdown reaching zero, and the network coming back all load the app's start URL (the sign-in page, which then redirects a signed-in user to the app home). The record the user was on is lost.
- The start URL is loaded into whichever tab is active at that moment, not the tab that showed the offline page. If the user has switched tabs while the offline tab counts down, or when the network comes back, the tab they are now working in is navigated away.
- "Open in browser" also opens the start URL, not the page that failed.

Screenshots from the 2026-09-18 desktop UI check: `offline-light-1280x800.png`, `offline-dark-1280x800.png`, `offline-light-1024x700.png`.

## Where

Screen: offline page shown inside a tab after a failed load.

- `apps/desktop/src/pages/offline.js:7-11`, `:18`, `:22-30`: the button, the `online` event and the countdown all call `ycDesktop.reload()`.
- `apps/desktop/src/core/ipc-handlers.ts:242-245`: `yc:reload` calls `loadStartUrl()` and ignores which tab sent it.
- `apps/desktop/src/main.ts:624-631`: `loadStartUrl` loads `config.startUrl` into `activeContents()`.
- `apps/desktop/src/core/ipc-handlers.ts:246-249`: `yc:open-in-browser` opens `config.startUrl`.
- `apps/desktop/src/main.ts:480-493`: `switchToTab` only detaches the old view, so an offline tab keeps its timer and `online` listener running in the background.

## Expected

- The offline page knows the URL that failed (pass the failed URL to `offline.html` when it is shown).
- Retry reloads that URL in the sender's own `webContents` (`event.sender`), and never touches the active tab or falls back to the start URL. The URL stays subject to the existing navigation allow-list.
- "Open in browser" opens the failed page.
- A test for `yc:reload` asserts that the sender's `webContents` is reloaded with the failed URL and the active tab is untouched.

aupyay:
Two functional children of the 2026-09-18 desktop UI check (#3304). Both are tab
content-view lifecycle bugs in the desktop shell.

## #3287 — a closed split view stayed on screen

`setSplitTab(null)` cleared `splitId` and re-ran the layout. `layoutContentPanes`
returned early when there was no split, so the view that had been the right-hand
pane was never passed to `removeChildView`. It was added *above* the primary
view, so it kept its old bounds through resizes, kept covering the right half of
the window, and kept receiving the clicks meant for the active tab.

Five places clear `splitId` (the menu/palette toggle, closing the split tab,
closing the active tab while split, leaving tab mode, and the window-reopen
reset), so fixing only the toggle would have left the rest. The fix goes in the
layout pass, which every one of them runs: it tracks the tab it mounted as the
right pane and detaches that view when the split closes or moves — unless that
tab is the one now attached, which is what happens when closing the active tab
promotes the split tab to primary.

The pane maths and the mount/detach decision moved to
`apps/desktop/src/ui/content-panes.ts` so the lifecycle is testable without
booting the app. No behaviour change to the split geometry itself.

## #3288 — offline "Try again" reloaded the wrong page in the wrong tab

`yc:reload` called `loadStartUrl()`, which loads `config.startUrl` into
`activeContents()`. All three retry paths — the button, the 20-second countdown
and the `online` listener — went through it, so:

- the record the user was on was replaced by the sign-in page, and
- because an offline tab keeps its timer running in the background, a countdown
  firing after the user switched tabs navigated away the tab they had moved to.

The shell now records the failed page against the `webContents` that is about to
show the offline page, and `yc:reload` reloads that page in `event.sender`. The
active tab is never touched.

Two deliberate choices:

- **The URL is never supplied by the renderer.** It is resolved against the
  existing navigation allow-list when it is *recorded*, so the retry path loads
  a value the shell already trusts. A blocked or missing URL records the start
  URL instead of widening the trust boundary. This is why the offline page is
  not given the URL as a query parameter, which is the shape the issue sketched.
- **A record only applies while that page is still showing.** `yc:open-in-browser`
  is shared with the welcome screen, where the button means "open the app", not
  "open whatever this window failed to load earlier".

The background timer itself is left alone. It is now harmless: it reloads its own
tab.

## Verification

Run in `apps/desktop` at the commit on this branch.

| Command | Result |
|---|---|
| `pnpm run type-check` | rc=0 |
| `pnpm run lint` | rc=0 |
| `pnpm test` | 75 suites / 1158 tests, rc=0, 0 skipped. Statements 96.33% → 98.4% |
| `pnpm run build && npx playwright test` | 44 passed, 1 skipped (the pre-existing `test.fixme` from #2249) |

### Mutation checks

Every new test was driven against a broken version of the code it guards, and the
tree was restored byte-identical (md5) afterwards.

| Mutation | Result |
|---|---|
| Remove the detach entirely | 3 unit failures |
| Detach without the "promoted to active tab" guard | 1 unit failure |
| Retry loads the start URL | 4 failures (unit + IPC) |
| Handlers act on the active tab again | 1 IPC failure |
| Drop the still-showing-the-offline-page scoping | 1 unit failure |
| Remove the detach, rebuild, run the e2e | e2e fails: `Expected: 1  Received: 2` mounted content panes after the split closed |

The last one is the reproduction: a new e2e in `tabs.e2e.ts` opens a split in the
real Electron app, closes it, and reads `contentView.children` from the main
process. Without the fix the old pane is still mounted and still half-width; with
it, one pane at `x=0` spanning the full content width.

Fixes #3287
Fixes #3288


ankit-yc (approved):
Approved at 04f89face.

Verified independently in a detached worktree at 04f89face:

- `apps/desktop` `pnpm test`: 75 suites / 1158 tests, 0 skipped, statements 98.4%. Matches the body.
- Mutation A - deleted the detach block from `layoutContentPanes`: 3 unit failures. Matches the body's row.
- Mutation B - `retry()` loads `config.startUrl.href` instead of `targetFor(sender)`: 4 failures across offline-retry and ipc-handlers. Matches.
- Tree restored byte-identical (md5 of both new modules before and after).

Fixing this in the layout pass rather than in `setSplitTab` is the right call: I traced the five sites that clear `splitId` and all of them re-run the layout, so `mountedSplitId` is the only handle that survives the clear. The promotion guard (`mountedSplitId !== attachedTabId`) is load-bearing for `closeActiveTab`, which calls `switchToTab(splitId)` while `splitId` is still set - that pass sees attached === mounted and correctly detaches nothing.

Resolving the retry URL at record time against the existing allow-list, rather than passing it to the offline page as a query parameter, is the safer shape. Nothing widens the trust boundary.

Two non-blocking notes:

1. `remember()` can discard a good record. If a second `did-fail-load` arrives for a tab that is already showing the offline page AND reports an empty `validatedUrl`, `candidate` falls back to `wc.getURL()`, which is then the `file://` offline page, which is not allow-listed - so the target is overwritten with the start URL and the page the user was actually on is lost. I confirmed this with a scratch test against the module: second record came back as the sign-in URL after a first record of a patient page. Narrow (Electron normally supplies `validatedUrl` on a retry failure, and the retry path re-records the good URL), so a note rather than a block. A guard like "ignore a candidate that is the offline page and keep the existing entry" would close it.

2. `yc:open-in-browser` changes meaning for the offline page: it used to open `config.startUrl` and now opens the recorded page. Deliberate and described, and the value was allow-list-validated at record time, so this is correct - flagging it only because it is a user-visible behaviour change that the issue did not ask for.

PR #3309, opened 2026-09-18
fix(repo): compile the desktop test suite in the type-check test arm

### LINKED ISSUE #3307: Desktop: the test arm of `pnpm run type-check` compiles zero test files
## What happens

`apps/desktop`'s `type-check` script is `tsc --noEmit && tsc --noEmit -p tsconfig.test.json`, and `apps/desktop/AGENTS.md` presents the second arm as the type-safety net for the test suite that `jest.config.js` deliberately gives up ("Transpile-only: type-safety is enforced separately by `tsc -p tsconfig.test.json`").

That arm compiles no test files at all. `tsconfig.test.json` sets `include: ["tests/**/*.ts", "src/**/*.ts"]`, but it extends `tsconfig.json`, which sets `exclude: ["node_modules", "build", "tests"]`. `exclude` is inherited and filters `include`, so every file under `tests/` is dropped from the program.

Reproduce from `apps/desktop/`:

```
pnpm exec tsc --noEmit -p tsconfig.test.json --listFiles | grep '/tests/' | wc -l
```

This prints `0`, while the same command without the `/tests/` filter lists several hundred files, so the program is built and it simply contains none of the suite.

## Why it matters

Type errors in test files are invisible to both gates: jest transpiles without checking, and this is the check that was supposed to cover it. There is at least one already on `dev`. Checking a single suite directly, with the project's own compiler options:

```
pnpm exec tsc --noEmit --strict --esModuleInterop --module Node16 \
  --moduleResolution node16 --target ES2021 --lib ES2021,DOM --types node,jest \
  tests/preload.test.ts
```

reports

```
tests/preload.test.ts(331,9): error TS2352: Conversion of type 'YcDesktop' to type 'Record<string, unknown>' may be a mistake
  because neither type sufficiently overlaps with the other.
  Index signature for type 'string' is missing in type 'YcDesktop'.
```

## Expected

The test arm of `type-check` compiles the suite. The likely fix is an explicit `"exclude"` in `tsconfig.test.json` that overrides the inherited one, rather than changing the app config.

Because the arm has never run, turning it on will surface whatever has accumulated across the 74 suites — the `TS2352` above is just the one found incidentally. Worth expecting a batch of fixes alongside the config change, not a one-line PR.

Found while working on #3303; deliberately left out of that PR's scope.

aupyay:
Closes #3307

## The defect

`apps/desktop/tsconfig.test.json` sets `include: ["tests/**/*.ts", "src/**/*.ts"]` but extends `tsconfig.json`, which sets `exclude: ["node_modules", "build", "tests"]`. An inherited `exclude` filters this file's `include`, so the second arm of `type-check` built a 526-file program containing **none** of the suite.

Measured on `dev` before the change, from `apps/desktop/`:

```
$ pnpm exec tsc --noEmit -p tsconfig.test.json --listFiles | wc -l
526
$ pnpm exec tsc --noEmit -p tsconfig.test.json --listFiles | grep -c '/tests/'
0
$ ls tests/*.test.ts | wc -l
76
```

Jest transpiles without checking (`isolatedModules`), and this was the check meant to cover that, so test files were unchecked by both gates.

## The change

`tsconfig.test.json` restates `exclude` so the inherited one stops filtering. That surfaced **416 errors over 47 files**, all fixed here. No strictness flag was relaxed — `strict` and `noUncheckedIndexedAccess` stay on for the test program.

236 of the 416 were `noUncheckedIndexedAccess` on mock call tuples and array reads (`calls[0].args`, `handlers['did-stop-loading']()`). Those were asserted with `!` at the exact span each diagnostic reported, via the TypeScript compiler API rather than a column heuristic, so the assertion lands on the expression the checker actually complained about.

The remaining 180 were real drift that nothing had been reading:

- `tests/cs-export.test.ts` — the `AuditLog` double returned `hash`/`previousHash` (the field names were renamed to `signature`/`prevSignature`), `verify()` returned `{ valid: true }` where the interface says `boolean`, and `verifyAll()` returned `[]` where it says `{ valid, tampered, otherKey }`.
- `tests/tab-manager.test.ts` — the initial-tab fixture is missing `audible` and `muted`; `TabManager` was not exported although it is the return type of the exported `createTabManager`.
- `tests/verify-macos-signature.test.ts` — the `VerifyModule` type had no `main`, though the script exports it and four tests call it.
- `tests/keyboard-shortcuts.test.ts` — a `find()` predicate typed `(c: string[])` over a `[accelerator: string, callback: () => void]` tuple.
- `tests/window-state.test.ts` — `save: (s) => saved.push(s)` returns `push()`'s `number` where the store contract says `boolean`.
- `tests/secondary-display.test.ts` — `DisplayInfo` used with no import.
- `tests/ipc-handlers.test.ts` — the services double never provided `idleUnlock`; spreading a `Partial<IpcServices>` had been making every member optional and hiding it.
- `tests/audit-key-preservation.test.ts` — `Array.prototype.at` is ES2022 and the app's `lib` is ES2021; changed to index access rather than widening the lib.
- The fs-shaped doubles in eight suites are narrower than the `typeof fs.readFileSync` seams they stand in for. `tests/helpers/fs-seam.ts` widens the mock and leaves the double's own body checked, matching what `tests/helpers/memory-fs.ts` already does.

One production change: `export interface TabManager` in `src/core/tab-manager.ts`. No emitted JS changes.

## Evidence

Mutation check — a deliberate `const mutationProbe: number = 'not a number';` appended to `tests/logger.test.ts`:

| tsconfig.test.json | `tsc --noEmit -p tsconfig.test.json` |
|---|---|
| with the explicit `exclude` (this PR) | `tests/logger.test.ts(87,7): error TS2322` — 1 error |
| with the inherited `exclude` (dev) | 0 errors |

The probe and the config revert were both removed afterwards; the tree is back to 0 errors.

Run from `apps/desktop/` at `77073ac46`:

```
pnpm run type-check   # exit 0 (both arms)
pnpm run lint         # exit 0
pnpm run archlint     # exit 0
pnpm run build        # exit 0
pnpm run test         # 76 suites passed, 1200 tests passed, 0 skipped
                      # statements 98.4% (10902/11079), branches 91.04%, functions 97.72%
```

Canary on the fixed config: the test-arm program now lists **86** files under `tests/` (82 spec files plus the two helpers and the e2e fixtures), against 0 before.


ankit-yc (approved):
Approved at 77073ac46a629550a101e451914ad2758a3484ca. Verified independently in a clean checkout of the head commit.

**Root cause is right.** `tsconfig.test.json` extends `tsconfig.json`, whose `exclude` lists `tests`; an inherited `exclude` filters this file's own `include`, so the test arm resolved zero test files. Restating `"exclude": ["node_modules", "build"]` is exactly the base list minus `tests` — no other path is newly or no-longer excluded.

**Measured both arms** (`tsc --noEmit -p apps/desktop/tsconfig.test.json --listFiles`, run from `apps/desktop`):

| arm | rc | files under `tests/` | total program |
|---|---|---|---|
| head | 0 | 86 | 726 |
| head with the `exclude` line removed | 0 | 0 | 525 |

**Mutation.** Appended `const __mutationProbe: number = 'not a number';` to `tests/tab-manager.test.ts`:
- with the fix: `rc=2`, `tests/tab-manager.test.ts(356,7): error TS2322`
- with the `exclude` line removed: `rc=0`, zero output

So the arm was dead, and it is the `exclude` restatement that makes it live.

**Gates at this sha:** `tsc --noEmit` (src arm) rc=0; `jest --ci` 76 suites / 1200 tests passed, lines 98.4%.

**No strictness relaxed.** `noUncheckedIndexedAccess` and `strict` stay on and no compiler flag changed. The 416 errors were absorbed as `[i]!` on index reads — 170 added assertion lines use `!`, zero use `?.`, which matters: `expect(xs[0]!.field)` still throws if the element is missing, whereas `?.` would have quietly passed. No `@ts-ignore`, `@ts-expect-error`, `@ts-nocheck`, `eslint-disable`, `.skip` or `.only` added anywhere in the diff, and `expect(` counts are 175 removed / 175 added — no assertion dropped.

**Two things I looked at and accept:**

- `tests/helpers/fs-seam.ts` returns `jest.Mock`, which does widen the double's call signature. It is the narrowest available concession: the production seams are typed as the full `fs` overload sets, which a hand-written double cannot implement, the callback body itself stays inferred and checked, and it follows the convention `tests/helpers/memory-fs.ts` already set. Production callers of those seams remain checked by the src arm.
- 22 added `as unknown as` casts, all in e2e `window.ycDesktop` access or in reading a typed value back as `Record<string, unknown>` for inspection. None of them cast away a check on the code under test.

`export interface TabManager` in `src/core/tab-manager.ts` is a visibility widening only — type-level, no runtime change.

PR #3310, opened 2026-09-18
fix(desktop): focus the idle lock and hold tab input until unlock

aupyay:
## PR Checklist

When the idle lock engages, the lock screen does not take keyboard focus: its Unlock and "Use password instead" buttons can only be reached with a click.

## What is the new behavior?

While the lock is up:

- **The lock page has focus.** It takes keyboard focus when it mounts. The card is a modal dialog (`showModal()`) with focus on Unlock, so Enter or Space unlocks without the pointer, and Escape does not dismiss it. The focused button uses the blue focus ring the other local pages use. The card's size and position are unchanged: before and after screenshots, light and dark, differ only by that ring.
- **Nothing else takes keyboard input.** Every web contents registers with the lock when it is created, and the lock drops its keyboard input (`before-input-event`). That covers every tab, including tabs opened or switched to during the lock, the tab bar and the window's own page. The lock page is the one exception.
- **The overlay stays on top of the whole window.** Tab attach over IPC (new, activate, duplicate, reopen, close) now runs the same layout pass as the menu path, which raises the lock last. A window closed while locked and reopened from the Dock comes back under the lock, signed in or out, with the lock page focused again.
- **Commands stand down.** Every app menu item with its own handler is inert until unlock (Quit still works). Role items such as Reload act on the focused contents, which is the lock page. The global shortcuts, command palette and tray commands, and the trackpad swipe do nothing until unlock.

On unlock the hold lifts and focus goes back to the active tab.

### Tests

Unit (Jest, mocked Electron objects):

- `idle-lock-overlay.test.ts`: the lock page is focused on show, and focus goes back to the workspace once the overlay is removed. A tab's keys pass before the lock, are dropped while it is up and pass again after. The lock page is exempt, a contents registered mid-lock is held, and only the current lock page is exempt.
- `app-menu.test.ts`: while locked only Quit acts; the same menu works again after unlock because the lock is read at click time. The Windows/Linux menu is covered too.
- `keyboard-shortcuts.test.ts`: no global shortcut acts while locked, and all of them do after unlock.
- `ipc-handlers.test.ts`: every IPC path that attaches a tab runs the shared layout pass.

E2E, `tests/e2e/idle-lock.e2e.ts`. It is macOS only because the overlay is on the Touch ID path. It launches with `--use-mock-keychain` and mocks idle time and the Touch ID prompt:

- A real idle lock engages. The lock page is focused, Unlock is the active element and the card is `:modal`. Native key presses (`sendInputEvent`, the same path as a keyboard) are not delivered to the tab, to a tab opened during the lock or to a tab switched to during the lock, and the overlay is topmost after each. Menu New Tab, palette `tab:new` and `pin-current`, and a swipe change nothing. Escape twice leaves the card open, and Enter then presses Unlock. When Touch ID resolves, the overlay is gone, the active tab is focused and it receives keys again.
- A window is closed while locked and reopened, signed in (tabs rebuilt) and signed out (welcome screen). Each reopened window is under the lock and the lock page is focused.

### Mutation checks

Each change below was applied, the suite run and the failure confirmed, then the file was restored with `cp` and checked identical with `cmp`:

| Change | Result |
| --- | --- |
| Lock page not focused on show | unit 3 failed; e2e focus check fails |
| Workspace not focused on hide | unit 1 failed; e2e focus check fails |
| Hold never drops input | unit 3 failed |
| Hold ignores the lock state | unit 2 failed |
| Hold also drops the lock page's input | unit 2 failed |
| Lock page not cleared on hide | unit 1 failed |
| Focus sent to a destroyed contents | unit 2 failed |
| Hold not registered on new web contents (`main.ts`) | e2e: tab received the key (1, expected 0) |
| Menu handlers not held | unit 2 failed; e2e tab count 3, expected 2 |
| Submenus not held | unit 2 failed |
| Quit held as well | unit 1 failed |
| Lock read once at menu build | unit 1 failed |
| Global shortcuts not held | unit 1 failed |
| IPC tab attach skips the layout pass | unit 5 failed; e2e topmost view is the new tab |
| Palette and tray commands not held | e2e tab count 3, expected 2 |
| Swipe not held | e2e 2 history moves, expected 0 |
| No refocus after the overlay is re-parented | e2e both reopen cases fail |
| Reopened window not covered | e2e signed-out reopen fails |
| Card not shown with `showModal()` | e2e `:modal` false |
| Escape not held on the lock page | e2e card closed |

### Validation

- `pnpm run type-check`, `pnpm run lint` and `pnpm test`, from `apps/desktop`: 76 suites, 1219 tests. `idle-lock-overlay.ts` is at 100% for statements, branches, functions and lines.
- The full desktop Playwright suite on macOS: 47 passed, 1 skipped (the existing skip in `packaged-app.e2e.ts`).
- `pnpm run archlint`: no findings beyond those already on `dev`.
- Aikido scan of the changed code: no issues.

## Related Issue(s)

No tracking issue.


ankit-yc (approved):
Read the full diff. The keyboard hold is the right shape: holdInput registers at web-contents-created so tabs opened or switched to mid-lock are covered, and the lockPage exemption is evaluated at event time rather than registration time, so the overlay's own contents (created before mount() returns) is still exempt. holdWhileLocked keys on identity of the mac quit item; the Windows/Linux quit is role:'quit' with no click, so it is untouched by construction. attachTabView delegating to services.layoutTabChrome matches the existing call at ipc-handlers.ts:1050 and ends in layoutLockOverlay, so a tab attached while locked lands under the overlay. All required checks green, desktop Sonar green. LGTM.

PR #3311, opened 2026-09-18
fix(desktop): scope app shortcuts to focus and stop them shadowing the menu

### LINKED ISSUE #3292: Desktop: shortcut list promises Cmd+Shift+T, Cmd+Shift+E, Cmd+Shift+B and Cmd+P, but global shortcuts take those keys
## What happens

The keyboard shortcut list (Cmd+/) shows these entries, which match the app menu:

- Cmd+Shift+T: Reopen closed tab
- Cmd+Shift+E: Toggle vertical tabs
- Cmd+Shift+B: Open in browser
- Cmd+P: Print

The app also registers the same four keys as system-wide global shortcuts for other actions: New appointment, Check in patient, Billing and the command palette. A global shortcut fires before the app menu, so pressing Cmd+Shift+T opens New appointment instead of reopening the tab, and so on for the other three.

Because they are registered system-wide at startup and only released on quit, these keys (and the other global shortcuts in the same list) are also taken away from every other app on the computer for as long as Yosemite Crew is running.

Two more entries only work in one place:

- Cmd+1 to Cmd+9 (jump to tab) and Cmd+/ (open the shortcut list) are handled only by key listeners inside the tab strip, so they do nothing while focus is in the page, which is most of the time.
- There is no menu item to open the shortcut list. `yc:show-cheatsheet` exists but nothing calls it.

Screenshots from the 2026-09-18 desktop UI check: `cheatsheet-light-1280x800.png`, `cheatsheet-dark-1280x800.png`, `cheatsheet-light-1024x700.png`.

## Where

Screen: keyboard shortcut list overlay, and the app menu.

- `apps/desktop/src/pages/tabbar.html:332`, `:341`, `:353`, `:371`: the shortcut list rows.
- `apps/desktop/src/ui/app-menu.ts:113`, `:129`, `:134`, `:218`: the menu accelerators for the same keys.
- `apps/desktop/src/ui/keyboard-shortcuts.ts:19-29`: the global registrations (`CommandOrControl+P`, `+Shift+E`, `+Shift+B`, `+Shift+T`, among others).
- `apps/desktop/src/main.ts:1994-2001`: registered with Electron `globalShortcut` at startup, never scoped to window focus; only unregistered on quit (`main.ts:2063`).
- `apps/desktop/src/pages/tabbar.html:719-740` and `:885-899`: Cmd+1 to 9 and Cmd+/ handled only inside the tab strip view.
- `apps/desktop/src/core/ipc-handlers.ts:1090`: `yc:show-cheatsheet` with no preload or menu caller.
- Also worth fixing in the same change: on Windows and Linux, the global `CommandOrControl+Shift+I` (Inbox, `keyboard-shortcuts.ts`) is the same key as the menu's Toggle Developer Tools (`Ctrl+Shift+I`, `app-menu.ts:204-205`).

## Expected

- Every shortcut in the list does what the list says.
- App shortcuts are menu accelerators, or are registered only while an app window is focused, on keys that do not collide with the tab and window keys. They are not held system-wide.
- Cmd+1 to 9 and Cmd+/ work from anywhere in the window (menu accelerators or a `before-input-event` handler).
- A Help menu item opens the shortcut list.
- A test fails if a global or window shortcut uses the same accelerator as a menu item.

aupyay:
Closes #3292

## What was wrong

Nine accelerators were registered with Electron's `globalShortcut` at startup (`main.ts`) and released only on quit, so they were taken away from every other application on the machine for as long as the app was running. A global shortcut also fires ahead of a menu accelerator, and four of them were the keys the shortcut list promises to the menu:

| Key | Shortcut list says | What actually ran |
|---|---|---|
| Mod+Shift+T | Reopen closed tab | New appointment |
| Mod+Shift+E | Toggle vertical tabs | Check in patient |
| Mod+Shift+B | Open in browser | Billing |
| Mod+P | Print | Command palette ("Quick Switch") |

`Mod+Shift+I` (Inbox) was the same key as Toggle Developer Tools on Windows and Linux. Two more list entries worked in one place only: Mod+1 to Mod+9 and Mod+/ were handled by a `keydown` listener inside the tab strip, which has focus almost never, and no menu item opened the shortcut list at all.

## The change

- **The app's own actions move to the `Mod+Alt` namespace.** The menu and the tab strip keep `Mod` and `Mod+Shift`. `Mod+Alt+I` is left alone because it is Toggle Developer Tools on macOS, so Inbox is `Mod+Alt+M`. The command palette keeps `Mod+K` and the menu's `Mod+Shift+P`; its duplicate `Mod+P` entry, which is what took Print, is dropped.
- **The keys are held only while one of the app's windows has focus** — taken on `browser-window-focus`, released on `browser-window-blur`. The blur decision is deferred a tick, because blur arrives before the focus of the window taking over: without that, moving from the main window to Settings would release them. `register()` is now idempotent so a second focus cannot double-register.
- **Mod+1 to Mod+9 are handled by the window** through `before-input-event`, so they work while the page has focus. The handler resolves the keystroke through the same exported accelerator list the conflict test checks, rather than a second copy of the digits.
- **Help > Keyboard Shortcuts**, on `Mod+/`. The tab-chrome page owns the overlay's open/closed state, so it exposes `__ycToggleCheatsheet` and the menu asks for it; the page's own `Mod+/` handler now calls the same toggle.

`before-input-event` is attached to every web contents the app creates, so Mod+1 works from a tab, the tab strip or a secondary window. That matches the existing menu accelerators, which also act on the main window from wherever they are pressed.

## Verification

From `apps/desktop/`:

- `pnpm run type-check` — rc=0.
- `pnpm test` — 78 suites, 1222 tests, rc=0. Statements 98.41, branches 91.12, functions 97.73, lines 98.41, all above the configured thresholds. (`dev` at `3effd3699` is 76 suites / 1200 tests.)
- `pnpm run lint` — rc=0. `pnpm run archlint` — rc=0, no new smell class.
- `npx prettier --check` on every changed file — clean.
- `npx playwright test` — 47 passed, 1 skipped (44 + 1 on `dev`, plus the three arms below).

`tests/shortcut-conflicts.test.ts` is the test the issue asks for: it builds the menu template for `darwin` and `win32`, collects every accelerator including the ones Electron supplies for menu **roles** (`undo`, `close`, `hideOthers`, …) which never appear in the template, and fails if a global or window accelerator lands on one, if two menu items share a key, or if two global shortcuts do.

Sixteen mutations were driven one at a time and reverted:

| Mutation | Result |
|---|---|
| `check-in` back on `Mod+Shift+E` | 2 conflict tests fail |
| window shortcut set to `Mod+T` | 2 conflict tests fail |
| two menu items given the same key | 5 fail |
| Help item's accelerator removed / its click unwired | 1 fails each |
| `register()` loses its idempotence guard | 1 fails |
| blur releases without checking focus | 2 fail |
| `start()` registers while unfocused | 1 fails |
| blur decision taken immediately instead of deferred | 2 fail |
| handler ignores the event type / the platform / Alt+Shift / the index / non-digit keys | 1 fails each |
| accelerator list no longer drives the handler's index | 1 fails |

Two e2e arms cover the wiring that unit tests cannot reach, and both were mutation-checked against a rebuilt app: removing the `before-input-event` registration fails the tab-jump arm, and removing the page's toggle export fails the Help-menu arm.

A third arm pins the unfocused half of the scoping: the launched app holds neither `Mod+K` nor `Mod+Alt+T` while it is not the frontmost application, and both focus and blur handlers are subscribed. Restoring the old unconditional `register()` at startup fails it. Playwright's Electron app never becomes frontmost — `win.focus()` leaves `isFocused()` false — so the focused half is covered by the unit tests over `createKeyboardShortcutManager` rather than end to end.

## Two things found on the way, not fixed here

- The macOS Quit item (`app-menu.ts`) is a plain click handler with no `accelerator` and no `role`, so nothing in the template binds Cmd+Q.
- Running the e2e suite leaves `playwright-report/`, which `eslint .` then lints: 1132 errors from bundled report assets. It is git-ignored but not eslint-ignored.


ankit-yc (changes requested):
Changes requested at a3af2a408327850de14a02e064e626a59c4733ee — one blocker, and it is entirely an artefact of timing rather than of the design, which I think is right.

## Blocker: this branch predates #3307, so its own new tests were never type-checked

`tsconfig.test.json` on this head is the pre-#3307 version, so the test arm of `pnpm run type-check` resolves **0** files under `tests/` here. The rc=0 in the description is real but vacuous for the six test files this PR touches — nothing in them was compiled.

#3307 merged to `dev` as `e73046f4` while this was open, and dev's tree now compiles all 86. Merging current dev into this branch locally (clean, no conflicts) and re-running the test arm gives **rc=2, 6 errors, all in files this PR adds or changes**:

```
tests/keyboard-shortcuts.test.ts(23,50): TS18048: 'windowEvents.focus' is possibly 'undefined'
tests/keyboard-shortcuts.test.ts(24,49): TS18048: 'windowEvents.blur' is possibly 'undefined'
tests/keyboard-shortcuts.test.ts(178,50): TS18048: 'windowEvents.focus' is possibly 'undefined'
tests/keyboard-shortcuts.test.ts(179,49): TS18048: 'windowEvents.blur' is possibly 'undefined'
tests/keyboard-shortcuts.test.ts(188,22): TS2532: Object is possibly 'undefined'
tests/e2e/tabs.e2e.ts(358,21): TS2532: Object is possibly 'undefined'
```

Control: dev's tree (`e73046f4^{tree}` = `99e3c1d1…`) is byte-identical to the tree I measured rc=0 on when reviewing #3307, so the six are attributable to this branch, not to dev.

Cause is `noUncheckedIndexedAccess` on the two `Record<string, Array<() => void>>` declarations at lines 10 and 168 — an index read off a `Record` is `… | undefined`, which is also what line 188's `deps.windowEvents[event]` trips on. Three lines fix it, and I verified rc=0 afterwards on the merged tree:

- lines 10 and 168: `const windowEvents: { focus: Array<() => void>; blur: Array<() => void> } = { focus: [], blur: [] };`
- `tests/e2e/tabs.e2e.ts:358`: `first.tabs![0]!.id`

Please merge or rebase onto current dev and push, so the arm that now exists actually runs over this diff. `core / Static (@yosemite-crew/desktop)` is still pending on this PR and is likely to land on exactly these six.

## The change itself is right, and the guard is real

Five mutations, each reverted before the next:

| mutation | result |
|---|---|
| `new-appointment` back to `CommandOrControl+Shift+T` | `darwin: global shortcuts are clear of the menu` fails |
| `CommandOrControl+W` appended to `WINDOW_SHORTCUTS` | `darwin: window shortcuts are clear of the menu` fails |
| Help accelerator off `CmdOrCtrl+/` | `the menu owns the keys the shortcut list promises for it` fails |
| `if (registered.length > 0) return;` removed | `a second focus does not register the same accelerators twice` fails |
| `event.preventDefault()` removed from the input handler | `Cmd+1 through Cmd+9 activate the tab at that position` fails |

So the conflict test the issue asked for is load-bearing in both directions — it fails on a clash and on the menu silently losing a promised key. Baseline on the merged tree: 78 suites / 1222 tests pass, src arm rc=0.

Two details I checked rather than assumed:

- `register()`'s early return and `unregister()`'s `registered.length = 0` are consistent, so the release-on-blur path can re-arm. If `unregister` had left the array populated the keys would never come back after the first blur; it does not.
- `Mod+9` means the ninth tab here, matching `activateByNumber` in the tab strip (`parseInt(key) - 1`). The window handler's 0-based index is the same tab, so the semantics did not shift.

## Non-blocking

1. `ROLE_ACCELERATORS` in the conflict test is a hand-maintained table, and a role that is not in it contributes no accelerator — so a future menu role could shadow a shortcut and the test would stay green. I enumerated the roles the template actually uses: `about`, `help`, `services` and `unhide` are the four missing from the table, and all four carry no accelerator, so there is no live gap today. Worth an assertion that every role in the template is either in the table or in an explicit no-accelerator list, so the detector reports what it could not analyse instead of passing over it.
2. The test covers global-vs-menu, window-vs-menu and global-vs-global, but not global-vs-window. Nothing collides now; a future `CommandOrControl+1` global would not be caught.
3. `tabbar.html`'s own `Cmd+1`–`9` branch is now unreachable — `before-input-event` calls `preventDefault()` before the key reaches the renderer. Harmless, but it is now a second copy of a rule that lives in `window-shortcuts.ts`.

ankit-yc (approved):
Approved at d5de6155fe3e882e3c56d822efbc5b19c330c53e. The blocker is resolved and all three non-blocking notes were taken.

## The blocker is gone, and the arm can now see this PR

`d5de6155f` merges current dev (`95691fb23`), so the test arm is live here. Measured at this head:

| check | result |
|---|---|
| `tsc --noEmit` (src arm) | rc=0 |
| `tsc --noEmit -p tsconfig.test.json` (test arm) | rc=0, 0 errors |
| test files the arm compiled | **89** (dev is 87 — the two this PR adds) |
| `jest --ci` | 78 suites / 1245 tests, rc=0 |

The 89 is the part that matters: rc=0 from an arm compiling nothing is exactly what let the six errors hide last time, so the count is what separates "passed" from "did not look."

## Both new guards were mutation-checked

| mutation | result |
|---|---|
| `{ role: 'zoom' }` added to the Help submenu — a real Electron role that is in neither table | `every menu role is either in the accelerator table or declared to have none` fails |
| `new-patient` moved to `CommandOrControl+1` | `no global shortcut takes a window key` fails |

Each reverted before the next; the file is clean. `ROLES_WITHOUT_ACCELERATORS` as an explicit list rather than an omission is the right shape — the detector now reports what it could not analyse instead of scoring it as no conflict. Collecting only roles that carry no explicit accelerator is correct, since one that does is already caught by the accelerator walk.

## One thing to watch before merging

Removing `activateByNumber` from `tabbar.html` leaves the main process's `before-input-event` handler as the *only* implementation of Mod+1–9. That is the right call — the comment you left in its place says exactly why the copy was unreachable — but it also means the new `tabs.e2e.ts` arm is the sole coverage for that key, and e2e does not run locally here. **Desktop E2E (macos-latest) and (windows-latest) are still pending on this PR.** Both need to be green before this goes in; if either fails on the Mod+1 arm, the handler is not attaching for the view under test and the key is now dead rather than duplicated.

Nothing else outstanding from my side.

PR #3313, opened 2026-09-18
fix(desktop): give the macOS Quit item its Cmd+Q accelerator

### LINKED ISSUE #3312: Desktop: the macOS Quit menu item has no Cmd+Q accelerator
## What happens

On macOS the application menu's Quit item is built as a plain click handler:

`apps/desktop/src/ui/app-menu.ts:94`

```ts
{ label: tr('menu.quit'), click: () => app.quit() },
```

It has neither an `accelerator` nor `role: 'quit'`, and no other menu item, global shortcut or window handler binds the key:

```
$ grep -rn "+Q'\|+Q\"" apps/desktop/src/     # no matches
$ grep -rn "quit" apps/desktop/src/ui/app-menu.ts
94:              { label: tr('menu.quit'), click: () => app.quit() },
168:        ...(isMac ? [] : [{ role: 'quit' as const }]),
```

So the Quit item shows no key next to it and Cmd+Q has nothing behind it. Every other macOS app quits on Cmd+Q, and the app menu is the only place this is declared.

Windows and Linux are unaffected: line 168 uses `role: 'quit'`, which carries its own accelerator.

## Why it is written that way

The mac item is not a role because it runs `app.quit()` through a translated label (`tr('menu.quit')`). A role would give the accelerator and the standard label but not the translation.

## Expected

The macOS Quit item carries `Cmd+Q`, keeping its translated label.

`accelerator: 'Cmd+Q'` on the existing item is the smallest change and keeps `tr('menu.quit')`. Worth checking against the idle-lock and window-close paths, since quitting is not the same as closing the last window on macOS.

Found while working on #3292; out of scope there, which is about keys that collide rather than a key that is missing.

aupyay:
Closes #3312

## Problem

The macOS application menu declared Quit as a plain click handler:

```ts
{ label: tr('menu.quit'), click: () => app.quit() },
```

No `accelerator`, no `role`. The item rendered with no key beside it and Cmd+Q had nothing behind it. Nothing else in `apps/desktop` bound the key — `/usr/bin/grep -rn -F -e "+Q'" -e '+Q"' apps/desktop/src/` returns rc=1 with no matches; the same scan for `+T'` returns 3 hits, so it can find one.

Windows and Linux were never affected: that template uses `role: 'quit'`, which carries its own accelerator.

## Fix

`accelerator: 'Cmd+Q'` on the existing item.

The item cannot become `role: 'quit'`: the role supplies the accelerator but also replaces the label, and this one is translated through `tr('menu.quit')` (`Quit Yosemite Crew PIMS` / `Salir de Yosemite Crew PIMS`). Declaring the accelerator restores the key and keeps the translation.

The click handler is unchanged, so the quit path is the same one the tray Quit item and the Windows/Linux role already use — `app.quit()`, which fires the existing `will-quit` teardown. `window-all-closed` still no-ops on darwin, which is the behaviour that makes Quit distinct from closing the last window.

### Composition with #3310

#3310 landed while this was open and hoisted the macOS Quit item into a `quit` const, so `holdWhileLocked` can exempt it by identity — Quit is the one item that still acts while the idle lock is up. The accelerator therefore goes on that declaration rather than on an inline item, and the two changes compose in the right direction: Cmd+Q reaches the item the lock exempts. A third test pins that, and fails if either the accelerator or the exemption is removed.

## Verification

Run on the branch head (`d2eebf63e`), from `apps/desktop`:

- `pnpm test` — 76 suites, 1222 tests, all passed
- `pnpm run type-check` — 0 errors on both arms (this needed #3317, which fixed the 8 pre-existing errors #3310 left in the arm #3309 had just switched on; see #3314)
- `pnpm run lint` — rc 0
- `pnpm run build` — rc 0
- the repo-wide pre-push gate (`turbo run lint` + `turbo run type-check`) passed on the pushed commit

Integration check against real Electron 44.1.1 on darwin, reading `Menu.getApplicationMenu()` back out of the compiled `build/ui/app-menu.js` — i.e. the accelerator Electron itself stored on the live menu item, after `holdWhileLocked` has mapped the template:

```
PROBE_PLATFORM darwin
PROBE_JSON [{"label":"Quit Yosemite Crew PIMS","role":null,"accelerator":"Cmd+Q"}]
PROBE_TOTAL_ACCELERATORS 33
```

Control arm — same probe with the accelerator line deleted from the build output:

```
PROBE_JSON []
PROBE_TOTAL_ACCELERATORS 32
```

## Test mutation checks

Every new assertion was driven to failure:

| mutation | result |
|---|---|
| `accelerator: 'Cmd+Q'` removed | **3 failed**, 6 passed |
| `item !== quit` dropped from `holdWhileLocked`, so the lock holds Quit too | **2 failed**, 7 passed |
| a second `Cmd+Q` added on another item | **1 failed** (`Expected length: 1 / Received length: 2`), 5 passed (measured pre-merge) |

## Not tested

Physically pressing Cmd+Q in the running app. Driving the macOS menu bar needs Accessibility automation permission for System Events, which cannot be granted in this non-interactive session; the AppleScript call hung on the permission prompt and was killed at 60s. The real-Electron probe above is what replaced it.


ankit-yc (approved):
Approving at 31d013520.

The PR's own green is against a head that **predates current dev** — `155052a04` is not an ancestor of `31d013520`, and the commits missing from it include #3311, which rewrote desktop shortcut scoping and added a menu-accelerator conflict guard. Since #3323 has not landed, CI here judged the branch tip, so nothing in this PR's checks has ever seen this accelerator next to #3311's guard. That is #3316's hazard with this PR as the live instance, so I merged dev locally and reviewed that tree instead.

**Merged tree** (`31d013520` + `origin/dev` `155052a04`, merge clean, no conflicts):

| check | result |
|---|---|
| `pnpm test` (`apps/desktop`, full package) | rc=0 — 78 suites, 1248 passed |
| `pnpm run type-check` | rc=0, 0 `error TS` |
| `pnpm run lint` | rc=0 |

1248 is dev's 1245 plus exactly the three tests this PR adds, so they are being collected rather than quietly filtered out. Branch tip alone was 76 suites / 1222 — the gap is the two suites #3311 brought.

**Composition with #3311 is the thing worth stating.** Its new `shortcut-conflicts.test.ts` collects `item.accelerator` across the whole template, normalizes, and asserts no duplicates and no clash with the registered shortcut list. `Cmd+Q` passes that on the merged tree, and it is not named in the guard's expected-accelerator list, so the two are independent rather than coupled.

**Mutation arms, driven on the merged tree** (baseline rc=0, 1248 passed; restored to rc=0, 1248 passed afterwards):

| mutation | result |
|---|---|
| `accelerator: 'Cmd+Q'` removed | rc=1 — **3 failed**, 1245 passed |
| `item !== quit` dropped from `holdWhileLocked` | rc=1 — **2 failed**: *"while the idle lock is up only Quit acts"* and *"Cmd+Q reaches the item the idle lock exempts"* |
| a second `Cmd+Q` (onto the `Cmd+Shift+P` item) | rc=1 — **5 failed across 2 suites**: this PR's *"declared once on macOS"* plus three from #3311's conflict guard |

The third arm is the one that pays for the merge. A duplicate accelerator is caught twice over on the merged tree — by this PR's own count assertion and independently by #3311's guard — and only one of those two covers exists on the branch tip CI actually judged.

Two things I am not claiming. I did not press Cmd+Q in the running app either; your real-Electron `Menu.getApplicationMenu()` probe with its 33-vs-32 control arm is better evidence than an AppleScript drive would have been, and I read it rather than repeating it. And `git merge` reporting no conflict is not by itself evidence of semantic compatibility — that is what the suite run and the third arm above are for.

Issue #3316, opened 2026-09-18
CI judges the PR head, not the merged tree, so a widened gate lands red on dev


aupyay:
`core / Static` (and `core`'s detect job) check out `github.event.pull_request.head.sha`:

```yaml
# .github/workflows/_core.yaml:291-293
- uses: actions/checkout@...
  with:
    ref: ${{ github.event.pull_request.head.sha || github.sha }}
```

So lint, `type-check` and `build` run on the branch tip, never on the merge of the branch
into `dev`. A PR is therefore green against a *tree that is not the tree being merged*, and
anything that only fails in combination lands on `dev` unseen.

This is not hypothetical. #3315 is exactly it: #3309 made the desktop `type-check` test arm
compile the suite (it had been compiling 0 test files), #3310 branched before that and added
test files with 8 type errors. On #3310's head the arm still compiled 0 test files, so
`core / Static (@yosemite-crew/desktop)` passed in 1m9s. The merged tree fails. The push run
on `dev` found it after the fact (run 35383477860).

The general shape: a PR whose branch predates a change that *widens what a gate reads* cannot
be judged by that widened gate, and GitHub does not re-run checks when the base advances. Any
gate that reads configuration from the tree has this property — a tsconfig `include`/`exclude`,
an eslint ignore list, a coverage threshold, a budget file.

Options, in rough order of cost:

1. Drop the `ref:` override on the jobs that judge the code, so `actions/checkout` uses the
   default `refs/pull/N/merge`. This is what the check is meant to be answering. Cost: the
   merge ref is recomputed by GitHub, so a run can be invalidated by a base push and needs a
   re-run to be trusted; jobs that legitimately need the head sha (anything reporting a
   status against a commit, or diffing against a base) must keep the override.
2. Require branches to be up to date with `dev` before merge (branch protection). Cost: a
   serialised merge queue in practice.
3. Keep head-sha checks and add a merge-queue run that judges the merged tree.

Worth a decision rather than a patch, because 1 and 2 have real costs and today's answer is
"`dev` finds out first".

aupyay:
Option 1 is implemented in #3323, scoped to the three `_core.yaml` jobs that judge the code
(`detect`, `build`, `Static`). Options 2 and 3 are untouched and still open — see below.

Confirmed in the run logs, which is the cleanest before/after this has:

```
#3310, pre-fix   core / Static (@yosemite-crew/desktop)
[...]

PR #3317, opened 2026-09-18
fix(desktop): type the idle-lock tests for the lib the suite compiles under

### LINKED ISSUE #3315: Desktop test arm fails type-check on dev: 8 errors in the idle-lock tests
`core / Static (@yosemite-crew/desktop)` is failing on `dev` at `0020fb937` with 8 type errors in the desktop test suite:

```
tests/e2e/idle-lock.e2e.ts(184,37): error TS2550: Property 'at' does not exist on type 'number[]'.
tests/e2e/idle-lock.e2e.ts(193,43): error TS2550: Property 'at' does not exist on type 'View[]'.
tests/e2e/idle-lock.e2e.ts(212,14): error TS2722: Cannot invoke an object which is possibly 'undefined'.
tests/e2e/idle-lock.e2e.ts(296,7):  error TS18048: 'win' is possibly 'undefined'.
tests/e2e/idle-lock.e2e.ts(297,7):  error TS18048: 'win' is possibly 'undefined'.
tests/e2e/idle-lock.e2e.ts(355,27): error TS18048: 'win' is possibly 'undefined'.
tests/e2e/idle-lock.e2e.ts(358,21): error TS18048: 'win' is possibly 'undefined'.
tests/idle-lock-overlay.test.ts(124,7): error TS2345: Argument of type 'number | undefined' is not assignable to parameter of type 'number | bigint'.
```

Reproduced locally with the same command CI runs, at the same commit:

```
git checkout 0020fb937
pnpm --filter "@yosemite-crew/desktop" run type-check   # exit 2, the same 8 errors
```

The desktop `lib` is `["ES2021", "DOM"]`, so `Array.prototype.at` is not available; the rest
are `noUncheckedIndexedAccess` on `BrowserWindow.getAllWindows()[0]`, an indexed method call
and a jest `invocationCallOrder[0]`.

### Why no PR check caught it

Neither contributing PR was wrong on its own. #3309 restated `exclude` in
`apps/desktop/tsconfig.test.json` so the test arm of `type-check` compiles the suite at all
(before it compiled 0 test files). #3310 branched before #3309 and added/edited these tests.
`core / Static` checks out `github.event.pull_request.head.sha`, not the merge ref, so on
#3310 the arm ran against a tree that still had no restated `exclude` — 0 test files, green
in 1m9s. The errors only exist in the merged tree, which no PR check compiled. The push run
on `dev` did catch it (run 35383477860), i.e. after merge.

The structural half of this is filed separately; this issue is the red gate on `dev`.

aupyay:
Closes #3315. `dev` is red: `core / Static (@yosemite-crew/desktop)` fails `type-check` at
`0020fb937` with 8 errors in the desktop test suite. This is the smallest change that clears
them without touching a strictness flag or the `lib` list.

### What was wrong

The desktop `lib` is `["ES2021", "DOM"]`, so `Array.prototype.at` is not in the program, and
`noUncheckedIndexedAccess` is on, so an indexed element is `T | undefined`.

| site | error | fix |
|---|---|---|
| `focused.at(-1)` | TS2550 | `focused[focused.length - 1]` |
| `children.at(-1)` | TS2550 | `views[views.length - 1]`, typed `| undefined` |
| `yc[m](...a)` | TS2722 | `yc[m]!(...a)` |
| `getAllWindows()[0]` ×2 | TS18048 | `getAllWindows()[0]!` |
| `invocationCallOrder[0]` | TS2345 | `invocationCallOrder[0]!` |

`.at(-1)` and `arr[arr.length - 1]` return the same value for a non-empty array and the same
`undefined` for an empty one, so no behaviour moves. The non-null assertions match the ones
the same file already makes on `wc!` and `app!`. Raising `lib` to ES2022 would clear the two
TS2550s, but it would also change the production arm, which is not what this red gate is about.

### Evidence, all at `0020fb937` + this commit

| command | before | after |
|---|---|---|
| `pnpm --filter "@yosemite-crew/desktop" run type-check` | exit 2, 8 errors | **exit 0**, 0 errors |
| `pnpm --filter "@yosemite-crew/desktop" run lint` | — | **exit 0** |
| `pnpm --filter "@yosemite-crew/desktop" run test` | — | **exit 0**, 76 suites / 1219 tests passed |

The "before" column is the same 8 errors CI reports on the `dev` push run `35383477860`,
character for character, and `type-check` at `e73046f47` (the commit before #3310 merged) is
exit 0 with 86 test files in the arm's program — so the errors arrived with the merge, not
with the arm.

The e2e file is not run locally; `Desktop E2E` on this PR is the check that exercises it.

### Why no PR check saw it

Neither contributing PR was wrong on its own. #3309 restated `exclude` in
`apps/desktop/tsconfig.test.json` so the test arm compiles the suite at all — before it
compiled **0** test files. #3310 branched before #3309 and added/edited these tests. Because
`core / Static` checks out `github.event.pull_request.head.sha` rather than the merge ref, the
arm on #3310 ran against a tree with no restated `exclude`: 0 test files, green in 1m9s. The
merged tree is the first tree in which these files are type-checked at all.

That hole is general — it applies to any gate that reads its own scope from the tree — and is
filed separately as #3316 with the options and their costs. This PR only clears the red.


ankit-yc (approved):
Approved at 1cd066e29b51d3f1000d4318f815e916866a8ecf. Verified independently; this unblocks `dev`, so it should go in ahead of the other open desktop PRs.

**Before/after, with the parent as the control.** This head's parent is `0020fb937`, which is the current `dev` tip, so the two states differ by this patch alone:

| tree | test arm | src arm |
|---|---|---|
| `0020fb937` (dev) | rc=2, 8 errors | rc=0 |
| `1cd066e2` (this head) | **rc=0, 0 errors** | rc=0 |

The 8 errors I measured on dev are the same 8 named in the body — the two `Array.prototype.at` calls against a `["ES2021", "DOM"]` lib, five possibly-undefined window/view reads, and the `number | undefined` argument in `idle-lock-overlay.test.ts`.

**Suite:** `jest --ci` at this head — 76 suites / 1219 tests pass, rc=0.

**Every rewrite is behaviour-preserving**, which matters because six of the seven changed lines are in a Playwright file that does not run locally:

- `focused.at(-1)` -> `focused[focused.length - 1]`: identical for a non-empty array, and both yield `undefined` on an empty one, so the `=== wc!.id` comparison cannot change verdict.
- `win?.contentView.children.at(-1)` -> `const views = win?.contentView.children ?? []; views[views.length - 1]`: same element, and the `| undefined` in the annotation is now honest about the no-window case that `?.` already allowed.
- `yc[m](...a)` -> `yc[m]!(...a)`, `getAllWindows()[0]` -> `[0]!`, `invocationCallOrder[0]` -> `[0]!`: `!` is erased at runtime. If the value were ever missing these still throw, so no assertion got weaker.

**No flag or lib touched**, which is the right call — raising `lib` to ES2022 to keep `.at` would have changed what the whole desktop program compiles against to satisfy two test lines.

Worth recording why this reached `dev` at all: #3310 merged alongside #3309, so it was reviewed and merged while the test arm still compiled zero test files and could not see it. Any branch cut before #3309 landed has the same exposure — I hit it on #3311 an hour ago for the same reason. A merge of current `dev` before merging is the check that catches it.

ankit-yc (approved):
Reproduced the red and the fix on dev head 0020fb937 in a clean worktree: pnpm --filter @yosemite-crew/desktop run type-check gave rc=2 with 8 errors before the patch and rc=0 with 0 after, same tree, same command. The 8 sites match the 8 CI reported one for one. Both rewrites are semantically identical for a non-empty array and for an empty one (a.at(-1) and a[a.length-1] are both undefined at length 0); the rest are non-null assertions in test code that change no runtime behaviour and cannot mask a failure - an undefined would still fail the assertion it feeds. Desktop E2E macOS passes here, which is also the second data point on the separate idle-lock e2e flake dev saw. LGTM.

PR #3319, opened 2026-09-18
fix(frontend): remove Harshvardhan Parmar from the About page core team cards

aupyay:
Removes Harshvardhan Parmar from the named core team cards on the About page ("the people who helped make it real"). The live GitHub contributor roster underneath is unchanged and still lists him, since it is built from GitHub rather than from this list.

- `About.tsx`: drops his `CORE_TEAM` entry. The grid is 4 tracks (2 on phones), so a single card sits in the first track with no gap to fill.
- `About.marketing.test.tsx`: moves him into the "departed members are not listed" checks (link and avatar absent).
- `About.stories.tsx`: the core-team grid expectation now lists Ankit only; the live-roster expectation keeps his GitHub card.

Checks: About test 9/9; putting the card back fails it (1 failed), restored byte-identical and green again; eslint clean on the three files.


ankit-yc (approved):
Approving. Owner-requested content change: Harshvardhan comes off the named core-team cards and stays in the live GitHub roster. The About test now asserts his card and avatar are absent (restoring the entry fails it), the story's core-team grid lists Ankit only while its live-roster expectation keeps his GitHub card, and CI is 47/47 including Storybook interactions.

PR #3321, opened 2026-09-18
test(desktop): bring a tab's pointer path up before trusting the click marker

### LINKED ISSUE #3318: Desktop: idle-lock e2e pre-lock pressKey marker is flaky on macOS
`tests/e2e/idle-lock.e2e.ts:233` (`takes focus, holds the workspace until unlock, then hands focus back`) failed on the dev push run for 0020fb937 on macos-latest, and passed on macos-latest for both the PR that introduced it (#3310) and the follow-up #3317, which is based on that same dev head.

## The failure

```
Error: expect(received).toBe(expected)
Expected: 1
Received: 0
Timeout 5000ms exceeded while waiting on the predicate
  at pressKey (apps/desktop/tests/e2e/idle-lock.e2e.ts:170)
  at apps/desktop/tests/e2e/idle-lock.e2e.ts:258
```

Run: https://github.com/YosemiteCrew/Yosemite-Crew/actions/runs/35383477330

Line 258 is the first `pressKey` on the tab under the lock. `pressKey` sends the key, then a click, then polls for the page's click count to increase — the click is the marker that says the page has finished handling the key. The poll timed out at 0, so the marker click never arrived within 5s; the lock assertion it guards never got to run.

## Why this reads as non-determinism rather than a regression

The only commits on dev that #3310's branch did not have are #3309, whose sole change to desktop runtime source is turning `interface TabManager` into `export interface TabManager`. Everything else it touched is test files and `tsconfig.test.json`. So the app binary the dev run exercised is materially the same one that passed on #3310 and on #3317.

## Suggested direction

The marker assumes a click sent immediately after a key is delivered and handled within the 5s default poll window while a modal overlay is being raised over the same window. That is a timing assumption about input delivery, not about the lock. Worth either raising the poll budget for this one assertion or replacing the click marker with something that does not depend on a second input event being delivered to a covered view.

Filed from the 2026-09-18 nightly QA pass. Not the same cause as #3314, which was the type-check red and is fixed by #3317.

aupyay:
Fixes #3318 — the idle-lock e2e failure that has had `Desktop E2E Required` red on dev since 0020fb937.

## What actually fails

Not the first `pressKey`. The trace from the failing run (`gh run download 35385721871 -n desktop-e2e-macos-latest`) carries a source line per action:

```
B 85422 Evaluate  :123    <- pressKey(firstTab) sends key + click
B 85442 Expect    :170 expected 1     <- PASSES, first try
B 85826 Expect    :258                <- /opened has loaded, keys == 0
B 85829 Evaluate  :123    <- pressKey(/opened) sends key + click
B 85832 Expect "poll toBe" :170 expected 1
  ERR x8, Received: 0 every time, 5s
```

Same run, same machine: the marker lands first try on the tab that was already open, and never lands on the tab created ~280ms earlier. `pageCount` returning `-1` then `0` on line 258 dates the load — the renderer was alive and `window.__clicks` existed. The mousedown is what did not arrive.

## Why

`newTab` calls `switchToTab`, so the view is attached, but it is seconds old and sits under a full-window overlay. A `mousedown` has to be hit-tested against the view's compositor data, which needs a layout and a paint; a `keydown` needs no hit test and goes to the focused document. The marker was therefore strictly more fragile than the keystroke it exists to guard — and on a slower machine it loses that race while the key does not. Nothing about it is a coin flip in the app; it is a paint that has or has not happened inside a 5s poll.

Raising the poll budget cannot help: the event was already dropped, and re-reading a count eight times will not bring it back.

## The change

One helper, used after a tab is opened or switched to (both re-attach a view): resend clicks until one is counted, then press. Clicks only, so the key counts it guards are untouched, and the `toBe(0)` / `toBe(1)` assertions are unchanged.

## Evidence

The failure reproduces locally once the dropped pointer event is modelled — the page swallows its first `mousedown` on `/opened` only:

| arm | tree | result |
|---|---|---|
| A | dev's spec + swallow first mousedown on `/opened` | **rc=1**, `Expected: 1 Received: 0`, `Timeout 5000ms`, `at pressKey (:170)` / `at :259` — byte-for-byte the CI failure |
| B | this branch + the same mutation | **rc=0**, 1 passed (45.2s) |
| C | this branch + `/opened` never counts a mousedown at all | **rc=1**, `no pointer input ever reached http://127.0.0.1:65277/opened`, `Timeout 20000ms`, `at waitForPointerInput (:184)` |

C is the must-fail arm: a pointer path that never comes up fails loudly with the gate's own message rather than letting `keysReceived === 0` pass vacuously.

Clean tree at 95691fb23 + this commit, in `apps/desktop`:

- `pnpm run type-check` rc=0 (both `tsc` passes)
- `pnpm run lint` rc=0
- `pnpm test` rc=0 — 76 suites, 1219 tests
- `pnpm exec playwright test` rc=0 — 47 passed, 1 skipped


ankit-yc (approved):
Approving at 06b360de5.

Checked the thing that would make this a weakened test rather than a fixed one: it isn't.

- The assertions are untouched. `keysReceived(...) === 0` under the lock and `=== 1` after unlock are the same lines they were.
- The warm-up sends `click()` only, never `keyPress`, so it cannot move the key counts it is there to protect.
- It fails rather than waits: `expect.poll` with a 20s bound and a named message, so a view that genuinely never takes pointer input still reds the spec.
- No straggler hazard in `pressKey`'s exact `toBe(clicks + 1)`. Input to one web contents is handled in order, so once the Nth warm-up click is counted, clicks 1..N-1 are already resolved (counted or hit-test-dropped) and cannot land after `pressKey` reads its baseline.

The two guard sites are the right two, and I checked the three `pressKey` calls you did not guard:

- L282 `newTab` and L293 `activateTab` are the fresh attaches — `activateTab` reaches `contentView.addChildView` in `ipc-handlers.ts`, so the "switching tabs re-attaches the view" comment is accurate.
- L278 and L350 are on `firstTab`, which is attached once at tab-mode entry and never re-attached by the lock: `layoutLockOverlay` and the overlay's `unmount` remove and re-add only `lockOverlayView`, not the tab view. So neither has the fresh-attach race, and L350 — the one asserting a key *does* arrive — is not a residual flake.

Merge after `Desktop E2E Required` lands; no failures on the PR as of this review, 5 checks still running.

PR #3255, opened 2026-09-17
fix(mobile): bump react-native-firebase/messaging to match app 24.1.1

aupyay:
## Summary
- Supersedes #3230. Dependabot bumped `@react-native-firebase/app` to 24.1.1 alone and left `@react-native-firebase/messaging` at 23.8.6; the two packages are lockstep-versioned and must match exactly, so `check-peer-deps.mjs` correctly reds the pair (`messaging@23.8.6 needs app@23.8.6, installed: app@24.1.1`).
- Bumps `messaging` to `^24.1.1` alongside `app` in `apps/mobileAppYC/package.json` and regenerates `pnpm-lock.yaml` off current `dev` (the original dependabot branch was 89h stale and carried unrelated lockfile drift).

## Test plan
- [x] `node scripts/ci/check-peer-deps.mjs` -> `peer dependencies: all satisfied`
- [x] `node scripts/ci/check-override-drift.mjs` -> no override contradicts a workspace declaration
- [x] `pnpm exec turbo run build --filter='./packages/*'` -> 8/8 successful
- [x] `pnpm --filter frontend run build` -> successful
- [x] `pnpm --filter mobileAppYC run type-check` -> clean
- [x] Root `pnpm run lint` and `pnpm run type-check` (pre-push hook) -> both green, pre-existing warnings only
- [x] Android native build: `./gradlew :app:assembleDebug` resolved and linked the new `react-native-firebase_app`/`react-native-firebase_messaging` 24.1.1 native modules through `writeDebugAarMetadata`; build stopped only at `processDebugGoogleServices` for a missing machine-local `google-services.json` (per-developer Firebase config, not present on this box and unrelated to this change)
- [ ] iOS native build not exercised - this box has Xcode Command Line Tools only, no full Xcode.app, so `pod install`/`xcodebuild` cannot run here

PR #3322, opened 2026-09-18
fix(mobile): pin the iOS pods to the Firebase 12.10.0 that messaging 24.1.1 requires

### LINKED ISSUE #3320: pod install fails on dev: Podfile.lock still pins Firebase/Messaging 12.8.0 after the 24.1.1 bump
## Symptom

`bundle exec pod install` in `apps/mobileAppYC/ios` exits 1 on `dev` (`95691fb23`):

```
[!] CocoaPods could not find compatible versions for pod "Firebase/Messaging":
  In snapshot (Podfile.lock):
    Firebase/Messaging (= 12.8.0)

  In Podfile:
    Firebase/Messaging

    RNFBMessaging (from `../node_modules/@react-native-firebase/messaging`) was resolved to 24.1.1, which depends on
      Firebase/Messaging (= 12.10.0)
```

## Cause

#3255 bumped `@react-native-firebase/app` and `@react-native-firebase/messaging` from `^23.5.0` to `^24.1.1`. `gh pr diff 3255 --name-only` lists exactly two files:

- `apps/mobileAppYC/package.json`
- `pnpm-lock.yaml`

`apps/mobileAppYC/ios/Podfile.lock` was not regenerated, so it still describes the 23.8.6 world:

```
Podfile.lock:36    - Firebase/Messaging (12.8.0):
Podfile.lock:65    - FirebaseMessaging (12.8.0):
Podfile.lock:3211  - RNFBApp (23.8.6):
Podfile.lock:3240  - RNFBMessaging (23.8.6):
Podfile.lock:3244    - Firebase/Messaging (= 12.8.0)
```

The JS side moved and the native side did not.

## Why this is not only a workstation problem

`mobile-release.yml` installs pods with `bundle exec pod install` and no `--repo-update`, inside a 5-attempt retry loop that resets `Pods`, `build` and the CocoaPods cache between attempts. Its own comment records the intent:

> Podfile.lock is deliberately left alone so resolution stays reproducible; everything derived from it is fair game.

Nothing that loop clears can move a constraint that comes from the lockfile snapshot, so this is deterministic rather than flaky: the next tagged release burns all five attempts and fails at the install step, before the archive.

The `mobile-gemfile-check.yml` gate added in #3265 does not cover this — it validates the Ruby bundle resolves and that `Gemfile.lock` is current, which it is. There is no equivalent check that `Podfile.lock` still matches the JS dependency tree.

## Fix

Regenerate `apps/mobileAppYC/ios/Podfile.lock` against `@react-native-firebase` 24.1.1 and commit it.

## Reproduction

From `apps/mobileAppYC` on `dev`, with Ruby >= 3.3 on PATH:

```
pnpm install --frozen-lockfile      # rc 0
bundle install                      # rc 0
cd ios && bundle exec pod install   # rc 1, output above
```

aupyay:
Closes #3320

## Problem

`bundle exec pod install` in `apps/mobileAppYC/ios` exits 1 on `dev` (`95691fb23`):

```
[!] CocoaPods could not find compatible versions for pod "Firebase/Messaging":
  In snapshot (Podfile.lock):
    Firebase/Messaging (= 12.8.0)

  In Podfile:
    Firebase/Messaging

    RNFBMessaging (from `../node_modules/@react-native-firebase/messaging`) was resolved to 24.1.1, which depends on
      Firebase/Messaging (= 12.10.0)
```

#3255 bumped `@react-native-firebase/app` and `@react-native-firebase/messaging` from `^23.5.0` to `^24.1.1`. It changed exactly two files — `apps/mobileAppYC/package.json` and `pnpm-lock.yaml` — and left `apps/mobileAppYC/ios/Podfile.lock` describing the 23.8.6 world. The JS side moved and the native side did not.

This is deterministic, not flaky. `mobile-release.yml` installs pods with `bundle exec pod install` and no `--repo-update`, in a five-attempt retry loop whose own comment reads "Podfile.lock is deliberately left alone so resolution stays reproducible; everything derived from it is fair game". Clearing `Pods`, `build` and the CocoaPods cache cannot move a constraint that comes from the lockfile snapshot, so the next tagged release would spend all five attempts on the same error and fail before the archive.

`mobile-gemfile-check.yml`, added in #3265, does not cover this. It checks that the Ruby bundle resolves and that `Gemfile.lock` is current — both of which are true here. Nothing checks that `Podfile.lock` still matches the JS dependency tree.

## Fix

`Podfile.lock` only, regenerated with `pod update` scoped to the Firebase family rather than a full re-resolution:

| pod | before | after |
|---|---|---|
| `Firebase` (CoreOnly, Messaging) and the `Firebase*` family | 12.8.0 | 12.10.0 |
| `GoogleDataTransport` | 10.1.0 | 10.1.1 |
| `GoogleUtilities` (and its subspecs) | 8.1.2 | 8.1.3 |
| `RNFBApp`, `RNFBMessaging` | 23.8.6 | 24.1.1 |

`Podfile` is untouched; `PODFILE CHECKSUM` is unchanged, which is the line that would move if the manifest had been edited.

Five pods also have a new `SPEC CHECKSUMS` entry with **no version change**: `ReactCodegen`, `RNGoogleSignin`, `RNReanimated`, `RNWorklets`, `react-native-blob-util`. None of them depends on any Firebase pod, so this is not a consequence of the pins above — the committed lock's values for them were already stale against the current `node_modules`. I did not chase why, and have not changed anything about those pods.

## Verification

All on this branch, `apps/mobileAppYC`, with Ruby 4.0.6 on `PATH` (the Gemfile needs >= 3.3):

Before, at `dev` `95691fb23` with the committed lock:

```
pnpm install --frozen-lockfile     rc 0
bundle install                     rc 0   Bundle complete! 9 Gemfile dependencies, 45 gems
cd ios && bundle exec pod install  rc 1   the error quoted above
```

After, with this lock and `Pods/` deleted first, so the install resolves from the lockfile rather than from an already-integrated tree:

```
rm -rf ios/Pods
cd ios && bundle exec pod install
  Pod installation complete! There are 133 dependencies from the Podfile and 163 total pods installed.
  rc 0
diff Podfile.lock <copy taken before the install>
  rc 0   (install is a no-op on the lock, so it is self-consistent)
```

The lockfile being unchanged by its own `pod install` is the part that matters: a lock that still needed resolving would rewrite itself here.

## Not covered

No simulator build or launch is included in this evidence. This PR is the install step only; the app build on top of these pods is the next thing to exercise and is not claimed here.


ankit-yc (approved):
LGTM

PR #3323, opened 2026-09-18
fix(ci): judge the merged tree in the core pipeline, not the branch tip

### LINKED ISSUE #3316: CI judges the PR head, not the merged tree, so a widened gate lands red on dev

aupyay:
Closes #3316.

## The defect

`detect`, `build` and `Static` in `.github/workflows/_core.yaml` all checked out
`github.event.pull_request.head.sha`, so lint, `type-check` and `build` judged the pull
request's branch tip — never the tree that merging it would produce. A branch that predates a
change to *what a gate reads* cannot be judged by that widened gate, and GitHub does not
re-run checks when the base advances.

## Reproduction, from the record

| | commit | job | result |
|---|---|---|---|
| #3309 widens the desktop test arm, merges 18:46:12Z | `e73046f473` | — | — |
| #3310 branched before it; CI at 18:52–18:53Z | head `f8aadd0b45` | `core / Static (@yosemite-crew/desktop)` | **success** |
| the merge of #3310 into dev, 19:00Z | `0020fb9370` | `core / Static (@yosemite-crew/desktop)` | **failure** (run 35383477860) |

Same job, same workspace, six minutes apart. The only difference is the tree. #3315 is what
that cost, and it was found by the push run on `dev` rather than by the pull request.

The widening itself is visible in `apps/desktop/tsconfig.test.json`: #3309 restated
`exclude`, because the inherited `exclude` from `./tsconfig.json` was filtering this file's
`include` and the arm was compiling zero test files.

## The fix

Drop the `ref:` override on the three jobs. `actions/checkout`'s default for a
`pull_request` event is the merge commit behind `refs/pull/N/merge` — the tree that would
exist on the base branch if this merged now, which is the question the check claims to
answer.

Every other trigger already resolved to `github.sha` through the `||` arm, so `push`,
`merge_group` and `workflow_dispatch` are byte-for-byte unchanged.

`scripts/ci/compute-base-sha.sh` needs no change. On a `pull_request` the base stays
`pull_request.base.sha` — the base branch tip at the time of the event, which is the merge
commit's first parent — so `git diff $BASE_SHA HEAD` is still precisely this pull request's
change and the affected matrix does not widen.

The invariant the fix rests on is `base.sha` == the merge commit's first parent — *not*
`base.sha` == the live `dev` tip, which is a different and weaker statement. Measured on three
open pull requests, including this one while `dev` was moving underneath it:

```
#3311  base.sha de74157c63   merge-ref parents: <merge> de74157c63 0152535347
#3313  base.sha de74157c63   merge-ref parents: <merge> de74157c63 31d0135207
#3323  base.sha 48ade8dec3   merge-ref parents: <merge> 48ade8dec3 d95a7454bf   (dev tip was
                                                                    already 155052a044)
```

The third row is the interesting one: `base.sha` had fallen behind the live `dev` tip and the
merge ref's first parent had fallen behind with it, in lockstep. That is the whole point — the
two are recomputed together, so `git diff $BASE_SHA HEAD` stays exact no matter how far the
base has since moved.

## What deliberately keeps the head sha

Anything that reports a status against a commit, or diffs against a base:

- `_sonar.yaml` — Sonar pull request analysis maps issues onto the pull request's own commits.
- `_test.yaml`'s added-line coverage gate — it diffs `BASE_SHA...HEAD_SHA` and matches the
  result against coverage line numbers. Moving the `test` job to the merged tree without
  re-basing that diff onto `HEAD^1...HEAD` would mismatch line numbers wherever the base and
  the branch touched the same file. That is a separate change with its own risk; called out
  below rather than smuggled in here.
- `secret-scan.yml`, `pr-governance.yml` commitlint, `story-coverage`, and the e2e workflows.

## Regression guard

`scripts/ci/core-checkout-ref.test.mjs` (runs in `pnpm run test:scripts`, which the `detect`
job runs on every event) fails if any checkout step in `_core.yaml` pins the head sha again.

Mutation-checked, four arms:

| arm | result |
|---|---|
| unmutated | green |
| head-sha `ref:` restored on `static` | red |
| head-sha `ref:` restored on `detect` | red |
| segmenter blinded (`uses:` + two spaces) | red |

The last arm is why the liveness control counts the action with a looser, independent needle:
the first version of the control shared the segmenter's anchor, so both counts fell together
and the blinded parser reported the gate as held.

## Not fixed here

The residual half of #3316: a run still does not refresh when the base advances *after* the
event. This change narrows the window from "the base moved any time since the branch forked"
to "the base moved since the last pull request event" — which is enough to have caught #3315,
whose event fired at 18:48:09Z, after #3309 merged at 18:46:12Z — but it does not close it.
Closing it is option 2 (`strict_required_status_checks_policy`) or option 3 (a merge queue),
both repository settings rather than a patch, and both with the serialisation cost the issue
names. Worth a separate decision.

## Verification

- `node --test scripts/ci/core-checkout-ref.test.mjs` — 1 pass, 0 fail; plus the four-arm
  mutation matrix above.
- `pnpm run test:scripts` — 369 pass, 4 fail. The four are `check-override-advisories`,
  `check-override-drift`, `check-peer-deps` and `check-release-upload-guards`, and they fail
  identically on unmodified `de74157c6` in the same worktree (`Cannot find module 'semver'` —
  a local dependency-resolution artefact of this checkout, not a code failure). The `dev` tip
  run 35391389034 is green, which is where `test:scripts` actually runs.
- The workflow parses: `yaml.parse` of `_core.yaml` yields `detect {fetch-depth: 0,
  persist-credentials: false}`, `build {persist-credentials: false}`, `static
  {persist-credentials: false}` — `ref` gone from all three, nothing else disturbed.
- This pull request touches `.github/workflows/`, so `detect` widens to `run_all` and the
  whole matrix exercises the change on its own merged tree.


ankit-yc (approved):
Approving at d95a7454b. This strengthens the gate rather than relaxing it, and I drove the guard's arms myself rather than taking the count on trust.

**The fix is correct, and correct *because* of the trigger.** `ci.yaml` fires on `pull_request: {}`, so dropping `ref:` makes checkout resolve its default — `refs/pull/N/merge`, the tree that would exist if this merged. Worth stating plainly for whoever edits this next: the same three-line deletion under `pull_request_target` would check out the **base branch** and every gate would then judge a tree containing none of the pull request's code — green on everything, silently. The trigger is load-bearing for this diff.

Blast radius is `pull_request` only. `push` (main/dev), `merge_group` and `workflow_dispatch` all previously fell through the `||` to `github.sha`, which is also what the default checkout resolves to on those events. `permissions: contents: read` and `persist-credentials: false` are retained, so running merged-tree code buys no token.

**Mutation arms, run on an isolated copy of the two files** (`node --test scripts/ci/core-checkout-ref.test.mjs`, baseline rc=0, 1 pass):

| arm | result |
|---|---|
| reintroduce `ref: ${{ github.event.pull_request.head.sha \|\| github.sha }}` on the detect checkout | rc=1 — *"a checkout step … pins the pull request head"* |
| that same expression present **only inside a comment** (2 occurrences) | rc=0 — stays green |
| a checkout in the inline `- uses:` form, which the segmenter's `^(\s*)uses:` anchor cannot see | rc=1 — *"segmented 3 checkout steps but … uses the action 4 times"* |
| no checkout steps at all | rc=1 — *"no checkout steps in …"* |
| restore | rc=0 |

The third arm is the one I'd have asked for if it weren't already here. Without the independent occurrence count, a reformat to the inline step form makes the segmenter return fewer steps and the loop passes over nothing — a guard that goes quiet exactly when the file is restructured. Counting the action a second way with a needle that shares only the action name closes that. The second arm matters for the same reason in the other direction: comment-stripping means the prose explaining this fix can neither satisfy the needle nor mask a real regression.

**The guard is wired, not inert.** `test:scripts` globs `scripts/ci/*.test.mjs` and the detect job runs `pnpm run test:scripts` — the job every event reaches. And it does not go blind on the regression it guards: if the pin came back, detect would check out the branch tip and read *that* tree's `_core.yaml`, which is the one carrying the pin, so it still fires.

Two residuals, neither blocking, and the first is the one you already flagged:

1. `pull_request.base.sha` is the base tip at event time; the merge ref's first parent is the base tip when GitHub last computed it. Normally identical. On a re-run against a stale payload after dev advances they can differ, and the divergence widens `git diff $BASE_SHA HEAD` — more workspaces tested, not fewer. Fail-safe, so I would not hold the PR for it.
2. I have **not** exercised what happens on a conflicted pull request, where `refs/pull/N/merge` is absent or stale. Flagging it as a question rather than a finding: branch protection will not merge a conflicted PR, and resolving pushes a commit that re-triggers CI, so I read the exposure as bounded — but I did not test it and am not claiming it.

Issue #3324, opened 2026-09-18
feat(mobile): lock the app with Face ID, Touch ID or fingerprint


aupyay:
## Summary

Let pet parents lock the mobile app behind the phone's own check: Face ID, Touch ID or Optic ID on iPhone, and fingerprint or the screen lock on Android. The app keeps no PIN of its own.

## Problem / motivation

Anyone holding an unlocked phone can open the app and read a pet's appointments, records, documents and messages. People who share a phone with family, or hand it over at the clinic desk, have asked for a way to keep the app closed without signing out each time.

## Proposed solution

- App lock is off by default. It is turned on in Preferences > Security.
- Once it is on, the app shows a lock screen on every cold start and after a chosen time away: Immediately, 1 minute (default), 5 minutes or 15 minutes.
- Unlocking reads a small marker from the keychain that the operating system only releases after Face ID, Touch ID, fingerprint or the device passcode.
- The lock is a cover drawn over the app, not a sign-out. Screens underneath stay mounted, so half-filled forms and navigation history survive.
- A native cover hides app content in the iOS app switcher, and Android 13 and later show a blank recents thumbnail.

## Scope

- Native module: privacy cover, a clock that keeps counting while the phone sleeps, and closing system sheets when the lock engages.
- Lock state, unlock rules and the keychain marker.
- Lock screen with ready, waiting, cancelled, failed, too-many-attempts and sign-in-again states.
- The lock stays with its owner across sign-out and sign-in.
- Preferences > Security: on/off switch and the time-away setting.
- Siri and Shortcuts answers pause while app lock is on.
- Detox and device QA.

### Out of scope

- Encrypting data at rest.
- An app-specific PIN.
- Desktop and web unlock (tracked separately).

## Current state / where it plugs in

- `react-native-keychain` 10 is already installed and holds the session item; the app-lock marker is a separate item.
- The native module follows the existing Assistant bridge pattern on iOS and Android.
- Redux Persist gains an `appLock` slice with a migration.
- No backend changes and no new dependencies.

## Acceptance criteria

- [ ] With app lock on, a cold start and a return after the time away both show the lock screen before any content.
- [ ] Only a successful operating-system check unlocks; cancel, failure and lockout keep the app locked.
- [ ] Changing the phone clock does not skip the time away.
- [ ] The app switcher and Android recents show no app content while app lock is on.
- [ ] Turning app lock off, or raising the time away, needs a successful unlock first.
[...]

PR #3326, opened 2026-09-18
fix(mobile): require an unlocked device for Siri and Shortcuts answers

aupyay:
## PR Checklist

The Siri and Shortcuts App Intents in `apps/mobileAppYC/ios/mobileAppYC/Assistant/Intents` do not set
`authenticationPolicy`, so they use the App Intents framework default.

## What is the new behavior?

Every App Intent declares `static var authenticationPolicy: IntentAuthenticationPolicy = .requiresAuthentication`:

- Read intents (`NextAppointmentIntent`, `VaccinationStatusIntent`, `UpcomingTasksIntent`) answer only
  once the device is unlocked.
- Hand-off intents (`AddCareTaskIntent`, `BookAppointmentIntent`, `LogExpenseIntent`) require the
  unlock before they run. They open the app, which needs the unlocked device anyway.

Why `.requiresAuthentication` and not `.requiresLocalDeviceAuthentication`: Apple documents the first
as requiring authentication before the intent runs, where any one unlocked device in the request is
enough (for example an unlocked Apple Watch). The second forces an unlock of the device running the
intent even when the request came from an authenticated watch, and is meant for data that is only
readable while that device is unlocked. These intents do not need that.

The property is available from iOS 16.0, the same floor as the `@available(iOS 16.0, *)` intents, so
the app's iOS 15.1 deployment target is unaffected.

New guard test `apps/mobileAppYC/__tests__/guards/appIntentsRequireUnlock.test.ts` fails when any
intent under `ios/mobileAppYC` has no unlock policy. That matters because a misspelled property name
compiles cleanly and silently leaves the default in place.

Android needs no change: the app declares no App Actions (`shortcuts.xml`/capabilities), no widgets,
no receivers and no `showWhenLocked` activity. Its dynamic launcher shortcuts carry generic action
labels and open `MainActivity`, which needs the device unlocked.

### Verification

- iOS simulator build, Xcode 27.0 (27A266a), `xcodebuild -workspace mobileAppYC.xcworkspace -scheme
  mobileAppYC -configuration Debug -sdk iphonesimulator -destination 'generic/platform=iOS Simulator'
  ... build`: `** BUILD SUCCEEDED **`, with no diagnostics in the Assistant sources.
  - Xcode 27 rejects the pod resource-bundle targets that still declare iOS 9.0-13.0 (for example
    `SDWebImage-SDWebImage`, `StripeCore-StripeCoreBundle`), so that build passed
    `IPHONEOS_DEPLOYMENT_TARGET=15.1` on the command line. CI pins Xcode 26.3 and is unaffected.
- Compiled App Intents metadata (`mobileAppYC.app/Metadata.appintents/extract.actionsdata`): all six
  actions record `authenticationPolicy: 1` on this branch and `0` on `dev` sources, rebuilt the same way.
- Jest: `npx jest __tests__/guards/appIntentsRequireUnlock.test.ts` 3 passed. It fails (exit 1) when
  given the `dev` intent sources, a misspelled property, or one intent reset to `.alwaysAllowed`.
- `pnpm --filter mobileAppYC run type-check` exit 0. ESLint and Prettier are clean on the new test file.

## Related Issue(s)

Related to #3324 (mobile app lock). This change stands on its own and does not depend on it.


ankit-yc (approved):
Reviewed: all six App Intents gate on an unlocked device, the compiled intent metadata confirms it, and the guard test fails when any intent loses the policy.

PR #3328, opened 2026-09-18
fix(mobile): hold react at the version React Native's renderer is built on

### LINKED ISSUE #3325: iOS release app crashes on launch: react 19.3.0 against React Native 0.81.6's 19.1.4 renderer
## Symptom

The iOS Release build off `dev` (`95691fb23`) installs and launches, then terminates before the first screen:

```
*** Terminating app due to uncaught exception 'RCTFatalException: Unhandled JS Exception:
Error: Incompatible React versions: The "react" and "react-native-renderer" packages must have
the exact same version. Instead got:
  - react:                  19.3.0
```

Reproduced on an iPhone 17 simulator (iOS 26.5), Release configuration with the embedded `main.jsbundle`. Firebase initialises first and does not throw — the crash is after `FirebaseApp.configure()`, in the JS bundle, at `findNodeHandle` during the first `__attach`.

## Cause

`aa113d220` (#3253, "bump the minor-and-patch group across 1 directory with 32 updates") moved `apps/mobileAppYC/package.json`:

```
-    "react": "19.1.4",
+    "react": "19.3.0",
```

`react-native@0.81.6` declares `peerDependencies.react: ^19.1.4`, so 19.3.0 satisfies the range and nothing refused the bump. But the renderer React Native 0.81.6 ships is built against React **19.1.4**, and React's own runtime guard requires an **exact** version match, not a semver-compatible one. The caret in the peer range is wider than the constraint the runtime enforces.

`react-native` itself stays resolved against 19.1.4 in the store, so the tree holds both:

```
node_modules/.pnpm/react-native@0.81.6_..._react@19.1.4
node_modules/.pnpm/react@19.3.0
```

and the app resolves `react` to 19.3.0.

## Why nothing caught it

- Jest runs against the same mismatched pair but never mounts through the native renderer, so the guard does not fire.
- `pnpm install --frozen-lockfile` is satisfied: the peer range is not violated.
- The release workflow builds and uploads; it does not launch the built app.

The first place it can surface is a device or simulator launch, which is why a dependency-group bump reached `dev` without it.

## Fix

Hold `react` at the version React Native 0.81.6's renderer is built against (19.1.4) until React Native itself moves, and keep the dependency-group bump from raising it again — the same shape as the `next` hold in #3272.

## Reproduction

```
pnpm install --frozen-lockfile
cd apps/mobileAppYC/ios && bundle exec pod install     # needs #3320 / #3322 first
xcodebuild -workspace mobileAppYC.xcworkspace -scheme mobileAppYC \
  -configuration Release -sdk iphonesimulator -derivedDataPath <dd> \
  CODE_SIGNING_ALLOWED=NO IPHONEOS_DEPLOYMENT_TARGET=15.1 build
simctl install <udid> <dd>/Build/Products/Release-iphonesimulator/mobileAppYC.app
simctl launch --console-pty <udid> com.yosemitecrew.mobile
```

The app needs a valid `ios/GoogleService-Info.plist`; with the placeholder from `config-templates/ios/` it dies earlier, in `+[FIRApp configure]`, and never reaches this.

aupyay:
Closes #3325

## Problem

The iOS Release build off `dev` installs and launches, then terminates before its first screen:

```
*** Terminating app due to uncaught exception 'RCTFatalException: Unhandled JS Exception:
Error: Incompatible React versions: The "react" and "react-native-renderer" packages must have
the exact same version. Instead got:
  - react:                  19.3.0
```

`aa113d220` (#3253, a 32-package minor-and-patch group) raised `react` and `react-test-renderer` from 19.1.4 to 19.3.0 in `apps/mobileAppYC/package.json`.

`react-native@0.81.6` declares `peerDependencies.react: ^19.1.4`. 19.3.0 satisfies that range, so nothing refused the bump - but React's runtime guard compares `react` against `react-native-renderer` for an **exact** match, and the renderer React Native 0.81.6 ships is built on 19.1.4. The caret in the peer range is wider than the constraint the runtime enforces, and that gap is the whole bug.

### Why no CI stage could see it

- jest passes: 509 suites / 8768 tests, and none of them mounts through the native renderer, so the guard never runs.
- `pnpm install --frozen-lockfile` passes: the peer range is not violated.
- `mobile-release.yml` builds and uploads the app. It does not launch it.

The first place this can surface is a device or simulator launch, which is why a dependency-group bump reached `dev` with the app unable to start.

## Fix

Both packages back to 19.1.4, plus a dependabot hold so the group cannot raise them again.

- `apps/mobileAppYC/package.json`: `react` and `react-test-renderer` to `19.1.4`, `@types/react-test-renderer` to `^19.1.0`
- `.github/dependabot.yml`: freeze `react`, `react-dom` and `react-test-renderer` for **every** update type. See "Revised after review" below - the first version of this PR held minors and majors only, and that is not sufficient.
- `apps/mobileAppYC/__tests__/config/reactRendererVersion.test.ts`: make the comparison the renderer makes at runtime, in jest.

The second commit carries three `SPEC CHECKSUMS` entries in `apps/mobileAppYC/ios/Podfile.lock` - `ReactCodegen`, `RNReanimated`, `RNWorklets`. pnpm lays reanimated and worklets out under a store directory whose name embeds the react version they resolved against, so the podspec CocoaPods reads changes with the pin; ReactCodegen is generated from the same tree. No pod versions change and the `Podfile` is untouched. Without it, the next person's `pod install` rewrites the lock under them.

## Verification

Everything below on this branch.

`apps/mobileAppYC`:

```
pnpm install --frozen-lockfile                                 rc 0
node -p "require('react/package.json').version"                19.1.4
node -p "require('react-test-renderer/package.json').version"  19.1.4
pnpm test                                                      rc 0   509 suites, 8768 tests, all passed
pnpm run type-check                                            rc 0   tsc --noEmit, no output
```

`apps/mobileAppYC/ios`, with Ruby 4.0.6 on PATH:

```
bundle exec pod install                    rc 0   133 dependencies, 163 total pods
bundle exec pod install   (second run)     rc 0   Podfile.lock byte-identical (diff rc 0)
```

Built and run on an iPhone 17 simulator, iOS 26.5, Release configuration with the embedded `main.jsbundle`:

```
xcodebuild ... -configuration Release -sdk iphonesimulator CODE_SIGNING_ALLOWED=NO   ** BUILD SUCCEEDED **
simctl install <udid> mobileAppYC.app                                                rc 0
simctl launch --console-pty <udid> com.yosemitecrew.mobile
```

**Before** (same command, same simulator, `react` at 19.3.0): the app terminates with the exception quoted above and the device returns to the home screen.

**After**: the app stays up, the onboarding carousel renders, and the system notification permission alert is presented. The process is still alive 30s after launch and the console log contains no `Terminating` line.

Screenshots of both arms are in the QA note rather than the PR.

## Not covered

- Granting the notification permission and reading back an FCM registration token. Tapping the system alert needs Accessibility automation for System Events, which cannot be granted in a non-interactive session.
- Android. This pin is not platform-specific, but only iOS was exercised.

## Note for anyone building iOS locally

`xcodebuild` on Xcode 27 rejects 23 CocoaPods **resource-bundle** targets whose `IPHONEOS_DEPLOYMENT_TARGET` is below the 15.0 floor (9.0 through 13.4, from their podspecs). `apps/mobileAppYC/ios/Podfile`'s `post_install` sets several build settings on every target but not that one, and `react_native_post_install` does not reach these generated targets. Passing `IPHONEOS_DEPLOYMENT_TARGET=15.1` on the xcodebuild command line clears all 23; the builds above used that override and no source change. CI runners are on Xcode 26, where the floor is lower, so this is latent rather than live - worth its own issue when the runner image moves.

## Revised after review

Two findings on the first version, both about the dependabot hold. The hold has been replaced rather than extended.

**A patch is as fatal as a minor.** The guard is not a semver range check. In `react-native@0.81.6`, `Libraries/Renderer/implementations/ReactNativeRenderer-{prod,dev,profiling}.js` all carry:

```js
var isomorphicReactPackageVersion = React.version;
if ("19.1.4" !== isomorphicReactPackageVersion)
  throw Error('Incompatible React versions: The "react" and "react-native-renderer" packages must have the exact same version...')
```

A literal string compare against a version baked into the shipped bundle. `react` 19.1.5 throws exactly as 19.3.0 did, so "patches keep flowing" was the wrong scope and the entries now carry no `update-types` at all. (The Fabric renderer builds do not carry the guard; only the three `ReactNativeRenderer-*` ones do.)

**`react-dom` is frozen alongside `react`.** There is one npm config for the whole pnpm workspace and ignores match on `dependency-name` only, so holding `react` also holds `apps/frontend`, which is on `^19.1.1`. `react-dom` carries the same exact-version guard against `react`; leaving it updatable while `react` is held is the one configuration that guarantees the pair drifts apart.

Ignore rules apply to dependabot **security** updates too, so a `react` advisory will not arrive as a PR while these entries stand. That is stated in the config comment. The mobile app cannot take a `react` bump without a `react-native` bump anyway, and the frontend lift is a manual PR raising `react` and `react-dom` together.

### A dependabot ignore is policy, not a control

`react-test-renderer` has been fully frozen in `dependabot.yml` since 2026-08-09 and still reached 19.3.0 on `dev` - by hand, in `e98197f30`, to match the `react` that #3253 raised. Config cannot stop a commit, so the new test makes the comparison the renderer makes at runtime, reading the same files a device loads:

- each `ReactNativeRenderer-*` build's baked-in version must equal the installed `react`
- the manifest must pin `react` and `react-test-renderer` to a single version, not a range
- if no renderer build carrying the guard is found, the suite **fails** - a moved directory or a renamed guard reads as an unread file, never as agreement

Mutation runs, each restored afterwards:

| mutation | result |
| --- | --- |
| `"react": "19.1.4"` -> `"^19.1.4"` in the manifest | 1 failed - `Expected: "19.1.4" / Received: "^19.1.4"` |
| renderer copies rewritten to `"19.3.0" !== isomorphicReactPackageVersion` | 3 failed, one per guarded build - `Expected: "19.1.4" / Received: "19.3.0"` |
| renderer directory pointed at an empty dir | 2 failed - `Expected: > 0 / Received: 0` |

`apps/mobileAppYC`: `npx jest --ci` -> rc 0, **510 suites / 8773 tests** passed (509/8768 before this commit). `npx tsc --noEmit` -> rc 0. Repo `lint` and `type-check` ran green in the pre-push hook.

### Not fixed here

`.github/dependabot.yml` is read from the **default branch**, which is `main`. `dev` already carries one hold `main` does not - the `next` hold from #3272, merged today - so that one is inert, and this one will be too until the file is mirrored. A `main`-targeted PR syncing the file follows once this merges; the pin in `package.json` and the new test do not depend on it.


ankit-yc (dismissed):
Reviewed at head `2e923a603da66f8588372081977c3a69e335c792`. I did not write this patch.

The fix itself is right and the evidence is the strongest kind available here: the Release app crashed before and renders after, on a real simulator. `react` and `react-test-renderer` move together to 19.1.4, the pin stays scoped to `apps/mobileAppYC` so `apps/frontend` keeps 19.3.0, and the three pod checksums are the ones that follow a React version change. Approving on that basis.

Two findings in the dependabot hold, both the same shape as the bug being fixed — a pair that must move together with only one side held. Neither is a defect in the current tree, so neither blocks the release fix.

### 1. Letting `react` patches flow re-arms #2096

The new entry holds `react` at minor and major only, and the comment says "Scoped to minor and major so patches keep flowing."

There is already an entry above it, at the `react-test-renderer` hold:

> react-test-renderer is fully frozen: @testing-library/react-native requires it to EXACTLY equal the react version RN pins, so even the 19.1.0 -> 19.1.9 PATCH failed every mobile suite (#2096).

That entry carries no `update-types`, so `react-test-renderer` is frozen against every update type including patch. This PR sets both packages to 19.1.4 and then leaves `react` free to take a patch. When React ships 19.1.5, Dependabot may raise `react` while `react-test-renderer` stays at 19.1.4, the exact-equality requirement that entry documents is violated, and the mobile suites go red the way #2096 records.

It would self-announce in CI rather than reach a release, which is why I am not blocking on it. But the cheap form is to drop `update-types` from the new `react` entry so it matches its partner's full freeze, and to reword the "patches keep flowing" sentence, which as written contradicts the entry twenty lines above it.

### 2. `react-dom` is not held, and the hold is workspace-wide

`react-dom` appears nowhere in `.github/dependabot.yml`.

There is a single `package-ecosystem: npm` entry, at `directory: /`, and its `ignore:` list is what this PR extends. That entry reaches every workspace manifest, not just the root: `aa113d220` — the bump that caused this bug — changed seven `package.json` files from it, including `apps/frontend/package.json` and `apps/mobileAppYC/package.json`.

So holding `react` also holds it for `apps/frontend`, which has no React Native renderer and no reason to be frozen. `react-dom` there stays free. `apps/frontend` currently carries `react: ^19.3.0` and `react-dom: ^19.3.0`; the reachable end state is `react` held at 19.3.x while `react-dom` moves to 19.4.x, and React requires those two to match.

Worth either adding `react-dom` alongside, or scoping the react hold so it does not silently freeze the web app. A note in the comment that the hold is workspace-wide would also help whoever lifts it later — the current comment reads as though it only concerns React Native.

### Smaller

- The new `react-test-renderer` entry at the bottom of the ignore list is a strict subset of the full freeze already present higher up, so it has no effect. Removing it keeps one rule per dependency and avoids the two comments disagreeing about whether patches flow.
- `@types/react` stays at `^19.3.0` in `apps/mobileAppYC/package.json`, with a root override pinning it to `19.3.0`, while the runtime drops to 19.1.4. Types only, and the override is shared with `apps/frontend` which legitimately runs 19.3.0, so lowering it is not obviously right — but it does mean `tsc` will accept a React API in mobile code that is not there at runtime.

aupyay:
Both review findings landed, and the first one is worse than it looked: the guard is a literal string compare in the shipped renderer bundle (`if ("19.1.4" !== isomorphicReactPackageVersion) throw`), so a patch is as fatal as a minor and `react` 19.1.5 would have crashed the app the same way. No `update-types` subset is safe - `react`, `react-dom` and `react-test-renderer` are now frozen outright, with `react-dom` held because the ignore is workspace-wide and the pair would otherwise drift in `apps/frontend`.

`react-test-renderer` was already fully frozen and still reached 19.3.0 on `dev` by hand (`e98197f30`), so the config alone is not a control. Added `apps/mobileAppYC/__tests__/config/reactRendererVersion.test.ts`, which makes the renderer's own comparison in jest and fails loudly if it cannot find the guard. Three mutation arms, full suite and type-check results are in the PR body.

Approval was dismissed by the push, so this needs a fresh one. Separately: `dev` and `main` have diverged in `dependabot.yml` again (the `next` hold from #3272), so both that hold and this one are inert until the file is mirrored - `main` PR to follow.

ankit-yc (changes requested):
Re-reviewed at `86c48dc824b540a811784b764f96605ac171d690`. Both findings from the previous round are addressed well — the renderer source quote is better evidence than my argument was, and the security-update consequence is called out rather than discovered later. The test is the part I want to keep; see below.

Requesting changes for one thing, which is mechanical.

## Blocker: the config is now invalid, so none of the ignores apply

The `.github/dependabot.yml` check is **failing** on this head with "Your .github/dependabot.yml contained invalid details". It has no row at all on the previous head `2e923a603` or on `dev`, so this commit introduced it.

`react-test-renderer` now appears twice in the same ignore list, and the two entries are byte-identical:

```
 93:      - dependency-name: 'react-test-renderer'
230:      - dependency-name: 'react'
231:      - dependency-name: 'react-dom'
232:      - dependency-name: 'react-test-renderer'
```

It is the only duplicated `dependency-name` in the file:

```
$ grep -oE "dependency-name: '[^']+'" .github/dependabot.yml | sort | uniq -d
dependency-name: 'react-test-renderer'
```

Last round the two differed — line 93 bare, the new one carrying `update-types` — so they were not duplicates and the file validated. Dropping `update-types` to make the freeze total made them identical, and the validator rejects that.

This matters more than a lint failure: an invalid `dependabot.yml` is not partially applied. Dependabot stops processing the file, so every ignore in it stops holding — including the react freeze this PR exists to install, and the `react-native` and `posthog-js` holds that predate it. The PR would land reading as a tightening and act as a repo-wide release of all holds.

Fix is to delete line 232. Line 93 already freezes `react-test-renderer` for every update type, with the #2096 history attached, so the new block only needs `react` and `react-dom`. Worth a line in the new comment pointing at line 93 so the next reader does not re-add it.

## The test earns its place — I drove it

I did not take the green suite on trust, and the check I care about is whether it can fail. It can, on the real defect rather than an arranged one.

The shared checkout still has the shipped broken state installed: `react` 19.3.0 against `react-native` 0.81.6, whose renderer is built on 19.1.4. Running this test file unchanged against that tree:

```
Tests:       3 failed, 2 passed, 5 total

● ReactNativeRenderer-prod.js is built on the installed react
    Expected: "19.3.0"
    Received: "19.1.4"
```

The three failures are `ReactNativeRenderer-{dev,prod,profiling}.js`. The two passes are the liveness case and the manifest-pin case — which is the discrimination I wanted: the suite was not failing by finding nothing, because the case asserting it found guarded files passed in the same run. Reading the literals straight out of the installed renderer agrees:

```
ReactFabric-dev.js          -> []
ReactFabric-prod.js         -> []
ReactFabric-profiling.js    -> []
ReactNativeRenderer-dev.js  -> [19.1.4]
ReactNativeRenderer-prod.js -> [19.1.4]
ReactNativeRenderer-profiling.js -> [19.1.4]
```

So the guard sits in three of six implementation files and the Fabric trio carries none. `guarded` filtering to the files that actually contain it is right, and `expect(guarded.length).toBeGreaterThan(0)` is what stops a rename or a moved directory from reading as agreement. That liveness case is the reason I am comfortable with this test, and it should not be dropped if the file gets tidied later.

One observation rather than a request: because the Fabric implementations carry no version guard, this test says nothing about a bundle that loads only those. It did not need to here — the crash came through `ReactNativeRenderer-prod.js` — but if the app moves fully to Fabric the guard this test reads may stop being the one that fires, without the test changing colour. A sentence in the file header would save someone that discovery later.

Everything else — the three pod checksums, the scoping to `apps/mobileAppYC`, the workspace-wide note, and freezing `react-dom` alongside `react` — looks right to me.

ankit-yc (approved):
Approving at `fe7bc615dedf195eeb9726d16611647e2c3f4328`.

The blocker is fixed and I checked it at the store rather than from the diff:

```
$ grep -oE "dependency-name: '[^']+'" .github/dependabot.yml | sort | uniq -d
(no output)

 93:      - dependency-name: 'react-test-renderer'
233:      - dependency-name: 'react'
234:      - dependency-name: 'react-dom'
```

`.github/dependabot.yml` check: **success**, completed 21:30:27Z at this head, titled "Dependabot config file validation" — against **failure** at `86c48dc82` and **no row at all** at `2e923a603` and on `dev`. The diff from the blocked head is the single deletion plus the comment, and nothing else moved. The comment naming the failure mode is the right place for it; that is what stops the next person re-adding the line, since the duplicate reads as an omission when you are looking at line 233.

## The freeze is inert on merge, and the exposure window is real

Confirming the point raised separately, because it changes what merging this accomplishes. Dependabot reads its config from the default branch, and this repo's default branch is `main`. Against `origin/main`:

```
react                ABSENT
react-dom            ABSENT
react-test-renderer  PRESENT
next                 ABSENT
```

So on merge to `dev` the new freeze does nothing. `react-test-renderer` is held today only because line 93 predates the split; the `next` hold from #3266/#3272 is inert for the same reason, and `main`'s copy of the file is three commits behind `dev`.

That is not a reason to hold this PR — the pin, the pod checksums and the test are all live on merge, and the config half cannot be fixed from this branch. It is a reason the follow-up main PR is load-bearing rather than tidy-up, and worth saying in its body that it carries three holds, not one.

One thing I did not verify, so treating it as an open question rather than reassurance: `apps/mobileAppYC/package.json` pins `react` exactly, so nothing but a Dependabot PR can move it during the window, and such a PR would be expected to run the new test and go red. Whether a manifest-only change actually selects the mobile jest suite depends on the affected-workspace detection, which was still in progress here, so I could not confirm the backstop fires. If it does not, the window is genuinely uncovered until `main` is synced.

The test, the renderer quote and the security-update trade-off I reviewed last round and still think are right.

PR #3331, opened 2026-09-18
fix(ci): restart the dev API by pm2 process name, not id

aupyay:
## PR Checklist

The dev target in `deploy-api.yml` restarts the API by pm2 id `1`, while production is already addressed by name (`yosemite-backend`, line 169).

The dev box now has a pm2 systemd unit (#2866). At boot it runs `pm2 resurrect`, and resurrect numbers processes from 0. When the unit first started the API, it came back as id 0, so `api-deploy.sh` would have failed at `pm2 restart "$PM2_TARGET"` (scripts/deploy/api-deploy.sh:330). The process was put back on id 1 by hand, but any reboot will renumber it again.

## What is the new behavior?

The dev target is the process name `yosemite-temp`, the same way production works. `pm2 id yosemite-temp` on the box resolves to exactly one process. A name does not change when pm2 renumbers, so a reboot no longer breaks the next deploy.

`api-deploy.sh` already accepts a name or an id (line 47), so the script is unchanged.

Checked: `prettier --check` and `actionlint` report nothing new for this file (the two SC2129 style notes are in the base file too), and `git-sync.test.sh`, `migrate.test.sh` and `controls.test.sh` all pass.

## Related Issue(s)

Refs #2866


ankit-yc (approved):
Single-line change to the dev arm of the deploy target resolver, matching the shape production already uses. The downstream consumer takes a name or an id (api-deploy.sh:47), and a name that does not exist fails loudly under set -euo pipefail at pm2 restart rather than silently restarting the wrong process. The workflow is workflow_dispatch only, so merging does not deploy. The process name itself is the one fact not checkable from the tree; the failure mode if it is wrong is a visible red step on a hand-triggered run. All 40 checks green at b65366d.

PR #3332, opened 2026-09-18
chore(repo): sync the dependabot config to the default branch

aupyay:
## Problem

Dependabot reads `.github/dependabot.yml` from the **default branch**, which is `main`. `target-branch: dev` only aims the PRs it opens at `dev`. The note at the top of the file says so, and asks that the two copies be kept byte-identical - it was added by #2079, which fixed a month of silently inert config changes.

`dev` has drifted ahead again. Every hold added there since is inert:

| hold | added by | state on `main` |
| --- | --- | --- |
| `next`, held on 15.x while Next 16 adoption is decided | #3272, merged today | absent |
| `react`, frozen for every update type | #3328 | absent |
| `react-dom`, frozen alongside `react` | #3328 | absent |

The `react` freeze is the one that matters most: `react-native@0.81.6`'s renderer compares `React.version` against a baked-in literal, so any bump - patch included - terminates the iOS app before its first screen (#3325). Without this sync, the next weekly group bump can raise `react` again.

## Change

`.github/dependabot.yml` copied verbatim from `dev`. Nothing else.

```
$ git diff --stat origin/main:.github/dependabot.yml <dev>:.github/dependabot.yml
 .github/dependabot.yml | 84 ++++++++++++++++++++++++++++++++++++++++++++++++++
 1 file changed, 84 insertions(+)
```

Additions only - no existing hold is changed or removed, and the two copies are byte-identical afterwards.

## Verification

- `yaml.safe_load` parses it: 3 update configs, 23 npm ignore entries, **no duplicate `dependency-name`** in any config. A duplicate fails the `.github/dependabot.yml` check and releases every hold in the file, which is how it failed once on #3328.
- The three holds land at lines 233 (`react`), 234 (`react-dom`) and 276 (`next`).
- GitHub's own `.github/dependabot.yml` check runs on this PR and is the authority on validity.

Opened while #3328 is green-pending on `dev`; the file here is that PR's approved head. I will confirm byte-identity against `dev` before this merges.


aupyay:
Applied `main-direct`. Stating the reasoning so it can be overridden rather than discovered.

`promotion-guard.yml` names this case in the comment beside the escape hatch: "a deliberate, human-applied escape hatch for the rare direct-to-main change (an emergency hotfix, or a dependabot.yml sync that has to take effect before the next promotion)".

It has to take effect before the next promotion because promotions here are rare and the next Dependabot run is not: `main` last moved on 2026-09-10, the last `dev` to `main` promotion was 2026-08-31, and the schedule in this file is `weekly, monday, 05:00 UTC`. Waiting for a promotion leaves the `react` freeze inert across at least one Monday run, and a group bump raising `react` again reships the crash in #3325 - which took a full mobile QA pass on a simulator to find, because no CI stage can see it.

The diff is additions only to `.github/dependabot.yml`, no other file, and `promotion-guard/dev-contains-main` is already `success`. If you would rather this waited for the promotion, drop the label and I will close it - the pin in `apps/mobileAppYC/package.json` and the new jest guard are on `dev` either way, and they are what actually stop the crash. The config only stops the PR being opened.

ankit-yc (approved):
Approving at `52ede29c7b8d9be628615f0a983a25adcbb0a771`. The label is right — keeping it.

## The label is the documented case, not a stretch

`promotion-guard.yml` names this exact use in the comment above the hatch:

> A deliberate, human-applied escape hatch for the rare direct-to-main change (an emergency hotfix, **or a dependabot.yml sync that has to take effect before the next promotion**).

and the check reads the same string this PR carries, `main-direct`, at `[[ ",${LABELS}," == *",main-direct,"* ]]`. The bot check sits above it, so the hatch cannot be applied to a bot PR. `Only dev may merge into main` and `promotion-guard/dev-contains-main` are both green here.

The timing justification holds too. The npm entry is `interval: weekly, day: monday, time: '05:00', timezone: UTC`; it is Friday 21:38Z as I write this, so the next run is Monday 05:00Z. Last promotion was 2026-08-31. Waiting for a promotion would miss it, which is the condition the hatch exists for.

## I did not take the clean check list as evidence the config is valid

Worth flagging because it nearly caught me: `.github/dependabot.yml` has **no row at all** on this head. On #3328 the same check was absent at `2e923a603`, red at `86c48dc82`, green at `fe7bc615d` — so it appears to post only once it has something to say. An absent validator reads exactly like a passing one, and here it is the check that matters most, because this is the branch Dependabot actually reads.

So the evidence is object identity instead. This PR's file is the **same blob** as the file the validator marked success on #3328:

```
$ git rev-parse fe7bc615d:.github/dependabot.yml pr3332:.github/dependabot.yml origin/dev:.github/dependabot.yml
a85d2641bf88ebdf7a596869862a2113e0a33361   <- #3328 head, validator success
a85d2641bf88ebdf7a596869862a2113e0a33361   <- this PR
3c8e8599d466768c6e18a9cbf00189db7b613048   <- dev today
```

Same hash, and the comparison is live — against `dev` it reports the expected 41 insertions. That also proves the sync is exactly complete rather than approximately: no drift, no omitted hold, nothing extra. `sort | uniq -d` over the dependency names on this head returns nothing, so the duplicate that broke the last round is not reintroduced, and all four holds are present — `react-test-renderer` 93, `react` 233, `react-dom` 234, `next` 276.

## One ordering point before merging

Merge **#3328 into dev first**, then this.

`back-merge.yml` fires on every push to `main` and says so explicitly — "whether it arrived by promotion or through the main-direct escape hatch". If this lands while #3328 is still open, the back-merge opens a PR into `dev` adding the `react` and `react-dom` entries, and #3328 adds those same lines to the same region of the same file. One of the two then conflicts, and it will be resolved by whoever happens to hit it rather than by whoever knows why the lines are there.

Merging #3328 first costs nothing — it is approved, and Monday 05:00Z is two and a half days out — and makes the back-merge content-identical instead. Either way `dev` will need that back-merge before the next promotion can pass `dev-contains-main`.

Nothing else on the diff: one file, additions only, and the freeze goes live the moment this merges rather than at the next promotion, which is the whole point of it.

PR #3333, opened 2026-09-18

ankit-yc:


aupyay (approved):
Back-merge is purely additive: the net diff is one file, `.github/dependabot.yml`, with 0 deletion lines, so nothing on dev is reverted.

Checked the hazard the file's own comment calls out — a duplicate `dependency-name` invalidating the whole config and voiding every hold in it. Parsed `.github/dependabot.yml` at the merge ref (794803b671a1f582c50f4d6ae53d1f49d09eff4b): it is valid YAML, the npm block has 23 ignore entries and 23 distinct names, and `react`, `react-dom`, `react-test-renderer` and `next` each appear exactly once. Mutation-checked that detector by injecting a second `react` entry — it reports the duplicate, so the clean result is falsifiable rather than a dead check.

CI green at the head: 42 pass, 23 skipping, 0 fail, 0 pending.

PR #3337, opened 2026-09-18
fix(ci): pin the Ruby version the mobile bundle resolves against

aupyay:
## What

Adds `apps/mobileAppYC/.ruby-version` (`3.3.9`) and lets the Mobile Gemfile check trigger on it.

## Why

The **Mobile Gemfile check has never passed.** It has run three times since it was added and failed all three, always at the same step:

```
Error: input ruby-version needs to be specified if no .ruby-version
or .tool-versions file exists
```

`ruby/setup-ruby` chdirs into `working-directory` and then looks for a version file there. With no `ruby-version` input it falls back to the literal default, and there is no `.ruby-version` or `.tool-versions` anywhere in the repo, so it throws before any bundler step runs.

So the red row on that job has never meant "the bundle does not resolve" - it means setup died first. `bundle install` and the stale-lockfile comparison have never executed. The job is not a required context, so nothing blocked on it and nothing surfaced it.

`mobile-release.yml` sets Ruby up the same way, also without a version input, so the release build carries the same defect - there with `bundler-cache: true`.

## The fix

One file fixes both call sites, because both already set `working-directory: apps/mobileAppYC`. `3.3.9` is what the lockfile records under `RUBY VERSION` and satisfies the Gemfile's `ruby ">= 3.3"`.

The paths filter gains the new file as well. That is deliberate: the Ruby version is an input the bundle resolves against, so changing it should re-resolve. Without it this check cannot run on this pull request and the fix would land with no evidence it works.

## Evidence

This PR's own `Bundle resolves` run is the test. A pass is the first time that job has got past setup; if `bundle install` then rewrites `Gemfile.lock` under a different bundler than the recorded `BUNDLED WITH 2.5.22`, the job will fail on the stale-lockfile comparison instead - which is a real finding the gate was previously unable to report.


ankit-yc (approved):
Approving. The version is not a guess: apps/mobileAppYC/Gemfile.lock already declares `RUBY VERSION ruby 3.3.9p170`, so 3.3.9 is exactly what the lockfile was resolved under, and it satisfies the Gemfile's `ruby ">= 3.3"` while staying below the 3.4 stdlib removals the Gemfile comments call out.

Worth recording that this fixes more than the check. mobile-release.yml carries the same `ruby/setup-ruby` block with `working-directory: apps/mobileAppYC` and no `ruby-version` input, and that block landed in 2ee12d1d on 2026-09-18 — after the last release run (mobile-v1.6.2, 2026-08-31). So the release job has never executed that step, and without this file the next tagged release would have failed at setup-ruby on macos-15, after the signing steps. Adding .ruby-version to the paths filter is also right: the file is now an input to the resolution the gate checks.

One thing this PR does not establish, and I am not asking it to block on: the assertion has never been observed failing. `Bundle resolves` passed here in 26s, which is its first end-to-end completion ever — before this, setup-ruby died first, so `bundle install` and `git diff --exit-code Gemfile.lock` have never actually been evaluated against a bad input. Green on a gate that has never been red is not yet evidence it has teeth. I will drive a deliberately stale Gemfile.lock through it on a throwaway branch and report the result; if it does not go red, that is a separate fix, not a reason to hold this one.

aupyay:
Followed up on the one thing I flagged in the review: the gate now has a demonstrated red arm.

Probe was #3341 — branched from this PR's head (96a1d72c) so `.ruby-version` was present and `setup-ruby` survived, since a red from setup-ruby dying would have proved nothing. The single added commit appended `gem 'rake'` to `apps/mobileAppYC/Gemfile` and left `Gemfile.lock` untouched.

`Bundle resolves` failed in 22s, in the `Lockfile matches the Gemfile` step — not in setup-ruby. `bundle install` ran to completion (gems installed into `./vendor/bundle`, post-install messages and all), then `git diff --exit-code Gemfile.lock` printed the regenerated entries:

```
+    rake (13.4.2)
...
+  rake
```

followed by the workflow's own error, `Gemfile.lock is stale. Run 'bundle install' in apps/mobileAppYC and commit the result.`, and exit code 1.

So the assertion distinguishes a good lockfile from a stale one, and the green on this PR is a real green rather than an untested one. #3341 is closed and its branch deleted.

PR #3341, opened 2026-09-18
ci(mobile): probe — deliberately stale Gemfile.lock, do not merge

aupyay:
Throwaway probe for the Mobile Gemfile check, opened to answer one question about #3337: now that the gate can reach its assertion, does that assertion actually fail on a bad input?

Based on #3337's head (96a1d72c) so .ruby-version is present and setup-ruby survives — a red from setup-ruby dying would prove nothing. The only added commit appends `gem 'rake'` to apps/mobileAppYC/Gemfile without regenerating Gemfile.lock, which is exactly the stale-lockfile condition the gate exists to catch.

Expected: `Bundle resolves` goes red at `git diff --exit-code Gemfile.lock`. Will be closed and the branch deleted as soon as the result is recorded. Not for merge.


aupyay:
Probe complete and recorded on #3337. The gate failed in the `Lockfile matches the Gemfile` step in 22s with `Gemfile.lock is stale`, which is the arm this was opened to observe. Closing; branch deleted.

Issue #3109, opened 2026-09-12
fix(auth): guard state-changing request replay after session refresh


aupyay:
## Summary

Resolve the stale-write replay remainder of #2168 with a scoped session-refresh policy that preserves drafts and current authorisation.

**Persona:** Any staff member saving work during session expiry. **Sequence:** Now - preserve intent and record integrity.

Delivery epic: [#3102](https://github.com/YosemiteCrew/Yosemite-Crew/issues/3102).

### Agent execution packet

**Start condition:** Start with a controlled reproduction. This is a security-sensitive shared-client change, so implement only after the real intercepted request path is established. It is not a good blind first task.

**Read order:** This issue, root AGENTS.md and the applicable per-workspace guide, the specific entry points below, then their callers/tests. Inspect current dev and overlapping PRs before editing. Packet paths/scripts were checked against `71ef4a3ddd68a76481f3d2d684fe9b3e718ebb0d`; follow renamed equivalents if dev advances. Do not read or implement the entire strategy epic.

#### Decisions already made

- Keep normal safe-read refresh. A delayed write cannot cross authenticated-user, organization, record or expected-version changes. If the implementation cannot prove same-context intent and safe exactly-once application, preserve an unsaved draft and require explicit resubmission.
- Do not fix only axios's existing transient-read retry loop: SuperTokens also intercepts fetch/XHR. Trace the installed SDK path and direct-fetch callers before selecting the shared interception point.
- Do not introduce another global HTTP client or disable session/MFA behavior. Never persist a sensitive draft for a different user's later session; same-user recoverability and shared-device isolation must both hold.
- If the suspected replay cannot be reproduced after tracing representative callers and exercising the real SDK boundary, report evidence and limit the change to demonstrated gaps. Do not add speculative retry machinery merely to produce a diff.

#### Ordered implementation checkpoints

1. Read authClient.ts, services/axios.ts, authStore/orgStore, useAuth logout/reset and representative appointment/form/clinical writers. Find direct fetch/XHR callers and the locked SuperTokens version.
2. Create a controlled delayed-refresh reproduction, recording original context and request identity. Existing authClient tests mock Session.init and cannot alone prove intercepted network behavior. First checkpoint: show the stale write attempt through the real supported interception path, or precisely document why it is not reproduced.
[...]

Issue #3103, opened 2026-09-12
test(repo): prove one complete clinic day across staff and owner surfaces


aupyay:
## Summary

Create one reproducible, synthetic clinic-day acceptance scenario that proves the handoffs between the existing staff PIMS, backend and pet-parent app.

**Persona:** Practice owner and release operator. **Sequence:** Now - release evidence.

Delivery epic: [#3102](https://github.com/YosemiteCrew/Yosemite-Crew/issues/3102).

### Agent execution packet

**Start condition:** Start the fixture and acceptance harness now. Missing linked features block their release stages, not fixture work; never mark a partial journey fully passed.

**Read order:** This issue, root AGENTS.md and the applicable per-workspace guide, the specific entry points below, then their callers/tests. Inspect current dev and overlapping PRs before editing. Packet paths/scripts were checked against `71ef4a3ddd68a76481f3d2d684fe9b3e718ebb0d`; follow renamed equivalents if dev advances. Do not read or implement the entire strategy epic.

#### Decisions already made

- Use one routine synthetic outpatient visit, two isolated practices, a clinician, receptionist, authorized parent/co-parent and unauthorized parent. Default fixture timezone Europe/Berlin and currency EUR; freeze the clock. Reuse existing valid clinical fixtures rather than inventing treatment guidance.
- The manifest owns only synthetic IDs and expected counts, not clinical state. All changes go through normal services/API/UI. Payment, signing and message providers may be stubbed at their external boundary for deterministic checks; label separate real provider/device checks.
- Use stage results PASS, FAIL, BLOCKED or NOT_RUN. Any required stage other than PASS prevents a ready result. Missing credentials must not become a skip that produces a green release signal.

#### Ordered implementation checkpoints

1. Read apps/frontend/e2e/auth-flow.spec.ts, e2e/support/auth.ts, playwright.config.ts, apps/mobileAppYC/e2e/smoke.e2e.js and workspace.prisma.service.ts. Inspect their credential/trace protection before reusing the harness.
2. Create one versioned synthetic fixture and expected-count manifest. Seed two practices and relevant roles with current supported setup; record IDs returned by APIs. First checkpoint: repeat setup and prove isolation without duplicate fixtures.
[...]

PR #3346, opened 2026-09-19
ci(repo): run the override advisory check nightly, fail-closed

aupyay:
## What

Adds `.github/workflows/override-advisories.yml`: a nightly, fail-closed run of
`scripts/ci/check-override-advisories.mjs` over **both `main` and `dev`**.

## Why

That script has never had a caller. It is the only thing in the repo that
answers *"does a pnpm override pin a version that still has an advisory"*, and
until now it ran only when somebody typed `pnpm check:overrides` by hand. That
is how dev came to be carrying two stale pins until #3344 was reviewed.

It cannot be a pull-request gate, and that part is deliberate — it needs the
registry, and a registry timeout on a required check is a flaky security check,
which gets routed around. `check-override-drift.mjs` already records that
reasoning where it explains why *it* can be a hard gate and this one cannot.

The half that was missing is the scheduled one. The script's own `--strict` flag
exists, in its words, *"so a scheduled security job can opt into fail-closed"* —
and that job was never written.

## Night one goes red on `main`, and it should

`main`'s `adm-zip` override pins **0.6.0** against **GHSA-7q85-xj36-vmfc**
(HIGH — uncontrolled memory allocation via the declared uncompressed size, DoS;
fixed in 0.6.1). The only acceptance recorded on `main` is for a *different*
advisory (GHSA-vwc7-r8mq-g2x9, moderate, no fix published), so this one is
unaccepted and open right now. Measured at `main` `b01f79dd7`:

```
check-override-advisories: 117 override entries checked
  advisories in tree:  9
  vulnerable pins:     2
  uncovered copies:    0
  accepted (baseline): 1
...
  adm-zip  [high]
    pinned at:     0.6.0
    advisory:      GHSA-7q85-xj36-vmfc
    fixed in:      0.6.1  (patched range >=0.6.1)
```

`dev` is clean at `bb46342c1`: 127 entries, 3 advisories, 0 vulnerable pins,
exit 0. This PR does not fix `main` — raising that pin is a change to `main`
and its own decision. The point of the workflow is that the finding stops
depending on someone thinking to look.

## Why both branches (review fixed this)

The first revision audited neither tree on purpose — it took whatever ref
triggered it. A schedule fires only from the default branch and
`actions/checkout` with no `ref` follows the triggering ref, so it read `main`
and never saw `dev`; **it would not have caught the stale pins #3344 fixed.**

Auditing only `dev` is wrong in the other direction. `main` drifts on its own:
dependabot SECURITY updates open against the default branch whatever
`target-branch` says — `promotion-guard.yml` says so in its own header — and
#2954 exists because `main` had accumulated stale-pinned advisories `dev` did
not have. The two baselines already differ today.

So: a `branch: [main, dev]` matrix with an explicit `ref:`, `fail-fast: false`
so a red `main` cannot cancel `dev`, and the branch carried in the **issue
marker** rather than only the body — otherwise `main`'s finding and `dev`'s
finding land on one issue and read as a single drifting problem.

## Why `--strict` is load-bearing

Without it a nightly would look green while verifying nothing. Measured with
`pnpm` removed from `PATH` (control: `command -v pnpm` → rc=1, so the arm really
was disarmed):

| arm | exit | output |
| --- | --- | --- |
| no `--strict` | 0 | `SKIPPED (no advisory data). Overrides were NOT verified.` |
| `--strict` | 2 | `--strict was passed, so this is a failure.` |

## Exit 1 and exit 2 are filed as different issues

`1` is an override actually pinning something vulnerable — a security finding.
`2` is `--strict` declining to pass without advisory data, which says nothing
about any override. Filing the second as the first is how a registry outage gets
triaged as a vulnerability; `tls-expiry.yml` was already fixed for that exact
mistake, and this follows its pattern (distinct marker, title and label, issue
opened on failure without swallowing the red).

The discriminator is the step's exit code rather than the report text, because
the exit code is what the script actually decided.

## Verification

Both branches run locally after `pnpm install --frozen-lockfile`, in separate
worktrees at the shas named above:

| tree | entries | advisories | vulnerable pins | exit |
| `main` `b01f79dd7` | 117 | 9 | 2 (1 accepted) | **1** |
| `dev` `bb46342c1` | 127 | 3 | 0 | **0** |

Must-fail arm, so the `dev` leg is not a check that can only ever be green — a
synthetic advisory against the **real** `protobufjs` `8.8.0` pin:

```
$ node scripts/ci/check-override-advisories.mjs --audit-json <synthetic> --no-baseline   # exit 1
  protobufjs  [high]
    override key:  "protobufjs@7.5.5"
    pinned at:     8.8.0
    installed:     8.8.0
    fixed in:      8.9.0  (patched range >=8.9.0)
```

`actionlint` passes on the new file.

## Known follow-up, please read before approving

Scheduled workflows and `workflow_dispatch` only fire from the **default
branch**, which here is `main` — currently 1144 commits behind `dev`. So this
file landing on `dev` does not start the job. A second, file-only PR into `main`
is needed for it to run at all, following the shape of #3332. That is a change
to `main` and needs an owner's call, not mine; it is called out here so nobody
reads the merge of this PR as the check being live.


ankit-yc (dismissed):
Approving at 8051db07a. The shape is right: fail-closed, off the critical path, exit 1 and exit 2 given different titles, markers and labels so a registry outage never gets triaged as a vulnerability, actions pinned by SHA, `permissions: {}` at the top with the narrowest grant on the job.

I confirmed the premise in the script rather than taking it from the PR body: `--strict` is exit 2 on unreachable advisory data (`check-override-advisories.mjs:68`, `:588-594`), and without it line 593 prints `SKIPPED (no advisory data). Overrides were NOT verified.` and returns 0. So the flag is load-bearing and the workflow passes it.

**1. The nightly audits `main`, not `dev`, and the two trees are materially different.**

Schedules fire from the default branch, and the checkout step takes no `ref:`, so the job runs against `main`. `main` is 1149 commits behind `dev`. I ran the checker against each tree (`git archive` into a temp dir, so the result is attributable to the tree that produced it):

| tree | override entries | advisories in tree | vulnerable pins | rc |
|---|---|---|---|---|
| `origin/dev` | 127 | 3 | 0 | 0 |
| `origin/main` | 117 | 9 | 2 (1 unaccepted) | 1 |

Different override block, different advisory set, different verdict. The incident that motivated this workflow — #3344 carrying two stale `adm-zip` pins on `dev` — is a `dev` finding, and the job as written would not have caught it. Worth a follow-up that runs both arms, e.g. a `matrix: ref: [main, dev]` with `ref: ${{ matrix.ref }}` on the checkout and the ref in the issue title. Not blocking: auditing `main` is strictly better than the nothing we have today, and the change is additive.

**2. The first scheduled run will fail, and it should. Measured, not predicted.**

Against `main`'s tree, `--strict` exits **1**:

```
adm-zip  [high]
  override key:  "adm-zip"    pinned at: 0.6.0    installed: 0.6.0
  advisory:      GHSA-7q85-xj36-vmfc
  title:         adm-zip: Uncontrolled memory allocation via the declared uncompressed size (DoS)
  vulnerable:    <0.6.1        fixed in: 0.6.1
```

`main`'s baseline accepts only the moderate `GHSA-vwc7-r8mq-g2x9`; this high one was never recorded. So night one opens a `security` issue whose body says "edit the `pnpm.overrides` block" — while `dev` already did exactly that in #3344. Please say this in the PR body before merge, otherwise the first person to see it triages a true finding as a workflow bug. A `dev` -> `main` back-merge clears it.

**3. The dedup works — verified against real data, not assumed.**

I did not want to take "search matches an HTML comment" on trust, so I found a planted hit that already exists: open issues #3103-#3109 carry `<!-- agent-execution-packet-2026-09-12 -->` in the body.

```
in:body "<!-- agent-execution-packet-2026-09-12 -->"  -> 7
in:body "agent-execution-packet-2026-09-12"           -> 7
```

Markers in HTML comments are indexed and matchable, so the comment-instead-of-duplicate path will fire. Note the two arms return the same count: the `<!--` and `-->` delimiters are stripped by the tokenizer, so the real discriminator is the bare slug. An issue body that merely mentions `override-advisories-tracker` in plain text — someone quoting this workflow, say — would be matched and commented on. Low likelihood, and the blast radius is a comment on the wrong issue. Filtering the search result by the `security` label, or re-checking the marker on the fetched body, closes it if you think it is worth the lines.

Nothing else. The `set +e` / `PIPESTATUS[0]` capture is correct for the `tee` pipeline, and routing an empty `steps.check.outputs.status` to the tooling branch is the right default for a step that died before it could set one.

ankit-yc (approved):
Re-approving at 4fded20f6. Both findings from the previous round are addressed, and addressed at the right level.

The matrix is the fix I asked for and the marker change is one I had not thought through: putting the branch in the marker rather than only in the body is what stops the two arms sharing an issue and reading as a single drifting problem. `fail-fast: false` is load-bearing for the same reason.

I checked the two citations added in the new comments rather than taking them as prose:

- `promotion-guard.yml` says it at lines 5-6 and 50-51 — "Dependabot SECURITY updates ignore `target-branch: dev` and always open against the default branch". The comment's claim is the repo's own documented position, not a new assertion.
- #2954 is real and is what the comment says it is: "fix(repo): close every stale-pinned security advisory on main", merged 2026-09-09.

Our two independent measurements of `main` agree: 117 override entries, 9 advisories in tree, 2 vulnerable pins, 1 accepted by the baseline, exit 1 on the unaccepted high. You measured at `b01f79dd7`, I measured `origin/main` via `git archive` into a temp dir. Same numbers from two desks and two trees.

**One caveat, not a change request, and it binds this PR and #3347 equally.** A schedule runs the workflow *file* from the default branch, so this workflow does not exist until `dev` reaches `main` — and `main` last took `dev` on 2026-08-31 (#2585, 1149 commits back). Until a promotion lands, merging this to `dev` arms nothing. Worth one line in the PR body next to the branch caveat you already added, so the absence of a nightly run is not read as the nightly passing.

aupyay:
Small correction to the closing caveat in my review, from the same stale-ref mistake I have just retracted on #3347: my clone's fetch refspec is dev-only, so `origin/main` was frozen at `a483c782d` while real main was at `b01f79dd7`.

Re-measured against real main:

- `main` is **1147** commits behind `dev`, not 1149.
- `.github/workflows/override-advisories.yml` is **absent** from `b01f79dd7` (`git ls-tree origin/main` returns nothing for it), so the caveat itself stands unchanged: a schedule runs the workflow file from the default branch, and this workflow is not there yet.
- The `main` advisory figures I quoted are unchanged — 117 entries, 9 advisories, 2 vulnerable pins, 1 accepted, exit 1 on the high `GHSA-7q85-xj36-vmfc`. That block did not move between the stale sha and the real one, so your `b01f79dd7` measurement and mine agree for a better reason than before.

Approval unaffected.

Issue #3345, opened 2026-09-18
dependabot: the react/react-dom hold does not cover security updates, contrary to the comment beside it


aupyay:
Found by the automated review on #3333 (P1), verified against the docs before filing. The P2 finding on the same PR was based on an `apps/dev-docs` workspace that does not exist in the tree and is not reproduced here.

## What the config claims

`.github/dependabot.yml` freezes `react` and `react-dom` for dependabot because react-native 0.81.6 embeds a renderer that does a literal string compare against `"19.1.4"`, so any react bump — patch included — kills the iOS release app on launch. The comment beside those entries states:

> Ignore rules apply to dependabot SECURITY updates too, so a react advisory will not arrive as a PR while these entries stand.

## Why that does not hold

The npm entry those `ignore` rules live in sets `target-branch: dev` (`.github/dependabot.yml` line 14), and this repo's default branch is `main`. The options reference says settings on a package-manager entry stop governing security-update PRs once `target-branch` points at a non-default branch — "also change how Dependabot creates pull requests for security updates, **except where `target-branch` is used**" — and the `ignore` section documents it as a version-update option, with security updates instead updating all vulnerable dependencies found in lock files.

So on a react advisory dependabot can still open a bump, `.github/workflows/promotion-guard.yml` retargets it to `dev`, and nothing downstream catches it: the peer range `^19.1.4` is satisfied, jest never mounts through the native renderer, and the release workflow builds and uploads the app without launching it. That is the same blind spot that shipped the crash originally — the hold that was supposed to close it is not armed for the advisory path.

The practical risk is bounded (it needs a react advisory to fire) but the failure mode is a release app that terminates before its first screen, and the comment currently tells the next reader the case is covered.

## Suggested direction

Not a config-only fix, since the exemption is a property of `target-branch`. Options, roughly in order of how much they actually enforce:

1. Enforce the constraint where the bot PR is already being handled — have `promotion-guard.yml` fail or close a PR that moves `react`/`react-dom` off the version react-native's bundled renderer pins.
2. Add a CI check that reads the renderer's embedded version literal and asserts it equals the installed `react` version, so any route to a mismatch is caught rather than just the dependabot route.
3. Isolate the mobile app's react so a frontend-driven bump cannot move it.

At minimum the comment should be corrected, because it currently documents a guarantee the configuration does not provide.

[...]

PR #3347, opened 2026-09-19
ci(repo): correct dependabot comments that claim security-update coverage

aupyay:
Comment-only change to `.github/dependabot.yml`. Refs #3345 — this is the "at minimum, correct the comment" half; the enforcement half stays open there.

## The problem

Three comments in the file tell the reader that an `ignore` entry suppresses dependabot **security** updates. It does not. Every block in this file sets `target-branch`, and the options reference carves security updates out wherever that is used, voiding the block as a unit — `ignore`, `cooldown`, `groups`, `labels` and `commit-message` alike. Security updates go to `main` with dependabot's defaults.

## Why this is measured, not inferred

The exemption leaves a mark on options that are visible from outside, so it does not need a CVE on a held package to observe. `labels: [dependencies]` and `commit-message.prefix: 'chore(repo)'` have been live on `main` since 519b33467 (2026-06-23); `target-branch: dev` since a05c62f1b (2026-07-03).

| PR | opened | base | labels | commit headline |
|----|--------|------|--------|-----------------|
| #1873 | 2026-07-18 | main | `dependencies`, `javascript` | `chore(deps-dev): bump adm-zip` |
| #2005 | 2026-07-25 | main | `dependencies`, `javascript` | `chore(deps): bump the npm_and_yarn group…` |
| #2020 | 2026-07-26 | main | `dependencies`, `javascript` | `chore(deps): bump the npm_and_yarn group…` |
| #2067 | 2026-08-08 | main | `dependencies`, `javascript` | `ci(deps): bump the npm_and_yarn group…` |

Version-update PRs from the same window (#1774, #1783, and every one since the retarget) carry `dependencies` alone and `chore(repo)`. Two independent signals:

- **Label.** A configured `labels:` list replaces the defaults, and this block lists `dependencies` only. No workflow applies labels (`grep -rn javascript .github/workflows/ | grep -i label` → no matches), so `javascript` can only appear if the block was not read.
- **Prefix.** `chore(deps-dev)` on #1873 is decisive alone — a fixed `prefix:` cannot produce a dev-dependency variant without `prefix-development`, which this file does not set.

#2131 (bundler, `dependencies`/`ruby`) is *not* evidence either way: the bundler block declares `ruby` itself, so config and default agree there.

## What changed

- **Header** — states the rule once, with the evidence, so the next reader does not re-derive it per entry, and says plainly that an `ignore` entry is a statement about weekly version updates only.
- **`react` / `react-dom`** — the comment claimed a react advisory "will not arrive as a PR while these entries stand". It will, against `main`, with no cooldown. This is the hold whose bypass matters most: the compare it guards is a runtime throw in the release app, jest never mounts through the native renderer, and an exact-pinned bump is a clean install — so no gate here sees it. The manual policy below it is unchanged.
- **`posthog-js`** — the comment claimed dependabot "will not open a remediation PR if an advisory lands against it". It will. The supply-chain scan stays as the standing control either way.

The cooldown comment already said cooldown does not apply to security updates, and is left alone — it was right.

## Verification

- `yaml.safe_load` of `HEAD:.github/dependabot.yml` and of `origin/dev:.github/dependabot.yml` compare equal (`json.dumps(..., sort_keys=True)` identical), so the parsed configuration is unchanged.
- The diff filtered to non-comment, non-blank lines is empty.
- Commit hooks ran: gitleaks (`no leaks found`), prettier, secretlint. Pre-push ran `pnpm run lint` and `pnpm run type-check`: 17 tasks, 17 successful.

No behaviour change, so nothing to exercise beyond the above. Per the file header this only takes effect once `dev` reaches `main`; since it is comment-only, there is nothing inert in the meantime.


ankit-yc (approved):
Approving at 6d69f5e7b — the comments now say what the config does, and the evidence paragraph is the right kind: #1873's `chore(deps-dev)` headline is decisive on its own, because a fixed `commit-message.prefix` cannot emit that variant.

**But the finding is bigger than the comment, and I think #3345 should be re-scoped rather than closed by this PR.**

The hold this PR is correcting the comment on is not in force at all. Not "bypassed for security updates" — absent.

```
$ git rev-parse origin/dev:.github/dependabot.yml
a85d2641bf88ebdf7a596869862a2113e0a33361
$ git rev-parse origin/main:.github/dependabot.yml
c4ff428f9ece28b21a19cfdc22a7e80c6e9b3a3a
$ git diff --stat origin/main:.github/dependabot.yml origin/dev:.github/dependabot.yml
 1 file changed, 84 insertions(+)
```

All 84 lines are dev-only, and they are the `#3325` block — the react/react-dom comment and both `- dependency-name:` entries. Dependabot reads this file from the default branch, which this file's own header says in its first sentence. So on the branch dependabot actually evaluates, `react` and `react-dom` carry no ignore entry, and the hold does not cover **weekly version updates** either. The security-update carve-out is the second reason it does not hold; the first is that it is not there.

That is precisely the failure the header warns about — "keep the copies on dev and main byte-identical (see PR #2079, which fixed a month of silently inert config changes)". #2079's failure mode has recurred, and the invariant is violated on `dev` right now, before this PR. This PR adds 21 more dev-only lines to the same file, which is correct as content and widens the divergence until a promotion.

Why it is urgent rather than tidy: `apps/mobileAppYC` pins `react` to an exact `19.1.4`, and the comment records why — RN's prod renderer does a literal `if ("19.1.4" !== React.version) throw`. A weekly version-update PR raising react is currently unopposed, and as the comment says, no CI stage can see the consequence: jest never mounts through the native renderer, the peer range is not violated so the install is clean, and the release workflow builds and uploads without launching.

`main` last took `dev` on 2026-08-31 (#2585); it is 1149 commits behind. #3325 landed after that, which is exactly why the hold never reached the branch that matters.

**Suggested split.** Merge this as-is, then let #3345 own the two things this PR cannot: (1) the promotion that puts the `#3325` entries on `main`, which is the difference between a documented hold and an actual one, and (2) the byte-identical invariant, which has now failed twice and is a check rather than a comment. If you would rather I open the second as its own issue so #3345 stays scoped to the security-update carve-out, say so and I will.

Nothing blocking in the diff itself. Parsed YAML unchanged is the right claim for a comment-only change, and it matches what I see.

aupyay:
On the review's wider point — the premise needs refetching, but the conclusion it leads to is right and I am acting on it.

**The 84 lines are on `main`.** PR #3332 (`chore/sync-dependabot-config-to-main`, base `main`, `.github/dependabot.yml +84 -0`) merged at 2026-09-18T22:23:00Z, and is `main`'s current head. Read from the API rather than a local ref:

```
$ for r in main dev; do gh api "repos/YosemiteCrew/Yosemite-Crew/contents/.github/dependabot.yml?ref=$r" --jq '.sha'; done
a85d2641bf88ebdf7a596869862a2113e0a33361
a85d2641bf88ebdf7a596869862a2113e0a33361
```

Same blob on both refs — byte-identical, 380 lines, and `dependency-name: 'react'` present on `main`. `git log origin/main -S"dependency-name: 'react'" -- .github/dependabot.yml` names 52ede29c7, 2026-09-18, "sync the dependabot config to the default branch".

So the #3325 hold is live for version updates. A stale `origin/main` reports exactly the same 84-line shortfall — it reported it to me an hour ago before I fetched, which is what makes the number worth double-checking rather than the finding careless. `main` really is 1147 commits behind `dev` (`compare/main...dev` → `ahead_by=1147`); it is just not behind on this file.

**The part that stands.** A dev-only edit to `.github/dependabot.yml` is inert, because dependabot reads the default branch — which is the rule this PR's own header states and #2079 was filed for. That applies to this PR. #3332 is the precedent for the shape of the remedy, and I am opening the matching sync-to-`main` PR rather than letting these corrections sit on `dev` for the next promotion.

The correction is comment-only either way, so nothing is mis-enforced in the meantime — the file just tells `main`'s reader the wrong thing until the sync lands.

aupyay:
**Correction — the main finding in my review above is wrong. Retracting it.**

The react/react-dom hold **is** on `main`. `.github/dependabot.yml` is byte-identical on both refs:

```
origin/main:.github/dependabot.yml  a85d2641bf88ebdf7a596869862a2113e0a33361
origin/dev:.github/dependabot.yml   a85d2641bf88ebdf7a596869862a2113e0a33361
```

#3332 synced it to the default branch and merged as `b01f79dd7`. So there is no 84-line gap, the `#3325` entries are live on the branch dependabot reads, and the byte-identical invariant is **held**, not violated. Everything I wrote above about #2079's failure mode recurring, and about a weekly version-update PR raising react being unopposed, is withdrawn.

**How I got it wrong**, because the mechanism is worth more than the retraction: this clone's fetch refspec is `+refs/heads/dev:refs/remotes/origin/dev` — dev only. I ran `git fetch origin main dev` before measuring, which looks like it covers main and does not: naming a branch on the command line writes `FETCH_HEAD`, but `refs/remotes/origin/main` is only updated if a configured refspec maps it. So `origin/main` stayed frozen at `a483c782d` from 2026-09-10 while real main was nine days and two merges ahead. Every `origin/main` figure in my review is from that stale ref. `git ls-remote origin refs/heads/main` returns `b01f79dd7` and is the form with no local ref anywhere in the path.

**What survives, re-measured against real `main` (`b01f79dd7`), not the stale ref.** The override-advisory numbers are unchanged, which is luck rather than diligence — that block did not move between the two shas:

```
117 override entries checked / advisories in tree: 9 / vulnerable pins: 2 / accepted (baseline): 1
rc=1 — adm-zip [high] GHSA-7q85-xj36-vmfc, pinned 0.6.0, fixed in 0.6.1
```

And `main` is 1147 commits behind `dev` (not 1149). Its last *promotion from dev* is still #2585 on 2026-08-31; #2954 and #3332 both landed on main directly, which is why the config sync arrived without the branch catching up.

So the approval stands on the diff itself, and the part of my review that holds is the narrow one: these comments are correct, and #3345's enforcement question is still open. The urgency I attached to it was manufactured by a stale ref.

PR #3348, opened 2026-09-19
chore(repo): sync the dependabot comment corrections to the default branch

aupyay:
Companion to #3347 (merged to `dev` as d69022434). Same shape and same reason as #3332. Refs #3345.

`.github/dependabot.yml` is read from the **default branch**, so a correction that only exists on `dev` says nothing to dependabot or to anyone reading `main`. That is the rule the file's own header states and what #2079 was filed for. `main` is 1147 commits behind `dev` (`compare/main...dev` → `ahead_by=1147`), so waiting for the next promotion is not a plan.

## What this carries

The comment corrections from #3347, verbatim. Three comments claimed that an `ignore` entry suppresses dependabot **security** updates. It does not: every block in the file sets `target-branch`, which carves security updates out of the block *as a unit* — `ignore`, `cooldown`, `groups`, `labels` and `commit-message` alike. Security updates go to `main` with dependabot's defaults.

Measured on this repo rather than taken from the docs. `labels: [dependencies]` and `commit-message.prefix: 'chore(repo)'` have been live on `main` since 519b33467 (2026-06-23); `target-branch: dev` since a05c62f1b (2026-07-03). Every npm security PR since carries dependabot's defaults instead:

| PR | opened | labels | commit headline |
|----|--------|--------|-----------------|
| #1873 | 2026-07-18 | `dependencies`, `javascript` | `chore(deps-dev): bump adm-zip` |
| #2005 | 2026-07-25 | `dependencies`, `javascript` | `chore(deps): bump the npm_and_yarn group…` |
| #2020 | 2026-07-26 | `dependencies`, `javascript` | `chore(deps): bump the npm_and_yarn group…` |
| #2067 | 2026-08-08 | `dependencies`, `javascript` | `ci(deps): bump the npm_and_yarn group…` |

Version-update PRs from the same window carry `dependencies` alone and `chore(repo)`. `chore(deps-dev)` on #1873 settles it alone — a fixed `prefix:` cannot emit a dev-dependency variant without `prefix-development`, which this file does not set.

The `react`/`react-dom` and `posthog-js` comments said the opposite of that, and the header now states the rule once so the next reader does not re-derive it per entry.

## Verification

- The file is taken verbatim from `origin/dev`: `diff <(git show origin/dev:.github/dependabot.yml) <(git show HEAD:.github/dependabot.yml)` is empty, so the two copies stay byte-identical as the header requires.
- Comment-only: `yaml.safe_load` of this revision and of the previous one compare equal, and the diff filtered to non-comment, non-blank lines is empty. Nothing dependabot evaluates changes.
- Commit hooks ran: gitleaks (`no leaks found`), prettier, secretlint. Pre-push ran `pnpm run lint` and `pnpm run type-check` green (the first attempt failed on `@yosemite-crew/database#build` with `TS2305` for an ungenerated Prisma client — a fresh-worktree setup gap, fixed by `pnpm --filter @yosemite-crew/database run prisma:generate`, not by skipping the hook).

Enforcement — a check that stops a `react` bump reaching a version react-native's bundled renderer does not expect — stays open on #3345. This PR only makes the file tell the truth on the branch that is read.


ankit-yc (approved):
Approving at a98bbab42, and applying the `main-direct` label so it can actually merge — see below.

**Verbatim is verified by blob hash, not by reading the diff.** The file this PR produces and the file on `dev` are the same object:

```
origin/pr3348:.github/dependabot.yml  60a46627a14c51794a081580aba9d51a7c499e7a
origin/dev:.github/dependabot.yml     60a46627a14c51794a081580aba9d51a7c499e7a
origin/main:.github/dependabot.yml    a85d2641bf88ebdf7a596869862a2113e0a33361
```

So merging this makes the two copies byte-identical, which is the invariant the file's own header asks for.

**Nothing else rides in, which is the risk specific to a main-targeted PR.** A sync branch cut from `dev` would drag 1147 commits into `main` behind a one-file title. This one is cut from `main`'s exact head:

```
merge-base(main, pr3348) = b01f79dd7 = main head
1 commit, 1 file, +43 -10
```

Authored `ankit@dunexploration.com`. Clean.

**The blocker, and the sanctioned way through it.** `Only dev may merge into main` is red, and it is the *sole* required status check in `main`'s ruleset (`gh api repos/.../rules/branches/main` -> `required_status_checks` contains exactly that one context). So this cannot merge in its current state. The failing step names its own remedy:

```
main only accepts merges from dev, but this PR's head is 'chore/sync-dependabot-comments-to-main'.
Retarget it at dev and let the usual dev-to-main promotion carry it.
If this genuinely must land on main first, add the 'main-direct' label.
```

The retarget suggestion is a no-op here — the content is already on `dev`, merged as #3347 (`d69022434`) — and the only other route is a full `dev` -> `main` promotion of 1147 commits, which a comment sync should not trigger. So `main-direct` is the right path, and it is the guard's own documented allowance rather than a weakening of it: with the label the first step exits 0 with `::notice:: 'main-direct' label present - a human has explicitly allowed this direct merge`, and the ancestry step has already reported `Head is 'chore/...', not dev - ancestry rule does not apply`.

Precedent is exact: #3332 was the same shape (base `main`, head `chore/sync-dependabot-config-to-main`), carried `main-direct`, and merged at 22:23Z.

I have applied the label as the reviewer, since that is the party the guard's message means by "a human". Re-run the check and it should go green; merge is yours.

PR #3349, opened 2026-09-19
feat(repo): fail CI when react and react-native's renderer disagree

aupyay:
The enforcement half of #3345. #3347/#3348 corrected the comments; this is the part that actually holds the line.

## The failure being guarded

`react-native` ships a renderer that compares `React.version` against a literal baked into its bundle:

```js
var isomorphicReactPackageVersion = React.version;
if ("19.1.4" !== isomorphicReactPackageVersion)
  throw Error('Incompatible React versions: The "react" and "react-native-renderer" packages must have the exact same version.')
```

Because the comparison is against a literal, a **patch** bump breaks it exactly as a major does, and nothing in CI sees it: the declared peer range is satisfied so the install is clean, jest never mounts through the native renderer, and the release workflow builds and uploads the app without launching it. The app installs, launches, and terminates before its first screen.

`.github/dependabot.yml` holds `react` for that reason, but an `ignore` entry is a version-update rule only — every block in that file sets `target-branch`, which carves security updates out of the block as a unit — so an advisory can still open the bump, and a hand-written change was never covered at all. This check does not depend on how the bump arrives.

## What it checks

`scripts/mobile/check-react-renderer-pin.mjs` reports three things:

1. **A pin that is not exact.** `^19.1.4` is how a hold stops holding without anyone editing the file — the next lockfile refresh moves react on its own.
2. **A resolved react the renderer was not built against.** The fatal one; the message names the startup throw.
3. **Drift between the pin and what is installed.** Suppressed when the pin is a range, because the range already explains it and reporting both sends the reader to fix the symptom.

It reads the **mobile workspace specifically**. That is load-bearing, not incidental: in this tree the root resolves `react` 19.3.0 while `apps/mobileAppYC` resolves 19.1.4, so a root-relative read would report a failure that is not there.

The version is located by anchoring on the thrown message and searching only the source immediately before it, and `rendererExpectation` returns `null` rather than a guess when that window holds none **or more than one** candidate. An ambiguous read is the dangerous outcome — picking either candidate leaves the gate reporting green against a literal that guards something else. `null` becomes exit 2, which fails the job for the same reason exit 1 does.

It reads `ReactNativeRenderer-prod.js`. `ReactFabric-prod.js` carries no equivalent compare (measured: 0 occurrences of the error text), so a check pointed at it could never fail — there is a test pinning the path.

## Placement

The `core` job, which every event reaches, rather than the mobile matrix leg, which only runs when the diff touched mobile. Same argument the hardcoded-colour gate beside it makes in its own comment. No extra setup: the install already populates `apps/mobileAppYC/node_modules`, which is where both the renderer bundle and the resolved react version live. The test file is picked up by `pnpm run test:scripts` through the existing `scripts/mobile/*.test.mjs` glob.

## Verification

**Live, against the real tree** (`apps/mobileAppYC/node_modules`, restored from a snapshot after each arm; `git status --porcelain` clean afterwards):

| arm | rc |
|-----|----|
| unmodified | 0 |
| installed react → 19.1.5 (the patch the hold exists for) | 1 |
| renderer's thrown message renamed | 2 |
| a second version compare inserted into the window | 2 |
| restored | 0 |

**Unit tests**: 12 pass. `pnpm run test:scripts` → `# tests 533 # pass 533 # fail 0`.

**Mutation matrix** — every mutation applied to the script, tests re-run, script restored from a snapshot and confirmed byte-identical afterwards:

| mutation | result |
|----------|--------|
| control | 12 pass, 0 fail |
| search the whole source instead of the window | 9 pass, **3 fail** |
| guess instead of refusing when ambiguous | 11 pass, **1 fail** |
| drop the range check | 10 pass, **2 fail** |
| invert the renderer mismatch comparison | 6 pass, **6 fail** |
| point at `ReactFabric-prod.js` | 11 pass, **1 fail** |
| drop the drift suppression for range pins | 10 pass, **2 fail** |
| restored | 12 pass, 0 fail |

The first of those started green. The original fixture had no second version comparison in it, so the window was not what made the test pass — the digit anchor was, and on the real 366 KB bundle there is exactly one match in the whole file. The fixture now carries an unrelated compare further down, which is what a real bundle offers a search that is not scoped, and the ambiguity refusal was added so the answer is never a guess between two candidates.

Pre-push ran `pnpm run lint` and `pnpm run type-check` green.

## Not covered

`react-dom` and `apps/frontend`. The renderer literal is a react-native constraint; the frontend's react/react-dom pairing is a different check and there is no observed failure behind it yet. Say so rather than widen this one on a guess.


ankit-yc (changes requested):
Not approving yet — two checks are red, and one of them is the gate that exists to validate exactly this PR's tests. The design is right and I have verified the hard part works; the blockers are wiring, not logic.

## What I verified myself, so you do not have to re-run it

I imported `rendererExpectation` and `problems` from your module and drove them against the real 365 KB bundle (`react-native 0.81.6`), not a fixture:

```
installed react-native: 0.81.6
rendererExpectation   : "19.1.4"          <- the real literal, read out of the real bundle
matching arm                       -> []
range pin "^19.1.4"                -> range problem fires
mutation: throw message renamed    -> null   (exit 2 path)
mutation: second candidate in window -> null (exit 2 path)
```

The mismatch arm is the one worth having: this desk's checkout has react **19.3.0** installed against a renderer built for 19.1.4, so the must-fire arm ran on a real disagreement I did not arrange, and produced both messages — the startup throw and the pin/lockfile drift. That is a stronger arm than a synthetic one.

I also checked the claim in the comment rather than trusting it. Occurrences of `Incompatible React versions` across the six renderer bundles:

```
ReactFabric-dev.js 0    ReactFabric-prod.js 0    ReactFabric-profiling.js 0
ReactNativeRenderer-dev.js 1  -prod.js 1  -profiling.js 1
```

Exactly as documented: Fabric carries no guard, so reading it would produce a check that can never fail. Choosing the Paper bundle is correct, and returning `null` rather than guessing when the window holds two candidates is the right call.

## Blocker 1 — `Tests Must Be Able To Fail` is red, and it is right to be

```
source files changed: 1
test files changed:   1
no runnable unit tests changed (e2e only, or outside a known workspace)
FAIL: The changed tests are all end-to-end specs, which this gate does not run.
```

The message misdescribes the cause — these are not e2e specs — but the gate's verdict is sound. Two rules in `scripts/ci/tests-must-be-able-to-fail.mjs` disagree about `scripts/`:

- `isSourceFile` excludes `^\.github\/` and `^scripts\/ci\/` only. `scripts/mobile/check-react-renderer-pin.mjs` is therefore counted as **source**.
- `workspaceOf` matches `^apps\/([^/]+)\/` and returns `undefined` for anything else, so `scripts/mobile/check-react-renderer-pin.test.mjs` is dropped before it can be run.

So the gate sees a source change with no verifiable test and fails closed. That is the carve-out for `scripts/ci/` never having been widened to the other three script directories, even though `test:scripts` has run all four for a while:

```
"test:scripts": "node --test scripts/ci/*.test.mjs scripts/security/*.test.mjs scripts/roadmap/*.test.mjs scripts/mobile/*.test.mjs"
```

Two ways out, and I have a preference:

1. **Minimal and consistent** — widen the `isSourceFile` exclusion from `^scripts\/ci\/` to `^scripts\/`, matching what `test:scripts` already covers. This makes the gate treat `scripts/mobile` exactly as it treats `scripts/ci` today.
2. **The real fix** — teach the gate to run `node --test` for `scripts/**` tests so they are genuinely mutation-checked instead of merely exempt. Bigger, and probably its own PR.

**Not acceptable here: the `no-behaviour-change` label.** It is the documented escape, and it would turn this green in one click, but it asserts the source change alters nothing observable — which is false for a PR whose entire point is new CI behaviour. That would be weakening a gate to pass it.

Your 533/533 `test:scripts` run is real and I am not questioning it; the gate simply never got to look at those twelve cases.

## Blocker 2 — Aikido is red

`Aikido Security: check code` reports **1 new LOW issue introduced** at `516564e68` (scan 197454165). I cannot see the finding body from the API — only the summary — so I am not going to guess at the site. Per the repo's standing rule it gets fixed rather than suppressed. If it turns out to be a non-literal `fs` path from `join(root, RENDERER)`, say so and we can decide whether a literal-path read is worth the restructure.

## One note, not a change request

`apps/mobileAppYC/android/gradle.properties:35` has `newArchEnabled=true`, and Fabric is the bundle with no version guard. That does not weaken the check — both bundles ship in the same `react-native` package and are built against the same react, so the Paper literal remains a correct statement of which react this RN expects. What could go stale is the *justifying sentence* ("the release app throws at startup") if the app ends up entirely off Paper. Worth a clause in the header comment so a future reader does not conclude the check is measuring something that can no longer happen.

Re-request me once the two checks are green and I will approve; nothing in the logic needs changing as far as I can tell.

ankit-yc (approved):
Approving at e776761876c6d19867b4059260bcdaabb970e563. Both blockers are properly fixed rather than labelled around, and the gate fix is wider than this PR needed it to be — `scriptTestsOf` covers `scripts/security` and `scripts/roadmap` as well as `scripts/mobile`, so the next author does not re-earn this.

**The decoy fixture now discriminates — I mutated it rather than taking the claim.** The old arm was `assert.equal(rendererExpectation(bundle('19.1.4')), '19.1.4')`, byte-identical to the test above it, so it could not fail for the reason it named. With the decoy moved before the throw and past the window, `WINDOW` is pinned from both sides:

```
WINDOW = 400  (as shipped)   12 pass, 0 fail
WINDOW = 900  decoy in range 11 pass, 1 fail
WINDOW =  60  gap too narrow 11 pass, 1 fail
```

Two mutants, opposite directions, one kill each. That is a real constant now, not a number with a comment next to it.

**`runChangedScriptTests` fails closed in the direction that matters, and the comment explaining why is the most valuable thing in the diff.** An empty changed `.test.mjs` exits 0, returns `true`, and that is the verdict which *rejects* the branch — the opposite of the jest side, where the same gap read as a pass and had to be caught by the shortfall reconciliation. The instruction not to "make this symmetrical" is correct and worth keeping verbatim; symmetry here would be a silent regression.

**One thing to be aware of, not a change request: this gate's PASS on this PR is weak evidence, and that is inherent, not a defect you introduced.** Reverting the source of a *new* module deletes the file, so the test fails at import — which is exactly what the log shows ("renderer-pin test failed on the missing import"). Any test file at all, including an empty one that merely imports the module, produces that same failure. So "PASS: the changed tests fail without the source change" is earned by a module-not-found here, and says nothing about the twelve cases. The same is true of the jest path for any new file, so this is the gate's standing semantics rather than something to fix in this PR — but nobody should read the green as evidence the tests are good. The evidence for that is the mutation matrix above and the live arms below.

**Re-confirming the substance from my first review, since the module is unchanged**: `rendererExpectation` returns `"19.1.4"` off the real 365 KB `react-native 0.81.6` bundle; the mismatch arm fires on this desk's genuine 19.3.0-vs-19.1.4 disagreement; renaming the throw message and planting a second candidate both return `null` rather than a guess; `Incompatible React versions` is 0 in all three Fabric bundles and 1 in each Paper bundle.

`Tests Must Be Able To Fail` and `Aikido Security: check code` are both green. Only `core / Build shared packages and frontend` and `Analyze (javascript-typescript)` were still running when I approved — merge on green.

On the collision: two desks writing the same file in one worktree is the shape that loses work rather than duplicating it, and it was caught here only because you happened to watch the file. Worth one line in the channel about which desk owns a PR before a second is dispatched onto it; nothing for this diff.

PR #3350, opened 2026-09-19

ankit-yc:


aupyay (approved):
Back-merge of main into dev. Diff is comment-only in .github/dependabot.yml (no changed line outside a comment), carrying the #3348 corrections back so the dev and main copies stay byte-identical. CI Required and Supply Chain Required both green at c0939a5; origin/main is still c0939a5. Approving as aupyay because ankit-yc opened this PR.

PR #3352, opened 2026-09-19
ci(repo): promote the override advisory nightly to the default branch

aupyay:
## Why

#3346 added `.github/workflows/override-advisories.yml` on `dev`. Its `schedule:`
trigger cannot fire from there — GitHub only runs scheduled workflows from the
**default branch**, and the file 404s on `main`:

```
GET /repos/YosemiteCrew/Yosemite-Crew/contents/.github/workflows/override-advisories.yml?ref=main
  -> 404 {"message":"Not Found"}
?ref=dev
  -> 6722281571946ab51e74685168048dcdfe70760a
```

So the nightly is inert until the file itself lands on `main`, independent of
what the job does once it starts. `workflow_dispatch --ref dev` is the only way
to run it today.

## Why the file alone is enough

The job checks out `ref: ${{ matrix.branch }}` and audits `main` and `dev` as
separate matrix legs, so it resolves its dependencies from **each matrix
branch**, not from the default branch. Both branches already carry all three:

| dependency | state |
|---|---|
| `scripts/ci/check-override-advisories.mjs` | blob `5e6a773` on `main` **and** `dev` — identical, and it parses `--strict` |
| `.github/actions/setup/action.yml` | present on `main` (`90580f5`) |
| `scripts/ci/override-advisory-baseline.json` | present on both, one per branch, as the workflow's own comment describes |

The promoted file is byte-identical to dev's copy (blob `6722281`), so this does
not open the kind of main/dev drift PR #2079 was about.

## What this does not claim

Whether the first run **passes**. The check needs the registry, so its verdict is
only knowable from a real run — I will dispatch it on `main` after merge and
report the result rather than wait for 06:40. A finding on the first run is the
workflow doing its job, not a defect in this PR.


ankit-yc (approved):
Promotion is the only way this workflow can ever fire: schedule triggers run from the default branch only, and the file 404s on main. Verified the three runtime dependencies resolve on both matrix legs - check-override-advisories.mjs is the same blob 5e6a773 on main and dev and parses --strict, .github/actions/setup and a per-branch override-advisory-baseline.json are present on both. File is byte-identical to dev's 6722281, so no main/dev drift. main-direct is the guard's own documented hatch for a change that must take effect before the next promotion, matching #3348.

PR #3353, opened 2026-09-19

ankit-yc:


aupyay:
Superseded by #3367, which carries this same merge with the two conflicts resolved. This PR cannot be fixed in place: its head branch is `main` itself, so there is nowhere to push a resolution. Closing once #3367 lands.

For the record, the red `sonar / Sonar scan (desktop|frontend)` here are push runs on `main`, not runs on this PR - this PR was CONFLICTING, so no `pull_request` workflow ever ran on it. They are `main`'s own checks on the shared head sha.

aupyay:
Closed as superseded by #3367, which merged the resolved ancestry restoration into `dev` as `fbf5a58bc8292c485db2a3e0ca42a2b670398eaf`.

PR #3355, opened 2026-09-19
fix(repo): raise the adm-zip override on main to the patched 0.6.1

### LINKED ISSUE #3354: A pnpm override on main pins a version with an open advisory
The nightly override advisory check reported a finding on `main`:

```
check-override-advisories: 117 override entries checked
  advisories in tree:  9
  vulnerable pins:     2
  uncovered copies:    0
  accepted (baseline): 1

Accepted drift (already recorded in the baseline):
  "adm-zip" -> 0.6.0 - GHSA-vwc7-r8mq-g2x9 [moderate] - no fix published

check-override-advisories: 1 override entry pins a version with a known advisory

  adm-zip  [high]
    override key:  "adm-zip"
    pinned at:     0.6.0
    installed:     0.6.0
    advisory:      GHSA-7q85-xj36-vmfc
    title:         adm-zip: Uncontrolled memory allocation via the declared uncompressed size (DoS)
    vulnerable:    <0.6.1
    fixed in:      0.6.1  (patched range >=0.6.1)
    url:           https://github.com/advisories/GHSA-7q85-xj36-vmfc

For a stale pin, raise the value in the pnpm.overrides block of package.json to the 'fixed in' version above, then re-run pnpm install. Bumping the dependent package alone will not work: the override pins resolution regardless.
If a fix is genuinely blocked, record it in scripts/ci/override-advisory-baseline.json with a reason.
```

An override beats every manifest, so dependabot cannot clear this on
its own: raising a manifest leaves the override pinning resolution
straight back to the vulnerable version. Edit the `pnpm.overrides`
block in the root `package.json` on `main`, or - if the pin is
deliberate - record it in `scripts/ci/override-advisory-baseline.json`
with the reason.

Run: https://github.com/YosemiteCrew/Yosemite-Crew/actions/runs/35414816390

aupyay:
Closes #3354

Raises the `adm-zip` pnpm override on `main` from 0.6.0 to 0.6.1. An override beats every manifest, so the pin is the only place this can be fixed.

Both adm-zip advisories in the tree clear at 0.6.1:

- `GHSA-7q85-xj36-vmfc` (high, uncontrolled memory allocation / DoS) - vulnerable `< 0.6.1`, first patched 0.6.1.
- `GHSA-vwc7-r8mq-g2x9` (medium, extraction follows destination symlinks) - vulnerable `>= 0.5.9, <= 0.6.0`. It has no `first_patched_version` recorded, but 0.6.1 is outside the vulnerable range.

The baseline acceptance for `GHSA-vwc7-r8mq-g2x9` was recorded when 0.6.0 was the latest release and there was nothing to raise the pin to. 0.6.1 exists now and leaves the range, so the acceptance is removed rather than carried forward. Removing it makes the check stricter, not weaker: if 0.6.1 were still in range the nightly would fail loudly instead of staying silent.


aupyay:
-

ankit-yc (approved):
Approving at `bcbfdd4`. Verified independently rather than re-reading the body.

**Lockfile is genuine, not hand-edited to a plausible hash.** A 7-line lock diff is the shape where a wrong integrity slips in, so I checked it at the registry:

```
curl -sS https://registry.npmjs.org/adm-zip/0.6.1
  integrity sha512-Xwrja8nx9e5o2N1my4DsKCeKpdrnACyr1wtbPxBDgGzKzKyE9kRtBFA8mWldI+RVlD7CBZNWY/wQ2+ydwOR6kQ==
  engines   {"node": ">=14.0"}
```

Byte-identical to `pnpm-lock.yaml:14698`, and `engines` matches, so this is a real resolution and not a retyped line. No `adm-zip@0.6.0` survives anywhere in the lock at this head — the only remaining `0.5.8` hits are `@types/adm-zip`, which carries no advisory.

**The baseline drop removes a different advisory from the one the nightly filed, and that is still correct.** Worth writing down because the two IDs are easy to conflate:

| advisory | severity | vulnerable range | first patched |
|---|---|---|---|
| GHSA-7q85-xj36-vmfc | high | `< 0.6.1` | `0.6.1` |
| GHSA-vwc7-r8mq-g2x9 | medium | `>= 0.5.9, <= 0.6.0` | `null` |

The nightly filed #3354 on the first. The baseline entry being deleted accepts the **second**, whose `first_patched_version` is still null — so "no patched version exists" was true when it was written and reads, at a glance, as still true. It is not: the vulnerable range is upper-bounded at `<= 0.6.0`, and `0.6.1` falls outside it. `pnpm audit` matches on the range, not on the patched field, so the finding stops being reported and the entry stops matching. `scripts/ci/check-override-advisories.mjs` keys an accepted entry on `package@pinned:advisory` (line 279), and `pinned` is now `0.6.1`, so the entry was already dead weight the moment the override moved. Deleting it matches the contract the file documents for itself: *"Raise the override ... and the entry stops matching."*

Dropping an acceptance is also the safe direction — it can only cause a finding to be reported, never suppressed. Nothing here weakens the gate.

**Two things for the follow-up, neither blocking.**

1. `apps/backend/package.json` and `packages/fhir/package.json` still declare `^0.6.0` on main where dev declares `^0.6.1`. The root override forces `0.6.1` either way and the lock proves it (`specifier: 0.6.1` on both importers), so there is no exposure — but the declared ranges now differ from dev and will surface as noise in the next promotion diff.
2. GHSA-vwc7-r8mq-g2x9 has no patched version upstream. If its range is later widened to cover `0.6.1`, the leg goes red on main with no accepted entry and no version to raise to. That is the correct outcome, just not a quiet one — expect it to need a fresh acceptance rather than a bump.

Gates at this head: Aikido, grype, Gitleaks, GitGuardian, Supply Chain Required, Review dependency changes all green; `main-direct` is on the PR and `promotion-guard.yml:109` exempts it from the dev-only ancestry rule. `back-merge.yml` carries this to dev automatically, so no manual back-merge is needed after the merge.

PR #3362, opened 2026-09-19
fix(backend): submit checkout lines at an amount the currency can represent

### LINKED ISSUE #3305: [Bug]: Stripe checkout submits a per-unit amount, so an unrepresentable unit price under-charges the invoice
## What happens

`buildCheckoutSessionLineItems` submits each invoice line to Stripe as a
per-UNIT `unit_amount` with a `quantity`, and Stripe charges the product of the
two. A unit price that is not exactly representable at the currency's precision
therefore loses its remainder once per unit, and the session collects less than
the invoice total.

Measured on the current `dev` behaviour (`apps/backend/src/services/finance/payment.ts`):

| line | invoice total | submitted to Stripe |
| --- | --- | --- |
| 1 x 8.165 USD | 817 | 816 (balance line: 817) |
| 2 x 8.165 USD | 1633 | 2 x 816 = 1632 |

The 2-unit row is the damaging one. The pre-tax guard immediately above the
line-item build compares the item sum against the invoice total to decide
whether to charge a single balance line instead; two 8.165 units sum to 16.33
under both roundings, so the guard finds nothing wrong, the session is
itemised, and the customer is charged a cent less than the invoice says they
owe. The checkout-completed handler then settles the invoice as paid in full.

The 1-unit row is caught today, because the two roundings disagree there and
the balance-line branch charges the exact amount.

## Why it is not a rounding curiosity

- The shortfall is per unit, so it scales with quantity.
- It is silent: no branch, no log, and the invoice reconciles as paid.
- It predates the exact-money work in #3153. Invoice pricing rounding at two
  decimals produced the same 16.33 total and the same 1632 charge.

## Suggested fix

Submit each line at an amount the currency can represent, rather than at a
per-unit price Stripe multiplies. Either send `quantity: 1` with the line's
own rounded total as `unit_amount`, or allocate the rounded line total across
units the way `allocateInvoiceDiscountAcrossLines` already allocates a discount,
so the submitted units sum to the posted total.

## Where it is pinned

`apps/backend/test/services/finance.payment.test.ts` holds both rows as
characterisation tests ("under-submits an itemised line whose unit price is not
representable" and "charges the balance line at the posted total when a tie
splits the two roundings"), so this behaviour cannot be changed without
changing a test.

Found while reviewing the first slice of #3153.

aupyay:
Fixes #3305.

## The defect

`buildCheckoutSessionLineItems` submitted each invoice line to Stripe as a per-UNIT `unit_amount` with a `quantity`, and Stripe charges the product. A unit price that is not exactly representable at the currency's precision therefore lost its remainder once per unit, so the session collected less than the invoice posted.

The pre-tax guard above the line-item build could not see it, because it summed the posted line totals while the session submitted something else. Two 8.165 USD units sum to 16.33 under either rounding, so the guard found nothing wrong, the session was itemised at `2 x 816 = 1632` against a 1633 invoice, and the checkout-completed handler settled the invoice as paid in full. Silent, and it scales with quantity.

## The fix

- A line whose units cannot reconstruct its posted total is submitted as a single unit at that total, with the quantity moved into the product description (`2 x Consult`), so the checkout still says how many were bought.
- A line whose units *do* reconstruct it — every unit price the currency can represent — keeps Stripe's per-unit presentation unchanged (`3 x $12.50`).
- The guard is now built from the line items themselves rather than from a second formula for them, and compared in minor units. Integers, and the same integers the session sends, so the guard weighs the money that will actually be collected.

The tie case (#3153's quantizer disagreeing with `roundMoney` on 8.165) still selects the balance line and still charges the posted total; that branch is unchanged.

## Verification

Run in `apps/backend` at `6a8c5defe`:

- `npx jest` — **511 suites, 12232 tests, all passed**, rc=0. Full package suite, not a scoped run.
- `npx tsc --noEmit` — no errors in `src/services/finance/payment.ts` lines 752-900 (the changed region). The workspace is not built locally, so unrelated `Cannot find module '@yosemite-crew/auth'` noise remains; the three pre-existing implicit-`any` reports in this file are at lines 236/239/1921, outside the diff.
- `npx prettier --check` on both files — clean.

### Mutation checks

The two characterisation tests in `test/services/finance.payment.test.ts` were rewritten, so each new assertion was checked against a deliberately broken implementation:

| mutation | result |
| --- | --- |
| `unitsReconstructLine = true` (restore the old always-per-unit behaviour) | 1 failed |
| `unitsReconstructLine = false` (collapse every line) | 2 failed |
| guard forced to `false` (never selects the balance line) | 3 failed |
| collapsed description drops the quantity | 1 failed |
| unmutated | 131 passed |

Both directions of the new branch fail when broken, so neither test is vacuous. The second test exists specifically so the fix cannot be implemented by collapsing every line.


ankit-yc (approved):
Approving. The defect is real, the fix is at the right level, and the two new tests fail in both directions of the branch they cover. Reviewed at `6a8c5defe`; verification below is mine, run in a fresh detached worktree at that sha rather than in the tree the PR was written in.

## The guard change, stated as what it now means

Worth writing down explicitly, because the PR body says the guard is "built from the line items themselves" without saying what that makes it equal to. By construction of the two branches:

- `unitsReconstructLine` true → `unit_amount * quantity` = `unitAmount * quantity` = `lineAmount`
- `unitsReconstructLine` false → `unit_amount * quantity` = `lineAmount * 1` = `lineAmount`

So `itemisedTotal` is always `Σ lineAmount_i` — the posted line totals, each rounded to minor units, then summed. The guard has therefore moved from **sum-then-round in major units** to **round-then-sum in minor units**, and that is the whole behavioural delta beyond the collapse.

That reclassifies invoices in both directions. Computed with this file's own `roundMoney` and `toStripeMinorUnits`:

| lines | pre-tax total | old | new |
| --- | --- | --- | --- |
| 16.325, 16.325 | 32.65 | itemised | **balance line** |
| 10.004, 10.004 | 20.00 | balance line | **itemised** |
| 37.50, 20.00 | 57.50 | itemised | itemised |
| 16.33, 16.33 (#3305) | 32.66 | itemised | itemised |

Both moves are safe, and for the same reason: the new predicate is exactly "the lines I am about to send collect the pre-tax total", so itemisation now happens if and only if it is exact, and everything else charges `summary.balance`. Row 1 is the "more multi-line invoices down the balance line" you flagged — it is a presentation loss on invoices whose line totals carry sub-cent precision, bought with the shortfall going away. Row 2 is the opposite and is an improvement nobody claimed: an invoice the old formula sent to the balance line purely because sum-then-round disagreed at the third decimal is now itemised, correctly.

I also checked the interaction with `disableAutomaticTax`, since it is derived from `useBalanceLine`. Row 1 turns it on only when `taxTotal > 0`, which is the tax-inclusive case the existing comment covers; row 2 turns it off, which is correct for itemised pre-tax lines. No new path through that flag.

## Mutation arms reproduced

All four of yours, restored after each, run against `test/services/finance.payment.test.ts`:

| mutation | your result | mine |
| --- | --- | --- |
| `unitsReconstructLine = true` | 1 failed | 1 failed, 130 passed |
| `unitsReconstructLine = false` | 2 failed | 2 failed, 129 passed |
| guard forced to `false` | 3 failed | 3 failed, 128 passed |
| collapsed description drops the quantity | 1 failed | 1 failed, 130 passed |
| unmutated | 131 passed | 131 passed, 131 total |

Exact agreement, and the second arm is the one that matters: it is what stops the fix being implemented by collapsing every line, and a reviewer has no way to know that from reading the test.

## One gap, non-blocking

I added a fifth arm you did not run — make `lineAmount` ignore the stored `total` and always recompute `unitPrice * quantity * (1 - discountPercent / 100)`:

```
Tests:       131 passed, 131 total
```

**It survives.** The preference for the posted `total` over the recomputed product is unpinned. In both new tests `total` equals the product exactly (16.33 and 8.165 x 2, 37.50 and 12.50 x 3), so nothing distinguishes them.

It is not a money bug as the code stands, and it is not one under the mutant either — `calculateLineDiscount` supports `FIXED_AMOUNT`, so a line's stored `total` can be `quantize(quantity x unitAmount) - fixedDiscount`, which no `discountPercent` reproduces; the recomputed sum would then miss the pre-tax total and the guard would divert the whole invoice to the balance line. Correct amount, itemisation silently lost for every fixed-discount invoice, and no test red. Since `Invoice.items` is a `Json` column, the shapes that can appear there are not constrained by the current writer either.

One case would close it: a line with a `FIXED_AMOUNT`-derived `total` that does not equal `unitPrice x quantity`, asserting the submitted amount follows `total`. Worth adding, not worth holding this.

Second, smaller: a collapsed line's `description` is replaced by `"2 x Consult"`, discarding whatever the invoice line said about itself. The comment justifies not appending on the grounds that a conditional prefix would have an unreachable empty arm, which is true of the *prefix* but not of the *suffix* — `typed.description` is genuinely optional, and your own second test pins that. The customer-visible effect is that a line both collapsed and described loses the description at checkout. Presentation only.

## Checks at `6a8c5defe`

```
CI Required                          SUCCESS
test / Merge coverage (backend)      SUCCESS
sonar / Sonar scan (backend)         SUCCESS
[backend] SonarCloud Code Analysis   SUCCESS
Aikido Security: check code          SUCCESS
CodeQL                               SUCCESS
Supply Chain Required                SUCCESS
Frontend / Desktop E2E Required      SUCCESS
```

Nothing failing, nothing pending. Sonar's new-code condition passes, so the rewritten characterisation test counts as coverage of the changed region.

Merging.

PR #3363, opened 2026-09-19
fix(desktop): make the self-drawn title bar behave like a native one

### LINKED ISSUE #3293: Desktop: maximise button never switches to Restore, and the title bar cannot restore a maximised window
## What happens

The window buttons the app draws itself (Windows and Linux) do not behave like native ones, and the macOS traffic lights are slightly out of line with the tab strip:

- When the window is maximised, the middle button still shows the single-square "Maximize" icon, tooltip and accessible name. It never changes to "Restore", so users cannot tell the state and screen readers announce the wrong action.
- Double-clicking the empty title-bar area does nothing, and dragging a maximised window does nothing. On a native Windows title bar, double-click toggles maximise and dragging restores the window under the pointer.
- The minimise icon renders blurred and faint at 100% display scaling (2.2:1 against the bar, where maximise and close reach 5.9:1).
- On macOS the traffic lights sit about 3px above the centre line of the 40px tab strip.

Screenshots from the 2026-09-18 desktop UI check: `caption-controls-maximised-light-2560x1364.png`, `caption-controls-maximised-hover-light-2560x1364.png`, `caption-controls-light-1280x800.png`, `caption-controls-dark-1280x800.png`, `tabs-light-1280x800.png`, `tabs-maximised-light-2560x1364.png`. The Windows-style buttons were checked on macOS with the Windows controls turned on; the markup does not change by platform, so a Windows build renders the same.

## Where

Screen: main window tab bar, window buttons.

- Restore state: `apps/desktop/src/pages/tabbar.html:281-285` hard-codes `aria-label`/`title` "Maximize" and a single `<rect>`. `tabbar.html:902-912` only wires the click. `apps/desktop/src/main.ts:1520-1525` toggles maximise, but no `maximize`, `unmaximize` or full-screen event, and no `isMaximized()` state, is ever sent to the tab bar (`apps/desktop/src/preload.ts:159` is send-only).
- Double-click: `.drag-spacer` (`tabbar.html:257`) has only pointer handlers (`tabbar.html:404-439`), no `dblclick`.
- Drag to restore: `moveMainWindowBy` returns early when the window is maximised (`main.ts:1499-1501`).
- Minimise icon: `tabbar.html:279` draws `<path d="M2 6h8">` with a 1.1 stroke (`tabbar.css:295-301`) on a pixel boundary, so it spreads across two half-covered rows.
- Traffic lights: `apps/desktop/src/shell/create-main-window.ts:123` sets `trafficLightPosition { x: 12, y: 10 }`, which was not tuned for `CHROME_STRIP_HEIGHT = 40` (`main.ts:182`).

## Checklist

- [ ] Maximise button shows a Restore icon and label while the window is maximised.
- [ ] Double-clicking the empty title-bar area toggles maximise.
- [ ] Dragging a maximised window restores it under the pointer and then moves it.
- [ ] Minimise icon renders at the same weight and contrast as maximise and close.
- [ ] macOS traffic lights are vertically centred on the 40px tab strip.

## Expected

- Main forwards `maximize`, `unmaximize`, `enter-full-screen` and `leave-full-screen` (plus the initial state) to the tab bar. While maximised the button shows the two-overlapping-squares Restore icon with label and tooltip "Restore", and switches back on unmaximise, including when the window is maximised by snapping or double-click.
- The empty title-bar area behaves like a native one for double-click and drag.
- The minimise path is snapped to the pixel grid (`M2 6.5h8`, matching the .5 offsets the maximise rect already uses) or uses `shape-rendering: crispEdges`.
- The traffic lights line up with the tab labels (about `y: 13`).

aupyay:
Closes #3293.

The window buttons the app draws itself did not reflect or change window state
the way a native title bar does. All five checklist items in the issue.

### Maximise button never became Restore

`main.ts` toggled maximise but never told the tab bar anything about window
state (`preload.ts` was send-only), so the middle button kept the single-square
icon, the "Maximize" tooltip and the "Maximize" accessible name while the window
was maximised — a screen reader announced the opposite of what the button does.

Main now forwards `maximize`, `unmaximize`, `enter-full-screen` and
`leave-full-screen`, plus the state at the moment the tab bar finishes loading
(a session can restore straight into a maximised window, before any event
fires). The listeners are wired once per window rather than once per
`enterTabMode`, because closing the last tab drops back to Welcome and opening
one enters tab mode again on the same window.

Both glyphs are in the markup and `applyMaxState` toggles `hidden`, so the page
gains no `innerHTML` sink and the geometry stays somewhere a stylesheet and a
reader can find it.

Because one Restore button now serves both states, `toggleMaximizeMainWindow`
leaves full screen as well as un-maximising — otherwise the button would be
inert in one of the two states it claims to undo.

### Double-click and drag on the empty title-bar area

`.drag-spacer` had only pointer handlers, so double-click did nothing; it now
toggles maximise. `moveMainWindowBy` returned early whenever the window was
maximised, so a drag did nothing at all; it now restores the window under the
pointer — same fraction across the narrower title bar, same offset below its top
edge — and moves normally from there. Full screen still returns early: that is a
mode, not a size.

The geometry is a pure function, `restorePositionUnderCursor`, so it is unit
testable without Electron's `screen` module; the caller still clamps the result
to the work area exactly as an ordinary drag does.

### Minimise icon and traffic lights

The minimise path was drawn at `y=6` with a 1.1px stroke, so it straddled two
device-pixel rows and rendered at 2.2:1 against the strip where maximise and
close reached 5.9:1. Snapped to the half-pixel grid the other two glyphs already
use. macOS `trafficLightPosition` y 10 → 13 centres the lights on the 40px tab
strip.

### Tests

- `restorePositionUnderCursor`: the pointer fraction, the vertical offset,
  pointers outside the maximised bounds, the zero-width guard, and rounding.
- Caption labels, and a glyph-geometry check that reads the shipped markup
  rather than a copy of it, so an off-grid coordinate fails. The close glyph's
  exemption is asserted rather than assumed — it is two diagonals, which no
  coordinate can snap.
- `onWindowMaximizedChanged` on the preload bridge, including unsubscribe.
- A new Playwright Electron spec drives the real window: the IPC push on
  maximise and unmaximise, double-click toggling maximise, and a drag on a
  maximised window leaving it narrower than the screen it filled. The Restore
  label is asserted on the Windows leg of the matrix, where these caption
  buttons actually render — macOS uses native traffic lights and keeps them
  hidden.

Verified on this branch at `1f682d82b`: `pnpm --filter @yosemite-crew/desktop run lint`,
`run type-check` and `run test:coverage` all exit 0 — 79 suites, 1267 tests,
98.45% statements. `run build` copies `window-caption.js` into `build/pages/`.
Each new test was mutation-checked: reverting the minimise path, flattening
`maximizeButton` to one state, renaming the Restore glyph, moving a rect
off-grid, and dropping each of the five behaviours in
`restorePositionUnderCursor` each turn the relevant test red.

The Playwright Electron harness does not start on my machine — a control run of
unmodified `tabs.e2e.ts` hangs in `electron.launch` the same way — so the new
spec's evidence is the `Desktop E2E` matrix on this PR, which is where it runs
on both macOS and Windows.


ankit-yc (approved):
LGTM — macOS E2E tests skip fixed

PR #3364, opened 2026-09-19
fix(frontend,desktop): focus the search fields without an autofocus attribute

aupyay:
Clears the two Sonar S9379 findings on `main` that are failing `sonar / Sonar scan (frontend)` and `sonar / Sonar scan (desktop)`, and with them the red checks on #3353.

## Why these are on a PR into main

The failing run on #3353 is not that PR's. #3353 has `main` as its **head** (it back-merges main into dev), so GitHub attaches main's own push run to it — run 35416166251, `event=push`, `branch=main`, the merge of #3355. `sonar-thresholds.mjs` reports `analysed: branch main` in that log. Both projects pass SonarCloud's own gate (`qualityGateStatus: OK`) and fail our stricter `--issues 0` ceiling with one code smell each:

| project | file | rule |
| --- | --- | --- |
| Frontend | `apps/frontend/src/app/features/chat/components/MessageSearch.tsx:96` | `typescript:S9379` |
| Desktop | `apps/desktop/src/pages/command-palette.html:32` | `Web:S9379` |

Read from the issues API at `sonarcloud.io`, unauthenticated — both projects are public, so this needs no token and is reproducible by anyone.

## Why not just delete the attribute

S9379 is about focus moving without the user asking for it, which is a page-load concern. Both of these are surfaces the user has *just opened* — a search popover behind a header button, and a command palette window opened with the accelerator — where the caret belonging to the field is the correct behaviour and removing it would be a regression. So the behaviour stays and only its expression changes.

**Frontend.** The input is focused from an effect keyed on `open`, which runs on exactly the render where the panel mounts.

**Desktop.** The attribute was already redundant: the last statement of `command-palette.js` is `searchEl.focus()`, reached on both the IPC branch and the no-IPC fallback (`loadActionsAndRecents` cannot throw synchronously in either). Only the attribute is removed.

## Verification

Run at `6fb7dedc9`.

**Frontend, `apps/frontend`:**

```
npx jest src/app/__tests__/components/chat
  Test Suites: 15 passed, 15 total
  Tests:       414 passed, 414 total

npx tsc --noEmit          no output, no errors
npx eslint <changed>      no output, 0 problems
npx prettier --check      All matched files use Prettier code style!
```

Mutation arms on the new jest case, each restored afterwards:

| arm | result |
| --- | --- |
| drop the focus effect | 1 failed, 10 passed |
| drop `ref={inputRef}` | 1 failed, 10 passed |
| **before-arm: `main`'s original source + the new test** | **11 passed** |
| unmutated | 11 passed, 11 total |

The third row is the one that matters. The new test passes against the *unmodified* `autoFocus` implementation as well, which is what makes it a statement about the behaviour rather than about this diff — and the first two rows show it is not vacuous. The storybook play function already asserted `document.activeElement` is the field before this change, so that pin is independent of anything added here; only its comment needed updating.

**Desktop.** No jsdom in this workspace and the palette e2e drives IPC rather than the palette DOM, so the focus line had no test to lean on. Measured directly instead, both pages served over a local HTTP server and driven with Playwright:

| arm | `autofocus` present | `document.activeElement.id` | script executed |
| --- | --- | --- | --- |
| `main`'s HTML | yes | `search` | yes |
| this PR's HTML | no | `search` | yes |

Identical otherwise — the diff between the two probe copies is exactly the one attribute line.

Worth recording how that measurement went wrong first, because the failure looked like a result: served as-is, the after arm reported `activeElement` = `BODY`, which reads as "removing the attribute broke focus". It had not. The page's own CSP is `script-src file:`, so over `http://` the browser blocked `command-palette.js` entirely and nothing ever called `focus()` — and the before arm still "passed", because `autofocus` needs no script. The control that caught it is the `script executed` column (`#results` is populated by `render()`, so a non-empty value means the IIFE reached its end). Both arms were then re-run with `file:` relaxed to `'self'` **in both probe copies identically**, leaving the attribute as the only difference. Neither the CSP nor anything else in the shipped file is changed by this PR.

## Not done here

Nothing is excluded or suppressed — `sonar-project.properties` is untouched in both apps.


ankit-yc (approved):
Checked against `origin/main f45175baf`.

- `apps/desktop/src/pages/command-palette.js:266` `searchEl.focus()` is the final statement of the IIFE, unconditional and after `loadActionsAndRecents()` — the IPC branch is a promise chain that returns immediately and the fallback branch is synchronous, so neither can skip it. Dropping the attribute leaves the field focused.
- `MessageSearch.tsx:23` `open` initialises to `false`, so the new `[open]` effect cannot fire on mount. It only runs on the render where the user opened the panel, which is what takes it out of S9379's scope rather than hiding from it.
- `ref={inputRef}` is on the input that is rendered inside the `open` branch, so the ref is populated by the time the effect commits.
- Nothing suppressed: both `sonar-project.properties` files are untouched in the diff.

One residual worth recording rather than fixing here: on desktop the focus now depends on the script executing, where the attribute did not. The shipped page loads over `file:` and its CSP admits the script there, so this only bites in a degraded load — which is the same condition under which the palette has no actions in it either.

LGTM.

PR #3365, opened 2026-09-19
test(backend): pin that a checkout line submits its posted total

aupyay:
Follow-up to #3362 (issue #3305). Test only — no production code changes.

## What the review found

The review of #3362 ran a fifth mutation arm I had not: replace

```ts
typeof typed.total === "number" ? typed.total : unitPrice * quantity * (1 - discountPercent / 100)
```

with the recomputed product alone, dropping the read of the invoice's stored line snapshot. It survived.

I reproduced it on `dev` at `4bc5471a9`: with the snapshot read removed, `npx jest` in `apps/backend` reported **511 suites / 12232 tests, all passed**. Nothing in the repository pinned that preference.

## Why it matters

The two expressions agree for every percentage-discounted line, which is every line `buildInvoiceLineSnapshots` writes today, so no existing row can tell them apart. They stop agreeing as soon as a line total is not the percentage product:

- `buildInvoiceLineSnapshots` (`apps/backend/src/services/invoice.service.ts:356`) computes `item.total ?? quantity * unitPrice - percentDiscount`, so a caller-supplied `total` passes through unvalidated — and `normalizeInvoiceLineItem` supplies one from the stored row on every re-save.
- `calculateInvoicePricing` already takes a `discountType`; this call site only ever emits `"PERCENTAGE"`, but a flat per-line discount is representable in the pricing layer.

In that case the recomputed amount disagrees with the posted invoice total, the pre-tax guard fires, and the session silently drops itemisation for the balance line. It charges the right money — this is a presentation regression, not a shortfall, which is why the review called it non-blocking — but it would happen with no test failing.

## The test

One line, `2 x 10.00` with a posted total of `15.00` (a flat 5.00 off), which the product does not reproduce. Asserting the line's own product name is what separates the two outcomes: the balance-line fallback also charges 1500 at quantity 1, so amount and quantity alone would pass under the mutation.

## Verification

Run in `apps/backend` at `28fd0a18b`:

- `npx jest` — **511 suites, 12233 tests, all passed**, rc=0. Full package suite.
- `npx prettier --check` on the touched file — clean.

Mutation check, same command, `test/services/finance.payment.test.ts`:

| | result |
| --- | --- |
| `lineAmount` recomputes the product (the surviving arm) | **1 failed** — `submits the posted line total rather than recomputing it from the unit price` |
| unmutated | 132 passed |

The arm is now killed, and by this test specifically rather than by collateral.


ankit-yc (approved):
Test-only change, reviewed against the mutation it claims to kill.

Confirmed the gap is real: `buildInvoiceLineSnapshots` writes `item.total ?? quantity * unitPrice - percentDiscount`, so a caller-supplied total passes through, and every existing row is percentage-discounted — which is exactly why no current fixture separates the snapshot read from the recomputed product. Asserting the line's own product name (rather than amount/quantity, which the balance-line fallback also satisfies at 1500 x 1) is the right discriminator.

CI is fully green: 511 suites / backend, Sonar backend quality gate passed, 0 new issues, 0 hotspots. No inline findings. Codex is out of repo credits, as on #3357/#3362/#3363.

Approving.

PR #3366, opened 2026-09-19
fix(backend): bound the public contact message at the SuperAdmin intake's limit

### LINKED ISSUE #3361: fix(backend): bound the public contact message at the SuperAdmin intake's limit so a submission cannot be accepted and then permanently refused by the mirror
## Summary

A web contact submission whose `message` is longer than the SuperAdmin panel's 5000-character limit, or whose `email` the panel cannot parse, is stored here, queued for the mirror, refused once with 400, and parked in `failedAt`. The CRM never receives it. Nothing in this repository stops such a submission being accepted in the first place.

Raised in review of #3357 (which implements #3329 and #3330).

## Problem / motivation

`createWebRequest` (`apps/backend/src/services/contact-us.service.ts:112-127`) validates presence only: `type`, `message.trim()`, `fullName.trim()`, `email.trim()`. There is no length bound on `message`, and the public forms do not impose one either — neither `ContactusPage.tsx` nor `AccessibilityReportClient.tsx` sets a `maxLength` on the message field, and `validateContactForm` checks only that it is non-empty.

The panel's intake caps `message` at 5000 and answers 400 when `parseSubmission` returns null. `sendOne` (`apps/backend/src/services/superadmin-contact.service.ts:259-272`) treats 400 as permanent and does not retry, which is correct — the same content cannot succeed on a second attempt. The result is that an over-long submission is visible here and absent from the CRM.

This is not a regression: before #3329 the forward was fire-and-forget and such a submission was not delivered either, and was not recorded as undelivered. The new behaviour records `failedAt`, `lastStatus: 400`, counts the row in the drain summary's `failed`, and logs the id. It is strictly better. It is still a gap.

The `email` variant is reachable in backfilled history even though the live path now requires `email`.

## Proposed solution

Make the limit the same on both sides of the mirror, closest to the visitor first:

1. A `maxLength` on the message field of both public forms, with the usual character counter, so a visitor cannot write past the limit unknowingly.
2. A matching length check in `createWebRequest`, rejecting with `ContactServiceError(400)` as the other field checks do, so the API holds the invariant even without the form.
3. The limit as one exported constant, referenced by both, so the two cannot drift.

Fixing this at the source rather than in the forward is deliberate: truncating in `buildIntakeBody` would deliver a submission the CRM believes is complete, and the product database is the documented source of truth for the full text.

## Scope

The two public forms, `createWebRequest`, the shared constant, and their tests.

### Out of scope

- The panel's own validation; its limit is the one being matched.
- Backfilled rows that already exceed the limit. They fail loudly with the id logged, which is what the follow-up below is for.
- Retry policy for 400. It is correctly permanent.

## Acceptance criteria

- [ ] A submission over the limit is refused with 400 by `createWebRequest`, before any row or forward is written.
- [ ] Both public forms bound the field at the same number and show the visitor the count.
- [ ] The number appears once in the tree, not once per site.
- [ ] A submission at exactly the limit is accepted and forwards successfully.

## Test plan

- [ ] `createWebRequest` rejects at limit+1 and accepts at the limit, asserting `prisma.contactRequest.create` is not called in the reject case.
- [ ] Form tests for the bound and the counter.
- [ ] Mutation checks: drop the length check (the reject test fails); set the two limits to different numbers (the shared-constant test fails).

## Dependencies & risks

- Follows #3357. No dependency on the rollout order in #3330.
- Historic rows already over the limit stay undeliverable; the drain's `failed` count and the logged ids are how they are found. Deciding what to do with them is separate work.

## Additional context

Tracker: YosemiteCrew/SuperAdmin#315.

aupyay:
Fixes #3361.

## What was wrong

A web contact submission longer than the SuperAdmin panel's intake accepts was stored here, queued for the mirror, refused once with 400, and parked in `failedAt`. Because 400 is permanent and correctly not retried, the CRM never received it. Neither public form stopped such a submission being accepted: `createWebRequest` validated presence only, and no `maxLength` existed on either message field.

## The fix

`CONTACT_MESSAGE_MAX_LENGTH` in `@yosemite-crew/types` is the only place the number lives.

- **`createWebRequest`** refuses past it with `ContactServiceError(400)` before `prisma.contactRequest.create` runs, so no row and no forward exist for a submission that cannot be delivered. The bound is on the *trimmed* message, because that is what is stored and what `buildIntakeBody` POSTs.
- **`/contact-us`** bounds its message field at that number and shows the count. The bound sits in `TextAreaField`, which is used only for this field, so all three enquiry types get it once.
- **`/accessibility/report`** does not send what the visitor types: the description is wrapped in a header (page URL, severity), and the limit applies to the whole message. Its description is bounded at what is left once that header is charged, so a full-length report is exactly at the limit rather than just over it. `validate` also refuses an over-long report on submit — `maxLength` stops new typing, it does not shorten text already in the field when the page URL grows. The URL field is bounded so it cannot eat the whole budget.

Fixing this at the source rather than truncating in `buildIntakeBody` is deliberate: a truncated forward would tell the CRM it had the complete submission.

`FormDesc` and `FormInput` gain an optional `maxLength`; when it is not passed their DOM is unchanged, so the other 40-odd consumers are unaffected.

## Evidence

Run on `04ca4395b`, branched from `a4336e3e` (current `dev`):

- `pnpm --filter backend run test` — rc=0, **514/514 suites, 12271 tests** (12268 on the base + 3 new).
- `pnpm --filter frontend run test` — 1072/1073 suites. The one failure is `AddInventory/FormSection.test.tsx` (2 date-format tests), **reproduced identically on the base commit `a4336e3e`** — a machine-timezone artefact, not this change.
- `type-check` rc=0 on `@yosemite-crew/types`, `backend` and `frontend` (real exit codes, not a pipe's).
- `lint` rc=0 on backend and frontend, 0 errors.

**Mutation checks — 8 mutants, all killed, baseline restored green after each:**

| Mutant | Test that fails |
|---|---|
| length guard deleted | over-long refusal + "no row written" |
| `>` → `>=` | at-limit acceptance |
| `.trim()` dropped from the measurement | untrimmed at-limit acceptance |
| status 400 → 500 | statusCode assertion |
| `maxLength` removed from `TextAreaField` | contact-form bound + counter |
| counter removed | contact-form counter |
| accessibility budget → raw limit | "full-length report is exactly at the limit" |
| `validate` over-length branch removed | "refuses rather than posting" |

**Driven in the running app** (`next dev`, both routes 200):

- `/contact-us`: textarea `maxLength` 5000, counter `115 of 5000 characters`, `aria-describedby` wired to it.
- `/accessibility/report` with a 45-char page URL: description `maxLength` **4892** — the 108-char header is charged — counter and `aria-describedby` wired; URL field `maxLength` 2048.
- Counter contrast, computed from `getComputedStyle` against the first opaque ancestor (every `opacity` on the chain is 1): contact form **6.29:1** light / **5.57:1** dark at 13px; report form **5.71:1** light / **6.41:1** dark at 12px. All above AA 4.5:1.
- 390px: counter 288px wide, no viewport overflow, `scrollWidth == clientWidth` (not clipped).

The counter is described rather than announced: a live region on a per-keystroke counter reads the whole number out again on every character typed.

Not covered: the 400 was exercised at the service layer and through the controller's existing error mapping, not against a running API with a database.


ankit-yc (approved):
Checked the bound against the real intake rather than the PR description. `apps/admin/src/app/features/contact/intake.ts` in SuperAdmin sets `LIMITS.message = 5000` and `parseSubmission` refuses when `rawMessage.trim().length > LIMITS.message` — so 5000-on-the-trimmed-text is an exact match at both ends, with no off-by-one and nothing accepted here that the intake would reject. Message is also the right field to bound: every other field goes through `optionalString`, which drops an over-long value rather than returning null, so `message` (and a malformed email) are the only permanent-400 paths.

The accessibility form is the part worth the care it got. `descriptionBudget` charges the header built by the same function that builds the submitted message, so the two cannot drift; the `validate` branch is genuinely reachable because `maxLength` does not retroactively shorten a description written before the URL grew; and `Math.max(0, ...)` keeps a 2048-char URL from producing a negative bound.

Fixing at the source rather than truncating in `buildIntakeBody` is the right call — a truncated forward would claim to be the whole submission.

`FormDesc`/`FormInput` keep their existing DOM when `maxLength` is absent, so the other consumers are untouched. Counter is `aria-describedby`, not a live region, which is correct for a per-keystroke count.

CI green across backend, frontend and mobile; three Sonar gates passed; react-doctor clean; no inline findings. Codex is out of repo credits, as on #3357/#3362/#3363.

Approving.

